PR #15, opened 2020-06-08
Adding Account UI + toast notifications

piyalbasu:


src/api/internal/index.ts
```diff
@@ -141,7 +141,10 @@ export const getAccountBalance = async (
   } catch (e) {
     console.error(e);
   }
-  return response.balances[0];
+  return response.balances.filter(
+    // eslint-disable-next-line camelcase
+    (balance: { asset_code?: string }) => !balance.asset_code,
+  )[0];
```
piyalbasu, line 147: looks like if there isn't an `asset_code`, i should assume it's XLM. If that's wrong lmk!

src/popup/views/Account/basics/Footer.tsx
```diff
@@ -0,0 +1,38 @@
...
+const FooterListItem = styled.li`
+  font-size: 0.75rem;
+`;
+
+const Footer = () => (
+  <FooterEl>
+    <FooterHeader>Use Lyra with</FooterHeader>
+    <FooterList>
+      <FooterListItem>
+        <a href="/">Laboratory</a>
```
piyalbasu, line 32: not putting the actual link in for now. Mostly bc Lyra is only integrated with a specific PR version of Lab and i'd like avoid confusion for the time being

Trying to avoid "are you on the right version of lab" type issues

src/popup/components/Toast/index.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useEffect } from "react";
```
piyalbasu, line 1: toast notifications are a v small use case right now. Maybe at some pt it's worth replacing this with a lib - but let's not do that til we actually think it's necessary

PR #16, opened 2020-06-09
Create password ui and integrate Formik form lib

piyalbasu:


src/popup/Router.tsx
```diff
@@ -96,10 +96,10 @@ const HomeRoute = () => {
 
     /* 
     We want to launch the extension in a new tab for a user still in the onboarding process.
-    In this particular case, open the tab with a query param that stops subsequent redirects.
+    In this particular case, open the tab if we are in the "popup" view.
     */
-    if (location.search !== "?sameTab=true") {
-      window.open(newTabHref("/?sameTab=true"));
+    if (window.innerWidth === POPUP_WIDTH) {
+      window.open(newTabHref("/"));
```
piyalbasu, line 102: this is slightly unrelated, but was causing a problem during work on this: 

creating a more deterministic way to figure out when to open in a new tab.

package.json
```diff
@@ -32,7 +34,8 @@
     "stellar-hd-wallet": "^0.0.10",
     "stellar-sdk": "5.0.1",
     "styled-components": "^5.0.1",
-    "typescript": "~3.7.2"
+    "typescript": "~3.7.2",
+    "yup": "^0.29.1"
```
piyalbasu, line 38: adding yup, a schema forming library.

formik validation is very open ended, and not using Yup makes using Formik almost not worth it

src/popup/components/form/basics/index.tsx
```diff
@@ -0,0 +1,111 @@
+import React from "react";
```
piyalbasu, line 1: this file is all generic form elements - inputs, checkboxes, etc.

i thought about splitting this file up, but it didn't quite make sense yet. I think if we add like 1 or 2 more elements it will be time to do that. Until then, it's helpful to each element colocated with its style

src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
```diff
@@ -36,82 +37,97 @@ const ClearButton = styled(Button)`
...
-const ConfirmMnemonicPhrase = ({
-  mnemonicPhrase,
-}: {
-  mnemonicPhrase: string;
-}) => {
+const convertToWord = (wordKey: string) => wordKey.replace(/-.*/, "");
+
+const ConfirmMnemonicPhrase = ({ words = [""] }: { words: string[] }) => {
   const dispatch = useDispatch();
-  const words = shuffle(mnemonicPhrase.split(" "));
```
piyalbasu, line 45: so we need to wait to fetch the mnemonic phrase, then once we have it, shuffle all the words into a random order. but we want to only do it ONCE and not re-do on re-render. I was doing this w ref's, but it was creating a bit of a mess, i just moved this logic to the parent

src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
```diff
@@ -1,12 +1,13 @@
-import React, { useRef, useState } from "react";
```
piyalbasu, line 1: convert this view to use Formik. Needed a bit of refactoring, unfortunately doesn't save us a whole lot of work bc we still have to do most of it manually

src/popup/views/CreatePassword/index.tsx
```diff
@@ -1,37 +1,55 @@
...
+    password: YupString()
+      .min(10, "Password must be at least 10 characters long")
+      .required("Password is required")
+      .matches(/.*\d/, "Must contain a number")
+      .matches(/.*[A-Z]/, "Must contain an uppercase letter")
+      .matches(/.*[a-z]/, "Must contain a lowercase letter"),
+    confirmPassword: YupString()
+      .oneOf([YupRef("password")], "Passwords must match")
+      .required("Password confirmation is required"),
+    termsOfUse: YupBool().oneOf([true], "Terms of Use are required"),
```
piyalbasu, line 52: this is the best reason to use Formik + Yup - makes validation rules v easy

Form/basics/index.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
```
piyalbasu, line 1: this file is all generic form elements - inputs, checkboxes, etc.

i thought about splitting this file up, but it didn't quite make sense yet. I think if we add like 1 or 2 more elements it will be time to do that. Until then, it's helpful to each element colocated with its style

src/popup/components/Form/basics/index.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
```
piyalbasu, line 1: this file is all generic form elements - inputs, checkboxes, etc.

i thought about splitting this file up, but it didn't quite make sense yet. I think if we add like 1 or 2 more elements it will be time to do that. Until then, it's helpful to each element colocated with its style

PR #21, opened 2020-06-15
quick fix for value handling on textarea in recoverAccount

piyalbasu:


src/popup/views/RecoverAccount/index.tsx
```diff
@@ -85,7 +85,12 @@ const RecoverAccount = () => {
           >
             <>
               <FormRow>
-                <FormTextField as="textarea" name="mnemonicPhrase" />
+                <FormTextField
+                  as="textarea"
+                  name="mnemonicPhrase"
+                  onChange={handleChange}
+                />
```
piyalbasu, line 92: I honestly have no idea why we need to manually pass `handleChange` when we do `as=""textarea"` on a styled <Field /> component. I spent a long time trying to figure it out, and gave up at the 1 hr mark. :/

vcarl: Weird!

PR #23, opened 2020-06-22
remove domain restriction by using postMessage for api

piyalbasu:


public/playground/js/index.js
```diff
@@ -8,7 +8,7 @@ document.querySelector("#lyra-connect").addEventListener("click", async () => {
   }
```
piyalbasu, line 8: just a little cleanup in this code as i was testing

src/api/helpers/index.ts
```diff
@@ -1,29 +1,37 @@
...
-  transactionStatus: string;
-  error: string;
-}
-
-export const sendMessageAndAwaitResponseExternal = (msg: {}): Promise<Response> =>
-  new Promise((resolve) => {
-    chrome.runtime.sendMessage(EXTENSION_ID, msg, (res: Response) =>
-      resolve(res),
-    );
+export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
```
piyalbasu, line 8: renamed these messages to be more indicative of what they are doing in the flow of the app

src/api/helpers/index.ts
```diff
@@ -1,29 +1,37 @@
...
+  window.postMessage(
+    { source: EXTERNAL_MSG_REQUEST, ...msg },
+    window.location.origin,
+  );
```
piyalbasu, line 12: the idea is for the client site to post a message to itself. content script is listening and takes the message and sends it on to Background

src/api/helpers/index.ts
```diff
@@ -1,29 +1,37 @@
...
+  return new Promise((resolve) => {
+    const messageListener = (event: { source: any; data: Response }) => {
+      // We only accept messages from ourselves
+      if (event.source !== window) return;
+      // Only respond to messages tagged as being from our content script
+      if (!event.data.source || event.data.source !== EXTERNAL_MSG_RESPONSE) {
+        return;
+      }
+      resolve(event.data);
+      window.removeEventListener("message", messageListener);
```
piyalbasu, line 22: Content Script takes Background's response, posts a message to itself. Since it shares the same window as Lyra API, it is able to listen (similar to above) and then resolve the API request with the data from Content Script

src/api/helpers/index.ts
piyalbasu: https://developer.chrome.com/extensions/messaging

src/background/index.ts
```diff
@@ -1,23 +1,15 @@
-import internalMessageListener from "./messageListener/internalMessageListener";
-import externalMessageListener from "./messageListener/externalMessageListener";
+import { popupMessageListener } from "./messageListener/popupMessageListener";
+import { lyraApiMessageListener } from "./messageListener/lyraApiMessageListener";
```
piyalbasu, line 2: renaming these to be a bit more clear

src/contentScript/helpers/index.ts
```diff
@@ -0,0 +1,34 @@
...
+        return;
+      // Forward the message on to Background
+      const res = await sendMessageToBackground(event.data);
+      // Send the response back to Lyra API
+      window.postMessage(
+        { source: EXTERNAL_MSG_RESPONSE, ...res },
+        window.location.origin,
+      );
+    },
+    false,
```
piyalbasu, line 32: this is the second part above the postMessage chain: it just listens for messages, then redirects them to Background. The response is posted to the shared window

PR #24, opened 2020-06-24
Added UI for <SignTransaction/>; removed memo for now; added BigNumber library

jeesunikim:
Ticket: [UI - Preview and sign a transaction](https://app.asana.com/0/1168666457741233/1168291088234574)

- Added `bignumber.js` library
- I had to `yarn build` in order to have the new styles applied even though I had no changes to `background` and `content script`. I wonder if it's because although this page is located in `/popup`, it can only viewed through `playground`. I am going to look at it why in more depth tomorrow morning.

Preview:
signintransaction-screenshot

### DIFF
--- a/src/popup/basics/index.tsx
+++ b/src/popup/basics/index.tsx
@@ -55,6 +55,7 @@ interface BackButtonProps {
 }
 
 const BackButtonEl = styled.button`
+  cursor: pointer;
   display: flex;
   align-items: center;
   background: ${COLOR_PALETTE.inputBackground};
--- a/src/popup/basics/index.tsx
+++ b/src/popup/basics/index.tsx
@@ -28,7 +28,7 @@ interface ButtonProps {
   children: React.ReactNode;
 }
 
-export const ButtonEl = styled.button<ButtonProps>`
+export const ButtonEl = styled(BasicButton)<ButtonProps>`
   width: ${(props) => (props.size === "small" ? "8.75rem" : "12.375rem")};
   display: ${(props) => (props.size === "small" ? "inline-block" : "block")};
   margin: 0 auto;
@@ -39,7 +39,6 @@ export const ButtonEl = styled.button<ButtonProps>`
   background: ${COLOR_PALETTE.primaryGradient};
   color: ${COLOR_PALETTE.white};
   border: none;
-  cursor: pointer;
   -webkit-appearance: none;
 `;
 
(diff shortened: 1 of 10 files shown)


src/popup/basics/index.tsx
```diff
@@ -47,6 +49,35 @@ export const Button = ({ size, children, onClick, ...props }: ButtonProps) => (
   </ButtonEl>
 );
 
+/* Back Button */
+interface BackButtonProps {
+  onClick: () => void;
+}
+
+const BackButtonEl = styled.button`
```
piyalbasu, line 57: should this extend `BasicButton` so we get the cursor, focus, etc. styles?

src/popup/views/SignTransaction/index.tsx
```diff
@@ -1,44 +1,96 @@
 import React from "react";
-import buffer from "buffer";
 import { useLocation } from "react-router-dom";
-import { get } from "lodash";
 import { useSelector } from "react-redux";
+import BigNumber from "bignumber.js";
```
piyalbasu, line 4: 👍

src/popup/basics/index.tsx
jeesunikim: Will add it to `ButtonEl` as well! Thanks!

src/popup/views/SignTransaction/index.tsx
```diff
@@ -51,58 +103,155 @@ const SignTransaction = () => {
...
+          <div>{truncatedPublicKey(publicKey)}</div>
+        </li>
+        {_fee ? (
+          <li>
+            <div>
+              <strong>Base fee</strong>
+            </div>
+            <div> {_fee}</div>
+          </li>
+        ) : null}
```
piyalbasu, line 244: since a lot fo these are constructed as
```
<li>
<div><strong>{title}</strong></div>
<div>{value}</div>
</li>
```

i wonder if it's worth creating a separate component for these 🤔

piyalbasu:
> * I had to `yarn build` in order to have the new styles applied even though I had no changes to `background` and `content script`. I wonder if it's because although this page is located in `/popup`, it can only viewed through `playground`. I am going to look at it why in more depth tomorrow morning.

Yup, this is correct - the SignTransaction/GrantAccess views are opened by the extension, so it opens whichever version of the view is packed inside of the extension. If you're looking at it in localhost, you'll see your changes live, but when the extension is programmatically opening these views, it will look at whichever version is packed inside of it in Chrome. So, you'll need to `yarn build` and refresh your version of the extension so it has access to your up to date code

src/popup/views/SignTransaction/index.tsx
jeesunikim: Updated

PR #29, opened 2020-07-05
open /recover-account in new window & fix border-box css errors in onboarding

piyalbasu:
open /recover-account in new window:
https://app.asana.com/0/1168666457741233/1181709640492793

Also fix some css problems with Onboarding. Setting * to box-sizing: border-box threw off styling for /recover-account, plus a couple other views

### DIFF
--- a/src/popup/views/UnlockAccount/index.tsx
+++ b/src/popup/views/UnlockAccount/index.tsx
@@ -38,7 +38,7 @@ const HeaderEl = styled.h1`
   margin: 0;
   margin-left: 1rem;
 `;
-const ImportButton = styled(BasicButton)`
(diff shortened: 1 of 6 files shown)


src/popup/views/UnlockAccount/index.tsx
```diff
@@ -39,7 +38,7 @@ const HeaderEl = styled.h1`
   margin: 0;
   margin-left: 1rem;
 `;
-const StyledLinkEl = styled(Link)`
+const ImportButton = styled(BasicButton)`
```
vcarl, line 41: Nit: `ImportButtonEl` since this isn't an exported style

PR #30, opened 2020-07-05
make submit async to auto reset isSubmitting state

piyalbasu:
https://app.asana.com/0/1168666457741233/1181709640492803

Reset /recover-account form to allow user to try again when an user gets an API error on submit


src/popup/views/Loading/index.tsx
```diff
@@ -12,7 +10,6 @@ const LoadingEl = styled.div`
 
 export const Loading = () => (
   <LoadingEl>
-    <FullscreenStyle />
```
piyalbasu, line 15: sorry @jeesunikim ! looks like this breaks the logic that opens a new window for <Welcome />
https://github.com/stellar/lyra/blob/master/src/popup/Router.tsx#L100

your PR had some good fixes for <Loading> that we can keep tho!

src/popup/views/RecoverAccount/index.tsx
```diff
@@ -54,9 +54,9 @@ const RecoverAccount = () => {
 
   const dispatch = useDispatch();
 
-  const handleSubmit = (values: FormValues) => {
+  const handleSubmit = async (values: FormValues) => {
     const { password, mnemonicPhrase } = values;
-    dispatch(recoverAccount({ password, mnemonicPhrase }));
+    await dispatch(recoverAccount({ password, mnemonicPhrase }));
```
piyalbasu, line 59: making `handleSubmit` async automatically resets the Form state after the function resolves

src/background/messageListener/lyraApiMessageListener.ts
```diff
@@ -3,12 +3,13 @@ import StellarSdk from "stellar-sdk";
...
 import { EXTERNAL_SERVICE_TYPES } from "statics";
 import { removeQueryParam } from "helpers";
+import { POPUP_WIDTH } from "popup/constants";
 import { Sender, SendResponseInterface } from "../types";
 import { responseQueue, transactionQueue } from "./popupMessageListener";
 import { store, publicKeySelector } from "../ducks/session";
 
 const WHITELIST_ID = "whitelist";
-const WINDOW_DIMENSIONS = "width=456,height=667";
+const WINDOW_DIMENSIONS = `width=${POPUP_WIDTH},height=667`;
```
piyalbasu, line 12: a little unrelated, but when i was revisiting how to switch between popup and new tab i noticed this was hardcoded when it should be a const

PR #31, opened 2020-07-06
use history.push for hash routing

piyalbasu:
https://app.asana.com/0/1168666457741233/1181709640492803

Fix pressing "Back"on the recovery screen show "Your file was not found" error


src/popup/views/RecoverAccount/index.tsx
```diff
@@ -81,7 +82,7 @@ const RecoverAccount = () => {
           <Onboarding
             header="Recover wallet from backup phrase"
             icon={icon}
-            goBack={() => window.location.replace("/")}
+            goBack={() => history.push({ pathname: "/" })}
```
piyalbasu, line 85: history.push is the proper way to nav using react-router

PR #32, opened 2020-07-06
only calculate BigNumber values if we have them

piyalbasu:


src/popup/views/SignTransaction/index.tsx
```diff
@@ -149,8 +149,6 @@ export const SignTransaction = () => {
         }: TransactionInfoResponse,
         i: number,
       ) => {
-        const formattedAmount = new BigNumber(amount).toFormat(2);
-        const formattedBuyAmount = new BigNumber(buyAmount).toFormat(2);
         const operationIndex = i + 1;
```
piyalbasu, line 152: on a non payment transaction (set options, for ex) we were hitting errors bc amount/buyAmount were undefined

PR #42, opened 2020-07-17
Flatten folder structure, fix styleguide issues

vcarl:
Overall this should hopefully be a little bit simpler to navigate and understand (bummer to wipe out the git blame by moving files tho)

This PR:
* Splits the background/ store creation from the sole duck,
* Changes structure so no imports are of _just_ `"constants"` or `"helpers"`
* Adds `El` suffix to all un-exported styled components
* Removes all default exports
* Changes lodash imports to be for specific functions
* Moves exports so no top-level folder imports from within another top-level folder

I left `api/*/index.js` because I'm not immediately clear what better names for them should be, and they're not imported by anything else anyway

Before/After of the file trees, mostly collapsed:
Screen Shot 2020-07-17 at 10 13 02 PM


vcarl:
Actually I realized there's some loose ends I left

src/popup/basics/Forms.tsx
```diff
@@ -0,0 +1,155 @@
...
+  background: ${COLOR_PALETTE.inputBackground};
+  border: 0.125rem solid ${COLOR_PALETTE.inputBackground};
+  border-radius: 0.625rem;
+  color: ${COLOR_PALETTE.primary};
+  cursor: pointer;
+  display: flex;
+  height: 2rem;
+  justify-content: space-evenly;
+  width: 2rem;
+
```
piyalbasu, line 137: Looks like you're reimplementing checkboxes, but this doesn't match the design anymore. Let's leave this PR as just styleguide changes and reimplement checkboxes separately
Screen Shot 2020-07-17 at 9 09 16 PM

vcarl: Oops! That was an exploration but I meant to back it out

PR #49, opened 2020-07-27
monorepo setup

piyalbasu:
in setting up the client facing API, i restructured the repo as a monorepo for 2 main reasons:

1) the API needed to be a separate npm package from the extension,. but they have some shared deps
2) i wanted to include docs here as they're fairly simple, but it is similar to the API in that it is its own project with its own build tools. Also, it needs to import the API, but it'd be better for it import a symlinked local version rather than the production version.

yarn workspaces were pretty handy as they're built into yarn, and our use case is pretty simple


@lyra/api/external.ts
```diff
@@ -1,15 +1,15 @@
-import { EXTERNAL_SERVICE_TYPES } from "constants/services";
-import { sendMessageToContentScript } from "api/helpers";
-import { ExternalRequest } from "api/types";
+import { EXTERNAL_SERVICE_TYPES } from "@lyra/constants/services";
```
piyalbasu, line 1: `@lyra` is a folder for shared resources that the extension/API both need

@lyra/api/external.ts
```diff
@@ -1,15 +1,15 @@
...
+import { sendMessageToContentScript } from "./helpers";
+import { ExternalRequest } from "./types";
 
-export const requestAccess = async (): Promise<{
+export const requestPublicKey = async (): Promise<{
```
piyalbasu, line 5: making this name more expressive of what it actually is

@stellar/lyra-api/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "svg-url-loader": "^5.0.0",
+    "ts-loader": "^6.2.2",
+    "tsconfig-paths-webpack-plugin": "^3.2.0",
+    "webpack": "4.41.5",
+    "webpack-cli": "3.3.10",
+    "webpack-dev-server": "3.10.3",
+    "webpack-merge": "4.2.2"
+  },
+  "scripts": {
+    "build": "../../node_modules/webpack/bin/webpack.js --config webpack.config.js"
```
piyalbasu, line 64: this is kind of an interesting (annoying) quirk of yarn workspaces. Because webpack is a shared dep, it's hoisted to the root, but unlike typescript, yarn scripts aren't smart enough to look into parent folders for the node module it needs

@stellar/lyra-api/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
piyalbasu, line 1: lyra-api is for public consumption, so we should compile w babel to make sure it works for everyone

docs/.docusaurus/client-modules.js
```diff
@@ -0,0 +1,5 @@
+export default [
```
piyalbasu, line 1: you can ignore all the `docs` files, this is all just docusaurus boilerplate. I included it here just to confirm everything works. The next pr will clean up this folder and put in the docs content i wrote

@stellar/lyra-api/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "@stellar/lyra-api",
+  "main": "build/index.min.js",
+  "types": "build/@stellar/lyra-sdk/src/index.d.ts",
```
piyalbasu, line 4: ick - i gotta fix this. wrestling a little with how tsconfig generates type declarations

extension/public/playground/js/index.js
```diff
@@ -1,8 +1,24 @@
+import { getPublicKey, signTransaction, isConnected } from "@stellar/lyra-api";
```
piyalbasu, line 1: the playground is going to be moved to docs in a subsequent PR

@lyra/api/external.ts
piyalbasu: note about `@lyra/constants` and `@lyra/statics`: 

the folder structure dumps all the files plus the package.json/tsconfig.json into the root. I spent a lot of time trying different approaches to this, but settled on this bc:

- i wanted to import files directly by path, rather than having one `main` file to import from 
```
import { services } from '@lyra/constants'; 
const {EXTERNAL_SERVICE_TYPES} = services;
// gross
```
- related to the above, I didn't want to have to compile each workspace every time a change was made. I just wanted this "npm package" to be treated as a regular folder and have the parent application compile it
- I wanted to import using a clean path like `@lyra/constants/services` rather than `@lyra/constants/src/services`

Conditional exports (https://nodejs.org/api/esm.html#esm_conditional_exports) would solve the issue of having to dump everything in root, but I could not get it to work for some reason 🤔 Will continue to look into it

docs/README.md
```diff
@@ -0,0 +1,33 @@
+# Website
```
piyalbasu, line 1: everything in the `docs/` folder is just docusaurus boilerplate - you can ignore. I just included it here to make sur everything works.

I'll be swapping in the content I wrote in the next PR

extension/src/contentScript/helpers/setIsConnected.ts
```diff
@@ -1,7 +1,6 @@
-export const embedLyraApi = () => {
-  const url = chrome.extension.getURL("lyraApi.min.js");
+export const setIsConnected = () => {
   const script = document.createElement("script");
-  script.src = url;
+  script.textContent = "window.lyra = true;";
```
piyalbasu, line 3: instead of putting the whole API on window, creating a flag just to let the client app know that Lyra is available

piyalbasu: They will check this flag by calling lyraApi.isConnected()

piyalbasu:
@vcarl ah good call - i did this all at once to make sure everything would work how i wanted it to. i can split up now

PR #50, opened 2020-07-28
Add lyra api

piyalbasu:


@lyra/api/external.ts
```diff
@@ -2,7 +2,7 @@ import { EXTERNAL_SERVICE_TYPES } from "@lyra/constants/services";
 import { sendMessageToContentScript } from "./helpers";
 import { ExternalRequest } from "./types";
 
-export const requestAccess = async (): Promise<{
+export const requestPublicKey = async (): Promise<{
```
piyalbasu, line 5: changing the name of this to be more descriptive of what it is

extension/public/playground/js/index.js
```diff
@@ -1,8 +1,24 @@
+import { getPublicKey, signTransaction, isConnected } from "@stellar/lyra-api";
```
piyalbasu, line 1: updating the playground to make sure it works - but this will be moving to docs in the next PR

extension/src/contentScript/helpers/setIsConnected.ts
```diff
@@ -1,7 +1,6 @@
-export const embedLyraApi = () => {
-  const url = chrome.extension.getURL("lyraApi.min.js");
+export const setIsConnected = () => {
   const script = document.createElement("script");
-  script.src = url;
+  script.textContent = "window.lyra = true;";
```
piyalbasu, line 3: now instead of putting the whole api lib on window, we're simply setting a flag that will be checked by lyra-api

@stellar/lyra-api/package.json
```diff
@@ -0,0 +1,50 @@
+{
```
piyalbasu, line 1: `lyra-api` is the sub-repo that contains an npm package that exports all the helper functions needed by a client

@stellar/lyra-api/package.json
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version"
+    ]
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@lyra/api": "1.0.0",
+    "typescript": "~3.7.2"
```
vcarl, line 26: Is TS needed as a dep for production? At first glance it seems like it'd be a devdep

@stellar/lyra-api/package.json
```diff
@@ -0,0 +1,50 @@
...
+      "last 3 firefox version"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version"
+    ]
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@lyra/api": "1.0.0",
```
vcarl, line 25: I think we might need to find a different way to consume this? Since this "package" comes from the extension, I'm not able to install this module from a packed tarball. It fails because `@lyra/api` doesn't exist on npm

piyalbasu: This should be fixed now!

@stellar/lyra-api/webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+const config = {
+  mode: "production",
+  entry: {
+    index: path.resolve(__dirname, "./src/index.ts"),
+  },
+  output: {
+    library: "lyra-api",
+    libraryTarget: "umd",
+    path: BUILD_PATH,
+    filename: "[name].min.js",
```
piyalbasu, line 15: this compiles the code in this subrepo with only the pertinent parts of `@lyra/api`. Comes in at a whopping 5kb 🎉

vcarl: I've been trying to test installing this package elsewhere, and i'm not 100% confident that it works atm. When I run `yarn pack`, the produced tarball still appears to want to install `@lyra/api` from npm but I'm not sure if there's a step I'm missing or something.

piyalbasu: @vcarl okay! I think i finally solved this. explanation in README

@stellar/lyra-api/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## `@lyraAlias` import
+
+Because this package needs to be packaged and published, it cannot have any
+dependencides that are not on the npm registry. Therefore, instead of using Yarn
+to import the `@lyra` workspace via symlinked dependency like in other repos,
+all references to `@lyra` must be relative paths.
+
+tsconfig.json rewrites `@lyraAlias` to a relative path to `@lyra` folders to
+prevent devs from manually having to write out a relative path.
```
piyalbasu, line 16: so this is kind of a blindspot with Yarn workspaces, unfortunately. They really do expect you to publish everything, which I don't want to do.

piyalbasu: I'd like to eventually use the symlinked dep during development, and at "prepack", swap out the symlinked dep for the relative path.

A couple of npm packages exist that sorta do this, but not to a level to my liking

piyalbasu: also open to renaming `@lyraAlias` to something else. Just didn't want to call it `@lyra` as it would lead someone to think it was a symlinked dep like in other repos

vcarl: Cool! I'll test it out tomorrow, that sounds like it hits all the issues I noticed though. Alias seems like a good solution.

PR #51, opened 2020-07-29
Add initial metrics implementation 

vcarl:
Needs some followup work to add events and some other stuff, but this is the functional metrics system itself


extension/src/helpers/metrics.ts
```diff
@@ -0,0 +1,108 @@
...
+ * of registered handlers and passes the current state and action. These are
+ * intended for metrics emission, nothing else.
+ */
+export function metricsMiddleware<State>(): Middleware<{}, State> {
+  return ({ getState }) => (next) => (action: AnyAction) => {
+    const state = getState();
+    // If the `?.` fails, this ends up as `null` which appears to make eslint
+    // grumpy
+    // eslint-disable-next-line no-unused-expressions
+    handlersLookup[action.type]?.forEach((handler) => handler(state, action));
```
piyalbasu, line 20: Might be better to do:
```
(handlersLookup[action.type] || []).forEach((handler) => handler(state, action));
```
not sure if that will fix the error

OR if you wanna get fancy: https://github.com/tc39/proposal-nullish-coalescing
But i think nullish-coalescing might require a babel plugin for older versions of chrome <80 (and maybe Brave?)

extension/src/popup/metrics/authServices.ts
```diff
@@ -0,0 +1,10 @@
+import { METRIC_NAMES } from "popup/constants/metricsNames";
```
piyalbasu, line 1: this file need to be imported somewhere to actually run this code and register the handler right?

extension/src/helpers/metrics.ts
vcarl: Since Chrome is evergreen I'm not too worried about older versions, good point tho. This seems like a corner case in the `no-unused-expressions` rule so I didn't feel bad about disabling it, but I'll swap to `|| []`

extension/src/popup/metrics/authServices.ts
vcarl: Ah good point. I meant this to be a complete example so I'll make sure this is used

piyalbasu: cool - just making sure i understood what the implementation would look like

extension/src/helpers/metrics.ts
piyalbasu: oh, looks like you haven't added this yet. Nbd, i approved. You can add when u have a sec

extension/src/popup/views/Account.tsx
```diff
@@ -19,6 +19,8 @@ import CopyColor from "popup/assets/copy-color.png";
 import StellarLogo from "popup/assets/stellar-logo.png";
 import { Footer } from "popup/components/Footer";
 
+import "popup/metrics/authServices";
```
vcarl, line 22: Couple points I wanted to highlight here:

In the past we've tied metrics logically with reducers, but that always felt strange to me. Tying them to actions makes the API for authoring the metrics emission a little plainer to me, but it means there isn't really anything to export--which I think is fine, but maybe unusual feeling. Since they don't have any impact on functionality, I think it makes sense for them to be something used invisibly.

This _could_ be imported in `App.tsx`, but I deliberately put it with a feature component instead. With code splitting (not a huge concern in an extension but something I wanted to consider for the pattern generally), `import`s can have dramatic impact on how much code is seen as "common" between modules. I find you have to be very deliberate about constructing your dependency tree to get maximum impact. We're not code splitting redux here so this is theoretical, but since this module pulls in the reducer, importing it in `App` would immediately mean all of our app state logic would have to live in the main bundle. For a medium-sized web app, that multiplies per feature and could be a relatively substantial bundle savings.

PR #52, opened 2020-07-29
remove 'server.submitTransaction' and return .toXDR()

jeesunikim:
[Ticket: Remove transaction submission from `Sign Transaction` method](https://app.asana.com/0/1168666457741233/1185300929163057)

* Assuming we will have to update `/playground`? What we have right now does not make sense. 1) Add XDR then return the signed transaction XDR. 2) Update the button's text from submit transaction to x

### DIFF
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -3,7 +3,6 @@ import StellarSdk from "stellar-sdk";
 // @ts-ignore
 import { fromMnemonic, generateMnemonic } from "stellar-hd-wallet";
 
-import { SERVER_URL } from "@lyra/constants/stellar";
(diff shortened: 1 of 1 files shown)


piyalbasu:
> [Ticket: Remove transaction submission from `Sign Transaction` method](https://app.asana.com/0/1168666457741233/1185300929163057)
> 
> * Assuming we will have to update `/playground`? What we have right now does not make sense. 1) Add XDR then return the signed transaction XDR. 2) Update the button's text from submit transaction to x

Yup - that's right. I actually didn't bother making a ticket for updating Playground as I'm moving it to docs right now. I'll just update it there as I'm doing the transition

jeesunikim:
> Oh actually - can you remove SERVER_URL from @lyra/constants ? I believe this is not used anywhere else

`getAccountBalance` in `@lyra/api/internal.ts` is using `SERVER_URL`.

PR #55, opened 2020-08-04
build lyra-api automatically on `yarn start` and `yarn build`

piyalbasu:
lyra-api needs to be built for docs to be able to import it


@stellar/lyra-api/webpack.config.js
```diff
@@ -10,6 +10,7 @@ const config = {
     index: path.resolve(__dirname, "./src/index.ts"),
   },
   output: {
+    globalObject: "this",
```
piyalbasu, line 13: this is needed for docusaurus to SSR lyra-api. Otherwise, it throws a `window` is not defined error

extension/package.json
```diff
@@ -18,7 +18,6 @@
     "@lyra/api": "1.0.0",
     "@lyra/constants": "1.0.0",
     "@reduxjs/toolkit": "^1.3.5",
-    "@stellar/lyra-api": "1.0.0-alpha",
```
piyalbasu, line 21: no longer needed in extension as playground has moved to docs

docs/docusaurus.config.js
```diff
@@ -3,15 +3,15 @@ module.exports = {
   url: "https://lyradocs.netlify.com",
   baseUrl: "/",
   onBrokenLinks: "throw",
-  favicon: "img/favicon.ico",
+  favicon: "./images/logo.png",
```
piyalbasu, line 6: these placeholder images needed to be moved into a folder called `static`, which apparently is a ⭐magic ⭐ folder docusaurus needs all static assets to be in

PR #59, opened 2020-08-06
lyra-api export default and package version bump

piyalbasu:


@shared/api/types.ts
```diff
@@ -21,7 +21,10 @@ export interface Response {
   url: string;
 }
 
-export interface ExternalRequest {
+export interface LyraApiRequest {
```
piyalbasu, line 24: weirdly i hadn't hit this TS error before, but just started getting it - so fixed!

@stellar/lyra-api/src/index.ts
```diff
@@ -3,3 +3,4 @@ import { signTransaction } from "./signTransaction";
 import { isConnected } from "./isConnected";
 
 export { getPublicKey, signTransaction, isConnected };
+export default { getPublicKey, signTransaction, isConnected };
```
piyalbasu, line 6: started testing on `lab` and realized only using named exports was a little annoying. importing a function called `isConnected` in a big app is weird bc it's not descriptive as to what it is and could lead to naming collisions.

In lab, what i really wanted is to be able to call `lyraApi.isConnected()`. My thought is to take the React route and export as both default and as named

vcarl: Nice, that makes sense to me

PR #61, opened 2020-08-07
update signTransaction to return { signedTransaction: string } + docs

piyalbasu:


extension/src/popup/ducks/access.ts
```diff
@@ -4,7 +4,7 @@ import {
   rejectAccess as internalRejectAccess,
   grantAccess as internalGrantAccess,
   signTransaction as internalSignTransaction,
-} from "@lyra/api/internal";
+} from "@shared/api/internal";
```
piyalbasu, line 8: whoops - caught this while making this change

@shared/api/external.ts
```diff
@@ -20,10 +20,10 @@ export const requestPublicKey = async (): Promise<{
 export const submitTransaction = async ({
   transactionXdr,
 }: LyraApiRequest): Promise<{
-  transactionStatus: string;
+  signedTransaction: string;
```
piyalbasu, line 23: since we're returning an actual transaction (instead of the response from Horizon like we were originally) let's change what this key is called

extension/src/popup/ducks/access.ts
vcarl: Possibly not based off latest master? Should have been fixed in https://github.com/stellar/lyra/pull/60/commits/dba296eb6b4db547a4a0e0b948c5b8bde1d2f74f

vcarl: Oh that was only merged like 2 hours ago, so yeah make sense

piyalbasu: oh maybe i gotta merge

PR #64, opened 2020-08-11
remove unnecessary permissions from manifest

piyalbasu:


extension/public/static/manifest.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "Lyra",
-  "version": "1.0",
-  "description": "Build an Extension!",
-  "permissions": ["activeTab", "declarativeContent", "storage"],
```
piyalbasu, line 5: `activeTab` and `declarativeContent` permissions aren't actually needed over anymore since i refactored the messaging system. 

The Chrome store makes you justify any permissions you use when you submit, so removing!

extension/public/static/manifest.json
```diff
@@ -1,8 +1,9 @@
...
+  "version": "1.0.0",
+  "version_name": "alpha.6",
```
piyalbasu, line 4: also need to use their style of versioning to be accepted

extension/public/static/manifest.json
```diff
@@ -30,6 +31,5 @@
     "48": "images/get_started48.png",
     "128": "images/get_started128.png"
   },
-  "key": "ncjenbmichjdlnepfekhijpmfinfhgdm",
```
piyalbasu, line 33: this is also a relic of a past messaging system

extension/public/background.ts
```diff
@@ -1,20 +1,5 @@
...
-          new chrome.declarativeContent.PageStateMatcher({
-            css: ["div"],
-          }),
-        ],
-        actions: [new chrome.declarativeContent.ShowPageAction()],
-      },
-    ]);
-  });
-});
-
```
piyalbasu, line 17: this is some extension boilerplate - we weren't using this, but we will need this `onInstalled` hook for this ticket: https://app.asana.com/0/1168666457741233/1187237204678441

PR #66, opened 2020-08-12
need to use browser_action to *always* open the popup

piyalbasu:


extension/public/static/manifest.json
```diff
@@ -16,7 +16,7 @@
     }
   ],
   "web_accessible_resources": ["lyraApi.min.js"],
-  "page_action": {
+  "browser_action": {
```
piyalbasu, line 19: this keeps the Lyra extension icon enabled on every page, unconditionally

piyalbasu: page_action was looking for a condition to load the popup, but I removed the condition yesterday. whoops!

PR #67, opened 2020-08-13
feature/emit-metrics-event 

jeesunikim:
Ticket: [Emit metrics that answer usage questions](https://app.asana.com/0/1168666457741233/1187261817032481)
[Metrics Spreadsheet](https://docs.google.com/spreadsheets/d/1lwSC0bkQ5bB3oovSIpAW3eWtZwL_cmbce-gLAp_Kj1Q/edit#gid=0)
- Added some comments for Bruno and I to go through together in the spreadsheet
- Would like to keep naming consistent
-- (`new wallet` vs. `create-password`)
-- (`share public` key vs. `grant-access`)
- Ones that need some coding update is `sign transaction` (row 27) with properties of `domain, # of operations, and operation type`

### DIFF
--- a/extension/src/popup/ducks/authServices.ts
+++ b/extension/src/popup/ducks/authServices.ts
@@ -18,20 +18,24 @@ interface ErrorMessage {
   errorMessage: string;
 }
 
-export const createAccount = createAsyncThunk(
-  "auth/createAccount",
-  async (password: string) => {
-    let res;
+export const createAccount = createAsyncThunk<
+  { publicKey: string },
+  string,
+  { rejectValue: ErrorMessage }
+>("auth/createAccount", async (password, thunkApi) => {
+  let res = { publicKey: "" };
 
-    try {
-      res = await createAccountService(password);
-    } catch (e) {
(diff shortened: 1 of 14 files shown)


extension/src/popup/ducks/authServices.ts
```diff
@@ -18,20 +18,23 @@ interface ErrorMessage {
...
-    return res;
-  },
-);
+  try {
+    res = await createAccountService(password);
+  } catch (e) {
+    console.error("Failed when creating an account: ", e.message);
+    return thunkApi.rejectWithValue({
+      errorMessage: e.message,
+    });
```
vcarl, line 34: 👍  We might want to reject with the entire error, instead of just the message?

jeesunikim: interface for ErrorMessage is `string`. Should I change it to object?

vcarl: I don't think this needs a change right now, but we might want to if we start feeling a need for more error details.

PR #68, opened 2020-08-13
Set up jest, add demo test

vcarl:


extension/package.json
```diff
@@ -59,15 +30,75 @@
...
+      "ts",
+      "tsx"
+    ],
+    "moduleDirectories": [
+      "node_modules",
+      "<rootDir>/src",
+      "<rootDir>/.."
+    ]
+  },
+  "dependencies": {
```
vcarl, line 83: I merged `devDependencies` and `dependencies`. Since this is an app there isn't really a situation where the division matters

extension/config/jest/setupTests.ts
```diff
@@ -0,0 +1,52 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import { JSDOM } from "jsdom";
+import fetch from "isomorphic-unfetch";
+
+// make a JSDOM thing so we can fuck with mount
+const jsdom = new JSDOM("<!doctype html><html><body></body></html>");
+const { window } = jsdom;
```
vcarl, line 10: This file is mostly copied from an old project (converted to TS). Nice to have on hand so we don't need to re-solve these environment problems, stuff like global mocks and overrides

extension/src/helpers/__tests__/stellar.test.ts
```diff
@@ -0,0 +1,11 @@
+import { truncatedPublicKey } from "../stellar";
+
+describe("truncatedPublicKey", () => {
+  it("truncates keys", () => {
+    expect(
+      truncatedPublicKey(
+        "GAJVUHQV535IYW25XBTWTCUXNHLQN4F2PGIPOOX4DDKL2UPNXUHWU7B3",
+      ),
+    ).toBe("GAJV…U7B3");
```
vcarl, line 9: I wanted the demo test to be a react testing library one, but for some reason that's causing the build to fail with linter errors?? Unclear but didn't want to block the PR on that

extension/src/helpers/stellar.ts
```diff
@@ -1,2 +1,2 @@
 export const truncatedPublicKey = (publicKey: string) =>
-  `${publicKey.slice(0, 4)}...${publicKey.slice(-4)}`;
+  `${publicKey.slice(0, 4)}…${publicKey.slice(-4)}`;
```
vcarl, line 2: Little stylistic thing here. It's better to have `“ ” ‘ ’ …` instead of `" " ' ' ...`

extension/package.json
```diff
@@ -59,15 +30,75 @@
...
+    "testURL": "http://localhost",
+    "transform": {
+      "^.+\\.(js|jsx|ts|tsx|mjs)$": [
+        "babel-jest"
+      ]
+    },
+    "moduleNameMapper": {
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
+      "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js"
+    },
```
vcarl, line 67: Transforms code with babel to strip types, and swaps out any style or static file imports with stubs

docs/package.json
```diff
@@ -6,7 +6,7 @@
     "buildDependencies": "yarn workspace @stellar/lyra-api build",
     "docusaurus": "docusaurus",
     "start": "yarn buildDependencies && docusaurus start",
-    "build": "yarn buildDependencies && docusaurus build",
+    "build": "docusaurus build",
```
piyalbasu, line 9: docusaurus can't build unless lyra-api has been built. try removing @stellar/lyra-api/build and run this. it'll error

extension/config/jest/setupTests.ts
piyalbasu: it'd be nice to either make this a separate npm package that we maintain or use an existing package that does this (esp the storage stuff https://www.npmjs.com/package/jest-localstorage-mock). Would rather not have to maintain all this setup stuff

docs/package.json
vcarl: The netlify build tho runs `yarn workspace @stellar/lyra-api build && yarn workspace docs build` so it will always already have lyra-api recently built

extension/config/jest/setupTests.ts
vcarl: Ah fair point. I'll see if there are good package for these mocks, that one looks solid. I'm wary of stuff like that being too heavy or complex

vcarl: Updated @piyalbasu!

docs/package.json
piyalbasu: yeah, i'd say running this locally for testing purposes should still work without knowing to have to build lyra-api though. 

the Netlify command should actually just be changed to `yarn workspace docs build`

vcarl: Chatted about it, I've got another branch in progress, gonna hash out build command changes there

PR #69, opened 2020-08-18
Add project-wide build/start commands

vcarl:
This has a couple of similar changes:

* Reorganize `package.json` files for consistency
* Use Apache 2.0 everywhere
* Add `yarn build` and `yarn start` as commands at the root of the project
  * `build` will produce production artifacts for the extension, docs, and `@stellar/lyra-api`
  * `start` runs file watchers for everything we might need to check while making changes. See the readme for more

There's still advantages to running individual components on their own, cuz if you're not touching docs then the watcher isn't helping you. I was finding that I'd forget to rebuild one part of it or another when trying to make modifications though, so this will at least make the default command a little more tolerant to user error.

### DIFF
--- a/@shared/constants/config.js
+++ b/@shared/constants/config.js
@@ -1,0 +1,20 @@
+const DEFAULT_STATS = {
+  // minimal
+  // can use `preset: "minimal"` once webpack 5 lands
+  all: false,
+  modules: true,
+  maxModules: 0,
+  errors: true,
+  warnings: true,
+  // our additional options
+  moduleTrace: true,
+  errorDetails: true,
+  assets: true,
+  excludeAssets: [/\.d\.ts/, /\.png/, /\.jpe?g/],
+  hash: true,
+  timings: true,
+};
+
+module.exports = {
+  DEFAULT_STATS,
+};
--- a/@shared/constants/config.js
+++ b/@shared/constants/config.js
@@ -6,7 +6,8 @@ const DEFAULT_STATS = {
   maxModules: 0,
   errors: true,
   warnings: true,
-  // our additional options
+  // We want to keep output narrowly focused when possible because there's a lot
+  // of different watchers started at once.
   moduleTrace: true,
   errorDetails: true,
   assets: true,
--- a/@stellar/lyra-api/package.json
+++ b/@stellar/lyra-api/package.json
@@ -11,7 +11,8 @@
   "main": "build/index.min.js",
   "scripts": {
     "prepare": "yarn build",
-    "build": "webpack --config webpack.config.js"
+    "build": "webpack --config webpack.config.js",
+    "start": "webpack --config webpack.config.js --watch"
   },
   "types": "build/@stellar/lyra-api/src/index.d.ts",
   "prettier": "@stellar/prettier-config",
--- a/@stellar/lyra-api/package.json
+++ b/@stellar/lyra-api/package.json
@@ -11,8 +11,8 @@
   "main": "build/index.min.js",
   "scripts": {
     "prepare": "yarn build",
-    "build": "webpack --config webpack.config.js",
-    "start": "webpack --config webpack.config.js --watch"
+    "build": "webpack --config webpack.config.js --mode production",
+    "start": "webpack --config webpack.config.js --watch --mode development"
   },
   "types": "build/@stellar/lyra-api/src/index.d.ts",
   "prettier": "@stellar/prettier-config",
--- a/@stellar/lyra-api/package.json
+++ b/@stellar/lyra-api/package.json
@@ -1,15 +1,20 @@
 {
   "name": "@stellar/lyra-api",
-  "main": "build/index.min.js",
-  "types": "build/@stellar/lyra-api/src/index.d.ts",
-  "prettier": "@stellar/prettier-config",
   "version": "1.0.0-alpha.2",
+  "license": "Apache-2.0",
+  "author": "Stellar Development Foundation <hello@stellar.org>",
   "description": "Utility functions to interact with Lyra extension",
   "repository": {
     "url": "git@github.com:stellar/lyra.git",
     "type": "git"
   },
-  "author": "Stellar Development Foundation <hello@stellar.org>",
+  "main": "build/index.min.js",
+  "scripts": {
+    "prepare": "yarn build",
+    "build": "webpack --config webpack.config.js"
+  },
+  "types": "build/@stellar/lyra-api/src/index.d.ts",
+  "prettier": "@stellar/prettier-config",
   "browserslist": {
     "production": [
       "last 3 chrome version",
@@ -20,7 +25,6 @@
       "last 1 firefox version"
     ]
   },
-  "license": "MIT",
   "lint-staged": {
     "src/**/*.ts?(x)": [
       "eslint --fix --max-warnings 0"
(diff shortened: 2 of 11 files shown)


vcarl:
I broke Jest's Babel setup, whoops

@stellar/lyra-api/package.json
```diff
@@ -1,31 +1,21 @@
...
-  "lint-staged": {
-    "src/**/*.ts?(x)": [
-      "eslint --fix --max-warnings 0"
-    ]
+  "main": "build/index.min.js",
+  "scripts": {
+    "prepare": "yarn build",
+    "build": "webpack --config webpack.config.js --mode production",
+    "start": "webpack --config webpack.config.js --watch --mode development",
+    "test": "jest `pwd`/src"
```
vcarl, line 16: This is set up such that there's only 1 jest "instance" for all workspaces, so by default `jest` will run all tests in all projects. This line scopes the test run to only tests under the specific workspace's directory.

@stellar/lyra-api/src/__tests__/index.test.js
```diff
@@ -0,0 +1,9 @@
+import LyraAPI from "../index";
+
+describe("lyra API", () => {
+  it("has keys", () => {
```
vcarl, line 4: Added this as a proof of concept that the configuration is working. Strikes me as a good idea to have Jest set up so it's easy to write tests against the public API for the module

@stellar/lyra-api/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "declarationDir": "build/",
     "outDir": "build/",
     "paths": {
-      "@sharedAlias/*": ["../../../@shared/*"]
+      "@shared/*": ["../../../@shared/*"]
```
vcarl, line 10: This supports Jest, which has a "modulesDirectories" config option so keeping the name in common with the folder simplifies config. It's quite a pain that our different tools need separate module resolution help

jest.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
vcarl, line 1: This is a little annoying. Jest doesn't seem to like being set up to run multiple sub-folders' tests, so there's some inconsistency. Each project has its own Jest dependency but relies on this shared config, and since the `rootDir` is the root of the monorepo, running `jest` in a workspace will run all tests for all workspaces.

There doesn't appear to be a way to solve that in config without duplicating a ton of configuration (anywhere `<rootDir>` is used), so each workspace is manually scoping its test run to only that workspace's files

package.json
```diff
@@ -1,10 +1,57 @@
...
     "@shared/api",
     "@shared/constants",
     "docs",
     "extension",
     "@stellar/lyra-api"
-  ]
+  ],
+  "scripts": {
+    "build": "run-p --print-label build:lyra-api build:docs build:extension",
+    "build:netlify": "run-s build:lyra-api build:docs",
```
vcarl, line 13: 👀

package.json
```diff
@@ -1,10 +1,57 @@
...
+    "start:docs": "yarn workspace docs start",
+    "start:extension": "yarn workspace extension start",
+    "start:unpacked": "yarn workspace extension start:unpacked-extension",
+    "install-if-package-changed": "git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD | grep --quiet yarn.lock && yarn install || exit 0",
+    "test:ci": "run-p test:ci:lyra-api test:ci:extension",
+    "test:ci:lyra-api": "yarn workspace @stellar/lyra-api test -ci",
+    "test:ci:extension": "yarn workspace extension test -ci",
+    "test": "run-p --print-label test:lyra-api test:extension",
+    "test:lyra-api": "yarn workspace @stellar/lyra-api test -o --watch",
+    "test:extension": "yarn workspace extension test -o --watch"
```
vcarl, line 28: I set it up so that `yarn test` runs a watcher that only runs tests for files that have changed since the last commit, and `yarn test:ci` runs all tests assuming a CI environment

package.json
```diff
@@ -1,10 +1,57 @@
...
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "concurrently 'pretty-quick --staged' 'lint-staged'",
+      "post-merge": "yarn install-if-package-changed",
+      "post-checkout": "yarn install-if-package-changed"
```
vcarl, line 34: This is a new addition, should cover another situation where deps can change

@shared/constants/config.js
```diff
@@ -0,0 +1,21 @@
+const DEFAULT_STATS = {
```
piyalbasu, line 1: nice! 👍

@stellar/lyra-api/tsconfig.json
piyalbasu: gotcha - i had originally done this to signal that this in this subrepo, @shared isn't sym linked - but i also documented this in the README, so if it makes things a little easier, go for it

piyalbasu: oh, actually - can you update the README in this repo?

package.json
```diff
@@ -1,10 +1,57 @@
...
+  ],
+  "scripts": {
+    "build": "run-p --print-label build:lyra-api build:docs build:extension",
+    "build:netlify": "run-s build:lyra-api build:docs",
+    "build:lyra-api": "yarn workspace @stellar/lyra-api build",
+    "build:docs": "yarn workspace docs build",
+    "build:extension": "yarn workspace extension build",
+    "start": "run-p --print-label start:lyra-api start:docs start:extension start:unpacked",
+    "start:lyra-api": "yarn workspace @stellar/lyra-api start",
+    "start:docs": "yarn workspace docs start",
```
piyalbasu, line 19: so I think we still have the problem of if I clone this repo fresh, and I just want to `start` docs, running this command will fail unless I've already built lyra-api

vcarl: I did a test clone, and it appears to finish building fast enough that it doesn't break. I definitely think it's a race condition that happens to work out, though

package.json
piyalbasu: nice

package.json
piyalbasu: if you just run `yarn start:docs`? shouldn't you need to do: `run-s build:lyra-api yarn workspace docs start` similar to the netlify build?

vcarl: It's working when I try it from scratch 🤷 
```sh
$ git clone ~/workspace/lyra/
$ cd lyra/
$ yarn
$ yarn start
```
goes off without a hitch locally

```

vcarl: Ah but you said specifically `yarn start:docs`. That does break, but I thought the happy path was good enough 🤷 I think it's relatively infrequent we'd only start docs

piyalbasu: yeah, it may be infrequent, but I think the start commands should work "out of the box" if we're including them

piyalbasu: oh i see - bc `yarn start` aliases these, you don't want to build every time

vcarl: Right exactly. Could work around that by having `docs:start` run the `@stellar/lyra-api` build first with like, `docs:start:quick` that skips that or something, but since it's infrequent and because `"scripts"` is getting pretty busy, I thought I'd avoid yet another script

jeesunikim:
I deleted all the `node_modules`, re-installed by `yarn`, started `yarn start` and they worked excellent!

@stellar/lyra-api/tsconfig.json
vcarl: Pushed @piyalbasu!

PR #72, opened 2020-08-20
Add integration test for mnemonic confirmation

vcarl:
This came out quite well! By clicking buttons and asserting metrics were called and that content loads, I feel pretty confident that this works whenever the test passes, even if we swap out the implementation.

### DIFF
--- a/extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
+++ b/extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
@@ -1,15 +1,19 @@
 import React, { useState } from "react";
+import { Redirect } from "react-router-dom";
 import styled from "styled-components";
 import { Formik } from "formik";
 import { useSelector, useDispatch } from "react-redux";
 
+import { APPLICATION_STATE } from "@shared/constants/applicationState";
 import {
   confirmMnemonicPhrase,
   authErrorSelector,
+  applicationStateSelector,
 } from "popup/ducks/authServices";
 
 import CloseIcon from "popup/assets/icon-close.svg";
 
+import { ROUTES } from "popup/constants/routes";
 import { COLOR_PALETTE } from "popup/constants/styles";
 import { Button } from "popup/basics/Buttons";
 import {
@@ -82,6 +86,7 @@ export const ConfirmMnemonicPhrase = ({
   );
   const [selectedWords, setSelectedWords] = useState<string[]>([]);
   const authError = useSelector(authErrorSelector);
+  const applicationState = useSelector(applicationStateSelector);
 
   const updatePhrase = (target: HTMLInputElement) => {
     if (target.checked) {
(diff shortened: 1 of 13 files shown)


extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
```diff
@@ -129,6 +121,10 @@ export const ConfirmMnemonicPhrase = ({
   const displaySelectedWords = () =>
     selectedWords.map((word) => convertToWord(word)).join(" ");
 
+  if (applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED) {
+    return <Redirect to={ROUTES.mnemonicPhraseConfirmed} />;
+  }
```
vcarl, line 126: This redirect was done with `history.push` in an action, but it was a little easier to test the outcome if I made it declarative like this.

extension/src/popup/constants/history.ts
```diff
@@ -0,0 +1,3 @@
+import { createHashHistory } from "history";
+
+export const history = createHashHistory();
```
vcarl, line 3: Had to peel a couple of functions out into new files for mocking purposes. This call had been in `App.tsx` so it was pulling in a ton of other stuff I didn't want to have to mock

extension/src/popup/views/__tests__/MnemonicPhrase.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    "@shared/constants/applicationState",
+  );
+  return {
+    confirmMnemonicPhrase: () =>
+      Promise.resolve({
+        isCorrectPhrase: true,
+        applicationState: APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED,
+      }),
+  };
+});
```
vcarl, line 58: Some of these mocks we might want to pull out into more reusable modules, or into a `__mocks__` folder, as we write more tests

extension/src/popup/views/__tests__/MnemonicPhrase.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+        await waitFor(() => screen.getByText(MNEMONIC));
+        // Click each word in the mnemonic
+        MNEMONIC.split(" ").forEach((word) =>
+          fireEvent.click(screen.getByLabelText(word)),
+        );
+        fireEvent.click(screen.getByTestId("confirm"));
+      });
+      await waitFor(() =>
+        screen.getByText(`redirect ${ROUTES.mnemonicPhraseConfirmed}`),
+      );
```
vcarl, line 111: Really like how the basic flow test came out. I think it's pretty readable and durable

extension/src/popup/views/__tests__/MnemonicPhrase.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    it("works", async () => {
+      render(
+        <Wrapper state={{ auth: { error: null, applicationState: "butts" } }}>
+          <MnemonicPhrase />
+        </Wrapper>,
+      );
+
+      const { emitMetric } = jest.requireMock("helpers/metrics");
+
+      // Reveal mnemonic
```
piyalbasu, line 81: I think we should break these down into subtests:
```
it('reveals mnemonic', () => {

};
```
so jest reports these as individual tests and so you can use `fit` and `xit` to run only specific tests when debugging

vcarl: I'm not sure we can split these up like that, tho I agree that'd be an improvement. They need to be called in sequence, the intermediate states between button clicks aren't saved anywhere

PR #79, opened 2020-08-24
Open onboarding on install

vcarl:


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "@stellar/tsconfig",
   "compilerOptions": {
     "importHelpers": true,
-    "target": "es5",
+    "target": "ES2019",
```
vcarl, line 5: Given our runtime (modern Chrome, Firefox on the horizon) we can target a much more recent version of JS without needing to transpile down. A big change here is that compiled output will natively use async/await, which should make debugging easier

extension/public/background.ts
```diff
@@ -1,11 +1,27 @@
-import { initMessageListener } from "../src/background";
+import { initMessageListener } from "background";
+import { browser } from "webextension-polyfill-ts";
```
vcarl, line 2: This appears to be the best way to use the webextension polyfill with TS, instead of as a global. Having this imported should make Firefox support basically a find/replace `chrome.` -> `browser.`

piyalbasu: nice!

PR #80, opened 2020-08-25
added url in sign transaction view and added metrics for sign transaction's options

jeesunikim:
Ticket: [In Sign Transaction confirmation modal, show requesting site's url](https://app.asana.com/0/1168666457741233/1185789836547227)

- UI updated, following this [model](https://projects.invisionapp.com/prototype/Lyra-Lite-ck60vjr6j003ky301z14bfmta/play/781cb375). Design isn't asking for page title so I removed it.
Screen Shot 2020-08-27 at 10 00 03 AM

- Added metrics for `loaded: sign transaction` and `loaded: grant access`.
metric_example

- [Punycode](https://app.asana.com/0/1167950234569163/1168342146125315) functionality to verify the domain url added.

### DIFF
--- a/extension/src/helpers/stellar.ts
+++ b/extension/src/helpers/stellar.ts
@@ -1,2 +1,22 @@
 export const truncatedPublicKey = (publicKey: string) =>
   `${publicKey.slice(0, 4)}…${publicKey.slice(-4)}`;
+
+export const getTransactionInfo = (search: string) => {
+  const decodedTransactionInfo = atob(search.replace("?", ""));
+  const transactionInfo = decodedTransactionInfo
+    ? JSON.parse(decodedTransactionInfo)
+    : {};
+
+  const {
+    tab: { title, url },
+    transaction,
+  } = transactionInfo;
+
+  const u = new URL(url);
+
+  return {
+    title,
+    transaction,
+    domain: u.origin,
+  };
+};
--- a/extension/src/helpers/stellar.ts
+++ b/extension/src/helpers/stellar.ts
@@ -13,9 +13,15 @@ export const getTransactionInfo = (search: string) => {
   } = transactionInfo;
 
   const u = new URL(url);
+  const { _operations } = transaction;
+  const operationTypes = _operations.map(
+    (operation: { type: string }) => operation.type,
+  );
 
   return {
     transaction,
     domain: u.hostname,
+    operations: _operations,
+    operationTypes,
   };
 };
--- a/extension/src/helpers/stellar.ts
+++ b/extension/src/helpers/stellar.ts
@@ -8,15 +8,14 @@ export const getTransactionInfo = (search: string) => {
     : {};
 
   const {
-    tab: { title, url },
+    tab: { url },
     transaction,
   } = transactionInfo;
 
   const u = new URL(url);
 
   return {
-    title,
     transaction,
-    domain: u.origin,
+    domain: u.hostname,
   };
 };
--- a/extension/src/popup/metrics/views.ts
+++ b/extension/src/popup/metrics/views.ts
@@ -33,7 +33,7 @@ registerHandler<AppState>(navigate, (_, a) => {
 
   // "/sign-transaction" and "/grant-access" require additionak metrics on loaded page
   if (search) {
-    const { transaction, domain } = getTransactionInfo(search);
+    const { domain } = getTransactionInfo(search);
     const METRIC_OPTION_DOMAIN = {
       domain,
     };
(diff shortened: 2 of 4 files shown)


extension/src/popup/views/SignTransaction.tsx
```diff
@@ -87,13 +90,26 @@ export const SignTransaction = () => {
...
+    (operation: { type: string }) => operation.type,
+  );
+
+  const METRIC_OPTIONS = {
+    domain: url,
+    number_of_operations: _operations.length,
+    operationTypes,
+  };
+
+  useEffect(() => {
```
jeesunikim, line 110: I'm not happy with how I implemented it so suggestions welcome.

vcarl: Is this needed? "View" metrics are being fired by `popup/metrics/views.ts`

piyalbasu: ooh is this needed bc you need to pass the metric a `body`?

piyalbasu: looks like this will double report `METRIC_NAMES.viewSignTransaction` since it will happen on page load from `popup/metrics/views` and then again on `useEffect`.

extension/src/popup/views/SignTransaction.tsx
```diff
@@ -229,7 +245,9 @@ export const SignTransaction = () => {
     <El>
       <BackButton onClick={() => window.location.replace("/")} />
       <HeaderEl>Confirm Transaction</HeaderEl>
-      <SubheaderEl>{title} is requesting a transaction</SubheaderEl>
+      <SubheaderEl>
+        {title} from {url} is requesting a transaction
```
piyalbasu, line 249: Would def like to hear what others think, but if we think the url origin is enough here we could do
```
const u = new URL (url); u.origin;
```
 That should hopefully support all domain suffixes

extension/src/popup/views/SignTransaction.tsx
jeesunikim: It is needed to fire the options. I can remove the one from `popup/metrics/views.ts`. It's just not pretty 😢

jeesunikim: I'm going to try a different approach.

jeesunikim: I thought of creating another slice for metrics but it isn't necessary since it isn't being used elsewhere other than for metrics. I also don't want to make an exception case by telling it to skip `/sign-transaction` in `popup/metrics/views.ts` 🤔

extension/src/helpers/stellar.ts
```diff
@@ -1,2 +1,22 @@
 export const truncatedPublicKey = (publicKey: string) =>
   `${publicKey.slice(0, 4)}…${publicKey.slice(-4)}`;
+
+export const getTransactionInfo = (search: string) => {
```
jeesunikim, line 4: Since this is being re-used, added in `stellar.ts` to re-use.

extension/src/popup/metrics/views.ts
```diff
@@ -22,10 +23,40 @@ const routeToEventName = {
...
+  const { pathname, search } = action.payload.location;
+
   const eventName = routeToEventName[pathname];
+
   if (!eventName) {
     throw new Error(`Didn't find a metric event name for path '${pathname}'`);
   }
-  emitMetric(eventName);
+
+  // "/sign-transaction" and "/grant-access" require additionak metrics on loaded page
```
piyalbasu, line 34: NIT: "additional"

extension/src/popup/metrics/views.ts
```diff
@@ -22,10 +23,40 @@ const routeToEventName = {
...
+      domain,
+    };
+
+    if (pathname === ROUTES.grantAccess) {
+      emitMetric(eventName, METRIC_OPTION_DOMAIN);
+    }
+
+    if (pathname === ROUTES.signTransaction) {
+      const { _operations } = transaction;
+      const operationTypes = transaction._operations.map(
```
piyalbasu, line 47: hm, maybe we can move this operation type mapping into `getTransactionInfo`. Might be nicer to do all the data manipulation there so all this file has to do is emit the proper metric

extension/src/popup/metrics/views.ts
vcarl:
```suggestion
  // "/sign-transaction" and "/grant-access" require additional metrics on loaded page
```

PR #81, opened 2020-08-25
return response as a string and throw error if needed

piyalbasu:
https://app.asana.com/0/1168666457741233/1188320836740117

`background` returns an object with the string you requested plus an error (if applicable)

Here, we're breaking that object down to return only what the user wants: either a string representing the data they asked for a thrown error indicating the user declined access

### DIFF
--- a/docs/docs/guide/usingLyra.md
+++ b/docs/docs/guide/usingLyra.md
@@ -51,18 +51,23 @@ if (isConnected()) {
 }
 
 const retrievePublicKey = async () => {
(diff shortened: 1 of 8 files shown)


docs/docs/guide/usingLyra.md
```diff
@@ -51,18 +51,23 @@ if (isConnected()) {
 }
```
piyalbasu, line 51: docs updated because lyra-api returns strings for these functions now instead of an obj

PR #82, opened 2020-08-27
/grant-access to display punycode url; separated the decoded search param code

jeesunikim:
grant-access

- `/grant-access` just needs url from its search params, not via transaction.
- Added punycode for `/grant-access` as well
- removed `{title}`

### DIFF
--- a/extension/src/popup/views/GrantAccess.tsx
+++ b/extension/src/popup/views/GrantAccess.tsx
@@ -4,7 +4,7 @@ import styled from "styled-components";
 import { useLocation } from "react-router-dom";
 import punycode from "punycode";
 
-import { parsedSearchParam } from "helpers/urls";
+import { parsedSearchParam, getUrlHostname } from "helpers/urls";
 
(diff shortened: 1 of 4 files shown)


extension/src/popup/views/GrantAccess.tsx
```diff
@@ -43,9 +46,9 @@ const RejectButtonEl = styled(Button)`
 export const GrantAccess = () => {
   const location = useLocation();
   const dispatch = useDispatch();
-  const decodedTab = atob(location.search.replace("?", ""));
-  const tabToUnlock = decodedTab ? JSON.parse(decodedTab) : {};
-  const { url, title } = tabToUnlock;
+  const { url } = parsedSearchParam(location.search);
+  const u = new URL(url);
+  const punycodedDomain = punycode.toASCII(u.hostname);
```
piyalbasu, line 51: hm, we have to convert to `new URL` and then grab this `hostname` a few times. Maybe we should create a helper that accepts a url string and returns the hostname

jeesunikim: Updated!

PR #84, opened 2020-08-31
make lyraApi.signTransaction accept string as param

piyalbasu:
https://app.asana.com/0/1168666457741233/1188069176357563

This is an improvement Iveta found while she was integrating with AV. This method just needs an xdr string. Instead of passing { transactionXdr: string; }, let's just pass it a string to make things simpler

### DIFF
--- a/@shared/api/external.ts
+++ b/@shared/api/external.ts
@@ -1,6 +1,5 @@
 import { EXTERNAL_SERVICE_TYPES } from "../constants/services";
 import { sendMessageToContentScript } from "./helpers";
-import { LyraApiRequest } from "./types";
 
 export const requestPublicKey = async (): Promise<string> => {
   let response = { publicKey: "", error: "" };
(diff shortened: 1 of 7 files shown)


docs/docs/guide/usingLyra.md
```diff
@@ -37,7 +37,7 @@ if (isConnected()) {
 
 ### getPublicKey
 
-#### `getPublicKey() -> <Promise<{ publicKey: string; error?: string; }>>`
+#### `getPublicKey() -> <Promise<string>>`
```
piyalbasu, line 40: this actually isn't related - just noticed this out of date content

@shared/api/external.ts
```diff
@@ -20,9 +19,9 @@ export const requestPublicKey = async (): Promise<string> => {
   return publicKey;
 };
 
-export const submitTransaction = async ({
-  transactionXdr,
-}: LyraApiRequest): Promise<string> => {
+export const submitTransaction = async (
```
vcarl, line 22: Might be nice to rename this function too?

piyalbasu: I purposely didn't call this `signTransaction` because I didn't want 2 functions with that name in the codebase. 

Also, the functions in this file are just passing off lyra-api requests to Background and mutating the response. The names are supposed to signify they're "attempting" the action called by lyra-api

PR #86, opened 2020-09-02
Set the extension's height to 600px which is Chrome extension's max height value

jeesunikim:
Ticket: [Extension popup shouldn't scroll left/right](https://app.asana.com/0/1168666457741233/1189755955263475)
- I couldn't find the case where the extension was scrolling left to right. @vcarl, do you remember which stage this was happening?
- I added its extension height to be chrome extension's max height (600px)


vcarl:
I don't remember, I think it was on all screens so seems like it's fixed!

PR #87, opened 2020-09-02
lyraApi.signTransaction accepts a string, but the Background listener still expects an object

piyalbasu:


extension/src/background/messageListener/lyraApiMessageListener.ts
```diff
@@ -64,7 +64,7 @@ export const lyraApiMessageListener = (
   };
 
   const submitTransaction = () => {
-    const transactionXdr = request;
+    const { transactionXdr } = request;
```
piyalbasu, line 67: by the time it gets here, the transactionXDR string has been added to an object

vcarl (approved):
Preview build works 👍

PR #88, opened 2020-09-03
add EXPERIMENTAL env variable for building the extension

piyalbasu:
https://app.asana.com/0/1168666457741233/1191515311689209


extension/package.json
```diff
@@ -5,7 +5,9 @@
   "prettier": "@stellar/prettier-config",
   "scripts": {
     "build": "webpack --config webpack.prod.js",
+    "build:experimental": "yarn build --env.EXPERIMENTAL=true",
     "start": "webpack-dev-server --config webpack.dev.js",
+    "start:experimental": "yarn start --env.EXPERIMENTAL=true",
```
piyalbasu, line 10: adding a flag called EXPERIMENTAL as this can be a long lived flag that lives beyond v2

extension/src/constants/featureFlag.ts
```diff
@@ -0,0 +1,3 @@
+declare const EXPERIMENTAL: string;
+const _EXPERIMENTAL = EXPERIMENTAL;
+export { _EXPERIMENTAL as EXPERIMENTAL };
```
piyalbasu, line 3: now you can import EXPERIMENTAL as a constant and show/hide features based off its value

extension/webpack.common.js
```diff
@@ -2,12 +2,13 @@ const CopyWebpackPlugin = require("copy-webpack-plugin");
...
 const TsconfigPathsPlugin = require("tsconfig-paths-webpack-plugin");
 const path = require("path");
+const webpack = require("webpack");
 
 const { DEFAULT_STATS } = require("../@shared/constants/config");
 
 const BUILD_PATH = path.resolve(__dirname, "./build");
 
-const commonConfig = {
+const commonConfig = (env = { EXPERIMENTAL: false }) => ({
```
piyalbasu, line 11: default to false if no flag passed

package.json
```diff
@@ -11,14 +11,16 @@
   "scripts": {
     "build": "run-p --print-label build:lyra-api build:docs build:extension",
     "build:netlify": "run-s build:lyra-api build:docs",
-    "build:extension-preview": "run-s build:lyra-api build:extension",
+    "build:extension-preview": "run-s build:lyra-api build:extension:experimental",
```
piyalbasu, line 14: let's *try* having all previews build with the experimental flag. We will likely find it necessary as we start reviewing new feature PR's

extension/package.json
vcarl: Hm maybe dev mode should always be experimental? I'm wary of stuff that adds in a decision point on what command to run to test changes. Production + experimental production makes sense, but dev mode seems like it should always have track the bleeding edge

extension/src/constants/featureFlag.ts
vcarl: My experience using env like this is that it means the code can't be 100% stripped from output when it's irrelevant, cuz dead code elimination doesn't trace imports for constant values. `DefinePlugin` overwrites the exact string variable name with the exact string output, if it's then used as an export then that value is still interpreted at runtime

I think that's probably fine, but something to be aware of

extension/package.json
piyalbasu: Ya I actually had it default to true in a previous version for the same reason, but my thought is if this is going to be an open source project and we want people to be able to jump in and mess around easily, the default behavior should be to mirror what people are seeing in the production version. That always helps me get my bearings in a new codebase

extension/src/constants/featureFlag.ts
piyalbasu: Interesting - this only being done to deal with some annoying TS issue. This can be changed if we run into issues

vcarl (approved):
Can keep an eye out for any issues, this looks good otherwise though!

PR #89, opened 2020-09-09
grant access > notify users to agree to provide their public key to a third party service for the first time

jeesunikim:
Ticket: [Notify users to agree to provide their public key to a third party service for the first time](https://app.asana.com/0/1168666457741233/1193011684416067)

We do not have design for this yet. It's based on the [wireframe](https://www.figma.com/file/uIoeq72R2XjiWjGyVGiPG6/Lyra-wireframes?node-id=410%3A142) that Bruno provided. **I think we should gather the components/pages that are only based on wireframes and not design**. There are quiet few I think.

**Asking users to agree to provide their public key to a third party service that they visited for the first time**
firsttime-domain

**Once the domain is whitelisted**
whitelisted-domain

### DIFF
--- a/extension/src/popup/views/GrantAccess.tsx
+++ b/extension/src/popup/views/GrantAccess.tsx
@@ -1,10 +1,14 @@
 import React, { useState } from "react";
 import { useDispatch } from "react-redux";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import { useLocation } from "react-router-dom";
 import punycode from "punycode";
 
-import { parsedSearchParam, getUrlHostname } from "helpers/urls";
+import {
+  removeQueryParam,
+  parsedSearchParam,
+  getUrlHostname,
+} from "helpers/urls";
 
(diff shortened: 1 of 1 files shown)


extension/src/popup/views/GrantAccess.tsx
```diff
@@ -66,21 +100,38 @@ export const GrantAccess = () => {
     <GrantAccessEl>
       <HeaderEl>Connection request</HeaderEl>
       <SubheaderEl>{punycodedDomain} wants to know your public key</SubheaderEl>
+      {!isDomainWhiteListed ? (
+        <WarningMessageEl>
+          <p>
+            <strong>
+              This is the first time you interact with this domain in this
+              computer.
```
vcarl, line 108: Grammar on this is funky, maybe
```suggestion
              This is the first time you‘ve interacted with this domain on this
              computer.
```

vcarl (approved):

PR #90, opened 2020-09-11
feature/reveal-backup-phrase > reveal backup phrase view added

jeesunikim:
Ticket: [Reveal backup phrase while logged in](https://app.asana.com/0/1168666457741233/1168342146125265)
- Implemented based on the wireframe (Design TBD). I made it to match `Sign Transaction` view since it also got a back button
- Just an FYI, I looked at to see if Metamask does allow users to see their backup phrase. It doesn't seem like they allow that.

**Big Changes Summary:**
- Components `<DisplayBackupPhrase/>` and `<UnlockBackupPhrase/>`
- added a `showBackupPhrase` in `api/internal.ts` and `popupMessageListener.ts` function to pass an error message if there is one when incorrect password is submitted
- `popupMessageListener.ts`: `extra.mnemonicPhrase` to be accessible when confirming Password so we can display it
- Added routes for `<DisplayBackupPhrase/>` and `<UnlockBackupPhrase/>`.
- Metrics added. Updated [Metrics Sheet](https://docs.google.com/spreadsheets/d/1lwSC0bkQ5bB3oovSIpAW3eWtZwL_cmbce-gLAp_Kj1Q/edit#gid=0) as well.

Design is TBD. But just an FYI that with a warning message, the view is longer than chrome extension allows so there will be a scroll.

**[Screenshot] Password Prompt Before Showing Backup Phrase**
show-backup-phrase

**[Screenshot] Backup Phrase Display**
show-backup-phrase-display


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -187,11 +198,14 @@ export const popupMessageListener = (
...
       console.error(e);
     }
+    let extra = { mnemonicPhrase: "" };
     let publicKey = "";
     let privateKey = "";
+
     if (keyStore) {
-      ({ privateKey, publicKey } = keyStore);
-      store.dispatch(logIn({ publicKey }));
+      ({ privateKey, publicKey, extra } = keyStore);
```
jeesunikim, line 206: Allowing `extra.mnemonicPhrase` to be accessible. Otherwise, it's just a blank.

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -176,6 +176,17 @@ export const popupMessageListener = (
...
   };
 
+  const showBackupPhrase = async () => {
+    const { password } = request;
+
+    try {
+      await keyManager.loadKey(localStorage.getItem(KEY_ID) || "", password);
+      sendResponse({});
+    } catch (e) {
+      sendResponse({ error: "Incorrect Password" });
```
jeesunikim, line 186: Sending error message if a user types in a wrong password

extension/src/popup/views/UnlockBackupPhrase.tsx
```diff
@@ -0,0 +1,122 @@
...
+  interface FormValues {
+    password: string;
+  }
+  const initialValues: FormValues = {
+    password: "",
+  };
+
+  const handleSubmit = async (values: FormValues) => {
+    const { password } = values;
+    const res = await showBackupPhrase(password);
```
jeesunikim, line 71: I decided to call this internal api within a component itself since on `fulfilled`, we aren't changing any state. All we need is an error message from `popupMessageListener.ts`

extension/src/popup/views/DisplayBackupPhrase.tsx
```diff
@@ -0,0 +1,137 @@
...
+export const DisplayBackupPhrase = () => {
+  const mnemonicPhrase = useMnemonicPhrase();
+  const [isCopied, setIsCopied] = useState(false);
+  const [isBlurred, setIsBlurred] = useState(true);
+
+  return (
+    <UnlockAccountEl data-testid="display-mnemonic-phrase">
+      <HeaderContainerEl>
+        <HeaderEl>
+          <BackButton onClick={() => window.location.replace("/")} />
```
jeesunikim, line 77: Using `window.location.replace` to go back to main `account` page. I wasn't sure if it needed to go back to password prompt view.

piyalbasu: I like this going back to Account straightaway.

There is an interesting issue here, though. This works in localhost, but does not work when you build and run the extension. That's become the internal url structure for the extension is a little different.

In the extension, the internal url for `/` is actually `chrome-extension://lnkacgnnjdohikfldmbgdkmbancciicl/index.html#/`

If you replace this with `history.push({ pathname: "/" })` it will work in both localhost and the extension itself.
In fact, I think we should probably always use `history.push` instead of `window.location` just to be safe

piyalbasu: I just opened a PR that creates a helper method that uses `history.push` to nav. Why don't I merge that PR and then you can just use the helper method here 

https://github.com/stellar/lyra/pull/92/files

jeesunikim: Sounds good, @piyalbasu. Thank you!

jeesunikim: Updated, @piyalbasu

piyalbasu (approved):

PR #91, opened 2020-09-17
feature/qr-code > QR code implementation

jeesunikim:
**Summary:**
- Using `qrcode.react` library
- It no longer is a modal. It's just a page view now
- metrics added
- Added a todo for switching to either `testnet` or `mainnet` for stellar.expert link

qr-copy-to-your-clipboard


piyalbasu:
@jeesunikim we've had some similar issues w `<Header />` in the past. Maybe it makes more sense to just include it with every view that needs it rather than making it a global.

For ex, maybe in `Router.tsx` you do:

```
        <PublicKeyRoute path={ROUTES.account}>
          <Header />
          <Account />
        </PublicKeyRoute>
```

vcarl:
Naming nitpick: "QR" isn't super descriptive without context of what data the QR code is encoding, maybe "viewAddress" or something?

jeesunikim:
> Naming nitpick: "QR" isn't super descriptive without context of what data the QR code is encoding, maybe "viewAddress" or something?

for `metrics` or as a component name in general? I'd like to include `QR` in its name. `viewQrAddress`?

vcarl:
As a general name, to me that it's a QR code seems like an implementation detail — what the user intends to do is access their Stellar address

extension/src/popup/views/QR.tsx
```diff
@@ -0,0 +1,155 @@
...
+  div {
+    border: 2px solid ${COLOR_PALETTE.primary};
+  }
+`;
+
+export const QR = () => {
+  const publicKey = useSelector(publicKeySelector);
+  const [isCopied, setIsCopied] = useState(false);
+
+  /* TODO: SHOULD BE BASED ON NETWORK SWITCH (MAINNET or TESTNET) */
```
piyalbasu, line 99: 👍

jeesunikim:
> As a general name, to me that it's a QR code seems like an implementation detail — what the user intends to do is access their Stellar address

Gotcha! Sounds good to me. Does it matter that `publickey` interchangeably with `address`?

vcarl:
Consistency is good! I don't know which we've used more in the code off the top of my head

piyalbasu:
We mostly use `public key` in the codebase and that's what `stellar-hd-wallet` calls it. Prob best to stick with `public key`

piyalbasu (approved):

PR #92, opened 2020-09-17
Add navigateTo helper method to make sure we always use right nav method

piyalbasu:


extension/src/popup/helpers/navigateTo.ts
```diff
@@ -0,0 +1,3 @@
+import { history } from "popup/constants/history";
+
+export const navigateTo = (pathname: string) => history.push({ pathname });
```
piyalbasu, line 3: I really wanted this to accept pathname as type `ROUTES` but we have some instances where we nav using `${ROUTES}?${queryParam}` which ruined my well laid plans

Doesn't appear to be a way to create a type to handle both

morleyzhi: What if `navigateTo` took a second optional param `queryParams`?

morleyzhi (approved):

extension/src/popup/helpers/navigateTo.ts
vcarl: Or maybe `route | LocationObject`?

jeesunikim (approved):
merci!

PR #96, opened 2020-09-22
Pubnet for beta

piyalbasu:
https://app.asana.com/0/1168666457741233/1193842482204548

We can pass a `production` flag to `extension:build` to switch to using Pubnet.
This means we  won't automatically fund a new account using friendbot and we'll check pubnet for account balance


@shared/api/internal.ts
```diff
@@ -136,6 +136,7 @@ export const getAccountBalance = async (
     response = await server.loadAccount(publicKey);
   } catch (e) {
     console.error(e);
+    return { balance: "0" };
```
piyalbasu, line 139: this error doesn't appear to return a status code, so we can't explicitly check for 404 (which likely means it's a pubnet acct that hasn't been funded). Let's fail gracefully for now, but we can polish this experience a bit in the future

@shared/constants/stellar.ts
```diff
@@ -1 +1,9 @@
-export const SERVER_URL = "https://horizon-testnet.stellar.org";
+declare const PRODUCTION: boolean;
+
+export const isTestnet = !PRODUCTION;
```
piyalbasu, line 3: `isTestnet` because we're adding steps in the flow based on this flag

extension/package.json
```diff
@@ -4,8 +4,9 @@
   "license": "Apache-2.0",
   "prettier": "@stellar/prettier-config",
   "scripts": {
-    "build": "webpack --config webpack.prod.js",
+    "build": "webpack --config webpack.extension.js",
```
piyalbasu, line 7: changing this to `extension` as build types are no longer a dev/prod binary

@shared/api/internal.ts
jeesunikim: Can we add TODO?

jeesunikim (approved):

PR #97, opened 2020-09-22
Add a dev page, surface "extension isn't installed" errors

vcarl:
Originally I was going to add a "reset extension" button, but there isn't existing logic to do that and digging into adding one seemed like it was going to be a relatively large effort, so this is a low priority placeholder for now. Also added an error to the situation where the extension isn't installed, by adding a 1s timeout to attempts to communicate with it


@shared/api/internal.ts
```diff
@@ -24,26 +24,14 @@ export const createAccount = async (
...
-    });
-  } catch (e) {
-    console.error(e);
-  }
-  return response;
-};
+}> =>
+  sendMessageToBackground({
+    type: SERVICE_TYPES.LOAD_ACCOUNT,
+  });
```
vcarl, line 34: As I started modifying this function to allow the errors to propagate, I realized there's always going to be a response (because `sendMessageToBackground` will either resolve or reject one) so the default wasn't useful, and any `catch` was just going to re-throw the error it got, so it could be removed

extension/src/popup/Router.tsx
```diff
@@ -96,6 +108,10 @@ const PrivateKeyRoute = (props: RouteProps) => {
 const HomeRoute = () => {
   const applicationState = useSelector(applicationStateSelector);
   const publicKey = useSelector(publicKeySelector);
+  const error = useSelector(authErrorSelector);
+  if (applicationState === APPLICATION_STATE.APPLICATION_ERROR) {
+    return <AppError>{error}</AppError>;
+  }
```
vcarl, line 114: This feels like there's an opportunity for abstraction, but I didn't dive into it. 3x copy-pasted data loading + early returns

extension/src/popup/constants/metricsNames.ts
```diff
@@ -1,4 +1,5 @@
 export const METRIC_NAMES = {
+  viewDebug: "loaded screen: debug",
```
vcarl, line 2: This should never actually make it into amplitude, but special casing it in the route metrics logic seemed bad

extension/src/popup/views/Debug.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+// Remove when not needed! 👇
+// eslint-disable-next-line arrow-body-style
```
vcarl, line 4: I figure this component would get something added that would necessitate `{}` basically the next time it's touched, so why force whitespace changes

jeesunikim:
Confirmed on dev environment. Nice to see error page when the extension isn't installed 👍

jeesunikim (approved):

piyalbasu (approved):

PR #102, opened 2020-09-25
use dynamic network name for viewPublicKey. Also,  change casing of the route

piyalbasu:


extension/src/popup/constants/routes.ts
```diff
@@ -12,5 +12,5 @@ export enum ROUTES {
   recoverAccountSuccess = "/recover-account-success",
   unlockBackupPhrase = "/unlock-backup-phrase",
   displayBackupPhrase = "/display-backup-phrase",
-  viewPublicKey = "/viewPublicKey",
+  viewPublicKey = "/view-public-key",
```
piyalbasu, line 15: just making this route match the same casing as the others.

Also, fun fact, I just learned that this type of casing is often called kebab-casing 🍖

extension/src/popup/views/ViewPublicKey.tsx
```diff
@@ -141,7 +139,8 @@ export const ViewPublicKey = () => {
         <LinkButton
           onClick={() => {
             window.open(
-              `https://stellar.expert/explorer/${CURRENT_NETWORK}/account/${publicKey}`,
+              
+              `https://stellar.expert/explorer/${NETWORK_NAME.toLowerCase()}/account/${publicKey}`,
```
piyalbasu, line 143: We can now get this network dynamically

jeesunikim (approved):

PR #103, opened 2020-09-25
bugfix/unlock-account-to-default > `/unlock-account` should redirect to `/` on no account found

jeesunikim:
**Summary:**
- `/unlock-account` should redirect to `/` on no account found
- Created a `UnlockAccountRoute` that checks to see if the account exists or not based on `applicationState` per Piyal's comment _"We don't know if the user is missing their public key because a) it’s in the keystore in localstorage and it needs to be extracted or b) the account doesn’t exist at all."_
-- If an account doesn't exist, go to the `<Welcome />` page; otherwise, go to `<UnlockAccount/>`
- Successful Page for **Recovery Account** and **Account Creator** should only happen when `publicKey` exist so I replaced its normal `Route` with `PublicKeyRoute`

### DIFF
--- a/extension/src/popup/Router.tsx
+++ b/extension/src/popup/Router.tsx
@@ -105,10 +105,33 @@ const PrivateKeyRoute = (props: RouteProps) => {
   return <Route {...props} />;
 };
 
+/*
+We don't know if the user is missing their public key because
+a) it’s in the keystore in localstorage and it needs to be extracted or b) the account doesn’t exist at all.
+We are checking for applicationState here to find out if the account doesn’t exist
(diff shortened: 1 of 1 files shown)


extension/src/popup/Router.tsx
```diff
@@ -105,10 +105,33 @@ const PrivateKeyRoute = (props: RouteProps) => {
...
+      <Redirect
+        to={{
+          pathname: "/",
+        }}
+      />
+    );
+  }
+  return <Route {...props} />;
+};
+
```
piyalbasu, line 129: nice! I was thinking something along these lines, too!

piyalbasu (approved):

PR #106, opened 2020-09-29
we don't want to use yarn workspaces for docs. we want to install from registry

piyalbasu:


docs/docusaurus.config.js
```diff
@@ -28,7 +28,6 @@ module.exports = {
       {
         docs: {
           // It is recommended to set document id as docs home page (`docs/` path).
-          homePageId: "docs/gettingStarted",
```
piyalbasu, line 31: this just fixes a warning i was getting from docusaurus - it's preferred to use slug to denote the homepage

package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "build": "run-p --print-label build:lyra-api build:docs build:extension",
-    "build:netlify": "run-s build:lyra-api build:docs",
+    "build:netlify": "cd docs && npm i && npm run build",
```
piyalbasu, line 13: using npm because using yarn defaults to using workspaces, which we don't want to use for docs. just install everything from the registry for docs

jeesunikim (approved):

PR #107, opened 2020-09-30
feature/allow-list > setting domain to be whitelisted if confirmed on signed transaction and checking if is whitelisted

jeesunikim:
**Summary:**
- renamed `whitelist` to `allowList`
- checking to see if the domain the user is on is within `ALLOWLIST_ID` on localStorage.
- `<GrantAccess/>` to display a warning message at all time

### DIFF
--- a/extension/src/background/messageListener/lyraApiMessageListener.ts
+++ b/extension/src/background/messageListener/lyraApiMessageListener.ts
@@ -8,17 +8,18 @@ import {
 } from "background/types";
 
 import { EXTERNAL_SERVICE_TYPES } from "@shared/constants/services";
-import { POPUP_WIDTH } from "constants/dimensions";
 import { NETWORK } from "@shared/constants/stellar";
 
-import { removeQueryParam } from "helpers/urls";
+import { POPUP_WIDTH } from "constants/dimensions";
+import { ALLOWLIST_ID } from "constants/localStorageTypes";
+
+import { getUrlHostname, getPunycodedDomain } from "helpers/urls";
 
 import { store } from "background/store";
 import { publicKeySelector } from "background/ducks/session";
 
 import { responseQueue, transactionQueue } from "./popupMessageListener";
 
-const ALLOWLIST_ID = "allowlist";
 const WINDOW_DIMENSIONS = `width=${POPUP_WIDTH},height=667`;
 
 export const lyraApiMessageListener = (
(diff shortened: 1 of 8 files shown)


extension/src/background/messageListener/lyraApiMessageListener.ts
```diff
@@ -18,7 +18,7 @@ import { publicKeySelector } from "background/ducks/session";
 
 import { responseQueue, transactionQueue } from "./popupMessageListener";
 
-const WHITELIST_ID = "whitelist";
+const ALLOWLIST_ID = "allowlist";
```
piyalbasu, line 21: I noticed this issue the other day, we redeclare this const in both `lyraApiMessageListener` and `popupMessageListener`. Why don't we make a `background/constants` file and store this there

extension/src/background/messageListener/lyraApiMessageListener.ts
```diff
@@ -65,18 +65,29 @@ export const lyraApiMessageListener = (
...
+    const { transactionXdr } = request;
     const transaction = StellarSdk.TransactionBuilder.fromXDR(
       transactionXdr,
       StellarSdk.Networks[NETWORK],
     );
 
     const { tab } = sender;
+    const tabUrl = tab?.url ? tab.url : "";
+
+    const sanitizedUrl = removeQueryParam(tabUrl);
```
piyalbasu, line 79: I actually really liked what you did with `punycode` in `popup`. Maybe we should create a helper method that takes a url and runs it through `getUrlHostname` and `punycode.toASCII` to get just the punycode domain and store that in the allowlist

That would also solve the issue of a different path on the same domain triggering the warning message again

jeesunikim: Updated @piyalbasu

extension/src/helpers/urls.ts
```diff
@@ -1,10 +1,19 @@
...
   return decodedSearchParam ? JSON.parse(decodedSearchParam) : {};
 };
+
 export const getUrlHostname = (url: string) => {
   const u = new URL(url);
   return u.hostname;
 };
+
+export const getPunycodedDomain = (url: string) => {
+  return punycode.toASCII(url);
```
piyalbasu, line 18: oooh interesting Github Actions warning. Yeah, let's fix this error

extension/src/background/messageListener/lyraApiMessageListener.ts
```diff
@@ -65,18 +67,30 @@ export const lyraApiMessageListener = (
...
+    const tabUrl = tab?.url ? tab.url : "";
+    const domain = getUrlHostname(tabUrl);
+    const punycodedDomain = getPunycodedDomain(domain);
+
+    const allowListStr = localStorage.getItem(ALLOWLIST_ID) || "";
+    const allowList = allowListStr.split(",");
+
+    if (allowList.includes(punycodedDomain)) {
+      isDomainListedAllowed = true;
+    }
```
piyalbasu, line 88: super small, but it's a little less error prone to do:
```
isDomainListedAllowed = allowList.includes(punycodedDomain);
```

jeesunikim: Updated, @piyalbasu

piyalbasu (approved):

PR #109, opened 2020-10-02
feature/detailed-transaction > adding more details for sign transaction

jeesunikim:
**Summary:**
- Updated its frontend to match the new design by Bruno
- Still needs `MEMO_TEXT`. I created a ticket for this sprint

**Screenshot:**
confirm-transaction

### DIFF
--- a/extension/src/popup/views/SignTransaction.tsx
+++ b/extension/src/popup/views/SignTransaction.tsx
@@ -1,6 +1,8 @@
 import React, { useState } from "react";
+import buffer from "buffer";
 import { useLocation } from "react-router-dom";
 import { useSelector, useDispatch } from "react-redux";
+import { get } from "lodash";
 import styled from "styled-components";
 import BigNumber from "bignumber.js";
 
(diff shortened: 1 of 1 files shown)


extension/src/popup/views/SignTransaction.tsx
```diff
@@ -95,7 +123,9 @@ export const SignTransaction = () => {
   );
 
   const punycodedDomain = getPunycodedDomain(domain);
-  const { _fee, _operations } = transaction;
+  const { _fee, _operations, _memo, _sequence } = transaction;
```
piyalbasu, line 126: ah, so I thought I remembered solving the issue with Memo text coming in as a buffer and I tracked down the PR where we removed it: https://github.com/stellar/lyra/pull/24

Looks like we had a convo about it but i cannot for the life of me recall what this was about :D
Do you remember?

jeesunikim: I'm not 100% sure but it looks like the very first [design](https://projects.invisionapp.com/prototype/Lyra-Lite-ck60vjr6j003ky301z14bfmta/play/781cb375) had the memo but the revised design after that didn't. Then orbitlens suggested including MEMO on 08/19.

jeesunikim: Added MEMO! Thanks @piyalbasu

piyalbasu (approved):

PR #111, opened 2020-10-05
this only gave users 1000ms to handle lyra-api modals

piyalbasu:


@shared/api/helpers.ts
```diff
@@ -4,17 +4,24 @@ import {
...
   window.postMessage(
     { source: EXTERNAL_MSG_REQUEST, ...msg },
     window.location.origin,
   );
   return new Promise((resolve, reject) => {
-    const timeout = setTimeout(() => {
-      reject(new TimeoutError());
-    }, 1000);
+    if (!window.lyra) {
+      reject(new NoExtensionInstalledError());
```
piyalbasu, line 22: if we want to short circuit based on Lyra not being installed, we can simply check to see if Lyra's content script has been installed

vcarl (approved):

PR #112, opened 2020-10-06
Show users tips to avoid scams on account creation

piyalbasu:
Screen Shot 2020-10-06 at 10 16 18 PM

### DIFF
--- a/extension/src/popup/App.tsx
+++ b/extension/src/popup/App.tsx
@@ -23,10 +23,14 @@ const GlobalStyle = createGlobalStyle`
   }
 
   body, html {
+    height: ${POPUP_HEIGHT}px;
     width: ${POPUP_WIDTH}px;
   }
-  body, html, #root {
-    height: ${POPUP_HEIGHT}px;
+
+  #root {
+    display: flex;
+    flex-flow: column;
+    height: 100%;
   }
 
   body * {
--- a/extension/src/popup/constants/styles.tsx
+++ b/extension/src/popup/constants/styles.tsx
@@ -35,3 +35,5 @@ export const FONT_WEIGHT = {
   normal: "400",
   bold: "600",
 };
+
+export const ROUNDED_CORNERS = "1.25rem";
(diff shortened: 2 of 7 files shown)


extension/src/popup/views/FullscreenSuccessMessage.tsx
```diff
@@ -50,27 +60,78 @@ const IlloContainerEl = styled.div`
...
+  margin: 0;
+`;
+
+const MessageListEl = styled.ul`
+  padding: 0 1.3rem;
+
+  li {
+    margin-top: 1rem;
+  }
+`;
```
piyalbasu, line 95: this alert message layout is very *similar* to what we do in `<WarningMessage />`, but just different enough that it doesn't really make sense to share any code. Kinda annoying.

This whole app would benefit from a design based on a system. Something I'll push for with the design team for v2

extension/src/popup/App.tsx
```diff
@@ -23,10 +23,14 @@ const GlobalStyle = createGlobalStyle`
...
+    height: ${POPUP_HEIGHT}px;
     width: ${POPUP_WIDTH}px;
   }
-  body, html, #root {
-    height: ${POPUP_HEIGHT}px;
+
+  #root {
+    display: flex;
+    flex-flow: column;
+    height: 100%;
```
piyalbasu, line 33: there were some issues with larger screen sizes. The blue background would stop abruptly in the middle of the page. This helps fix that

extension/src/popup/views/FullscreenSuccessMessage.tsx
```diff
@@ -25,7 +29,7 @@ const WrapperEl = styled.div`
   background: ${COLOR_PALETTE.primary};
   display: flex;
   flex-direction: column;
-  height: calc(100vh - ${HEADER_HEIGHT}px);
+  flex-grow: 1;
```
piyalbasu, line 32: this makes the div extend to the bottom of the page

extension/src/popup/constants/styles.tsx
```diff
@@ -35,3 +35,5 @@ export const FONT_WEIGHT = {
   normal: "400",
   bold: "600",
 };
+
+export const ROUNDED_CORNERS = "1.25rem";
```
jeesunikim, line 39: TY!

jeesunikim (approved):

PR #113, opened 2020-10-09
use formik's `dirty` prop to disable submit btn until user interaction

piyalbasu:
Screen Shot 2020-10-08 at 9 42 12 PM

### DIFF
--- a/extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
+++ b/extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
@@ -128,7 +128,7 @@ export const ConfirmMnemonicPhrase = ({
   return (
     <>
       <Formik initialValues={initialWordState} onSubmit={handleSubmit}>
-        {({ setValues, values, isSubmitting, handleChange }) => (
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase.tsx
```diff
@@ -128,7 +128,7 @@ export const ConfirmMnemonicPhrase = ({
   return (
     <>
       <Formik initialValues={initialWordState} onSubmit={handleSubmit}>
-        {({ setValues, values, isSubmitting, handleChange }) => (
+        {({ dirty, setValues, values, isSubmitting, handleChange }) => (
```
piyalbasu, line 131: `dirty` checks if all fields have been altered from their initial state. Basically, has the user filled out the form yet
https://formik.org/docs/api/formik#dirty-boolean

piyalbasu: not a great/intuitive prop name. Also, kind of a bummer we need to use this render props pattern because we need to remember to pass this value every time we create a form

jeesunikim (approved):

PR #114, opened 2020-10-09
feature/opt-out-event-tracking > component and redux toolkit slice added

jeesunikim:


@shared/api/helpers.ts
```diff
@@ -13,22 +13,24 @@ declare global {
 }
 
 export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
+  const MESSAGE_ID = Date.now();
```
piyalbasu, line 16: the issue before was that we were instantiating 2 message listeners. They would each post a message, but both instances would respond and terminate as soon as the first message came back. So, as soon as `loadAccount` responded, both listeners would terminate and there was nothing left to respond to `loadSettings`. 

This tags each instance with a unique id so they will only respond to messages that they sent

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -294,6 +295,26 @@ export const popupMessageListener = (
...
+
+    localStorage.setItem(DATA_SHARING_ID, JSON.stringify(isDataSharingAllowed));
+
+    sendResponse({
+      isDataSharingAllowed,
+    });
+  };
+
+  const loadSettings = () => {
+    const dataSharingValue = localStorage.getItem(DATA_SHARING_ID) || "true";
```
piyalbasu, line 310: defaulting to "true". a user can opt out after account creation has been completed, per ticket

jeesunikim:
@piyalbasu, it looks good! Thank you for tackling it. It says I cannot approve this PR because I'm the author 😄.

piyalbasu (approved):

PR #115, opened 2020-10-16
clear an API error after failing to recover from mnemonic phrase

piyalbasu:
I explored a few different ways to make this scale a bit better, but it would require a fair amount of refactoring. Since we only need to do this in one place, I'll hold off for now and take the easy way out :D

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,12 @@ module.exports = {
         "import/no-extraneous-dependencies": [0, { devDependencies: false }],
       },
     },
+    {
+      files: ["*/**/ducks/*.ts"],
+      rules: {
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -12,6 +12,12 @@ module.exports = {
         "import/no-extraneous-dependencies": [0, { devDependencies: false }],
       },
     },
+    {
+      files: ["*/**/ducks/*.ts"],
+      rules: {
+        "no-param-reassign": "off",
+      },
+    },
```
piyalbasu, line 20: this is something we discussed in Slack a while back. Disabling this rule allows us to use immer in reducers

vcarl (approved):

PR #117, opened 2020-10-20
complete favicon for client site in /grant-access

piyalbasu:
Screen Shot 2020-10-20 at 7 23 18 PM


extension/src/popup/helpers/getSiteFavicon.ts
```diff
@@ -0,0 +1,3 @@
+export const FAVICON_URL = "https://www.google.com/s2/favicons?domain=";
+
+export const getSiteFavicon = (url: string) => `${FAVICON_URL}${url}`;
```
piyalbasu, line 3: google returns a generic icon if the domain is not found 🎉

jeesunikim (approved):

PR #118, opened 2020-10-23
cleaning up <Loading /> implementation to prevent jumpy pageload

piyalbasu:


extension/src/popup/App.tsx
```diff
@@ -59,7 +66,10 @@ export function App() {
   return (
     <Provider store={store}>
       <GlobalStyle />
-      <Router />
+      <RouteWrapperEl>
+        <Router />
+      </RouteWrapperEl>
+      <Loading />
```
piyalbasu, line 72: keep the <Loading /> component behind the Router component so if the Route hasn't resolved, the loading screen will show

extension/src/popup/Router.tsx
```diff
@@ -55,9 +53,11 @@ const PublicKeyRoute = (props: RouteProps) => {
   if (applicationState === APPLICATION_STATE.APPLICATION_ERROR) {
     return <AppError>{error}</AppError>;
   }
+
   if (applicationState === APPLICATION_STATE.APPLICATION_LOADING) {
-    return <Loading />;
+    return null;
```
piyalbasu, line 58: returning `null` will show the <Loading /> component

extension/src/popup/Router.tsx
```diff
@@ -193,25 +194,21 @@ export const Router = () => {
       <RouteListener />
       <Switch>
         <PublicKeyRoute path={ROUTES.account}>
-          <Header />
```
piyalbasu, line 196: It's a better implementation to just include <Header /> inside views that need it.

extension/src/popup/components/Header.tsx
```diff
@@ -42,12 +38,6 @@ type HeaderProps = {
 };
 
 export const Header = ({ className, ...props }: HeaderProps) => {
-  const applicationState = useSelector(applicationStateSelector);
-
-  if (applicationState === APPLICATION_STATE.APPLICATION_LOADING) {
-    return null;
-  }
```
piyalbasu, line 49: this was a dumb idea i tried near the beginning and was preventing views that required the Header from just including <Header /> in their render

extension/src/popup/views/Account.tsx
```diff
@@ -164,5 +166,5 @@ export const Account = () => {
       </AccountEl>
       <Footer />
     </>
-  );
+  ) : null;
```
piyalbasu, line 169: Now we'll just be seeing the Loading screen until this component returns something

extension/src/popup/views/AccountCreator.tsx
```diff
@@ -65,62 +66,65 @@ export const AccountCreator = () => {
...
-      header="Create a password"
-      icon={CreatePasswordIllo}
-      goBack={() => navigateTo(ROUTES.welcome)}
-    >
-      <Formik
-        initialValues={initialValues}
-        onSubmit={handleSubmit}
-        validationSchema={AccountCreatorSchema}
+    <>
+      <Header />
```
piyalbasu, line 70: much nicer to include Header here

extension/src/popup/views/Loading.tsx
```diff
@@ -8,8 +18,39 @@ const LoadingEl = styled.div`
...
+  }
+
+  @keyframes lds-dual-ring {
+    0% {
+      transform: rotate(0deg);
+    }
+    100% {
+      transform: rotate(360deg);
+    }
+  }
```
piyalbasu, line 46: this is an open source css loading spinner. This is placeholder until design comes up with something custom to implement

extension/src/popup/App.tsx
vcarl: Ah interesting, so this is like a backdrop that gets covered up if there isn't something else to show?

extension/src/popup/views/Loading.tsx
vcarl: Maybe nice to include a link to the source of it, in that case

vcarl (approved):

extension/src/popup/App.tsx
piyalbasu: yup, that's right

PR #119, opened 2020-10-27
fix to make "blue background" reach the bottom on full screen success messages

piyalbasu:
This also makes zooming in and out retain the desired layout

### DIFF
--- a/extension/src/popup/components/Header.tsx
+++ b/extension/src/popup/components/Header.tsx
@@ -37,11 +37,9 @@ type HeaderProps = {
   className?: string;
 };
 
-export const Header = ({ className, ...props }: HeaderProps) => {
-  return (
-    <HeaderEl className={className} {...props}>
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/Header.tsx
```diff
@@ -37,11 +37,9 @@ type HeaderProps = {
...
-export const Header = ({ className, ...props }: HeaderProps) => {
-  return (
-    <HeaderEl className={className} {...props}>
-      <H1El>Lyra</H1El>
-      <NetworkEl>{NETWORK_NAME}</NetworkEl>
-    </HeaderEl>
-  );
-};
+export const Header = ({ className, ...props }: HeaderProps) => (
+  <HeaderEl className={className} {...props}>
```
piyalbasu, line 41: this fixes a linting warning

PR #120, opened 2020-10-28
Add links to client sites to footer

piyalbasu:
Screen Shot 2020-10-28 at 6 32 40 PM


extension/src/popup/basics/Images.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+interface RetinaImgProps {
+  retina: string;
+  src?: string;
+  alt: string;
+}
+
+export const RetinaImg = ({ retina = "", src = "", alt }: RetinaImgProps) => (
+  <img src={src || retina} srcSet={`${retina} 2x`} alt={alt} />
```
morleyzhi, line 10: Why offer the choice? Convention is usually better over configuration.

piyalbasu: Hm. Not sure I understand. Are you asking why include `src` as a prop?

vcarl: I read this as, `src` + `retina` are already follow a `x@2x` naming convention, so it'd be simpler to assume the `@2x` exists and skip the `retina` prop altogether

PR #137, opened 2020-11-10
remove unused `storage` permission

piyalbasu:


extension/public/static/manifest.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.0.0",
   "version_name": "beta.1",
   "description": "Freighter is a non-custodial wallet extension that enables you to sign Stellar transactions via your browser.",
-  "permissions": ["storage"],
```
piyalbasu, line 6: this was flagged by the Chrome store as a violation as we're not using it in the extension

PR #141, opened 2020-11-13
Show options params

piyalbasu:
Screen Shot 2020-11-13 at 5 34 28 PM

### DIFF
--- a/extension/src/popup/components/signTransaction/Operations.tsx
+++ b/extension/src/popup/components/signTransaction/Operations.tsx
@@ -14,7 +14,8 @@ interface TransactionInfoResponse {
   asset: { code: string };
   signer: {
     ed25519PublicKey?: string;
-    sha256Hash?: { data: [] };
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/signTransaction/Operations.tsx
```diff
@@ -0,0 +1,260 @@
...
+            <OperationBoxHeaderEl>
+              {operationIndex}. {OPERATION_TYPES[type]}
+            </OperationBoxHeaderEl>
+            <OperationsListEl>
+              {destination ? (
+                <KeyValueList
+                  operationKey="Destination"
+                  operationValue={truncatedPublicKey(destination)}
+                />
+              ) : null}
```
piyalbasu, line 135: this pattern of checking a value and then showing a custom label/value pair is a little duplicative, but each operation param is _just_ different enough that this appears to be the cleanest, most readable way to show these

PR #149, opened 2020-12-02
install webextension-polyfill and switch callback message listeners to use promises

piyalbasu:
This makes the extension compatible with Firefox


@shared/api/helpers.ts
```diff
@@ -39,13 +40,14 @@ export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
   });
 };
 
-export const sendMessageToBackground = (msg: {}): Promise<Response> => {
+export const sendMessageToBackground = async (msg: {}): Promise<Response> => {
```
piyalbasu, line 43: `webextension-polyfill` uses a promise based API rather than the callback based API we were using, so all these methods had to be mildly refactored

extension/src/background/index.ts
```diff
@@ -1,13 +1,26 @@
...
+  browser.runtime.onMessage.addListener(async (request, sender) => {
+    // todo this is kinda ugly
+
+    let res;
+    if (Object.values(SERVICE_TYPES).includes(request.type)) {
+      res = await popupMessageListener(request);
+    }
+    if (Object.values(EXTERNAL_SERVICE_TYPES).includes(request.type)) {
+      res = await freighterApiMessageListener(request, sender);
+    }
```
piyalbasu, line 22: in an upcoming PR, i'm going to make this a little nicer as we're now doing this check here and then again inside of the messageListener

extension/src/background/index.ts
```diff
@@ -1,13 +1,26 @@
...
 import { freighterApiMessageListener } from "./messageListener/freighterApiMessageListener";
 
 export const initMessageListener = () => {
   // returning true is very important in these message listeners. It tells the listener that the callback
   // could possibly be async, so keep the channel open til we send a reponse.
-  chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
-    popupMessageListener(request, sender, sendResponse);
-    freighterApiMessageListener(request, sender, sendResponse);
+  browser.runtime.onMessage.addListener(async (request, sender) => {
+    // todo this is kinda ugly
```
piyalbasu, line 14: will be addressed in an upcoming PR

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -25,7 +21,6 @@ const WINDOW_DIMENSIONS = `width=${POPUP_WIDTH},height=667`;
 export const freighterApiMessageListener = (
   request: Request,
   sender: Sender,
-  sendResponse: (response: SendResponseInterface) => void,
```
piyalbasu, line 28: `webextension-polyfill` uses a promise based API rather than the callback based API we were using, so all these methods had to be mildly refactored

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -46,11 +42,7 @@ interface StellarHdWallet {
   getSecret: (number: Number) => string;
 }
 
-export const popupMessageListener = (
-  request: Request,
-  _: Sender,
-  sendResponse: (response: SendResponseInterface) => void,
-) => {
```
piyalbasu, line 53: same. removing `sendResponse`

extension/webpack.dev.js
```diff
@@ -14,6 +15,14 @@ const devConfig = {
     new webpack.DefinePlugin({
       DEVELOPMENT: true,
     }),
+    new ModuleReplaceWebpackPlugin({
+      modules: [
+        {
+          test: /webextension-polyfill-ts/,
+          replace: "./src/shims/webextension-polyfill.ts",
+        },
```
piyalbasu, line 23: oh boy - so `webextension-polyfill` tries to be helpful and will throw a breaking error if you try to run it in a browser (instead of an extension). Unfortunately, for our dev flow, we need to open the extension in a browser. We don't _use_ the webextension in dev mode, but just importing the lib crashes the app. Very annoying. Very lame solution

PR #150, opened 2020-12-02
Firefox new tab

piyalbasu:
firefox doesn't like `window.open` evidently. We'll create a tab using the webextension lib


extension/src/shims/webextension-polyfill.ts
```diff
@@ -1 +1,5 @@
-export const browser = {};
+export const browser = {
+  tabs: {
+    create: (url: string) => window.open(url),
```
piyalbasu, line 3: in dev mode, let's still use window.open

PR #168, opened 2020-12-20
Add new svg's and styled heights for each

piyalbasu:


extension/src/popup/components/Onboarding.tsx
```diff
@@ -98,9 +98,7 @@ export const Onboarding = ({
       {goBack && !isNewTabSession ? <BackButton onClick={goBack} /> : null}
       <Screen>
         <HeadingEl>
-          <IlloContainerEl>
-            <img src={icon} alt="Onboarding Illustration" />
-          </IlloContainerEl>
+          <IlloContainerEl>{icon}</IlloContainerEl>
```
piyalbasu, line 101: since each icon needs to be styled, let's just have the parent handle that and pass in a fully formed React element to this component

PR #170, opened 2020-12-21
fix broken footer and doc content

piyalbasu:


docs/docs/guide/usingFreighter.md
```diff
@@ -135,5 +135,44 @@ const userSignTransaction = async (xdr: String) => {
   return signedTransaction;
 };
 
-const userSignedTransaction = userSignTransaction();
+const xdr = ""; // replace this with an xdr string of the transaction you want to sign
+const userSignedTransaction = userSignTransaction(xdr);
+```
+
+freighter-api will return a signed transaction xdr. Below is an example of how you might submit this signed transaction to Horizon using `stellar-sdk` (https://github.com/stellar/js-stellar-sdk):
+
```
piyalbasu, line 143: a user on keybase was unclear about what to do with a signed transaction XDR so I added this bit to spell out what a flow might look like

docs/docusaurus.config.js
```diff
@@ -21,6 +21,19 @@ module.exports = {
...
+          title: "Questions or comments?",
+          items: [
+            {
+              label: "Let us know!",
+              to: "https://stellarform.typeform.com/to/r4FiNpX1",
+            },
+          ],
+        },
+      ],
+    },
```
piyalbasu, line 36: the latest minor version of docusaurus breaks if you don't have a footer, so I guess now we need footer links. Throwing in the typeform link, as it could be useful here

PR #181, opened 2021-02-01
Add background functionality of adding a new account from existing mnemonic

piyalbasu:


extension/src/background/ducks/session.ts
```diff
@@ -56,6 +63,10 @@ export const mnemonicPhraseSelector = createSelector(
   sessionSelector,
   (session) => session.mnemonicPhrase,
 );
+export const allAccountsSelector = createSelector(
+  sessionSelector,
+  (session) => session.allAccounts,
+);
```
piyalbasu, line 69: adding an array of all public keys associated with the mnemonic

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -25,7 +26,7 @@ import {
   publicKeySelector,
 } from "background/ducks/session";
 
-const KEY_ID = "keyId";
+let DEFAULT_ACCOUNT_ID = 0;
```
piyalbasu, line 29: WIP: need to track which iteration of the mnemonic the user is on

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -84,21 +93,21 @@ export const popupMessageListener = (request: Request) => {
       console.error(e);
     }
 
+    const keyIdListStr = localStorage.getItem(KEY_ID_LIST) || "[]";
+    const keyIdListArr = JSON.parse(keyIdListStr);
+    keyIdListArr.push(keyStore.id);
+
+    localStorage.setItem(KEY_ID_LIST, JSON.stringify(keyIdListArr));
```
piyalbasu, line 100: we're storing an array of keyId's in localStorage. We will keep a concept of an "active" keyId, meaning the one the user is currently working with

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: This will likely need to be store in localStorage to keep it from being wiped on extension refresh

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -187,30 +230,59 @@ export const popupMessageListener = (request: Request) => {
...
+      keyIdList.push(localStorage.getItem(KEY_ID));
+      localStorage.setItem(KEY_ID_LIST, JSON.stringify(keyIdList));
     }
 
+    const unlockedAccounts = [] as Array<string>;
+    let selectedPublicKey = "";
+    let selectedPrivateKey = "";
+    let accountMnemonicPhrase;
+
+    await Promise.all(
```
piyalbasu, line 248: because we don't story any public keys in plain text, to get all the accounts the user is working with, we need to iterate over all the key stores we have

PR #184, opened 2021-02-03
add ducks and simple ui toggle to enable switching between accounts

piyalbasu:


extension/src/popup/views/Account.tsx
```diff
@@ -166,6 +173,26 @@ export const Account = () => {
           </div>
         </AccountDetailsEl>
       </AccountEl>
+      <ul>
```
piyalbasu, line 176: unstyled test ui to test dispatching action

PR #190, opened 2021-02-08
fixing potential public/private key mismatches

piyalbasu:


extension/src/background/helpers/session.ts
```diff
@@ -8,13 +8,17 @@ const SESSION_LENGTH = 5;
...
   DURATION = 1000 * 60 * SESSION_LENGTH;
+  runningTimeout: null | ReturnType<typeof setTimeout> = null;
   constructor(duration?: number) {
     this.DURATION = duration || this.DURATION;
   }
 
   startSession(key: { privateKey: string }) {
+    if (this.runningTimeout) {
+      clearTimeout(this.runningTimeout);
+    }
```
piyalbasu, line 19: if we had set a timer previously, disregard it and setup a new one

PR #191, opened 2021-02-09
Store key derivation number to prevent data issues on extension refresh

piyalbasu:


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -295,9 +299,10 @@ export const popupMessageListener = (request: Request) => {
 
     if (wallet) {
       const keyPair = {
-        publicKey: wallet.getPublicKey(DEFAULT_ACCOUNT_ID),
-        privateKey: wallet.getSecret(DEFAULT_ACCOUNT_ID),
+        publicKey: wallet.getPublicKey(0),
+        privateKey: wallet.getSecret(0),
```
piyalbasu, line 303: on recoverAccount, just grab the primary key. User can branch off from there

PR #192, opened 2021-02-11
change shape of allAccounts to incorporate a custom account name

piyalbasu:


@shared/api/types.ts
```diff
@@ -22,10 +22,16 @@ export interface Response {
...
 export interface ExternalRequest {
   transactionXdr: string;
   type: EXTERNAL_SERVICE_TYPES;
 }
+
+export interface Account {
+  publicKey: string;
+  name: string;
+  imported: boolean;
+}
```
piyalbasu, line 37: this is the main thing to see in this PR: rather an array of publicKeys, allAccounts will now be an array of type Account so the FE has all the information about the account in a contained object

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -344,13 +369,21 @@ export const popupMessageListener = (request: Request) => {
...
+      const keyId = localStorage.getItem(KEY_ID);
+      if (keyId) {
+        keyIdList.push(keyId);
+        localStorage.setItem(KEY_ID_LIST, JSON.stringify(keyIdList));
+        localStorage.setItem(KEY_DERIVATION_NUMBER_ID, "0");
+        localStorage.setItem(
+          ACCOUNT_NAME_LIST_ID,
+          JSON.stringify({ [keyId]: "Account 1" }),
+        );
+      }
```
piyalbasu, line 381: if you have a legacy account, just set the account name as "Account 1"

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -58,26 +66,36 @@ export const popupMessageListener = (request: Request) => {
...
+            imported,
+          },
+        ],
       }),
     );
 
     const keyMetadata = {
       key: {
-        extra: { mnemonicPhrase },
+        extra: { imported, mnemonicPhrase },
```
piyalbasu, line 98: this somewhat tangentially related: we need to differentiate between imported/generated accounts and include that info in allAccounts.

PR #201, opened 2021-02-19
clear api errors after correcting fields

piyalbasu:


extension/src/popup/views/AddAccount/AddAccount.tsx
```diff
@@ -46,17 +50,27 @@ export const AddAccount = () => {
     }
   };
 
+  const clearAddAccountError = (e: React.ChangeEvent<any>) => {
+    if (authError && e.target.value === "") {
+      dispatch(clearApiError());
+    }
+  };
```
piyalbasu, line 57: this code is a bit duplicative, but I wasn't able to figure out a way to reuse a hook on callback quickly. I've created a ticket for further investigation

PR #205, opened 2021-02-24
make UI flexible enough to handle long custom names gracefully

piyalbasu:
Screen Shot 2021-02-24 at 12 28 53 AM
Screen Shot 2021-02-24 at 12 29 03 AM


piyalbasu:
also includes new yarn.lock after re-yarn'ing

PR #215, opened 2021-03-05
make sure to preserve order of account

piyalbasu:


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -429,30 +429,32 @@ export const popupMessageListener = (request: Request) => {
...
-            unlockedAccounts.push({
-              publicKey,
-              name: accountNameList[keyId] || `Account ${keyIdList.length}`,
-              imported,
-            });
-          }
-        }),
-      );
+
+      // eslint-disable-next-line no-plusplus
```
piyalbasu, line 433: all that async iteration talk got me thinking maybe I should ensure order here afterall just to account for any weird edge cases :)

quietbits: Sweet! 🙂 Simple `for` loop for the win! 👍

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -565,6 +568,33 @@ export const popupMessageListener = (request: Request) => {
...
+  //   var arr = [prom1, prom2];
+  //   await Promise.all(
+  //     arr.map(async (num) => {
+  //       var a = await num();
+  //       hold.push(a);
+  //     }),
+  //   );
+
+  //   console.log(hold);
+  // };
```
quietbits, line 596: Do we need to keep this?

quietbits (approved):

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: lol whoops - no we do not

PR #217, opened 2021-03-10
show correct warning for malicious and unsafe accounts

piyalbasu:


extension/src/popup/components/warningMessages/FlaggedWarningMessage.tsx
```diff
@@ -16,22 +16,22 @@ const DangerousAccountWarning = ({
   isUnsafe: boolean;
   isMalicious: boolean;
 }) => {
+  if (isMalicious) {
```
piyalbasu, line 19: if account `isMalicious`, show malicious warning message. But only show unsafe warning message if account is not also malicious

PR #218, opened 2021-03-11
update eslint and fix all remaining errors

piyalbasu:


.eslintrc.js
```diff
@@ -12,16 +12,13 @@ module.exports = {
         "import/no-extraneous-dependencies": [0, { devDependencies: false }],
       },
     },
-    {
-      files: ["*/**/ducks/*.ts"],
-      rules: {
-        "no-param-reassign": "off",
-      },
-    },
```
piyalbasu, line 20: this rule was added to stellar's ts-config recently

.eslintrc.js
```diff
@@ -12,16 +12,13 @@ module.exports = {
...
   ],
   rules: {
-    "no-console": 0,
```
quietbits, line 23: Looks like we might need to put the `no-console` rule back, it's not happy about it in some places.

quietbits (approved):
Nice! 👏

PR #226, opened 2021-03-24
add transaction history view

piyalbasu:
https://user-images.githubusercontent.com/6789586/112391445-4d1f9880-8cce-11eb-9b23-42645a45860a.mov


@shared/api/internal.ts
```diff
@@ -193,31 +199,41 @@ export const getAccountDetails = async (
...
   let isFunded = null;
+  let operations = [] as Array<HorizonOperation>;
 
   try {
     ({ balances } = await dataProvider.fetchAccountDetails());
-    const transactionData = await dataProvider.fetchPayments({
-      limit: TRANSACTIONS_LIMIT,
-    });
-    payments = transactionData?.records || null;
+    isFunded = true;
```
piyalbasu, line 208: evidently `isAccountFunded ` just calls `fetchAccountDetails`and returns `false` if that call fails. So, instead of doing 2 networks requests, i'll do that myself here and get both on 1 request

@shared/api/internal.ts
```diff
@@ -193,31 +199,41 @@ export const getAccountDetails = async (
...
+      balances,
+      isFunded: false,
+      operations,
+    };
   }
 
   try {
-    isFunded = await dataProvider.isAccountFunded();
+    const server = new StellarSdk.Server(NETWORK_URL);
+    const operationsData = await server
```
quietbits, line 220: If we don't need non-payment operations, we can use `stellar/wallet-sdk`s `fetchPayments()` method, which simplifies this call a bit. [Example ](https://github.com/stellar/account-viewer-v2/blob/master/src/ducks/txHistory.ts#L28:L41)in AV.

quietbits (approved):

@shared/api/internal.ts
piyalbasu: Ah, unfortunately, we do need all types of operations

PR #233, opened 2021-04-02
mainnet test toggle background

piyalbasu:
This PR changes the network toggle from being a build time configuration to a run time configuration that defaults to mainnet


@shared/api/internal.ts
```diff
@@ -279,7 +294,7 @@ export const retryAssetIcon = async ({
   } catch (e) {
     return assetIcons;
   }
-  const icon = await getIconUrlFromIssuer({ key, code });
+  const icon = await getIconUrlFromIssuer({ key, code, networkDetails });
```
piyalbasu, line 297: the only downside to not having a static network config is that in non-redux environments, we need to pass `networkDetails` as a param to any func that needs to access Horizon

piyalbasu: We could make the UI redux store a global, but it's a little messy to do so right now bc of monorepo structure. In the future, if passing around values becomes annoying, we can do some more work to do that

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -74,9 +75,10 @@ export const freighterApiMessageListener = (
 
   const submitTransaction = async () => {
     const { transactionXdr } = request;
+    const { network, networkUrl } = getNetworkDetails(getIsTestnet());
```
piyalbasu, line 78: this is basically how it works now - figure out if we're on testnet based on localStorage, and then get the appropriate config

extension/src/popup/ducks/settings.ts
```diff
@@ -55,26 +59,28 @@ const settingsSlice = createSlice({
...
-        const { isDataSharingAllowed } = action?.payload || {
-          isDataSharingAllowed: false,
+        const { isDataSharingAllowed, networkDetails } = action?.payload || {
+          ...initialState,
         };
 
         return {
           ...state,
           isDataSharingAllowed,
+          networkDetails,
```
piyalbasu, line 83: for UI, ask the Background for network config and then put it into redux

quietbits (approved):

PR #238, opened 2021-04-09
add settings to control stellar.expert validation

piyalbasu:
Screen Shot 2021-04-08 at 10 15 07 PM

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -35,8 +35,8 @@ export const addAccountName = ({
 export const getIsTestnet = () =>
   JSON.parse(localStorage.getItem(IS_TESTNET_ID) || "false");
 
-export const getIsValidatingMemo = () =>
+export const getIsMemoValidationEnabled = () =>
   JSON.parse(localStorage.getItem(IS_VALIDATING_MEMO_ID) || "true");
 
-export const getIsValidatingSafety = () =>
+export const getIsSafetyValidationEnabled = () =>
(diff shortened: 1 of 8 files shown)


extension/src/background/helpers/account.ts
```diff
@@ -32,3 +34,9 @@ export const addAccountName = ({
 
 export const getIsTestnet = () =>
   JSON.parse(localStorage.getItem(IS_TESTNET_ID) || "false");
+
+export const getIsValidatingMemo = () =>
+  JSON.parse(localStorage.getItem(IS_VALIDATING_MEMO_ID) || "true");
+
+export const getIsValidatingSafety = () =>
+  JSON.parse(localStorage.getItem(IS_VALIDATING_SAFETY_ID) || "true");
```
piyalbasu, line 42: dealing with localstorage is getting a bit duplicative, but i'm hoping to possibly replace all of this soon with IndexedDB

quietbits (approved):
Looks great! 🎉 When I saw `isValidatingMemo` and `isValidatingSafety` variables, they sounded like status variables (is something pending), but it got clear once I saw how they are used. Just thought I'll mention that. 😊 I don't have a better suggestion for the names. 😅

piyalbasu:
> Looks great! 🎉 When I saw `isValidatingMemo` and `isValidatingSafety` variables, they sounded like status variables (is something pending), but it got clear once I saw how they are used. Just thought I'll mention that. 😊 I don't have a better suggestion for the names. 😅

This is a good point! I'll think about that 🤔

PR #239, opened 2021-04-13
update freighter-api to retrieve current network setting

piyalbasu:


@stellar/freighter-api/src/signTransaction.ts
```diff
@@ -1,4 +1,6 @@
 import { submitTransaction } from "@shared/api/external";
 
-export const signTransaction = (transactionXdr: string) =>
-  submitTransaction(transactionXdr);
+export const signTransaction = (
+  transactionXdr: string,
+  network?: "PUBLIC" | "TESTNET"
```
quietbits, line 5: Not sure how the file structure works out, but maybe we could use enum for the network.

quietbits (approved):

@stellar/freighter-api/src/signTransaction.ts
piyalbasu: Ah yeah, I initially did that - but the error message a client using this lib would have seen would be something like `argument of type "BAD_INPUT" is not assignable to parameter of type NETWORKS | undefined` which I think is slightly less useful for people trying to implement this function as they'd have to hunt through this repo to find what NETWORKS is. 

I think I'd like for them to clearly see that they need to pass the string "PUBLIC" or "TESTNET" even if it's a bit duplicative on our part

PR #240, opened 2021-04-13
add fund with friendbot button for testnet accounts

piyalbasu:
Screen Shot 2021-04-13 at 1 22 10 PM


extension/src/popup/App.tsx
```diff
@@ -11,7 +11,7 @@ import { POPUP_WIDTH, POPUP_HEIGHT } from "constants/dimensions";
 import { metricsMiddleware } from "helpers/metrics";
 
 import { FONT_FAMILY, COLOR_PALETTE } from "popup/constants/styles";
-import { reducer as auth } from "popup/ducks/authServices";
+import { reducer as auth } from "popup/ducks/accountServices";
 import { reducer as settings } from "popup/ducks/settings";
```
piyalbasu, line 15: a lot of the changes in this PR are simply renaming `authServices` to be more descriptive

extension/src/popup/components/account/AccountDetails.tsx
```diff
@@ -81,7 +84,7 @@ export const AccountDetails = () => {
     };
 
     fetchAccountDetails();
-  }, [publicKey, networkDetails]);
+  }, [publicKey, networkDetails, isAccountFriendbotFunded]);
```
piyalbasu, line 87: force a refetch of account details from Horizon if we've funded with friendbot

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -134,7 +134,9 @@ export const popupMessageListener = (request: Request) => {
     });
   };
 
-  const _fundAccount = async (publicKey: string) => {
+  const fundAccount = async () => {
+    const { publicKey } = request;
```
piyalbasu, line 138: instead of automatically funding with friendbot on Testnet, a user has to manually trigger this from the UI now

leighmcculloch:
I think it could be worth adding copy to the text box that says the account doesn't exist that this is on the test network so it can be funded by friendbot, to make sure it isn't misunderstood this is somehow a pubnet thing. I realize it says test network up in the top right, but I think it's worth calling it out in the text too so that the text leads into the call to action button below.

>### This Stellar address is not funded
>
>To create this account, fund it with a minimum of 1 XLM. On the Test Network, Friendbot can fund new accounts.

I don't feel strongly about this.

extension/src/popup/components/account/NotFundedMessage.tsx
```diff
@@ -40,27 +46,61 @@ const NotFundedHeaderEl = styled.h3`
...
+}: {
+  isTestnet: boolean;
+  publicKey: string;
+  setIsAccountFriendbotFunded: (isAccountFriendbotFunded: boolean) => void;
+}) => {
+  const dispatch = useDispatch();
+
+  const handleFundAccount = async () => {
+    await dispatch(fundAccount(publicKey));
+    setIsAccountFriendbotFunded(true);
```
quietbits, line 62: Could we run into issues with timing if we trigger fetching of the account info before funding is done?

quietbits (approved):

extension/src/popup/components/account/NotFundedMessage.tsx
piyalbasu: That's a good question. `fundAccount()` doesn't resolve until the funding request has completed, so `setIsAccountFriendbotFunded` is only triggered once we're done. In a Slow 3G test, for ex, the UI stays in a loading state while we're making the funding request and then finally re-fetches once that promise has resolved

piyalbasu:
> I think it could be worth adding copy to the text box that says the account doesn't exist that this is on the test network so it can be funded by friendbot, to make sure it isn't misunderstood this is somehow a pubnet thing. I realize it says test network up in the top right, but I think it's worth calling it out in the text too so that the text leads into the call to action button below.
> 
> > ### This Stellar address is not funded
> > To create this account, fund it with a minimum of 1 XLM. On the Test Network, Friendbot can fund new accounts.
> 
> I don't feel strongly about this.

@leighmcculloch I talked with Bruno about this and we thought it was a good idea. I just added this line to the copy for Testnet

PR #242, opened 2021-04-22
Basic implementation of Sponsorships and Claimable Balances

piyalbasu:
Screen Shot 2021-04-22 at 3 33 31 PM

### DIFF
--- a/extension/src/constants/transaction.ts
+++ b/extension/src/constants/transaction.ts
@@ -30,6 +30,6 @@ export enum CLAIM_PREDICATES {
   claimPredicateAnd = "And",
   claimPredicateOr = "Or",
   claimPredicateNot = "Not",
-  claimPredicateBeforeRelativeTime = "Relative Time Before",
+  claimPredicateBeforeRelativeTime = "Before Relative Time",
   claimPredicateBeforeAbsoluteTime = "Before Absolute Time",
 }
--- a/extension/src/constants/transaction.ts
+++ b/extension/src/constants/transaction.ts
@@ -1,24 +1,35 @@
 export enum OPERATION_TYPES {
   accountMerge = "Account Merge",
   allowTrust = "Allow Trust",
+  beginSponsoringFutureReserves = "Begin Sponsoring Future Reserves",
   bumpSequence = "Bump Sequence",
   changeTrust = "Change Trust",
   createAccount = "Create Account",
+  createClaimableBalance = "Create Claimable Balance",
   createPassiveSellOffer = "Create Passive Sell Offer",
+  endSponsoringFutureReserves = "End Sponsoring Future Reserves",
   manageBuyOffer = "Manage Buy Offer",
   manageData = "Manage Data",
   manageSellOffer = "Manage Sell Offer",
   pathPaymentStrictReceive = "Path Payment Strict Receive",
   pathPaymentStrictSend = "Path Payment Strict Send",
   payment = "Payment",
-  setOptions = "Set Options",
-  beginSponsoringFutureReserves = "Begin Sponsoring Future Reserves",
-  endSponsoringFutureReserves = "End Sponsoring Future Reserves",
   revokeAccountSponsorship = "Revoke Account Sponsorship",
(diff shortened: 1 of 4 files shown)


extension/src/popup/components/signTransaction/Operations.tsx
```diff
@@ -19,34 +23,59 @@ interface Path {
...
-  selling: { code: string };
-  buyAmount: string;
-  price: string;
-  inflationDest: string;
-  setFlags: number;
-  clearFlags: number;
-  masterWeight: number;
-  lowThreshold: number;
-  medThreshold: number;
-  highThreshold: number;
```
piyalbasu, line 46: alphabetizing for my own sanity

extension/src/popup/components/signTransaction/Operations.tsx
```diff
@@ -421,6 +470,45 @@ export const Operations = ({
...
+                ? claimants.map(({ _destination }, index) => (
+                    /* eslint-disable react/no-array-index-key */
+                    <div key={`${_destination}${index}`}>
+                      {/* eslint-enable */}
+                      <SubHeaderEl>Claimant {index + 1}</SubHeaderEl>
+                      <KeyValueWithPublicKey
+                        operationKey="Destination"
+                        operationValue={_destination}
+                      />
+                      {/* TODO: Add appicable predicate UI */}
```
piyalbasu, line 501: I spent about a day to trying to algorithmically decode and show all the possible predicates for a Claimable Balance, but the data is shaped in a way that does not make it very easy to do in a UI (check the type for `PredicateValue`, it's pretty wild).

Definitely not impossible and something we should do in the future, but it will take a bit more time and some design input to figure out how to show this information in a usable way.

extension/src/popup/components/signTransaction/Operations.tsx
```diff
@@ -421,6 +470,45 @@ export const Operations = ({
...
+                    </div>
+                  ))
+                : null}
+              {Object.entries(rest).map(([k, v]) =>
+                React.isValidElement(v) ? (
+                  <div key={k}>
+                    <KeyValueList key={k} operationKey={k} operationValue={v} />
+                  </div>
+                ) : null,
+              )}
```
piyalbasu, line 511: This bit will simply show a key/value for any operations we've missed in `OPERATION_TYPES`

extension/src/popup/components/signTransaction/Operations.tsx
```diff
@@ -19,34 +23,59 @@ interface Path {
...
+interface Predicate {
+  _switch: PredicateSwitch;
+  _value?: PredicateValue;
+}
+
+interface Claimant {
+  _destination: string;
+  _predicate: Predicate;
+}
+
```
piyalbasu, line 45: these Predicate types are a bit more typing than we actually need for this current implementation, but this is useful code that was a pain to figure out and will be beneficial in the future

extension/src/constants/transaction.ts
```diff
@@ -19,3 +23,13 @@ export enum TRANSACTION_WARNING {
...
 }
+
+export enum CLAIM_PREDICATES {
+  claimPredicateUnconditional = "Unconditional",
+  claimPredicateConditional = "Conditional",
+  claimPredicateAnd = "And",
+  claimPredicateOr = "Or",
+  claimPredicateNot = "Not",
+  claimPredicateBeforeRelativeTime = "Relative Time Before",
+  claimPredicateBeforeAbsoluteTime = "Before Absolute Time",
```
leighmcculloch, line 34: 💡 The first one should probably lead with before, otherwise it won't make much sense when combined with Not and And.

leighmcculloch (approved):
One suggestion (💡), but this is great regardless. I'm not proficient in typescript so I defer to others.

quietbits (approved):

extension/src/constants/transaction.ts
piyalbasu: Yes, good thinking. Also, better symmetry with Absolute times

PR #341, opened 2022-01-24
Feature/add sds input

piyalbasu:


extension/src/popup/views/UnlockAccount.tsx
```diff
@@ -108,15 +100,19 @@ export const UnlockAccount = () => {
...
+                      error={authError}
+                      placeholder="Enter Password"
+                      type="password"
+                      {...field}
+                    />
+                  )}
+                </Field>
               </FormRow>
-              <ApiErrorMessage error={authError} />
               <ButtonRowEl>
```
piyalbasu, line 116: This should be a pretty good pattern to follow when implementing an SDS Input that's connected to Formik

piyalbasu: We can remove <ApiErrorMessage /> as SDS Input will handle displaying the error 🎉

acharb (approved):
lgtm!

quietbits (approved):
:tada:

Issue #329, opened 2022-01-20
Reskin: Onboarding Flow


piyalbasu:
* Apply new designs and new copy to Onboarding flow: Account creation -> confirm backup phrase 
* Remove unused svg's from repo

This is the flow you enter when you have NO wallet info stored in `background` (a brand new user, for ex). 

An easy way to get into this flow is to open the DevTools for `chrome-extension://eiodnmeacfjkeaficifjdjomimneiona/_generated_background_page.html` and go to Application -> Local Storage and clear all the data in here. Nav'ing to anywhere in the app should now bring you int the Onboarding flow.

/views involved:
```
<Welcome>
<MnemonicPhrase>
<FullScreenSuccessMessage>
```

acharb:
looks like AccountCreator.tsx also is in this flow (unless it's included on another ticket I missed). Will do as part of this ticket too

acharb:
closed after https://github.com/stellar/freighter/pull/361

PR #352, opened 2022-01-25
popup/views: reskin Welcome page

acharb:
**WHAT**
re-skin welcome page

part of https://github.com/stellar/freighter/issues/329. Want to make sure I'm doing this correct before changing other views

Screen Shot 2022-01-25 at 3 46 50 PM


acharb:
@piyalbasu the buttons from SDS are slightly different than the designs (eg. I think border radius, shadow, color?). Is the idea to use the SDS imports as is, or try to mimic the designs completely?

piyalbasu:
> @piyalbasu the buttons from SDS are slightly different than the designs (eg. I think border radius, shadow, color?). Is the idea to use the SDS imports as is, or try to mimic the designs completely?

Good catch. Let's use the SDS <Button> as is to match our other apps! I'll let Charles know

piyalbasu (approved):

PR #353, opened 2022-01-26
Reskin/sign transaction- Operations and Warning Messages

piyalbasu:
Cleaning up Operations and adding Warning messages

(There isn't an official design for path payments and claimable balances yet, but this is similar to the old design, refactored to use scss)

Screen Shot 2022-01-26 at 1 49 08 PM
Screen Shot 2022-01-26 at 1 48 06 PM


extension/src/popup/components/identicons/KeyIdenticon/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+          "--Icon-padding": "0.2rem",
+          "--Icon-dimension": "1.5rem",
+        }
+      : {}),
+  } as React.CSSProperties;
+
+  return (
+    <div className="KeyIdenticon">
+      <div
+        className="KeyIdenticon--icon KeyIdenticon--icon"
```
quietbits, line 37: Duplicate class name `KeyIdenticon--icon`

extension/src/popup/components/identicons/KeyIdenticon/styles.scss
```diff
@@ -0,0 +1,18 @@
+.KeyIdenticon {
+  --Icon-padding: 0.25rem;
+  --Icon-dimension: 2rem;
+
+  align-items: center;
+  display: flex;
+
+  &--icon {
+    background: var(--pal-info-on);
+    border: 1px solid #d2d5db;
```
quietbits, line 10: Could we pick something from the palette that's close to this color?

extension/src/popup/components/signTransaction/Operations/styles.scss
```diff
@@ -1,5 +1,42 @@
 .Operations {
+  display: flex;
+  flex-direction: column;
+  gap: 2rem;
+
   &--header {
     color: var(----pal-text-primary);
```
quietbits, line 7: Looks like this typo got through from the previous PR. Extra `--` for the var.

quietbits (approved):

PR #357, opened 2022-01-28
popup/views/AccountCreator: create a password page reskin

acharb:
Screen Shot 2022-01-27 at 5 41 28 PM


extension/src/popup/components/PasswordRequirements.tsx
```diff
@@ -1,28 +1,30 @@
...
+  color: var(--pal-text-primary);
+  opacity: 0.8;
   list-style-type: disc;
   list-style-position: inside;
-  font-size: 0.9rem;
+  font-size: 1rem;
   padding: 0;
   margin: 0;
+
+  // TODO reconcile with SDS
```
acharb, line 17: the SDS has different bullets than the figma designs, so went with the designs for now. lmk if I should change

extension/src/popup/views/AccountCreator.tsx
```diff
@@ -76,70 +94,95 @@ export const AccountCreator = () => {
...
+                            </TextLink>
+                          </span>
+                        }
+                        {...field}
+                      />
+                    )}
+                  </Field>
                 </FormRow>
-                <FormRow>
+                {errors.termsOfUse && touched.termsOfUse
```
acharb, line 169: also the checkbox input from SDS doesn't have an `error` prop like the `Input` one does. Should that be added to SDS? 
cc @quietbits @piyalbasu

quietbits: Checkbox component intentionally is built to be flexible. We can't really add an error to every individual checkbox component because they can be in a group, so then we'd want to show an error message for the group.
Maybe we could create a new component `CheckboxWithError` (or a new prop that does this magic internally) or something like that to handle this properly. :thinking: Thoughts?

extension/src/popup/views/AccountCreator.tsx
```diff
@@ -76,70 +94,95 @@ export const AccountCreator = () => {
...
+                  ? errors.termsOfUse
+                  : null}
+                <ButtonRowEl>
                   <SubmitButton
```
quietbits, line 173: Should we use SDS `Button` component here?

extension/src/popup/views/AccountCreator.tsx
```diff
@@ -12,30 +12,47 @@ import {
...
 
 import { Header } from "popup/components/Header";
-import { Onboarding, HalfScreen } from "popup/components/Onboarding";
+import { HalfScreen } from "popup/components/Onboarding";
 import { PasswordRequirements } from "popup/components/PasswordRequirements";
 
-import CreatePasswordIllo from "popup/assets/illo-create-password.svg";
+import { Input, Checkbox, TextLink } from "@stellar/design-system";
+
+const HeaderEl = styled.div`
```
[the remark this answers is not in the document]
acharb: We're doing css from my understanding, I forgot here I'll change that

extension/src/popup/components/PasswordRequirements.tsx
```diff
@@ -1,28 +1,30 @@
...
-import { COLOR_PALETTE } from "popup/constants/styles";
-
 const El = styled.div`
   padding: 1.25rem 0 2.5rem;
 `;
 
 const ListEl = styled.ul`
-  color: ${COLOR_PALETTE.secondaryText};
+  color: var(--pal-text-primary);
+  opacity: 0.8;
```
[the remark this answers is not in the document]
acharb: I took that from the figma designs, I can remove though

extension/src/popup/views/AccountCreator.tsx
acharb: good catch I wasn't sure about that. @piyalbasu or thoughts on changing `SubmitButton` to use the SDS button? since they all should be using SDS from my understanding. I can change in this PR if you agree

extension/src/popup/views/AccountCreator.tsx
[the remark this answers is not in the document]
piyalbasu: @quietbits Yup, the plan is for everything to be moved to CSS before reskin phase is complete.

@acharb That conversion to CSS can be done either in this PR or in a followup PR

extension/src/popup/views/AccountCreator.tsx
piyalbasu: Hmmm interesting - yeah `CheckboxWithError` seems like the way to go. 

There are a couple of other checkboxes that need to be validated in Freighter. I'll keep this in mind as we convert the other ones and come up with a good API for `CheckboxWithError` and add to SDS when I get back from vaca

extension/src/popup/views/AccountCreator.tsx
piyalbasu: @acharb Yeah you should use the SDS Button. In fact, by the end of the reskin, we'll be removing `SubmitButton` completely from the repo as we'll have replaced all instances with the SDS Button.

Instead of making any changes to `SubmitButton`, I would just import `Button` from SDS in this component, similar to what I did here: https://github.com/stellar/freighter/blob/release/2.0.0/extension/src/popup/views/SignTransaction.tsx

extension/src/popup/views/AccountCreator.tsx
acharb: sgtm! I'll add as a TODO and leave for now

extension/src/popup/views/AccountCreator.tsx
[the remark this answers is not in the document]
acharb: I can do here 👍 
https://github.com/stellar/freighter/pull/357/commits/f55ab7c7b76bda5c16402f7457e78b6c58383069

extension/src/popup/views/AccountCreator.tsx
acharb: https://github.com/stellar/freighter/pull/357/commits/8be89e5cda99382b15082fbaefdd954ad22e5006

extension/src/popup/views/AccountCreator/styles.scss
```diff
@@ -0,0 +1,40 @@
...
+
+  &--half-screen {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    padding: 0;
+    padding-top: 2rem;
+    padding-right: 0;
+    padding-bottom: 2rem;
+    padding-left: 1.55rem;
```
piyalbasu, line 37: This can be written shorthand as `padding: 2rem 0 2rem 1.55rem`;

extension/src/popup/views/AccountCreator/styles.scss
```diff
@@ -0,0 +1,40 @@
+.AccountCreator {
+  margin-bottom: 1rem;
+
+  &--header {
+    font-size: 2.125rem;
+    line-height: 1.2rem;
+    font-weight: normal;
```
piyalbasu, line 7: there's actually a var for this `--font-weight-normal`

piyalbasu (approved):
Looking good! Just a couple of small comments before you merge

extension/src/popup/views/AccountCreator/styles.scss
acharb: da2065e

extension/src/popup/views/AccountCreator/styles.scss
acharb: da2065e

PR #360, opened 2022-02-04
popup/views/RecoverAccount: reskin RecoverAccount

acharb:
Screen Shot 2022-02-03 at 4 12 35 PM


extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+              <div className="RecoverAccount__half-screen">
+                <div className="RecoverAccount__header">
+                  Import wallet from recovery phrase
+                </div>
+                <FormRow>
+                  <div className="RecoverAccount__mnemonic-input">
+                    <Field name="mnemonicPhrase">
+                      {({ field }: FieldProps) => (
+                        <>
+                          <textarea
```
acharb, line 94: @quietbits thoughts on this textarea input using the Card class? Should we add to SDS?

quietbits: Agreed that we should add textarea component to SDS. We'll need to check with the Design about styling, though. We haven't used this element anywhere else, so not sure if the style should always be like this. Good workaround in this case! :+1:

quietbits (approved):

Issue #330, opened 2022-01-20
Reskin: Password Confirmation


piyalbasu:
Apply new design and new copy to password confirmation screen

This can be reached by nav'ing directly to: `/#/unlock-account`

In the real world, this is a screen that is reached when your authentication has "timed out" (after 5 min in the current implementation)

PR #362, opened 2022-02-07
reskin UnlockAccount view

acharb:
closes https://github.com/stellar/freighter/issues/330

Screen Shot 2022-02-07 at 2 59 50 PM
Screen Shot 2022-02-07 at 2 59 59 PM


extension/src/popup/views/AccountCreator/index.tsx
```diff
@@ -80,9 +80,10 @@ export const AccountCreator = () => {
                           placeholder="New password"
                           type="password"
                           error={
-                            authError || (errors.password && touched.password)
```
acharb, line 83: also fixing these parentheses in some other onboarding forms

extension/src/popup/views/UnlockAccount/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    await dispatch(confirmPassword(password));
+    navigateTo(destination, queryParams);
+  };
+
+  return (
+    <>
+      <Header />
+      <SubviewWrapper>
+        <div className="UnlockAccount__header">
+          A <strong>Stellar</strong> wallet for <br />
```
quietbits, line 48: Do we need a line break there? It's usually not great to use them with text because it can break "not nicely" on smaller screen sizes. Maybe we could set max-width for the text container to control the text break?

quietbits (approved):
Looks good, just left a question about the line break.

extension/src/popup/views/UnlockAccount/index.tsx
acharb: good point `max-width` works better, changed in some other parts as well
https://github.com/stellar/freighter/pull/362/commits/8e188291b3212a16afafcf145a479ed41df242ec

PR #365, opened 2022-02-15
Lift account details

piyalbasu:
Small PR to lift the `fetchAccountDetails` call into `Account`. This should make it easier to have access to balances in the Payments module


acharb (approved):
lgtm! I was going to get your thoughts on doing this while working on the Account view. So looks like the answer is yes :)

PR #368, opened 2022-02-18
Reskin: account dropdown

acharb:


extension/src/popup/components/account/AccountHeader/styles.scss
```diff
@@ -0,0 +1,134 @@
+$account-row-height: 5rem;
+// TODO - import from views/Account
```
acharb, line 2: question, what would be the best way of sharing constants between .scss files?

piyalbasu (approved):

PR #369, opened 2022-02-22
Reskin/preferences

piyalbasu:
https://user-images.githubusercontent.com/6789586/155219925-5aca2a1d-7f93-446a-b9a3-c232594bcdf8.mov

This isn't the best UI, but I needed to show a little something once Save completed. I'll work with Charles on how to make this look nicer


extension/src/popup/basics/Toggle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+interface ToggleComponent {
+  labelPosition: typeof LabelPosition;
+}
+
+interface ToggleProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  id: string;
+  checked: boolean;
+  customInput?: React.ReactElement;
```
piyalbasu, line 16: @quietbits What do you think about this `customInput` implementation for `Toggle`? I want to be able to pass it a Formik Input (which automatically tracks state change

extension/src/popup/basics/Toggle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  useEffect(() => {
+    setCheckedValue(checked);
+  }, [checked]);
+
+  const additionalClasses = [...(disabled ? ["Toggle--disabled"] : [])].join(
+    " ",
+  );
+
+  const renderLabel = () => (
+    <span>{checked ? labelOn : labelOff ?? labelOn}</span>
```
piyalbasu, line 45: I also made a small change here where you only have to pass it `labelOn` if you don't want the label to change on state change

extension/src/popup/basics/Toggle/styles.scss
```diff
@@ -0,0 +1,70 @@
+.Toggle {
```
piyalbasu, line 1: this is unchanged from SDS

extension/src/popup/basics/Toggle/index.tsx
quietbits: I think it works nicely. :+1:

extension/src/popup/basics/Toggle/index.tsx
quietbits: This is fine if we want to have the label unchanged between the states.

I just noticed that I made a mistake in not rendering the `span` conditionally if there isn't `labelOn` or `labelOff`. Right now it will always render a `span` there. I'll make sure we update it when we add this change to SDS.

extension/src/popup/Router.tsx
```diff
@@ -243,7 +243,7 @@ export const Router = () => {
         <PublicKeyRoute path={ROUTES.unlockBackupPhrase}>
           <UnlockBackupPhrase />
         </PublicKeyRoute>
-        <PublicKeyRoute path={ROUTES.settings}>
+        <PublicKeyRoute path={ROUTES.settings} exact>
```
piyalbasu, line 246: @acharb I found that this is needed if you want to nav to the parent of a nested route (in this case, nav'ing to `/settings/` rather than `/settings/preferences`)

extension/src/popup/basics/PopupWrapper/styles.scss
```diff
@@ -0,0 +1,5 @@
+.PopupWrapper {
+  max-height: 512px; // screen height minus nav height
```
acharb, line 2: 💭  we could probably just use the `--popup--height` var you added in `:root`?

extension/src/popup/styles.scss
```diff
@@ -1,3 +1,7 @@
 :root {
   --z-index--settings: 10;
+
+  // popup dimensions
+  --popup--width: 456px;
```
acharb, line 5: ❓ where is 456 from? in figma it says 360px?

extension/src/popup/basics/PopupWrapper/styles.scss
```diff
@@ -0,0 +1,5 @@
...
+  overflow-y: scroll;
```
acharb, line 3: 💭 thoughts on adding `max-width: var(--popup--width)` as well?

acharb (approved):
lgtm! had a couple thoughts/questions, but non-blocking

extension/src/popup/styles.scss
piyalbasu: Ah good catch - `456` is the existing width of the popup in v1. I'll change to `360`

extension/src/popup/basics/PopupWrapper/styles.scss
piyalbasu: Good call, that's better

PR #372, opened 2022-02-25
payment flow skeleton

acharb:
Just the general structure of the payment flow without additional features (like tx fee recommendation, recent destinations etc) and styling. 
Want to make sure the below foundational parts look good before adding on top (in order of importance).
- background script and duck additions
- routes
- transaction input state management

https://user-images.githubusercontent.com/30449853/155661967-0532763f-dc87-41cb-8970-8e2d9d1f645e.mov


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -544,6 +544,23 @@ export const popupMessageListener = (request: Request) => {
     }
   };
 
+  const signTransactionXDR = () => {
```
acharb, line 547: open to naming suggestions. Trying to figure the best way to differentiate this method name from the `signTransaction` above, even though they're signing but differently

extension/src/popup/basics/Buttons.tsx
```diff
@@ -94,8 +95,12 @@ const BackButtonEl = styled(BasicButton)`
...
-  <BackButtonEl onClick={onClick} {...props}>
-    <Icon.ArrowLeft /> BACK
+export const BackButton = ({
+  onClick,
+  isPopup = false,
+  ...props
+}: BackButtonProps) => (
+  <BackButtonEl onClick={onClick} isPopup={isPopup} {...props}>
+    <Icon.ArrowLeft /> {!isPopup && "BACK"}
   </BackButtonEl>
```
piyalbasu, line 105: This is fine for this iteration. I'm working on some better reusable components, including Back buttons, etc. that we can use when we start adding styles

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+          publicKey,
+          networkDetails,
+        });
+        setAccountBalances(res);
+      } catch (e) {
+        console.error(e);
+      }
+    };
+    fetchAccountBalances();
+  }, [publicKey, networkDetails]);
```
piyalbasu, line 48: Since we fetch fetchAccountBalances in `Account/index.tsx`, and we always end up on this View from Account, we can pass those fetched balances if use <Link state>:

https://v5.reactrouter.com/web/api/Link

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </PublicKeyRoute>
+        <PublicKeyRoute exact path={ROUTES.sendPaymentSettings}>
+          <SendSettings
+            transactionFee={transactionFee}
+            setTransactionFee={setTransactionFee}
+            memo={memo}
+            setMemo={setMemo}
+          />
+        </PublicKeyRoute>
+        <PublicKeyRoute exact path={ROUTES.sendPaymentConfirm}>
```
piyalbasu, line 74: This route actually needs to be wrapped in `PrivateKeyRoute`. My private key had "expired" in background and I tried to dispatch `signTransactionXDR`. In `signTransactionXDR`, it couldn't find my private key, was unable to sign, and returned an empty string as the signedTransaction (instead of throwing the "Session timed out" error)

Not necessary for this particular PR, but in the next version, we should surface this error

extension/src/popup/components/sendPayment/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+import { navigateTo } from "popup/helpers/navigate";
+import { ROUTES } from "popup/constants/routes";
+
+import { PopupWrapper } from "popup/basics/PopupWrapper";
+
+import { BackButton } from "popup/basics/Buttons";
+
+import "./styles.scss";
+
+interface CodeBalanceInterface {
```
piyalbasu, line 21: anyway we can make this interface name a bit more descriptive?

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: hmmmmm how about something like: `signUserTransactionXDR`?

I think we want it to be clear that Freighter is the one generating this XDR and it's not coming from a client website

I'm gonna create a ticket to change the name of `SignTransaction` to something like `SignClientTransaction` to make it more clear

acharb: what about `signFreighterTransaction` vs `signClientTransaction` ?

extension/src/popup/views/SendPayment/index.tsx
acharb: til!
https://github.com/stellar/freighter/pull/372/commits/26cd683656a1e493a5f9d7a1c45bba85ab708d9e

acharb: note: had to stringify it or else React non-serialized data error pops up because of BigNumber balances

extension/src/popup/views/SendPayment/index.tsx
acharb: oh nice I had a mental note to look into that. I changed them all to be PrivateRoute, I think it would be weird to go through the whole flow only at the confirm step need to re-sign in and go through the flow again
https://github.com/stellar/freighter/pull/372/commits/7ae726e432ad5f766c04fe5ab0296f42aa835749

there's also a weird UX where when you re-sign in you go to the `/sendPayment` page and it's missing the `accountBalances` data so you have to go back to `/account` and then `/sendPayment`. I'll fix that on next iteration

extension/src/popup/components/sendPayment/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+    Array<CodeBalanceInterface>
+  >([]);
+  const [selectedAsset, setSelectedAsset] = useState({
+    code: "",
+    balance: "0",
+    canonical: "",
+  });
+
+  useEffect(() => {
+    const filtered: Array<CodeBalanceInterface> = [];
```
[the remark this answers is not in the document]
acharb: ya  I thought it was messier when I tried only looping through once before, but agreed it is simpler now using your method
https://github.com/stellar/freighter/pull/372/commits/a34a6c5e7a893fdbd95ffe13c7fb8a01deb8fc01

having a dynamic key field name in the [wallet sdk for Balances](https://github.com/stellar/js-stellar-wallets/blob/ea375ff473f7a69ad7851777e164a4e790d3fa9c/src/types/data.ts#L160) is pretty annoying/confusing when looping through imo ( have to do `Object.entries` and then `.map`). Maybe something we'd want to change in the new wallet-sdk cc @quietbits

extension/src/popup/components/sendPayment/index.tsx
acharb: was able to remove after this change 
https://github.com/stellar/freighter/pull/372/commits/a34a6c5e7a893fdbd95ffe13c7fb8a01deb8fc01

extension/src/background/messageListener/popupMessageListener.ts
acharb: I changed it to `signFreighterTransaction`. If you hate it though lmk 🙂
https://github.com/stellar/freighter/pull/372/commits/dd01ed99eb967aabe294f288742bd9c7734b22db

piyalbasu (approved):

PR #373, opened 2022-03-01
Reskin/security

piyalbasu:
Screen Shot 2022-03-01 at 3 03 37 PM
Screen Shot 2022-03-01 at 3 03 55 PM
Screen Shot 2022-03-01 at 3 04 07 PM


extension/package.json
```diff
@@ -51,10 +51,10 @@
     "prop-types": "^15.7.2",
     "punycode": "^2.1.1",
     "qrcode.react": "^1.0.0",
-    "react": "^16.12.0",
+    "react": "^17.0.2",
     "react-copy-to-clipboard": "^5.0.2",
-    "react-dom": "^16.12.0",
-    "react-redux": "^7.2.0",
+    "react-dom": "^17.0.2",
+    "react-redux": "^7.2.6",
```
piyalbasu, line 57: updating React versions to be up to date with SDS

extension/src/popup/basics/PopupWrapper/styles.scss
```diff
@@ -1,6 +1,6 @@
 .PopupWrapper {
   max-height: calc(var(--popup--height) - var(--bottom-nav--height));
   max-width: var(--popup--width);
-  overflow-y: scroll;
+  overflow-y: overlay;
```
acharb, line 4: ❓ noob css-er question: was trying to see what `overlay` does and didn't see it on the [mozilla css docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y), will all modern browsers support?

acharb (approved):
lgtm! had 1 non-blocking ❓

extension/src/popup/basics/PopupWrapper/styles.scss
piyalbasu: Ah, good catch - I didn't mean to include this. This is only supported in Chrome. It makes the scrollbar sit on top of the content rather than appearing off to the side (as it usually does). `Overlay` is the first one
Screen Shot 2022-03-01 at 4 43 00 PM
Screen Shot 2022-03-01 at 4 43 20 PM

PR #375, opened 2022-03-02
Reskin: receive funds/qr code page

acharb:
https://user-images.githubusercontent.com/30449853/156270064-3ae3a075-fd70-42a7-a6d3-2bb39f6061bc.mov


extension/src/popup/views/ViewPublicKey/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+              height: "10rem",
+            }}
+          />
+        </div>
+        <div className="ViewPublicKey__address-copy-label">Wallet Address</div>
+        <div className="ViewPublicKey__address-copy">
+          {truncatedPublicKey(publicKey)}
+        </div>
+        <div className="ViewPublicKey__copy-btn">
+          <CopyToClipboard text={publicKey} onCopy={() => handleCopy()}>
```
piyalbasu, line 117: Could you replace this with `CopyText` from SDS?
https://design-system.stellar.org/component/copyTexts

extension/src/popup/views/ViewPublicKey/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+              <div>{accountName}</div>
+              <button
+                className="ViewPublicKey--account-name-btn"
+                onClick={() => setIsEditingName(true)}
+              >
+                <Icon.Edit />
+              </button>
+            </>
+          )}
+        </div>
```
piyalbasu, line 101: Can you make the `Back` arrow lineup with the Account name and `Edit/Check` button so they're all in a line?

piyalbasu: Also, pretty NIT: but I noticed when the Account Name isn't being edited, it's center aligned, but when you click Edit, the Account Name is left aligned in the input box. Can you make it centered in the input box, please? I think it'll make the experience a little less jumpy when moving between Edit and Save

extension/src/popup/views/ViewPublicKey/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  return (
+    <PopupWrapper>
+      <div className="ViewPublicKey">
+        <BackButton isPopup onClick={() => navigateTo(ROUTES.account)} />
+        <div className="ViewPublicKey__header">
+          {isEditingName ? (
+            <Formik initialValues={initialValues} onSubmit={handleSubmit}>
+              <Form className="ViewPublicKey__form">
```
piyalbasu, line 71: I would actually replace this `Form` with `import { Form } from "formik"`;

That way, you won't have to overwrite any styles with `!important` in `&__form`

extension/src/popup/views/ViewPublicKey/index.tsx
acharb: did not know I could use a button in there, awesome 😄 
https://github.com/stellar/freighter/pull/375/commits/0055fde0828f27f96d5c255f99c459cd15ba8a0e

extension/src/popup/views/ViewPublicKey/index.tsx
acharb: >Can you make the Back arrow lineup with the Account name and Edit/Save button so they're all in a line?

I was gonna wait for the[ better BackButton](https://github.com/stellar/freighter/pull/372#discussion_r815024523) you're adding before making sure it looked exactly right. What about a TODO for when that component's added? (or has it already been added?)

>Can you make it centered in the input box, please? I think it'll make the experience a little less jumpy when moving between Edit and Save

yup I agree. I also changed the display height to match the input which I think makes less jumpy as well. Take a look if you get a sec
https://github.com/stellar/freighter/pull/375/commits/7c2650980c6c57954caa70031921affd8a93513e

extension/src/popup/views/ViewPublicKey/index.tsx
acharb: 👍  https://github.com/stellar/freighter/pull/375/commits/58945962b1510042c937b5bca05a73fc89b0494e

extension/src/popup/views/ViewPublicKey/index.tsx
acharb: caused some styling errors fixed here: https://github.com/stellar/freighter/pull/375/commits/58945962b1510042c937b5bca05a73fc89b0494e

piyalbasu: re: better backButton

Good call - we can wait on this. Let's add a TODO so we don't forget :)

piyalbasu (approved):

PR #376, opened 2022-03-02
finish Welcome and Header reskin

piyalbasu:


extension/src/popup/views/Welcome/styles.scss
```diff
@@ -4,4 +4,61 @@
     justify-content: space-around;
     padding-top: 1rem;
   }
+
+  &--column {
```
acharb, line 8: I started using [BEM naming conventions](http://getbem.com/naming/) that Iveta pointed me to, where these would be elements (`__` prefix) instead of modifiers (`--` prefix). Thoughts on starting to use these? (even though the project wouldn't be 100% consistent yet 😅 )

acharb (approved):
lgtm! 1 thought, non-blocking

extension/src/popup/views/Welcome/styles.scss
piyalbasu: Ah, you're right. I'll make that change here and start fixing this going forward 💯

PR #377, opened 2022-03-02
Reskin not funded view

acharb:
Screen Shot 2022-03-02 at 3 14 36 PM
Screen Shot 2022-03-02 at 3 14 23 PM


extension/src/popup/views/Account/styles.scss
```diff
@@ -37,16 +37,20 @@
     display: flex;
     align-items: center;
     justify-content: center;
+  }
 
-    div {
-      width: 1rem;
-      height: 1rem;
+  &__qr-icon svg {
```
acharb, line 42: also fixed these icon colorings to be white

piyalbasu (approved):

extension/src/popup/views/Account/index.tsx
```diff
@@ -175,12 +187,6 @@ export const Account = () => {
           />
         )}
       </div>
-      <div>
-        {/* TODO - handle click */}
-        <Button fullWidth variant={Button.variant.tertiary}>
-          Manage Assets
-        </Button>
```
piyalbasu, line 182: Should we be removing this?

piyalbasu: My understanding is we're still adding the Trustlines module, unless I missed this

acharb: I kept it but just moved it a couple lines up 😄  (to be in the `isFunded` view since it was showing up in the `nonFunded` view)

piyalbasu: Ah, lol - my bad. 👍

PR #378, opened 2022-03-03
Reskin/account creator css

piyalbasu:


extension/src/popup/basics/Forms/index.tsx
```diff
@@ -213,3 +209,26 @@ export const RadioField = ({
...
+export const SubmitButtonWrapper = ({ children }: SubmitButtonWrapperProps) => (
+  <div className="SubmitButtonWrapper">{children}</div>
+);
+
+interface ErrorMessageProps {
+  children: React.ReactNode;
+}
+export const FormError = ({ children }: ErrorMessageProps) => (
+  <div className="FormError">{children}</div>
+);
```
piyalbasu, line 234: this is the start of adding useful reusable components for the new design. Not perfect yet, but the idea is we'd keep adding css styled components here and slowly deleting the styled components from the old design up above

extension/src/popup/App.tsx
```diff
@@ -17,6 +17,7 @@ import { reducer as settings } from "popup/ducks/settings";
 import { Router } from "./Router";
 
 import "./styles/global.scss";
+import "./styles/fullScreen.scss";
```
acharb, line 20: 💭 thoughts on importing in the FullscreenStyle.ts file? (I _think_ that would work) To keep the style coupled with the component

extension/src/popup/basics/BackButton/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Icon } from "@stellar/design-system";
+import { useHistory } from "react-router-dom";
+
+import "./styles.scss";
+
+interface BackButtonProps {
+  hasBackCopy?: boolean;
+}
+
+export const BackButton = ({ hasBackCopy }: BackButtonProps) => {
```
acharb, line 11: 🙌 
❓ can we remove the old BackButton in basics/Buttons now?

extension/src/popup/views/AccountCreator/index.tsx
```diff
@@ -131,12 +131,12 @@ export const AccountCreator = () => {
...
                       )}
                     </Field>
-                  </FormRow>
+                  </div>
                   {/* TODO - add error to Checkbox in SDS */}
-                  {errors.termsOfUse && touched.termsOfUse
-                    ? errors.termsOfUse
-                    : null}
-                  <div className="AccountCreator__button-row">
+                  {touched.termsOfUse ? (
```
acharb, line 136: ❓  are we still adding the error field to the Checkbox in SDS? Or does adding a <FormError> work (in which case can you remove the TODO a couple lines up?)

acharb (commented):
💯  had a thought and some Qs

extension/src/popup/basics/BackButton/index.tsx
piyalbasu: We still have a *couple* more instances of the old one being used. My plan is to systematically go through all the views converting things, and once all the instances of the old Back Button (and old components, in general) are all replaced, I'll be deleting them!

extension/src/popup/App.tsx
piyalbasu: Good call - that's better!

extension/src/popup/views/AccountCreator/index.tsx
piyalbasu: I would still like to move this to SDS. We needed <FormError> in a couple of other non-checkbox places, so I figured I should keep this aligned until we're able to make that change in SDS

acharb (approved):
lgtm! pending the FullScreen style import move

PR #382, opened 2022-03-04
Reskin/mnemonic phrase css

piyalbasu:


extension/src/popup/components/Onboarding/styles.scss
```diff
@@ -2,3 +2,25 @@
...
+  max-width: 57rem;
+  width: 100%;
+  margin: auto;
+}
+
+.OnboardingHalfScreen {
+  display: flex;
+  padding: 2rem 0 2rem 1.55rem;
+  width: 27rem;
+}
```
piyalbasu, line 26: Adding these allows us to omit some duplicative styles and makes sure all the Onboarding screen stay consistent

extension/src/popup/components/Onboarding/index.tsx
```diff
@@ -7,18 +7,67 @@ import "./styles.scss";
 
 export const Onboarding = ({
   goBack,
+  hasGoBackBtn,
   children,
 }: {
   goBack?: () => void;
```
acharb, line 13: 💭  could we add a comment here saying something like 
`// if null defaults to the BackButton goback`

took me a minute to figure out that was happening (and why both a hasGoBackBtn and a goBack var are needed)

acharb (approved):
lgtm! had 1 thought

extension/src/popup/components/Onboarding/index.tsx
piyalbasu: Good callout. This is not clear. I'll fix this

acharb (approved):
lgtm!

PR #386, opened 2022-03-08
changes to payment skeleton

acharb:


extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <div className="SendTo">
+        <div className="header">Send To</div>
+        <input
+          className="SendTo__input"
+          value={destination}
+          onChange={(e: React.ChangeEvent<any>) =>
+            setDestination(e.target.value)
+          }
+        />
+        {/* TODO - use form validation */}
```
acharb, line 84: I played around with this for a bit but couldn't get it working right. Left these ugly `&&` statements for now but will fix when adding the styling

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+  fail = "fail",
+  view = "view",
+}
+const initialState = {
+  state: SUBMIT_STATE.idle,
+  response: null,
+  error: null,
+};
+
+const reducer = (
```
acharb, line 39: @piyalbasu thoughts on this? I didn't think the submission state stuff should be in redux, but having two different dispatch methods in the same component does feel a bit odd. But too complicated for just a useState

piyalbasu: Hmmmm so I totally understand your apprehension in adding submission state to redux. It feels a bit weird to have UI state stored in memory, but it actually is often easier to store it there because we're often funneling async calls through redux. Since submission state is tied to the state of the async call, it's better to have the thing that monitors the async call (redux) also monitor the submission state. In fact, the redux lib we use in Freighter, redux-toolkit, standardizes this as an approach using built in pending/fulfilled/rejected states: https://redux-toolkit.js.org/api/createAsyncThunk#parameters

I would recommend creating a new `ducks` file called `internalTransaction` (or something along those lines), that has your existing `signFreighterTransaction` as well as a new `submitFreighterTransaction` action. You can use the pending/fulfilled/rejected cases to set submission state. Then, here in the UI, all you have to do is alter the UI based on changes in the Redux state. This new `duck` would be useful in the future when we want to sign/submit a different kind of transaction; we'd likely be able to reuse some of the logic here.

Take a look at https://github.com/stellar/freighter/blob/release/2.0.0/extension/src/popup/ducks/accountServices.ts and you can see an example of how to do this and also how something like `applicationState` can change based on the outcome of an async call. (We're not using `pending` in this file, but it's the same concept)

extension/src/popup/components/sendPayment/SendTo/index.tsx
piyalbasu: No prob - we can use Formik validators to clean this up a little in the next iteration

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
piyalbasu: Here's a simpler example of using redux toolkit to monitor async call status: https://github.com/stellar/account-viewer-v2/blob/master/src/ducks/account.ts

acharb: 👍 , changed to use redux lmk what you think
https://github.com/stellar/freighter/pull/386/commits/d0556addb5b71cac1acac38916bdf786713d156c

@shared/api/internal.ts
```diff
@@ -367,17 +367,23 @@ export const signFreighterTransaction = async ({
...
 }): Promise<{ signedTransaction: string }> => {
-  try {
-    const { signedTransaction } = await sendMessageToBackground({
-      transactionXDR,
-      network,
-      type: SERVICE_TYPES.SIGN_FREIGHTER_TRANSACTION,
-    });
-    return { signedTransaction };
-  } catch (e) {
-    console.error(e);
```
acharb, line 378: @piyalbasu note: I switched these to not catch the errors here. They were getting caught too early and wouldn't trigger the thunk reject

piyalbasu: Cool - no problem. The try/catch's in the functions in this file are a little redundant anyway. Errors will be caught in the duck

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+  amount: string;
+  asset: string;
+  destination: string;
+  transactionFee: string;
+  memo: string;
+}) => {
+  const dispatch: AppDispatch = useDispatch();
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const submission = useSelector(transactionSubmissionSelector);
+  const [isViewOnly, setIsViewOnly] = useState(false);
```
piyalbasu, line 43: Hm, can we call this `isViewingDetails`? I was a little confused by what this meant

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+      await dispatch(
+        submitFreighterTransaction({
+          signedXDR,
+          networkUrl: networkDetails.networkUrl,
+        }),
+      );
+    }
+  };
+
+  const TransactionDetails = () => (
```
piyalbasu, line 102: Seems like TransactionDetails and handleSend could live in their own file? Might be nice to have this file just handle showing the right view

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
acharb: I changed it to `isSendComplete`, I think having "view" in there makes it confusing since you're viewing either way. If you think of a better name though lmk and I can change it

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
acharb: good call
https://github.com/stellar/freighter/pull/386/commits/4fe51d025f6b3c2ee1cd958f10b8fad3252d6b7b

PR #396, opened 2022-03-11
sendPayment: lift transactionData to redux

acharb:
makes form data management easier in the sendPayment sub-components


extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -13,15 +13,19 @@ import { SendSettingsFee } from "popup/components/sendPayment/SendSettings/Trans
...
-  // keep state separate for now, combine later if needed
-  const [amount, setAmount] = useState("");
-  const [asset, setAsset] = useState("native");
-  const [destination, setDestination] = useState("");
-  // TODO - use lumens instead of stroops
-  const [transactionFee, setTransactionFee] = useState("100");
-  const [memo, setMemo] = useState("");
+  const transactionData = useSelector(transactionDataSelector);
+
+  // TODO - load from redux in the child components
```
acharb, line 21: i'll remove these as I finish each sendPayment sub-component

extension/src/popup/ducks/transactionData.ts
```diff
@@ -0,0 +1,36 @@
...
+    // TODO - add for each field
+  },
+});
+
+export const { saveDestination } = transactionDataSlice.actions;
+export const { reducer } = transactionDataSlice;
+
+export const transactionDataSelector = (state: {
+  transactionData: TransactionData;
+}) => state.transactionData;
```
piyalbasu, line 36: Hm, what do you think of coupling this with the `transactionSubmissionSlice`? They're tied together as we should only have one transaction in process at a time (ideally). 

for ex:
```
initalState = {
status,
data: {amount, asset, etc. },
xdr,
}
```

Once the the transaction has terminated, you could clear the whole slice to reset.

acharb: I like that idea a lot, moved here:
https://github.com/stellar/freighter/pull/396/commits/bb0ff06a4a5af591fd1bbe335818743b4be73b48

piyalbasu (approved):

PR #397, opened 2022-03-14
Reskin/account history

piyalbasu:
https://user-images.githubusercontent.com/6789586/158246209-2ed1c2c9-9bdb-463b-a24c-59dd2ccb1e41.mov


extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+      operations.forEach((operation) => {
+        if (isPayment(operation.type)) {
+          if (operation.source_account === publicKey) {
+            segments[SELECTOR_OPTIONS.SENT].push(operation);
+          }
+          if (operation.to === publicKey) {
+            segments[SELECTOR_OPTIONS.RECEIVED].push(operation);
+          }
+        }
+      });
```
piyalbasu, line 151: we'll add Claimable Balances here when the time comes

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+            >
+              {option}
+            </div>
+          ))}
+        </div>
+        <ul className="AccountHistory__list">
+          {historySegments[SELECTOR_OPTIONS[selectedSegment]].map(
+            (operation: HorizonOperation) => (
+              <HistoryItem
+                key={operation.id}
```
acharb, line 189: 💭  thoughts on passing to `HistoryItem` the type of payment? So you don't have to check again in that component.

Also a (silly) edge case was my account had a couple payments to itself. So the icon was a "receive" icon even though it was in the send payment group (which I think makes sense it's in both send and receive)

acharb (commented):
overall lgtm, had 1 💭

extension/src/popup/views/AccountHistory/index.tsx
piyalbasu: I addressed the first part of your comment.

Good point about self payments. There really isn't a good solve for this besides flipping the icon based on view, and that I think adds some unnecessary complexity for a pretty small edge case. I'm gonna pass on that one for now and see if it ends up being an issue in the future!

PR #399, opened 2022-03-17
Reskin/account history detail

piyalbasu:


extension/src/popup/basics/BackButton/index.tsx
```diff
@@ -5,17 +5,25 @@ import { useHistory } from "react-router-dom";
...
+}: BackButtonProps) => {
   const history = useHistory();
 
   return (
     <button
       className={`BackButton ${hasBackCopy ? "BackButton--has-copy" : ""}`}
       onClick={() => {
-        history.goBack();
+        if (customBackAction) {
+          customBackAction();
```
piyalbasu, line 23: adding the custom back prop as discussed with @acharb

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+          headerTitle: isPayment
+            ? `${recipientLabel} ${operationAssetCode}`
+            : operationString,
+          operationText: isPayment
+            ? `${paymentDifference}${amount} ${operationAssetCode}`
+            : operationString,
+          externalUrl: `${url}/op/${id}`,
+          setIsDetailViewShowing,
+        });
+        setIsDetailViewShowing(true);
```
piyalbasu, line 95: I explored ways to implement this for a while (probably too long, tbh). This is a bit circuitous, as we're setting props for `TransactionDetails` here, but actually instantiating the component in a different file (`AccountHistory`).

The benefit here is that we only ever make one call to Horizon to get all operations. Then you can click back and forth between multiple transactions without ever having to make another call for those operations again. We also only need to figure out things like `isRecipient` in one place.

We're going to need to do some work on `TransactionDetail` when we implement Claimable Balances, so if we end up hating how I did this, we can always do some refactoring then.

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -166,7 +101,9 @@ export const AccountHistory = () => {
     fetchAccountHistory();
   }, [publicKey, networkDetails]);
 
-  return (
+  return isDetailViewShowing ? (
+    <TransactionDetail {...detailViewProps} />
```
piyalbasu, line 105: This is where we're controlling show/hide of TransactionDetail and pass the props that we constructed in `HistoryItem`

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+      setNetworkIconUrl(iconUrl);
+    };
+
+    if (assetIssuer) {
+      fetchIconUrl();
+    }
+  }, [assetCode, assetIssuer, isRecipient, networkDetails]);
+
+  return assetIssuer && !networkDomain ? null : (
```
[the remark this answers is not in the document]
piyalbasu: Yup, good question. In the case of a non-XLM payment, it takes a second to hit Horizon and get the `networkDomain` of the asset. This was causing an annoying UX where it would show the rest of the detail data, and then the `networkDomain` would flash in. 

This is a workaround that's basically saying, wait til we have *all* the data before we show anything 

Ideally, in the future, we will be caching not only the asset icon, but its `networkDomain` as well so we won't need to hit Horizon at all. We will already have all the data available to us. The TODO here will take care of that for us.
https://github.com/stellar/freighter/pull/399/files#diff-dacdc62063df120813b20d41a256cb1a25bb61b226443a9c6cbd1225bedc4023R74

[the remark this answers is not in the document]
acharb: got it, thanks for explaining!

PR #400, opened 2022-03-17
sendAmount view of send payment flow

acharb:
https://user-images.githubusercontent.com/30449853/158901150-0407fd4a-4610-44a2-ab54-34a42249d24e.mov


extension/src/popup/App.tsx
```diff
@@ -52,6 +57,11 @@ const RouteWrapperEl = styled.div`
   height: 100%;
 `;
 
+// .isBigNumber() not catching correctly, so checking .isBigNumber
+// property as well
+const isSerializable = (value: any) =>
+  value?.isBigNumber || BigNumber.isBigNumber(value) || isPlain(value);
```
acharb, line 63: this was a weird bug. The balances values returned from the wallet sdk fail the .isBigNumber() method test. The BigNumber source code wasn't clear why it fails. I _think_ the source of it comes from `js-stellar-base` is still using [bignumber.js version 4](https://github.com/stellar/js-stellar-base/blob/13e45406f4d67e2ec488c3247bff32f6037b29df/package.json#L107), and isBigNumber method was [added in version 6](https://github.com/MikeMcl/bignumber.js/blob/master/CHANGELOG.md#600).

Regardless, this workaround shouldn't create any issues

acharb: cc @quietbits for wallet-sdk v2 knowledge

piyalbasu: Ah weird - good to know though!

extension/src/popup/App.tsx
```diff
@@ -60,8 +70,14 @@ const rootReducer = combineReducers({
...
-  middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().prepend(metricsMiddleware<AppState>()),
+
+  middleware: [
+    ...getDefaultMiddleware({
+      serializableCheck: {
+        isSerializable,
+      },
+    }),
+  ].concat(metricsMiddleware<AppState>()),
```
piyalbasu, line 80: Thanks for adding this. Been thinking about this for a while 👍

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -1,93 +1,156 @@
...
     }
   }, [asset, accountBalances]);
 
+  const handleContinue = (values: { amount: string }) => {
+    dispatch(saveAmount(values.amount));
+    dispatch(saveAsset(assetInfo.canonical));
+    navigateTo(ROUTES.sendPaymentSettings);
+  };
+
+  const handleAssetSelect = (e: React.ChangeEvent<any>) => {
```
piyalbasu, line 54: Sorry I didn't catch this before: You can pass the input type to `React.ChangeEvent` instead of `any`

 you can use `React.ChangeEvent<HTMLSelectElement>` here, for example

extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -32,18 +32,42 @@ import {
 
 import "../styles.scss";
 
+export const AccountDoesntExistWarning = () => (
+  <div className="SendTo__info-block">
+    <InfoBlock className="SendTo__info-block">
+      The destination account doesn't exist. Send at least 1 XLM to create
```
piyalbasu, line 38:
```suggestion
      The destination account doesn’t exist. Send at least 1 XLM to create
```
Super NIT: but we try to use smart apostrophes/quotes in text strings so they're not the same as a code '

I'm kind of bad about this, too, so call me out on it if you see me do it!
https://github.com/stellar/product-conventions/blob/master/STYLEGUIDE.md#text

extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -32,18 +32,42 @@ import {
...
+      </TextLink>
+    </InfoBlock>
+  </div>
+);
+
+const InvalidAddressWarning = () => (
+  <div className="SendTo__info-block">
+    <InfoBlock variant={InfoBlock.variant.warning}>
+      <strong>INVALID STELLAR ADDRESS</strong>
+      <p>Addresses are uppercase and begin with letters "G" or "M".</p>
```
piyalbasu, line 56:
```suggestion
      <p>Addresses are uppercase and begin with letters “G“ or “M“.</p>
```

piyalbasu:
NIT: I know you're still working on this, but  wanted to add the loading spinner on Send To doesn't reset properly. Before you enter the public key, it's already spinning. Also, the `continue` is lowercase. 

If you're already aware, you can fix in a followup PR, but just thought I'd bring it up!

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendAmount/index.tsx
acharb: ah cool  https://github.com/stellar/freighter/pull/400/commits/5216670a4a3675872ea41c608f817dbd0fe6d2e2

acharb:
good catch on capital C. For loading the spinner on start I actually thought that's how we'd want it (ie "loading recent contacts") 😅 . But changed it and agreed it does feel like better UX now
https://github.com/stellar/freighter/pull/400/commits/81e1ad433e2f9a390d5f526a24a2e1d6ec0f0029

PR #403, opened 2022-03-22
sendConfirm pt 1

acharb:
includes some pretty miscelaneous changes related to sendConfirm, so pointed them out in the comments.

part 2 will be the Success, Failed, and Pending views

Screen Shot 2022-03-22 at 9 00 46 AM

### DIFF
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -367,11 +367,16 @@ export const signFreighterTransaction = async ({
   transactionXDR: string;
   network: string;
 }): Promise<{ signedTransaction: string }> => {
-  const { signedTransaction } = await sendMessageToBackground({
+  const { signedTransaction, error } = await sendMessageToBackground({
     transactionXDR,
     network,
     type: SERVICE_TYPES.SIGN_FREIGHTER_TRANSACTION,
   });
+
+  if (error || !signedTransaction) {
+    throw new Error(error);
+  }
+
   return { signedTransaction };
 };
 
--- a/extension/src/popup/components/account/AccountAssets/index.tsx
+++ b/extension/src/popup/components/account/AccountAssets/index.tsx
@@ -1,13 +1,17 @@
-import React from "react";
+import React, { useState } from "react";
+import { useSelector } from "react-redux";
+
 import { BigNumber } from "bignumber.js";
 
 import { AssetIcons } from "@shared/api/types";
+import { retryAssetIcon } from "@shared/api/internal";
 
 import StellarLogo from "popup/assets/stellar-logo.png";
+import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
 
 import "./styles.scss";
 
-const AssetIcon = ({
+export const AssetIcon = ({
   assetIcons,
   code,
   issuerKey,
@@ -20,7 +24,7 @@ const AssetIcon = ({
 }) =>
   assetIcons[code] || code === "XLM" ? (
     <img
-      className="AssetWrapper--asset-logo"
+      className="AccountAssets--asset-logo"
       alt={`${code} logo`}
       src={code === "XLM" ? StellarLogo : assetIcons[code] || ""}
       onError={() => {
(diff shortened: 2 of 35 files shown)


extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -20,44 +24,71 @@ const AssetIcon = ({
...
-        </div>
-        <div className="AssetWrapper__copy-right">
-          <div>
-            {new BigNumber(total).toString()} <span>{code}</span>
+}) => {
+  const [assetIcons, setAssetIcons] = useState(inputAssetIcons);
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const [hasIconFetchRetried, setHasIconFetchRetried] = useState(false);
+
+  const retryAssetIconFetch = async ({
```
acharb, line 49: moved the `retryAssetIconFetch` method to inside `AccountAssets`, so that don't have to re-define when using

extension/src/popup/components/sendPayment/SendConfirm/index.tsx
```diff
@@ -6,41 +6,31 @@ import {
...
     if (isSendComplete) {
-      return <TransactionDetails {...transactionDetailsProps} />;
+      return <TransactionDetails isSendComplete />;
     }
     switch (submission.status) {
       case ActionStatus.IDLE:
-        return <TransactionDetails {...transactionDetailsProps} />;
+        return <TransactionDetails />;
       case ActionStatus.PENDING:
-        return <SubmitPending />;
```
acharb, line 37: the "pending" view is now just the loader overlaid the TransactionDetails view

extension/src/popup/views/Account/index.tsx
```diff
@@ -158,13 +134,12 @@ export const Account = () => {
             </div>
           </div>
         </div>
-        <div>
+        <div className="AccountView__assets-wrapper">
```
acharb, line 137: changed the AccountAssets "wrapper" to the component using it so can control the row spacing there

@shared/api/internal.ts
```diff
@@ -367,11 +367,16 @@ export const signFreighterTransaction = async ({
   transactionXDR: string;
   network: string;
 }): Promise<{ signedTransaction: string }> => {
-  const { signedTransaction } = await sendMessageToBackground({
+  const { signedTransaction, error } = await sendMessageToBackground({
```
acharb, line 370: better error handling if signing fails, moves the `transactionSubmission` slice to failed

acharb: TIL learned if the method in background `popupMessageListener` throws it returns the error in the response instead of propagating to the popup methods

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: Good call - it looks like we don't even need `assetIcons` at all in `views/Account/index.tsx`. Should we move all that logic here?

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -1,120 +1,205 @@
...
+          })
+            .addOperation(
+              StellarSdk.Operation.payment({
+                destination,
+                asset: horizonAsset,
+                amount,
+              }),
+            )
+            .addMemo(StellarSdk.Memo.text(memo))
+            .setTimeout(180)
```
piyalbasu, line 97: jc - what does setting a timeout of 180 do here?

extension/src/popup/views/Account/index.tsx
```diff
@@ -158,13 +134,12 @@ export const Account = () => {
...
           {isFunded ? (
             <>
               <AccountAssets
                 sortedBalances={sortedBalances}
                 assetIcons={assetIcons}
```
piyalbasu, line 142: to clarify what I was saying up above: do we need to bother setting `assetIcons` here and passing to `AccountAssets`? Should we just fetch/retry `assetIcons` from within `AccountAssets`?

extension/src/popup/components/account/AccountAssets/index.tsx
acharb: ya  I can do that, do you mind if I do in a next PR? it'll take some refactoring in views/Account

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
acharb: setting the [timebounds](https://developers.stellar.org/docs/glossary/transactions/#time-bounds) for the transaction. (I noticed we aren't using it in all repos atm but according to our docs it's good practice). 180 is what we use in our example docs so I went with that, not opposed to changing it though. wdyt?

piyalbasu: Oh cool - I hadn't caught that in the docs before. I wonder if `180` is the standard and if we should set that as a const in the codebase 🤔 Not something we need to solve now!

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: Yeah, totally.

Actually, if it'll be a bit of a lift, let's ignore for now and focus on Send Payment. We can come back to this later

piyalbasu (approved):

extension/src/popup/components/account/AccountAssets/index.tsx
acharb: sounds good, ya I think in this refactor we'd probably want to go further down the rabbit hole and also refactor `getAssetIcons` to not depend on Balances, I'll create a ticket to do later 👍 
https://github.com/stellar/freighter/issues/404

Mateusz9611: 94951a70c4198e4b9ef10ce0fc5cd195c6c5a2be

PR #407, opened 2022-03-23
sendPayment: send confirm pt2

acharb:


extension/src/popup/components/sendPayment/styles.scss
```diff
@@ -1,4 +1,26 @@
-// TODO - separate style sheet
```
acharb, line 1: I kept the sendPayment styling in one style sheet. imo it made more sense because of the shared classes ( eg. .btn-continue and .header). Open to discussing separating though

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -10,26 +10,25 @@ import { SendSettings } from "popup/components/sendPayment/SendSettings";
 import { SendSettingsFee } from "popup/components/sendPayment/SendSettings/TransactionFee";
 import { SendConfirm } from "popup/components/sendPayment/SendConfirm";
 
-// TODO - enforce can't move to next route data not given
```
acharb, line 13: I changed my mind on this, happy to discuss if needed

extension/src/popup/components/sendPayment/styles.scss
```diff
@@ -1,4 +1,26 @@
...
+:root {
+  --bottom-btn-margin: 24px;
+
+  --continue-btn-height: 2.9rem;
```
acharb, line 4: created a ticket for better bottom button spacing: https://github.com/stellar/freighter/issues/408

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -10,26 +10,25 @@ import { SendSettings } from "popup/components/sendPayment/SendSettings";
...
+      <SendTo previous={ROUTES.account} />
     </PrivateKeyRoute>
     <PrivateKeyRoute exact path={ROUTES.sendPaymentAmount}>
-      <SendAmount />
+      <SendAmount previous={ROUTES.sendPaymentTo} />
     </PrivateKeyRoute>
     <PrivateKeyRoute exact path={ROUTES.sendPaymentSettings}>
-      <SendSettings />
+      <SendSettings previous={ROUTES.sendPaymentAmount} />
     </PrivateKeyRoute>
```
piyalbasu, line 26: I may be missing something, but the Back Button should default to history.goBack (meaning, basically just press back in the browser window). Is that not sufficient? Looks like you're moving linearly through these screens

I see that going back on `sendPaymentConfirm` should take you back 2 screens to `sendPaymentSettings`, so that custom `previous` makes sense to me

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -1,59 +1,140 @@
...
+      return <AccountDoesntExistWarning />;
+    }
+
+    // no trustline
+    if (asset !== Asset.native().toString()) {
+      let found = false;
+      Object.keys(destinationBalances.balances || {}).forEach((key) => {
+        if (asset === key) {
+          found = true;
+        }
```
piyalbasu, line 88: pretty small - but you should be able to use `.some` here. It will return a boolean signifying if an element in the array was found. Will save you having to see set `found` manually:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

extension/src/popup/views/SendPayment/index.tsx
acharb: it's really just for the edge case of if you're in the middle of the sendPayment flow and your session times out so get redirected to `/unlock-account`. So after logging in goBack will take you back to `/unlock-account`

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -135,11 +146,11 @@ export const TransactionDetails = ({
   return (
     <div className="TransactionDetails">
       {submission.status === ActionStatus.PENDING && (
-        <div className="SendConfirm__proccessing">
+        <div className="TransactionDetails__processing">
           <Loader /> <span>Processing transaction</span>
```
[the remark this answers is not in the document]
piyalbasu: And you can add an optional string to `components/Loading` to include `Processing transaction`

extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -32,6 +33,23 @@ import {
...
+) => {
+  if (
+    !isFunded &&
+    (new BigNumber(amount).lt(baseReserve) ||
+      assetID !== Asset.native().toString())
+  ) {
+    return true;
+  }
+  return false;
+};
```
piyalbasu, line 51: can this just be 
```
export const shouldAccountDoesntExistWarning = (
  isFunded: boolean,
  assetID: string,
  amount: string,
) => (
    !isFunded &&
    (new BigNumber(amount).lt(baseReserve) ||
      assetID !== Asset.native().toString())
  )
```

It should still return the boolean values you're looking for

extension/src/popup/components/sendPayment/styles.scss
piyalbasu: A convention i've been following in this repo is to create `basics` for shared things.

So in this case, you might make a folder called `basics/SendTransaction` (assuming these buttons/headers will be reused when we want to Sign/Submit other transactions)

You can take a look at `basics/Forms` as a template. You would probably export something like `export const BtnContinue = () => <div className="SendTransaction__btn-continue`>{children"</div>`.

The danger of just having a class called `btn-continue` rather than something like `SendTransaction__btn-continue` is that you could run into a naming collision.  You can imagine `btn-continue` is  a name that would come up a lot bc we're not very creative thinkers when it comes to class naming lol

extension/src/popup/views/SendPayment/index.tsx
piyalbasu: Yeah, this is a bit of an edge case since the user won't have access to a traditional Back/Forward button. It'd be *nice* to have, but might be a bit hairy to enforce. You can see get an idea of how gnarly it can get by looking at Router.tsx

I'm fine to table this until it becomes a problem

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
acharb: 👍  https://github.com/stellar/freighter/pull/407/commits/f6d1547be86d5c95f6fac60d5deb39b5502702d6

extension/src/popup/components/sendPayment/SendTo/index.tsx
acharb: good call https://github.com/stellar/freighter/pull/407/commits/17f8a597599f23da662c476d79d5380db4f252ed

extension/src/popup/components/sendPayment/styles.scss
acharb: good point on the naming collisions, I changed it to `SendPayment__btn-continue` and `SendPayment__header`. https://github.com/stellar/freighter/pull/407/commits/557f4beef936ce430302c92773834467ebb0a96f

I don't think these should live in `basics` though, I left them in `sendPayment` folder. They're pretty specific to this send payment flow (eg. the header I don't see being used elsewhere and the btn-continue is just a space wrapper)

lmk what you think

piyalbasu: Ok cool, then then I'm fine with this!

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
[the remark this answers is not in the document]
piyalbasu: Oh I see. Yeah, it doesn't darken the bg. Didn't realize this was an overlay. Fine to leave as is!

piyalbasu (approved):

kgaogelomatshwi (approved):

PR #409, opened 2022-03-24
send payment: federation / muxed addressese changes

acharb:
- hide the memo for Muxed Addresses (slack convo [here](https://stellarfoundation.slack.com/archives/C016CPSLVLG/p1648070784448519))
- display the muxed or federation address instead of its G Address
- send transaction to muxed addresses (so needed to update stellar-sdk)

### DIFF
--- a/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
@@ -44,7 +44,14 @@ export const TransactionDetails = ({
   const submission = useSelector(transactionSubmissionSelector);
   const {
     destinationBalances,
-    transactionData: { destination, amount, asset, memo, transactionFee },
+    transactionData: {
+      destination,
+      federationAddress,
+      amount,
+      asset,
+      memo,
+      transactionFee,
+    },
   } = submission;
 
   const transactionHash = submission.response?.hash;
@@ -135,7 +142,9 @@ export const TransactionDetails = ({
         );
 
         if (submitFreighterTransaction.fulfilled.match(submitResp)) {
-          await dispatch(addRecentAddress({ publicKey: destination }));
+          await dispatch(
+            addRecentAddress({ publicKey: federationAddress || destination }),
+          );
         }
       }
     } catch (e) {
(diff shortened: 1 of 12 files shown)


extension/src/popup/views/Account/index.tsx
```diff
@@ -111,26 +110,23 @@ export const Account = () => {
...
           </div>
           <div className="AccountView__send-receive-display">
-            <div
-              className="AccountView__send-receive-button"
-              onClick={() => navigateTo(ROUTES.viewPublicKey)}
-            >
-              <span className="AccountView__qr-icon">
-                <Icon.QrCode />
-              </span>
+            <div className="AccountView__send-receive-button">
```
acharb, line 113: also changed these icons to be NavButtons

extension/src/popup/styles/global.scss
```diff
@@ -35,6 +35,7 @@ a {
 
 #root {
   position: relative;
+  height: 100%;
```
acharb, line 38: the PopupWrapper needed this to be full height (so can center content correctly). Looks like it lost it when we removed [this wrapper](https://github.com/stellar/freighter/pull/405/files#diff-d9ab29cfb513dab891d9a6d041b202d627e8895ec8d8dbaad90514e1eab61f5dL97-L99)

piyalbasu: Good catch! I removed this, looked around, and things *appeared* to be fine. Classic 🤦

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -166,16 +175,24 @@ export const TransactionDetails = ({
...
+          {federationAddress ? (
+            <div>
+              <span>{federationAddress}</span>
+            </div>
+          ) : (
+            <div className="TransactionDetails__identicon">
+              <IdenticonImg publicKey={destination} />
+              <span>{truncatedPublicKey(destination)}</span>
+            </div>
+          )}
```
piyalbasu, line 187: I see that this is very similar to 
```
      {federationAddress ? (
        <div>
          <span>{federationAddress}</span>
        </div>
      ) : (
        <div className="SubmitResult__identicon">
          <IdenticonImg publicKey={destination} />
          <span>{truncatedPublicKey(destination)}</span>
        </div>
      )}
```

We should consolidate this repeated logic into a reusable component. I know the styles aren't exactly the same for each, but there's a good amount of overlap. Looks like it's just adding a border to `SubmitResult`. It'd be nice to reuse this as a I can see needing to use this whenever we're dealing with federation addresses.

acharb: good point, moved it to basics with a new sendPayment folder https://github.com/stellar/freighter/pull/409/commits/5f85e7682491eb14d2ed2f0cf955da1417c5e0b2

piyalbasu (approved):

PR #411, opened 2022-03-26
add path payment send pt1

acharb:
talked with Charles about simplifying the styling for this initial release, which is why the UI looks a little simpler than the figma screens.

https://user-images.githubusercontent.com/30449853/160445993-6fbddff0-25a0-4227-b6c5-d760a21e06de.mov

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
@@ -14,6 +14,11 @@ import {
 
 import "./styles.scss";
 
+enum PAYMENT_TYPES {
+  REGULAR = "REGULAR",
+  PATH_PAYMENT = "PATH_PAYMENT",
+}
+
 interface RadioCheckProps {
   name: string;
   title: string;
@@ -55,11 +60,17 @@ export const SendType = () => {
       </div>
       <div className="SendPayment__header">SendType</div>
       <Formik
-        initialValues={{ isPathPayment: String(destinationAsset !== "") }}
+        initialValues={{
+          paymentType:
+            destinationAsset === ""
+              ? PAYMENT_TYPES.REGULAR
+              : PAYMENT_TYPES.PATH_PAYMENT,
+        }}
         onSubmit={(values) => {
+          // path payment flag is a non empty string in redux destinationAsset
           dispatch(
             saveDestinationAsset(
-              values.isPathPayment === "true" ? "native" : "",
+              values.paymentType === PAYMENT_TYPES.PATH_PAYMENT ? "native" : "",
             ),
           );
         }}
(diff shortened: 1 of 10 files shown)


extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+            ),
+          );
+        }}
+      >
+        <Form className="SendType__form">
+          <AutoSaveFields />
+          <RadioCheck
+            name="isPathPayment"
+            title="Same source and destination asset"
+            value="false"
```
acharb, line 72: til react radio inputs dont like boolean values

extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+      >
+        <Icon.X />
+      </div>
+      <div className="SendPayment__header">SendType</div>
+      <Formik
+        initialValues={{ isPathPayment: String(destinationAsset !== "") }}
+        onSubmit={(values) => {
+          dispatch(
+            saveDestinationAsset(
+              values.isPathPayment === "true" ? "native" : "",
```
acharb, line 62: destinationAsset set to something other than `""` in redux is the flag for using path payment

extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
piyalbasu: Yup! Kinda weird, but they want you to assign a string to it, similar to an `input`

Since `value` is going to be a string (and not a boolean), I think it'd be best not to use `isPathPayment` with `true` or `false` strings. It's a little misleading since you're expecting a boolean, but you're just matching a string.

I think it might be good to just treat it how we would treat it as any other string value. 

Something like:

```
enum PaymentTypes {
  REGULAR_PAYMENT = "REGULAR_PAYMENT", // better naming lol
  PATH_PAYMENT = "PATH_PAYMENT"
}

<RadioCheck
            name="paymentType"
            title="Same source and destination asset"
            value="PaymentTypes.REGULAR_PAYMENT"
```

for ex: https://github.com/stellar/freighter/blob/release/2.0.0/extension/src/popup/views/Preferences/index.tsx#L101

piyalbasu:
Just a thought I had (and you prob just haven't gotten to it yet!) but it'd be nice to disable the `Continue` button while calculating the conversion

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendAmount/SendType/index.tsx
acharb: ya  I tried making it "simple" with the boolean and React said nope lol, makes sense for enum
https://github.com/stellar/freighter/pull/411/commits/97ff8160efae70796e03842de89807b8f1a8677b
(and added the continue btn disable if loading)

PR #412, opened 2022-03-29
Manage assets

piyalbasu:
https://user-images.githubusercontent.com/6789586/160508679-626cf6c3-a28e-4e67-bc34-07d9dc66cda4.mov


extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -99,6 +104,21 @@ export const getDestinationBalances = createAsyncThunk<
...
+>(
+  "auth/getAssetIcons",
+  ({
+    balances,
+    networkDetails,
+  }: {
+    balances: Balances;
+    networkDetails: NetworkDetails;
+  }) => getAssetIconsService({ balances, networkDetails }),
+);
```
piyalbasu, line 120: I'm adding `assetIcons` to the redux store because we always fetch all these icons when we land on `Account`, and we need them for subsequent views like Manage Assets and Send Payment. I figure we can save ourselves reaching back out for toml files by saving the url's in redux for reuse

Ideally, this will hold an asset's home domain, as well, at some point

extension/src/popup/components/manageAssets/TrustlineError/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (balance) {
+      setAssetBalance(
+        `${new BigNumber(balance.available).toString()} ${balance.token.code}`,
+      );
+    }
+  }, [balances, errorAsset]);
+
+  const errorState: TRUSTLINE_ERROR_STATES = error
+    ? mapErrorToErrorState(getResultCode(error))
+    : TRUSTLINE_ERROR_STATES.UNKOWN_ERROR;
```
piyalbasu, line 103: parse the result_code we get back from Horizon and map it to a Error UI

piyalbasu:
The UX is still WIP, so I tried not to go too deep until we finalize some details

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+import "./styles.scss";
+
+export type ManageAssetCurrency = CURRENCY & { domain: string };
+
+interface ManageAssetRowsProps {
+  assetRows: ManageAssetCurrency[];
+  setErrorAsset: (errorAsset: string) => void;
+}
+
+const getCodeIssuerStr = (assetCode: string, assetIssuer: string) =>
```
acharb, line 35: ❓ can we call this `getCanonicalFromAsset` or something similar? there's the opposite func in the [helpers](https://github.com/stellar/freighter/blob/6cae2173abd40ab35d82038ffc30644e69f240ad/extension/src/helpers/stellar.ts#L41) now, could we move this one next to it in that file? (I'm sure we'll re-use this at some point)

I've been using the term `canonical` when referring to `<asset code>:<issuer>`, I've noticed we're not 100% consistent referring to it that way across projects but I think it's a good way to refer to it, and is used in [SEPs](https://github.com/stellar/stellar-protocol/pull/313) (thought not all idt)

extension/src/popup/components/manageAssets/TrustlineError/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { transactionSubmissionSelector } from "popup/ducks/transactionSubmission";
+
+import { getResultCode, RESULT_CODES } from "popup/helpers/parseTransaction";
+
+import { Balances } from "@shared/api/types";
+
+import "./styles.scss";
+
+export enum TRUSTLINE_ERROR_STATES {
+  UNKOWN_ERROR = "UNKNOWN_ERROR",
```
acharb, line 16:
```suggestion
  UNKNOWN_ERROR = "UNKNOWN_ERROR",
```

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -32,10 +40,7 @@ export const submitFreighterTransaction = createAsyncThunk<
   Horizon.TransactionResponse,
   { signedXDR: string; networkUrl: string },
   {
-    rejectValue: {
-      errorMessage: string;
-      response: Horizon.TransactionResponse;
-    };
+    rejectValue: ErrorMessage;
```
acharb, line 43: nice, thanks 😄

extension/src/popup/views/Account/index.tsx
```diff
@@ -72,17 +74,10 @@ export const Account = () => {
     });
```
acharb, line 74: ❓ can you use the new sortBalances helper in `helpers/account.ts` to replace the lines above? (line 68-73)

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { PrivateKeyRoute } from "popup/Router";
+import { ROUTES } from "popup/constants/routes";
+
+export const ManageAssets = () => {
+  const { accountBalances } = useSelector(transactionSubmissionSelector);
+  const [errorAsset, setErrorAsset] = useState("");
+
+  const { balances } = accountBalances;
+
+  if (!balances) {
```
acharb, line 20: nit: should we also disable the ManageAssets button if no assets? otherwise it will just load this view and then redirect back (or not sure if you were thinking of doing something different for next PR)

acharb (commented):
💯  left some comments, mostly little things

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
piyalbasu: Yup, good call

extension/src/popup/views/Account/index.tsx
piyalbasu: Argh, yes, good catch!

extension/src/popup/views/ManageAssets/index.tsx
piyalbasu: We do actually hide the button based on the `isFunded` which, now that I'm thinking about it, means the same thing as `!balances`. I'll create a ticket to address that as `isFunded` shouldn't be needed

acharb (approved):
lgtm!

PR #414, opened 2022-03-31
Fix padding issues

piyalbasu:


extension/public/index.html
```diff
@@ -14,7 +14,7 @@
     </style>
     <title>Freighter</title>
   </head>
-  <body>
+  <body class="dark-mode">
```
piyalbasu, line 17: making sure dark mode stays switched on

extension/src/popup/views/About/styles.scss
```diff
@@ -1,17 +1,27 @@
 .About {
+  display: flex;
+  flex-direction: column;
+  height: 600px;
```
acharb, line 4:
```suggestion
  height: var(--popup--height);
```

piyalbasu: bah, yes - good catch!

PR #417, opened 2022-04-01
Qa/account

piyalbasu:
Screen Shot 2022-04-01 at 2 22 06 PM
Screen Shot 2022-04-01 at 2 34 49 PM

### DIFF
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -14,7 +14,7 @@ import { APPLICATION_STATE } from "../constants/applicationState";
 import { sendMessageToBackground } from "./helpers/extensionMessaging";
 import { getIconUrlFromIssuer } from "./helpers/getIconUrlFromIssuer";
 
-const TRANSACTIONS_LIMIT = 15;
+const TRANSACTIONS_LIMIT = 100;
 
 export const createAccount = async (
   password: string,
--- a/extension/src/popup/components/identicons/KeyIdenticon/index.tsx
+++ b/extension/src/popup/components/identicons/KeyIdenticon/index.tsx
@@ -7,6 +7,10 @@ import { IdenticonImg } from "../IdenticonImg";
 import "./styles.scss";
 
 interface IdenticonWrapperElProps {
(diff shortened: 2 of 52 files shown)


@shared/api/internal.ts
```diff
@@ -14,7 +14,7 @@ import { APPLICATION_STATE } from "../constants/applicationState";
 import { sendMessageToBackground } from "./helpers/extensionMessaging";
 import { getIconUrlFromIssuer } from "./helpers/getIconUrlFromIssuer";
 
-const TRANSACTIONS_LIMIT = 15;
+const TRANSACTIONS_LIMIT = 100;
```
piyalbasu, line 17: we had this limit on because we had limited space in the old UI. Having a long list of transactions shouldn't be an issue now

extension/src/popup/components/identicons/KeyIdenticon/index.tsx
```diff
@@ -27,6 +32,12 @@ export const KeyIdenticon = ({
           "--Icon-dimension": "1.5rem",
         }
       : {}),
+    ...(customSize
+      ? {
+          "--Icon-padding": customSize.padding,
+          "--Icon-dimension": customSize.dimension,
+        }
```
piyalbasu, line 39: this is not great - but we have a one-off version of the Identicon in transaction details.

We can circle back later to see if we have any other sizes and come up with a better API for this

extension/src/popup/styles/global.scss
```diff
@@ -15,9 +15,14 @@
 
   --back--button-dimension: 1rem;
   --back--button-z-index: 2;
+
+  // TODO: migrate to SDS
+  --background-color: #1f212a;
+  --button-bg--tertiary: #303448;
```
piyalbasu, line 21: talked with Charles. Looks like the existing SDS colors weren't quite working. We'll override for now

acharb (approved):
lgtm!

I'll change the send flow button to use the pill button (I think there's just 1)

PR #420, opened 2022-04-04
QA bugs; fix scrollbars; make account sessions 24 hour

piyalbasu:
A significant change here is how scrollable div's are presented:

https://user-images.githubusercontent.com/6789586/161629169-1459d266-83ad-44fc-9aad-17cc8bc2dbe3.mov

### DIFF
--- a/extension/src/popup/components/account/AccountAssets/index.tsx
+++ b/extension/src/popup/components/account/AccountAssets/index.tsx
@@ -1,8 +1,6 @@
 import React, { useState } from "react";
 import { useSelector } from "react-redux";
 import { BigNumber } from "bignumber.js";
-import SimpleBar from "simplebar-react";
-import "simplebar-react/dist/simplebar.min.css";
 
 import { AssetIcons } from "@shared/api/types";
 import { retryAssetIcon } from "@shared/api/internal";
@@ -73,7 +71,7 @@ export const AccountAssets = ({
   };
 
   return (
-    <SimpleBar className="AccountAssets">
+    <>
       {sortedBalances.map(({ token: { issuer, code }, total }) => (
         <div className="AccountAssets__asset" key={code}>
           <div className="AccountAssets__copy-left">
@@ -92,6 +90,6 @@ export const AccountAssets = ({
           </div>
         </div>
       ))}
-    </SimpleBar>
+    </>
   );
 };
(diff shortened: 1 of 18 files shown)


extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -91,17 +92,23 @@ export const AddAsset = ({ setErrorAsset }: AddAssetProps) => {
...
-                      setErrorAsset={setErrorAsset}
-                    />
+                    <div
+                      className="AddAsset__results__rows"
+                      ref={ManageAssetRowsWrapperRef}
+                    >
+                      <ManageAssetRows
+                        assetRows={assetRows}
+                        setErrorAsset={setErrorAsset}
+                        parentRef={ManageAssetRowsWrapperRef}
```
acharb, line 109: 💭  could we just pass in the height as a prop? eg:
```suggestion
                        maxHeight={ManageAssetRowsWrapperRef.current?.clientHeight}
```
it would simplify how `ManageAssetRows` is used (assuming we don't need the parentRef for anything else)

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { useState } from "react";
 import { useSelector } from "react-redux";
-
 import { BigNumber } from "bignumber.js";
+import SimpleBar from "simplebar-react";
```
acharb, line 4: ❓ sgtm to use, I haven't used before though so jc. Is it used for simpler scroll styling?

acharb (approved):
lgtm! had a thought and a question but not critical

extension/src/popup/components/manageAssets/AddAsset/index.tsx
piyalbasu: Good idea - that's more straightforward

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: It does 2 things that are pretty useful for us: 1) it overlays the scrollbar on top of your content without pushing the layout (as a built in div scrollbar does) and 2) it disappears when you're not hovering

These are both good for making the UI match the figma design, regardless of content length. The normal implementation of a div scroll bar just causes so many headaches because it has `display: block` and pushes your content around depending on if a scrollbar is necessary or not

acharb: ahh I see nifty!

Issue #408, opened 2022-03-23
change the send payment bottom button styling to use flex


acharb:
instead of `position: absolute`. We may also want to change for the PopupWrapper and bottomNav as well to be consistent

PR #421, opened 2022-04-04
qa css changes

acharb:
- fix UnlockAccount moving when error shows up
- remove the arrow for path payments UI for this v1 version
- adding icon colors to send payment success / fail
- other misc things like capitalizing or spacing


extension/src/popup/views/UnlockAccount/styles.scss
```diff
@@ -10,7 +14,7 @@
   }
 
   &__import-account {
-    margin-top: 7.2rem;
+    margin-top: auto;
```
acharb, line 17: til using margin auto with flex > `position: absolute`  🕺 

I'll use that for https://github.com/stellar/freighter/issues/408

piyalbasu (approved):

PR #422, opened 2022-04-06
all: use custom freighter button

acharb:
a new custom freighter button that overrides the SDS button css


extension/src/popup/basics/buttons/Button/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      : SDSButton.variant.tertiary;
+  const sds = SDSButton({
+    variant: sdsVariant,
+    fullWidth,
+    children,
+  });
+
+  return (
+    <SDSButton
+      className={`BasicButton BasicButton--${variant} ${sds?.props?.className}`}
```
acharb, line 43: kind of weird, but didn't want to wrap the SDSButton in a div that could mess up styling (eg. like [here](https://github.com/stellar/freighter/blob/126ef9dd2ca89111a7c0f180d3ecbf3b98eb3501/extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss#L69)). So creating the SDS button and then adding styling to it instead. This way if we move back to SDS we can just change the imports again

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/styles.scss
```diff
@@ -13,6 +13,11 @@
     .Card {
       padding: 1rem;
     }
+
+    img {
```
acharb, line 17: also resizing icon images

extension/src/popup/basics/buttons/Button/index.tsx
piyalbasu: This is great! I like this is strategy.

Let's definitely work under the assumption that these will be in SDS sooner rather than later

extension/src/popup/basics/buttons/Button/index.tsx
```diff
@@ -0,0 +1,54 @@
+import React from "react";
```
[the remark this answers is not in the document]
piyalbasu: Good call!

piyalbasu (approved):

PR #423, opened 2022-04-06
Qa/account settings

piyalbasu:


extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -262,8 +262,12 @@ const transactionSubmissionSlice = createSlice({
       state.status = ActionStatus.SUCCESS;
       state.response = action.payload;
     });
+    builder.addCase(getAccountBalances.pending, (state) => {
```
acharb, line 265: ❗  This `status` should be only for submitting the transaction. (`status` might not have been the best name, `submitStatus` maybe is a better description ). The send flow assumes the status == `IDLE` means a new send is happening (Eg. if we set it to success on get account balances the sendConfirm will automatically go to the successful tx view bc it thinks the sending was successful.)

what about a new field in redux `accountBalanceFetchStatus`? maybe kind of wordy but something similar

piyalbasu: Ah got it. Yup, I'll fix this

acharb (approved):
lgtm!

Issue #410, opened 2022-03-24
[SPIKE] getIconUrlFromIssuer should also return homeDomain


piyalbasu:
In the new design, we often need to show the `homeDomain` alongside the asset icon. Getting the `homeDomain` is part of the getIconUrlFromIssuer flow, but it is not cached anywhere because it was never needed before. Make getIconUrlFromIssuer both cache and return `homeDomain`

piyalbasu:
This may not be needed when we switch to using the new wallet-sdk

piyalbasu:
From @leighmcculloch: 

```
We should be cautious displaying the domain from the asset issuers Stellar account home domain field, because anyone can put any domain into their accounts home domain. For example, I could put stellar.org in my personal domain. Or a wallet may decide to put the services domain into the home domain field of all their users.

Do we verify that the Stellar account that has the domain in the home domain field is mentioned as a CURRENCY in the stellar.toml of the domain to confirm there's a two-way relationship?
```

leighmcculloch:
We started defining a protocol for asset issuer home domain verification in [SEP-32](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0032.md). SEP-32 defines a process to verify the two-way relationship and may be useful for Freighter to implement.

Keeeeeeeks:
Closed - has changed significantly since issue was written.

PR #431, opened 2022-04-11
Qa/asset fixes

piyalbasu:
Screen Shot 2022-04-11 at 4 36 18 PM

### DIFF
--- a/extension/src/popup/components/account/AccountAssets/index.tsx
+++ b/extension/src/popup/components/account/AccountAssets/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useSelector } from "react-redux";
 import { BigNumber } from "bignumber.js";
 
@@ -46,6 +46,10 @@ export const AccountAssets = ({
   const [assetIcons, setAssetIcons] = useState(inputAssetIcons);
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const [hasIconFetchRetried, setHasIconFetchRetried] = useState(false);
+
+  useEffect(() => {
+    setAssetIcons(inputAssetIcons);
+  }, [inputAssetIcons]);
 
   const retryAssetIconFetch = async ({
     key,
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -3,8 +3,9 @@ import { useDispatch, useSelector } from "react-redux";
 import debounce from "lodash/debounce";
 import { BigNumber } from "bignumber.js";
 import { useFormik } from "formik";
-
+import { Types } from "@stellar/wallet-sdk";
 import { Select, Icon, Loader } from "@stellar/design-system";
+import StellarSdk from "stellar-sdk";
 
 import { InfoBlock } from "popup/basics/InfoBlock";
 import { Button } from "popup/basics/buttons/Button";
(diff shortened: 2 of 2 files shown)


extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -47,6 +47,10 @@ export const AccountAssets = ({
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const [hasIconFetchRetried, setHasIconFetchRetried] = useState(false);
 
+  useEffect(() => {
+    setAssetIcons(inputAssetIcons);
+  }, [inputAssetIcons]);
+
```
piyalbasu, line 53: this was why USDC icons (for ex) weren't showing up. We needed to reset the state when the passed prop updates

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -65,6 +66,41 @@ const ConversionRate = ({
   </div>
 );
 
+const BalanceOption = ({
```
piyalbasu, line 69: tried to make this pretty standalone as I can see this being used elsewhere later

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -65,6 +66,41 @@ const ConversionRate = ({
...
+    const fetchAssetDomain = async () => {
+      let homeDomain = "";
+      // https://github.com/stellar/freighter/issues/410
+      try {
+        ({ home_domain: homeDomain } = await server.loadAccount(assetIssuer));
+      } catch (e) {
+        console.error(e);
+      }
+
+      setAssetDomain(homeDomain);
```
leighmcculloch, line 89: We should be cautious displaying the domain from the asset issuers Stellar account home domain field, because anyone can put any domain into their accounts home domain. For example, I could put stellar.org in my personal domain. Or a wallet may decide to put the services domain into the home domain field of all their users.

Do we verify that the Stellar account that has the domain in the home domain field is mentioned as a `CURRENCY` in the stellar.toml of the domain to confirm there's a two-way relationship?

We started defining a protocol for this type of asset domain verification in [SEP-32](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0032.md) although it is just a proposal at this stage.

piyalbasu: Thanks @leighmcculloch this is a good callout. We don't do any type of verification now, but issue #410 will be a follow up to clean up this type of home_domain fetching. I'll make sure to add that there

extension/src/popup/components/account/AccountAssets/index.tsx
acharb: ahh weird, I would think it would get re-rendered when redux loads the assetIcons? 🤔  but nice fix!

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -65,6 +66,41 @@ const ConversionRate = ({
...
+    };
+
+    if (balance.token.type !== "native") {
+      fetchAssetDomain();
+    }
+  }, [assetIssuer, server, balance.token.type]);
+
+  return (
+    <option key={key} value={key}>
+      {balance.token.code} &bull; {assetDomain}
```
acharb, line 99: 💭  thoughts on no bullet if no assetDomain?
```suggestion
      {balance.token.code} {assetDomain && "&bull;"} {assetDomain}
```

acharb (approved):
lgtm! had 1 💭

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: Yeah, I thought the same, too. It appears that once `useState`'s default state is set, it doesn't update again until you manually set it. I guess it makes it more deterministic and prevents too many re-renders

PR #437, opened 2022-04-18
Qa/send fixes

piyalbasu:


extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -119,10 +120,38 @@ export const SendAmount = ({ previous }: { previous: ROUTES }) => {
     destinationAsset,
   } = transactionData;
 
+  const { recommendedFee } = useNetworkFees();
+
+  const calculateAvailBalance = useCallback(
```
piyalbasu, line 125: this fixes 3 issues:

1) we were deducting base reserve from non-XLM assets. Now we only do that for XLM
2) we take into account the recommendedFee when showing what your Max available to send is (otherwise, you could click Send Max, try to submit to Horizon, and it would fail bc the fee would put you over your limit)
3) we centralize this "Max Available" logic so in the header it shows the same amount that will be populated into the field when you click "Send Max"

extension/src/popup/components/sendPayment/SendSettings/TransactionFee/index.tsx
```diff
@@ -16,44 +15,13 @@ import {
...
-        } else if (ledgerCapacityUsage > 0.75) {
-          setNetworkCongestion(NetworkCongestion.HIGH);
-        } else {
-          setNetworkCongestion(NetworkCongestion.LOW);
-        }
-      } catch {
-        // use default values
-      }
-    })();
-  }, [networkUrl]);
```
piyalbasu, line 56: moving all this logic into a custom hook so we can use the network congestion/fee info in multiple places

extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -225,6 +225,7 @@ export const SendTo = ({ previous }: { previous: ROUTES }) => {
                       <button
                         onClick={async () => {
                           setIsLoading(true);
+                          setValidatedPubKey(pubKey);
```
piyalbasu, line 228: this fixes an issue where if you clicked a recent address, it would say the account was unfunded because it didn't know it needed to lock up the destination account's balance on Ln 186

acharb (approved):
🔥  lgtm!

PR #438, opened 2022-04-19
ask for password less; UI fixes

piyalbasu:


@shared/api/internal.ts
```diff
@@ -47,13 +47,23 @@ export const fundAccount = async (publicKey: string): Promise<void> => {
...
   let allAccounts = [] as Array<Account>;
+  let hasPrivateKey = false;
 
   try {
-    ({ allAccounts, error, publicKey } = await sendMessageToBackground({
+    ({
+      allAccounts,
+      error,
+      publicKey,
+      hasPrivateKey,
```
piyalbasu, line 65: the biggest change in this PR is starting the private key "timer" whenever a user enters their password.

So, for ex, if a user had to enter their password to import a new account, save the private key so we don't need to show the "unlock account" prompt for the next 24h

piyalbasu: `hasPrivateKey` is the trigger for showing "unlock account" on screens like Manage Assets and Send Payment

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -219,11 +219,14 @@ export const popupMessageListener = (request: Request) => {
 
     store.dispatch(timeoutAccountAccess());
 
+    sessionTimer.startSession({ privateKey: keyPair.privateKey });
```
piyalbasu, line 222: start the session and let the UI know we have by passing back `hasPrivateKey`

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -102,7 +102,7 @@ export const HistoryItem = ({
         <div className="HistoryItem__operation">
           {isPayment ? operationAssetCode : operationString}
           <div className="HistoryItem__date">
-            {recipientLabel} • {date}
+            {isPayment ? `${recipientLabel} \u2022` : null} {date}
```
piyalbasu, line 105: UI fix to only show the `Received/Sent` label if it's a payment. Otherwise, just show the date

extension/src/popup/components/manageAssets/TrustlineError/index.tsx
```diff
@@ -70,9 +82,20 @@ const renderError = (
...
-        </InfoBlock>
+        <>
+          <InfoBlock variant={InfoBlock.variant.error}>
+            <div>
+              <p className="TrustlineError__title">
+                This transaction could not be completed.
+              </p>
+              {resultCodes ? (
+                <p className="TrustlineError__subtitle">
+                  Error code: {resultCodes}
```
piyalbasu, line 93: we're having some issues identifying *what* went wrong when a `change trust` fails. So, for now, we'll just show the `result_codes` from Horizon. In the future, we'll have a custom design for every error

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -65,9 +66,8 @@ export const RecoverAccount = () => {
   };
 
   useEffect(() => {
-    if (publicKey) {
+    if (publicKey && publicKey !== publicKeyRef.current) {
       navigateTo(ROUTES.recoverAccountSuccess);
-      window.close();
```
piyalbasu, line 70: why'd I do this? nav and then close? makes no sense lol

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -65,9 +66,8 @@ export const RecoverAccount = () => {
   };
 
   useEffect(() => {
-    if (publicKey) {
+    if (publicKey && publicKey !== publicKeyRef.current) {
```
piyalbasu, line 69: we can't just check the existence of a public key. We need to make sure it's different than the one we already have

@shared/api/internal.ts
acharb: I thought the `hasPrivateKey` was available to the popup from the[ `auth` field ](https://github.com/stellar/freighter/blob/018c263420c71f3bc2f1e64a2a1cb3f7247c47a3/extension/src/popup/ducks/accountServices.ts#L227)in the redux store. Why does it need to be returned from these methods as well? Is it because it doesn't update in the popup redux after calling these methods (ie addAccount, recoverAccount)?

piyalbasu: Yup, you got it. Previously, addAccount/recoverAccount/etc didn't start the timer. Now that they do, these methods will return a new `hasPrivateKey` value that will be used in the popup redux duck to update the field. For example: [extension/src/popup/ducks/accountServices.ts](https://github.com/stellar/freighter/pull/438/files#diff-3d5d228c5caf387ac8dbfe41239ed0a40dd86588386fb9674839819681053905)

acharb: ah I see, cool thanks for explaining

acharb (approved):
lgtm!

PR #439, opened 2022-04-19
Qa/send sign

piyalbasu:


extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -220,29 +220,31 @@ export const SendTo = ({ previous }: { previous: ROUTES }) => {
...
-                          if (isFederationAddress(pubKey)) {
+                          if (isFederationAddress(address)) {
                             const fedResp = await FederationServer.resolve(
-                              pubKey,
+                              address,
                             );
-                            handleContinue(pubKey, fedResp.account_id);
+                            const publicKey = fedResp.account_id;
+                            setValidatedPubKey(publicKey);
+                            handleContinue(publicKey, address);
```
piyalbasu, line 235: we had a bug (on the old ln 234) where we sent the params in the wrong order: param 1 is supposed to be the public key, while param 2 is supposed to be the federated address. I think it was a little confusing because the var `pubKey` could either be a public key or a federated address. 

I renamed it to something a little more generic (`address`) so it was clear it could be public key *or* federated address. I created this var `publicKey` to make it clear `fedResp.account_id` is actually a public key

extension/src/popup/components/signTransaction/Operations/index.tsx
```diff
@@ -85,6 +92,10 @@ interface TransactionInfoResponse {
   sponsoredId: string;
   trustor: string;
   type: keyof typeof OPERATION_TYPES;
+  value: {
+    type: string;
+    data: Buffer;
+  };
```
piyalbasu, line 98: here we're just adding some possible operation fields that we had missed in the past

extension/src/popup/components/sendPayment/styles.scss
```diff
@@ -35,8 +36,7 @@
     font-size: 0.875rem;
     line-height: 1.375rem;
     color: var(--pal-text-tertiary);
-    position: absolute;
```
acharb, line 38: ahh nice, glad we could get rid of this 😄

acharb (approved):
lgtm!

PR #441, opened 2022-04-21
don't round account history number and trim trailing zeros in detail

piyalbasu:


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -89,7 +89,9 @@ export const HistoryItem = ({
             ? `${recipientLabel} ${operationAssetCode}`
             : "Transaction",
           operationText: isPayment
-            ? `${paymentDifference}${amount} ${operationAssetCode}`
+            ? `${paymentDifference}${new BigNumber(
+                amount,
+              )} ${operationAssetCode}`
```
acharb, line 94: do you need a .toString() here for the BigNumber?

I think you shouldn't but iirc I had to add it under certain scenarios otherwise would get a "react cant render an object" runtime error, so just making sure

acharb (approved):
lgtm!

had 1 question that if looks good you can ignore

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
piyalbasu: Good callout - this didn't cause an issue because I believe a JS template string automatically coerces anything within the `${}` into a string

piyalbasu: It actually looks like `.toString()` wasn't need up above either

PR #449, opened 2022-04-25
Qa/misc fixes

piyalbasu:


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -153,11 +151,12 @@ export const ManageAssetRows = ({
               </div>
               <div className="ManageAssetRows__button">
                 <PillButton
-                  isLoading={isSubmitting}
+                  disabled={isActionPending}
```
piyalbasu, line 154: disable all buttons if there is an action in process

acharb (approved):
lgtm!

PR #451, opened 2022-04-25
update button in both transaction details

piyalbasu:


extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -3,7 +3,6 @@ import { useSelector } from "react-redux";
 import StellarSdk from "stellar-sdk";
 
 import { Button } from "popup/basics/buttons/Button";
-import { PopupWrapper } from "popup/basics/PopupWrapper";
```
piyalbasu, line 6: `PopupWrapper` actually isn't super useful anymore. It was good when we had that bottom nav on a lot of the screens, but now it's just making the placement of the bottom button inconsistent :/

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -250,6 +250,9 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
         </div>
       </div>
       <div className="TransactionDetails__bottom-wrapper">
+        <div className="TransactionDetails__bottom-wrapper__copy">
+          {isPathPayment && "The final amount is approximate and may change"}
+        </div>
```
piyalbasu, line 255: the style [here](https://github.com/stellar/freighter/blob/v2.0.0-beta.3/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/styles.scss#L84) was always expecting this div `TransactionDetails__bottom-wrapper__copy` to be on screen.

In a future update, I'll be removing all instances of `PopupWrapper` and styling these SendPayment bottom buttons in the same manner as I did up above. I'm punting on that for now as this will affect a lot of screens and I don't want to accidentally break the UI at this stage

acharb (approved):
lgtm!

PR #464, opened 2022-04-27
update version and use code splitting

piyalbasu:


extension/webpack.extension.js
```diff
@@ -6,6 +6,12 @@ const prodConfig = (env = { PRODUCTION: false }) => ({
   mode: "production",
   optimization: {
     minimize: env.PRODUCTION,
+    splitChunks: {
+      // ignore non-index.min.js chunk
+      chunks: (chunk) => chunk.name === "index",
+      // Firefox addon store has a max file size of 4mb
+      maxSize: 4000000,
+    },
```
piyalbasu, line 14: we're introducing code splitting because Firefox rejected our generated `index.min.js` file for being >4mb

This is sort of a rudimentary attempt at this just to get it working for this submission. I'll continue to evaluate how we do this. Also, 4mb is HUGE for a js file. Something's up here 🤔

extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "extension",
-  "version": "1.2.0",
+  "version": "2.0.0",
```
acharb, line 3: 🍾

acharb (approved):
lgtm!

PR #482, opened 2022-05-11
add additional tracking for payments and manage assets

piyalbasu:


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -158,6 +160,16 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
             addRecentAddress({ publicKey: federationAddress || destination }),
           );
         }
+
+        if (isPathPayment) {
```
acharb, line 164: 💭  should we add these `emitMetric` lines to w/in the `if (submitFreighterTransaction.fulfilled.match(submitResp)) { ... }` scope? I don't _think_ it matters because I think if submitting fails an error gets thrown and takes you to the catch block, but just to be safe?

acharb (approved):
lgtm! had 1 non blocking thought

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
piyalbasu: Good idea - i'll do that!

PR #483, opened 2022-05-11
adding first attempt at beta pipeline

piyalbasu:
there's like a 99.9999999% chance this fails on the first try 🙃


.github/workflows/submitBeta.yml
```diff
@@ -0,0 +1,127 @@
...
+          file: ./extension/public/static/manifest.json
+          field: version_name
+          value: ${{ github.event.inputs.release }}-beta.${{ github.event.inputs.release }}
+      - name: Update manifest.json version
+        uses: jossef/action-set-json-field@6e6d7e639f24b3955ef682815317b5613ac6ca12 #v1
+        with:
+          file: ./extension/public/static/manifest.json
+          field: version
+          value: ${{ github.event.inputs.release }}
+      - name: Build extension
```
acharb, line 40: ❓ I noticed the `submitProduction.yml` doesn't have this `Build extension for Firefox` command, but submitBeta does, even though looks like both are submitting to firefox, is that on purpose?

.github/workflows/submitBeta.yml
```diff
@@ -0,0 +1,127 @@
...
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./extension/build/build.zip
+          asset_name: build-${{ github.event.inputs.version }}.zip
+          asset_content_type: application/zip
+      - name: Build extension for Firefox
```
acharb, line 82: ❓  I noticed the submitProduction.yml doesn't have this `Build extension for Firefox` job, but submitBeta does, even though looks like both are submitting to firefox, is that on purpose?

.github/workflows/submitBeta.yml
```diff
@@ -0,0 +1,127 @@
...
+        id: firefox-build
+        uses: kewisch/action-web-ext@a28b4ba2f491cc36c67a714a716ca54218440056 #v1
+        with:
+          cmd: build
+          source: ./extension/build
+          filename: ${{ github.event.inputs.release }}-beta.${{ github.event.inputs.release }}.xpi
+      - name: Sign extension for Firefox
```
acharb, line 89: ❓ related to my Q above, is this submitting to firefox? In submitProduction.yml the job name is `Submit extension to Firefox` even though here it says `Sign`, but the jobs look similar?

acharb (commented):
had 2 Qs, mostly because I'm not familiar with these GHAs

.github/workflows/submitBeta.yml
piyalbasu: Ah good questions: I'm realizing now that I should have annotated this PR to explain what's going on!

So, the idea here is that we're going to build an `.xpi` file, submit to Firefox sign off on it, and then host the build artifact ourselves. 

Self-hosting this .xpi file is Mozilla's solution for beta testing as anyone can install this file by drag/dropping it onto their local Firefox. It's *slightly* easier than unzipping a zip file and loading a temporary add on, thought not as good as Chrome's ability to have a private listing

piyalbasu: In submitProduction.yml, we're building one zip file and sending them off to Firefox and Chrome to generate whatever file types they want and host it. Here, we actually need to build the xpi file ourself since we're self-hosting. When Firefox is hosting, they do this step behind the scenes

.github/workflows/submitBeta.yml
piyalbasu: So these are very similar, but the big difference here is `channel: unlisted`. In submitProduction.yml, we're using `channel: listed`.

As I alluded to above, what this means is that submitProduction.yml is just sending our zipped build folder to Firefox to sign, generate, and then host the extension. 

I'm calling this `Sign extension for Firefox` as we're *just* having Firefox sign our .xpi file and then return it to us (used in the Upload Release Asset step). 

These look really similar (and maybe the API shouldn't call both cmds `sign`, but that's on this API's designer 😄  ) but they actually are doing different things for us

acharb:
got it, thanks for explaining!

acharb (approved):
lgtm!

PR #484, opened 2022-05-11
when fetching balances filter out liquidity pool shares

acharb:
wallet-sdk doesn't filter these out, and freighter currently isn't supporting, so filter them out after retrieving using the wallet-sdk.

currently because the shares don't have a `token` field it throws a runtime error in the send flow

closes https://github.com/stellar/freighter/issues/485


piyalbasu (approved):
Good catch!

Issue #494, opened 2022-05-16
cached asset icons use the asset code as the cached object key


acharb:
Let's use the canonical asset representation as the key to fix.

acharb:
closed with https://github.com/stellar/freighter/pull/495

Issue #477, opened 2022-05-11
Path payments starts at 0 and causes "no paths found" error on load


piyalbasu:

acharb:
closed with https://github.com/stellar/freighter/pull/495

PR #495, opened 2022-05-16
use asset canonical ID for key in cache

acharb:
**WHAT**
uses canonical ID when caching the asset image url, instead of the asset code.

**WHY**
so storing assets with same code don't conflict

also, don't show "no path found" if send amount input is 0

closes https://github.com/stellar/freighter/issues/494, https://github.com/stellar/freighter/issues/477

### DIFF
--- a/@shared/api/helpers/getIconUrlFromIssuer.ts
+++ b/@shared/api/helpers/getIconUrlFromIssuer.ts
@@ -42,7 +42,7 @@ export const getIconUrlFromIssuer = async ({
   try {
     /* First, check our localStorage cache in Background to see if we've found this url before */
     ({ iconUrl } = await sendMessageToBackground({
-      assetCode: code,
+      assetCanonical: `${code}:${key}`,
       type: SERVICE_TYPES.GET_CACHED_ASSET_ICON,
     }));
     if (iconUrl) {
@@ -88,7 +88,7 @@ export const getIconUrlFromIssuer = async ({
           iconUrl = image;
           /* And also save into the cache to prevent having to do this process again */
           await sendMessageToBackground({
-            assetCode: code,
+            assetCanonical: `${code}:${key}`,
             iconUrl,
             type: SERVICE_TYPES.CACHE_ASSET_ICON,
           });
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -338,7 +338,7 @@ export const SendAmount = ({ previous }: { previous: ROUTES }) => {
             <div className="SendAmount__input-amount__asset-copy">
               {getAssetFromCanonical(formik.values.asset).code}
             </div>
-            {destinationAsset && (
+            {destinationAsset && formik.values.amount !== "0" && (
               <ConversionRate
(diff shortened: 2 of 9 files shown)


extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -338,7 +338,7 @@ export const SendAmount = ({ previous }: { previous: ROUTES }) => {
             <div className="SendAmount__input-amount__asset-copy">
               {getAssetFromCanonical(formik.values.asset).code}
             </div>
-            {destinationAsset && (
+            {destinationAsset && formik.values.amount !== "0" && (
```
acharb, line 341: just 1 line change for the "no path found" fix

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -205,7 +208,10 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
             assetIcons={assetIcons}
             sortedBalances={[
               {
-                token: { issuer: sourceAsset.issuer, code: sourceAsset.code },
+                token: {
```
acharb, line 211: related fix :), AccountBalances expects the issuer to be nested one more level under the `key` field

@shared/api/helpers/getIconUrlFromIssuer.ts
```diff
@@ -42,7 +42,7 @@ export const getIconUrlFromIssuer = async ({
   try {
     /* First, check our localStorage cache in Background to see if we've found this url before */
     ({ iconUrl } = await sendMessageToBackground({
-      assetCode: code,
+      assetCanonical: `${code}:${key}`,
```
leighmcculloch, line 45: Does the canonical form here handle the case where the code is an ASSET_TYPE_ALPHANUM12 but only contains 3 non-zero bytes?

acharb: could you give an example of what you mean by `only contains 3 non-zero bytes`?

Shaptic: I think it's a reference to this subtlety in SEP-11, https://stellar.org/protocol/sep-11#alphanum4-alphanum12:

> Stellar disallows assets of type ASSET_TYPE_CREDIT_ALPHANUM12 that have fewer than 5 bytes, but such assets can be represented in binary XDR, and so in txrep an AlphaNum12 value with fewer than 5 bytes in its assetCode is rendered with trailing \x00 (escaped NUL bytes) to as to make the length 5---e.g., the 12-byte asset code ABC is rendered ABC\x00\x00. Note that stellar-core disallows non ASCII bytes in AssetCode fields, so the primary use of this feature is to construct or examine invalid transaction test cases.

leighmcculloch: On re-read, I think it is not relevant because of:
>Note that stellar-core disallows non ASCII bytes in AssetCode fields, so the primary use of this feature is to construct or examine invalid transaction test cases.

leighmcculloch (approved):

PR #502, opened 2022-05-19
add custom select inputs

acharb:
https://user-images.githubusercontent.com/30449853/169355634-fdad7ecb-1a50-46b0-9db3-917d958e7cbc.mov

closes https://github.com/stellar/freighter/issues/476

### DIFF
--- a/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from "react";
 import { useSelector } from "react-redux";
 import StellarSdk from "stellar-sdk";
 import { Link } from "react-router-dom";
+import { Icon, Loader } from "@stellar/design-system";
 
 import { Button } from "popup/basics/buttons/Button";
 import { ROUTES } from "popup/constants/routes";
@@ -20,16 +21,25 @@ import "./styles.scss";
 interface ChooseAssetProps {
   balances: Balances;
   setErrorAsset: (errorAsset: string) => void;
+  selectingSourceAsset?: boolean;
+  selectingDestAsset?: boolean;
 }
 
-export const ChooseAsset = ({ balances, setErrorAsset }: ChooseAssetProps) => {
+export const ChooseAsset = ({
+  balances,
+  setErrorAsset,
+  selectingSourceAsset = false,
+  selectingDestAsset = false,
+}: ChooseAssetProps) => {
   const { assetIcons } = useSelector(transactionSubmissionSelector);
   const { networkUrl } = useSelector(settingsNetworkDetailsSelector);
   const [assetRows, setAssetRows] = useState([] as ManageAssetCurrency[]);
   const ManageAssetRowsWrapperRef = useRef<HTMLDivElement>(null);
+  const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
     const fetchDomains = async () => {
+      setIsLoading(true);
       const collection = [] as ManageAssetCurrency[];
       const sortedBalances = sortBalances(balances);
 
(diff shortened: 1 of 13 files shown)


extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from "react";
 import { useSelector } from "react-redux";
 import StellarSdk from "stellar-sdk";
 import { Link } from "react-router-dom";
+import { Icon, Loader } from "@stellar/design-system";
```
acharb, line 5: also the ChooseAsset view w/ a large amount of assets takes a second to load, so I added a loader to it while waiting

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -143,7 +163,23 @@ export const ManageAssetRows = ({
           const isActionPending = submitStatus === ActionStatus.PENDING;
 
           return (
-            <div className="ManageAssetRows__row" key={canonicalAsset}>
+            <div
```
piyalbasu, line 166: I think it's smart to reuse UI elements ManageAssetsRows for this view, but I think it's a bit confusing to have `ManageAssetsRows` be both a view for editing trustlines and a view for selecting an asset for a payment since they're two pretty dissimilar use cases

My suggestion would be to break out pieces of the UI like `ManageAssetRows__content` and `ManageAssetRows__row` into their own `basics` file with a more generic name. (Check out popup/basics/Forms/index.tsx for an example) Then you could create a new `component` that imports the needed UI elements from that `basics` file with only the logic for dealing with asset selection. Wdyt?

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleSelectAsset = () => {
+    navigateTo(
+      ROUTES.manageAssets,
+      `?${ASSET_SELECT.QUERY_PARAM}=${ASSET_SELECT.SOURCE}`,
+    );
+  };
+
+  return (
+    <div className="">
```
piyalbasu, line 35: NIT: empty className

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const truncateLongAssetCode = (code: string) => {
+    if (code.length >= 5) {
+      return `${code.slice(0, 5)}...`;
+    }
+    return code;
+  };
+
+  return (
+    <div className="">
```
piyalbasu, line 90: NIT: empty className

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/styles.scss
```diff
@@ -0,0 +1,73 @@
+.AssetSelect {
+  &__wrapper {
+    height: 3.5rem;
+    border: solid 1px var(--pal-background-secondary);
+    border-radius: 8px;
```
piyalbasu, line 5: .5rem

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: thought it might be slightly over-abstracting at first thought, but true the name ManageAssetRows is really a different use case. I'll do some refactoring here 👍

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
acharb: ah good 👁️  https://github.com/stellar/freighter/pull/502/commits/2365119bdc719840490ac09d8bfb08fcebfbcca5

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
acharb: https://github.com/stellar/freighter/pull/502/commits/2365119bdc719840490ac09d8bfb08fcebfbcca5

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/styles.scss
acharb: https://github.com/stellar/freighter/pull/502/commits/2365119bdc719840490ac09d8bfb08fcebfbcca5

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: added a new component similar to `ManageAssetRows`. I started abstracting the `ManageAssetRows__content` and importing into both but it gets hairy since the row would need access to things like trustlines data for `ManageAssetRows` that the `SelectAssetRows` doesn't use. I think since only used in these two spots this is fine and can re-asses if this UI is used in more spots. Lmk what you think
https://github.com/stellar/freighter/pull/502/commits/dac5c20b3999736d663ae28d41e424bf53729cf0

piyalbasu: Yeah the repeated styles isn't the best, but if it's too messy to share the styles, that's fine. I think having these 2 components that have clearer purposes is a lot easier to parse 👍 

I would at least make 

```
domain
                    ? domain.replace("https://", "").replace("www.", "")
                    : "Stellar Network"
```

into a helper, though.

acharb: good idea https://github.com/stellar/freighter/pull/502/commits/3ede4987a0db420640239f6ece2ff7710c100dc8

piyalbasu: Btw, I think we just need to roll back some of the changes which shouldn't be needed anymore in this file, then we're good to approve! 😄

acharb: oof ya (I blame it being friday 😄 ), changed back
https://github.com/stellar/freighter/pull/502/commits/ee458a1c69b6913a7c00e1cdc0e1eb185682ecfa

piyalbasu (approved):

PR #507, opened 2022-05-23
use mode of max fee instead of fee charged mode

piyalbasu:
Using a slightly higher transaction fee to reduce the number of transactions that timeout during high congestion


leighmcculloch (approved):

acharb (approved):
lgtm!

acharb (approved):
lgtm!

PR #515, opened 2022-05-24
add swap routes, and make SendAmount changes

acharb:
part 1 of adding swap support:
- routes
- sendAmount component changes

https://user-images.githubusercontent.com/30449853/170337196-6ff565cc-2f10-4797-b963-2dce2f1abd83.mov


extension/src/popup/styles/global.scss
```diff
@@ -45,6 +45,10 @@ a {
   height: 100%;
 }
 
+.simplebar-content {
```
acharb, line 48: annoyingly this simple bar inner container doesn't grow to its parents height. I put it in global because I'm pretty sure we'd want this for any simplebar use

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -257,127 +271,155 @@ export const SendAmount = ({ previous }: { previous: ROUTES }) => {
...
+                        balance={formik.values.amount}
+                      />
+                      <PathPayAssetSelect
+                        source={false}
+                        assetCode={
+                          getAssetFromCanonical(formik.values.destinationAsset)
+                            .code
+                        }
+                        issuerKey={
+                          getAssetFromCanonical(formik.values.destinationAsset)
```
piyalbasu, line 393: might be useful to create const's for ` getAssetFromCanonical(formik.values.destinationAsset)` and `getAssetFromCanonical(formik.values.asset)` as we're calling these 2 methods a few time

extension/src/popup/components/sendPayment/styles.scss
```diff
@@ -75,6 +101,10 @@
     width: 100%;
     transition: opacity 0.15s ease-in-out;
 
+    &__full-height {
+      margin-top: 2rem !important;
```
piyalbasu, line 105: jc - why is `important` needed here? Often I find if I have to use this, something can be improved to not need it

extension/src/popup/views/Account/index.tsx
```diff
@@ -57,6 +58,9 @@ export const Account = () => {
   const { balances, isFunded } = accountBalances;
 
   useEffect(() => {
+    // reset to avoid any residual data eg switching between send and swap or
+    // previous stale sends
+    dispatch(resetSubmission());
```
piyalbasu, line 63: Good call. This is a pretty good place for this. Wonder if this will catch everything 🤔 Let's keep an eye on this, too

piyalbasu (approved):

extension/src/popup/components/sendPayment/styles.scss
acharb: it's not 😄 . removed
https://github.com/stellar/freighter/pull/515/commits/fac9066a12c0871f45222808d13e29bd83c16349

extension/src/popup/styles/global.scss
acharb: 🤔  could be ya . I ran through those views w/ this change and didn't see any noticeable differences, but good point will keep an eye on it

PR #517, opened 2022-05-31
add a search asset view

piyalbasu:
Screen Shot 2022-05-31 at 12 24 25 PM

### DIFF
--- a/extension/src/popup/components/manageAssets/SearchAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/SearchAsset/index.tsx
@@ -67,6 +67,12 @@ export const SearchAsset = ({ setErrorAsset }: SearchAssetProps) => {
     </div>
   );
 
+  interface AssetRecord {
+    asset: string;
+    domain?: string;
+    tomlInfo?: { image: string };
+  }
+
   const handleSearch = useCallback(
     debounce(async ({ target: { value: asset } }) => {
       let res;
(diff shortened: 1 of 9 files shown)


extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export const SearchAsset = ({ setErrorAsset }: SearchAssetProps) => {
+  const { isTestnet } = useSelector(settingsNetworkDetailsSelector);
+  const [assetRows, setAssetRows] = useState([] as ManageAssetCurrency[]);
+  const [maxHeight, setMaxHeight] = useState(0);
+  const [isSearching, setIsSearching] = useState(false);
+  const [hasNoResults, setHasNoResults] = useState(false);
+  const ResultsRef = useRef<HTMLDivElement>(null);
+
+  const ResultsHeader = () => (
```
acharb, line 50: nit - any reason `ResultsHeader` needs to be defined w/in SearchAsset? Could we define outside of SearchAsset to be consistent with the other sub-component `AddManualAssetLink` (or both defined w/in, I think outside reads easier though)

acharb (approved):
lots of fun testing out the search asset functionality 🕺 , lgtm! just had 1 nit

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
piyalbasu: Ah yes, good catch!

Issue #519, opened 2022-06-02
Can not override interacting with accounts flagged as malicious


hanseartic:
In the settings there is a toggle that allows to block interaction with malicious domains or accounts. Even when toggled off
![grafik](https://user-images.githubusercontent.com/1752217/171598882-8ebe6c98-b01c-421c-a561-cc06a096d6dd.png)
such interaction is blocked. This prevents legit cases as burning spam claimable balances.

### What version are you using?
`v.2.1.2`

### What did you do?
I tried to burn a claimable balance of an asset issued by account `GBBIIQAP7H6H5WIGUUFTAHJMH7FDHYAEYQN2F2XUUWN5BCTYOYAPPEND` which is marked as malicious on [stellar.expert](https://stellar.expert/explorer/public/account/GBBIIQAP7H6H5WIGUUFTAHJMH7FDHYAEYQN2F2XUUWN5BCTYOYAPPEND) through [stellar claim](https://balances.lumens.space/claim/GAHX7SQFYIGF5HWNDDPYTP2FBD3XY2FIEHRMEHLLN77BGOXZDHLJPEND).

To reproduce either visit https://balances.lumens.space/claim/GA2DVDC4DYS2FPHCYT4DHOYIGNLUWGSXR7XAJ6ANAL4J6TMU24NT6KTZ (a random account that has a pending claimable balance `PENDULUM:GBBIIQAP7H6H5WIGUUFTAHJMH7FDHYAEYQN2F2XUUWN5BCTYOYAPPEND` and try to burn the **Pendulum Chain** balance (click 'Clean selected balances`):
![grafik](https://user-images.githubusercontent.com/1752217/171605090-9c198b83-997f-40c0-aad3-76c6f2af3e89.png)

[...]

PR #524, opened 2022-06-10
freighterApiMessageListener: update the collectedTags array if isValidatingSafety is false

acharb:
right now the array isn't updating, so turning isValidatingSafety to false still shows the warnings

closes https://github.com/stellar/freighter/issues/519


piyalbasu (approved):
Oof - can't believe I did that 🤦

PR #525, opened 2022-06-13
adding default and loading icons

piyalbasu:
https://user-images.githubusercontent.com/6789586/173400161-e0d4de07-4541-4bd0-a383-3b55470b932d.mov


extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -166,10 +166,8 @@ export const SearchAsset = ({ setErrorAsset }: SearchAssetProps) => {
                 ref={ResultsRef}
               >
                 {isSearching ? (
-                  <div className="SearchAsset__overlay">
-                    <div className="SearchAsset__loader">
-                      <Loader />
-                    </div>
```
piyalbasu, line 172: removing the dark bg for the loading spinner

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
```diff
@@ -47,14 +47,21 @@
...
       margin-right: 0.5rem;
     }
 
+    .AccountAssets__asset--error {
+      img {
+        width: 0.4375rem;
+        height: 0.4375rem;
+      }
+    }
+
```
piyalbasu, line 64: I went through and adjusted styles in the Senf/Swap flow that didn't play nice with the new structure. Let me know if anything looks crazy!

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
```diff
@@ -47,14 +47,21 @@
...
 
-    img {
-      width: 1rem;
-      height: 1rem;
+    .AccountAssets__asset--logo {
+      display: flex;
+      width: 0.875rem;
+      height: 0.875rem;
       border-radius: 1rem;
-      background: var(--pal-brand-primary-on);
```
piyalbasu, line 54: I couldn't quite figure out why this background was needed. Let me know if this needs to come back!

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
```diff
@@ -47,14 +47,21 @@
     border-radius: 100px;
     padding: 0.25rem 0.5rem;
 
-    img {
-      width: 1rem;
-      height: 1rem;
```
piyalbasu, line 52: the designs had this icon at 14px, so I sized this down in my changes

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -21,25 +23,40 @@ export const AssetIcon = ({
...
+      className={`AccountAssets__asset--logo ${
+        hasError ? "AccountAssets__asset--error" : ""
+      } ${isLoading ? "AccountAssets__asset--loading" : ""}`}
+    >
+      <img
+        alt={`${code} logo`}
+        src={isXlm ? StellarLogo : imgSrc}
+        onError={() => {
+          if (retryAssetIconFetch) {
+            retryAssetIconFetch({ key: issuerKey, code });
```
acharb, line 44: ❓ definitely an edge case, but if the retry function fails, then `hasError` will stay as false right?

not sure what the best fix would be 🤔  we could have retryAssetIconFetch return something that could be used to check if failed/succeeded

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
piyalbasu: Ah right!

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: Good catch - `setHasError(true)` should always be called. If retry works, it'll re-render and skip right over onError to onLoad

`setIsLoading(false)` actually doesn't really need to be called onError bc if the icon is never found, it just skips down to line 54 where isLoading doesn't have any effect

acharb (approved):
lgtm!

PR #527, opened 2022-06-13
make sendSettings rows clickable, add pin extension view

acharb:
clickable items https://stellarorg.atlassian.net/browse/WAL-234
pin extension view https://stellarorg.atlassian.net/browse/WAL-277

Screen Shot 2022-06-13 at 1 11 50 PM


extension/src/popup/views/FullscreenSuccessMessage/index.tsx
```diff
@@ -26,38 +50,16 @@ const MnemonicPhraseConfirmedMessage = () => (
...
-      </InfoBlock>
-    </div>
+    <AvoidScamsWarningBlock />
     <SubmitButtonWrapper>
       <Button
         fullWidth
         onClick={() => {
           emitMetric(METRIC_NAMES.accountCreatorFinished);
-          window.close();
+          navigateTo(ROUTES.pinExtension);
```
piyalbasu, line 59: hmm is it worth only nav'ing to ROUTES.pinExtension in Chrome since this doesn't appear to apply on Firefox

Sorry! I should have caught this beforehand

acharb: nice catch updated 👍  using the old images / copy if it isn't chrome

piyalbasu (approved):

PR #531, opened 2022-06-15
add connecting with ledger UI

acharb:
https://user-images.githubusercontent.com/30449853/173958301-f730261e-d47c-4445-9ce2-0b7bcde31a1d.mov


extension/src/popup/views/Account/styles.scss
```diff
@@ -1,9 +1,9 @@
 :root {
-  --account-view-padding-left: 1.5rem;
+  --account-view-padding-side: 1.5rem;
```
acharb, line 2: took me a solid min to figure out why we were using only `left` for the side padding 😆  so I changed to `side`

extension/src/popup/components/hardwareConnect/LedgerConnect/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  if (message.indexOf("Incorrect length")) {
+    return LEDGER_ERROR.NOT_OPEN;
+  }
+  return LEDGER_ERROR.OTHER;
+};
+
+export const LedgerConnect = ({ goBack }: { goBack?: () => void }) => {
+  const [isConnecting, setIsConnecting] = useState(false);
+
+  // TODO - move to redux with bipPath
```
acharb, line 40: will most probably move the connect stuff into redux, wasn't sure of the best API yet so doing in later PR

extension/package-lock.json
```diff
@@ -1,13 +0,0 @@
-{
```
acharb, line 1: this isn't needed right?

extension/src/popup/ducks/accountServices.ts
```diff
@@ -238,15 +238,17 @@ interface InitialState {
   applicationState: APPLICATION_STATE;
   hasPrivateKey: boolean;
   publicKey: string;
+  connectingWalletType: string;
```
piyalbasu, line 241: pretty small - but should the type here be `WalletType`? Would make it clear that we're looking for something from that enum here

extension/src/popup/components/hardwareConnect/LedgerConnect/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const [connectError, setConnectError] = useState("");
+
+  const handleConnect = async () => {
+    setIsConnecting(true);
+    try {
+      const transport = await TransportWebUSB.request();
+      const ledgerApi = new LedgerApi(transport);
+      const response = await ledgerApi.getPublicKey(defaultStellarBipPath);
+
+      setPublicKey(response.publicKey);
```
piyalbasu, line 52: Awesome that their API looks pretty easy to use!

piyalbasu (approved):

extension/package-lock.json
piyalbasu: Yup! Can be deleted, thanks!

extension/src/popup/ducks/accountServices.ts
acharb: ya  good call https://github.com/stellar/freighter/pull/531/commits/b1bba6d7b161390d23d023bc7ad0fd87021d1dec

PR #537, opened 2022-06-24
Demonic Vulnerability Part 1

piyalbasu:
Screen Shot 2022-06-24 at 3 47 45 PM
 
A very much unstyled prototype of what the experience of recovering an account by mnemonic phrase could look like


extension/src/popup/components/mnemonicPhrase/MnemonicDisplay/index.tsx
```diff
@@ -1,8 +1,50 @@
...
+}
+
+export const generateMnemonicPhraseDisplay = ({
+  mnemonicPhrase,
+}: generateMnemonicPhraseDisplayProps) =>
+  mnemonicPhrase.split(" ").map((word: string) => {
+    const randomNumber = random(1, 10);
+    const randomWordArr = generateMnemonic().split(" ");
+    const randomWordIndex = random(0, randomWordArr.length);
+    const randomWord = randomWordArr[randomWordIndex];
```
piyalbasu, line 20: when we show the mnemonic phrase, we're going to add some hidden, random words throughout the phrase to obscure the actual words: https://halborn.com/disclosures/demonic-vulnerability/

extension/src/popup/helpers/useMnemonicPhrase.ts
```diff
@@ -19,6 +19,10 @@ export const useMnemonicPhrase = () => {
     };
 
     fetchMnemonicPhrase();
+
+    return () => {
+      setMnemonicPhrase("");
+    };
```
piyalbasu, line 25: dereference the var as soon as we're done with it

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -26,41 +31,90 @@ import { PasswordRequirements } from "popup/components/PasswordRequirements";
...
+  const publicKeyRef = useRef(publicKey);
   const RecoverAccountSchema = YupObject().shape({
-    mnemonicPhrase: mnemonicPhraseValidator,
     password: passwordValidator,
     confirmPassword: confirmPasswordValidator,
     termsOfUse: termsofUseValidator,
   });
 
   const dispatch = useDispatch();
+  const phraseLengthOptions = [12, 24];
```
piyalbasu, line 105: I imagine this list will be longer - I've only tested with 12 and 24 length phrases so far

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -71,15 +125,32 @@ export const RecoverAccount = () => {
...
   }, [publicKey]);
 
+  useEffect(() => {
+    const phraseInputsArr = [];
+
+    // eslint-disable-next-line no-plusplus
+    for (let i = 1; i <= maxPhraseLength; i++) {
+      phraseInputsArr.push(`MnemonicPhrase-${i}`);
+    }
+    setPhraseInputs(phraseInputsArr);
```
piyalbasu, line 135: programmatically create the maximum number of fields needed

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -71,15 +125,32 @@ export const RecoverAccount = () => {
...
+  }, [maxPhraseLength]);
+
+  const handleMnemonicInputChange = (value: string, i: number) => {
+    const arr = [...mnemonicPhraseArr];
+    arr[i] = value;
+
+    setMnemonicPhraseArr(arr);
```
piyalbasu, line 142: as the user makes changes to a phrase input, construct the mnemonic phrase in memory by placing the input in the correct index of the string

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -26,41 +31,90 @@ import { PasswordRequirements } from "popup/components/PasswordRequirements";
...
+}: PhraseInputProps) => {
+  const [isTextShowing, setIsTextShowing] = useState(false);
+  const [inputValue, setInputValue] = useState("");
 
+  return (
+    <div
+      key={phraseInput}
+      className={`RecoverAccount__phrase-input ${
+        isDisabled ? "RecoverAccount__phrase-input--disabled" : ""
+      }`}
```
piyalbasu, line 55: the strategy here is to create the max number of inputs possible for a mnemonic phrase (in this case, 24) and then just show/hide the needed fields depending on how many the user selects

extension/src/popup/components/mnemonicPhrase/MnemonicDisplay/index.tsx
```diff
@@ -1,8 +1,50 @@
...
+// @ts-ignore
+import { generateMnemonic } from "stellar-hd-wallet";
+import { random } from "lodash";
 
 import "./styles.scss";
 
+interface generateMnemonicPhraseDisplayProps {
+  mnemonicPhrase: string;
+}
+
```
acharb, line 12: can we add a comment here explaining the reason for the random words? For future readers who may not have context

extension/src/popup/components/mnemonicPhrase/DisplayMnemonicPhrase/index.tsx
```diff
@@ -50,26 +45,6 @@ export const DisplayMnemonicPhrase = ({
...
-                filename: "freighterMnemonicPhrase.txt",
-                content: mnemonicPhrase,
-              });
-              emitMetric(METRIC_NAMES.accountCreatorMnemonicDownloadPhrase);
-            }}
-          >
-            <IconButton preset={IconButton.preset.download} />
-          </button>
-          <CopyText textToCopy={mnemonicPhrase} showTooltip>
-            <IconButton preset={IconButton.preset.copy} />
```
acharb, line 70: maybe a question for bruno/charles too, but are we sure we want to remove the download/copy options? Since the vulnerability is about the browser saving the phrase unknowingly to users, where as download/copy is an active action by the user.

not saying we shouldn't remove, but just curious of your thoughts

acharb (approved):
makes sense. Had a question but since this is still a prototype approving for now

extension/src/popup/components/mnemonicPhrase/MnemonicDisplay/index.tsx
piyalbasu: Good thinking!

extension/src/popup/components/mnemonicPhrase/DisplayMnemonicPhrase/index.tsx
piyalbasu: So for the copy method, Halborn specifically mentions preventing adding the mnemonic to the clipboard as it can leak into memory:

```
Prevent (and warn) users from directly copying/pasting the Mnemonic Phrase into the browser. The browser’s clipboard is also responsible for leaking data into memory. Therefore, it is important to warn the user not to fill the browser’s clipboard with the Mnemonic Phrase,```

They advocate for completing removing the ability to copy in any capacity.

Download is a little more of a gray area. I still haven't been able to determine how risky it is, but at this stage, I'm leaning towards being a bit more conservative. I think we'll probably learn a bit more about this vulnerability as Metamask/Phantom open-source their solutions. If at that point we determine any of these are not as risky as we thought, we can definitely bring them back

acharb: gotchya, makes sense 👍

PR #538, opened 2022-06-24
include hardware wallet account in background functions

acharb:


extension/src/background/helpers/session.ts
```diff
@@ -14,11 +11,10 @@ export class SessionTimer {
     this.DURATION = duration || this.DURATION;
   }
 
-  startSession(key: { privateKey: string }) {
+  startSession() {
     if (this.runningTimeout) {
       clearTimeout(this.runningTimeout);
     }
-    store.dispatch(grantAccountAccess(key));
```
acharb, line 21: I decoupled starting a session w/ setting the private key. Since now we'll start a session if confirming password while using a hardware wallet, but there's no private key

for this change I also changed the name from `grantAccountAccess` to `setPrivateKey`

jest.config.js
```diff
@@ -8,6 +10,7 @@ module.exports = {
   transform: {
     "^.+\\.(js|jsx|ts|tsx|mjs)$": ["babel-jest"],
   },
+  transformIgnorePatterns: [`/node_modules/(?!${esModules})`],
```
acharb, line 13: for some reason jest + babel doesn't like the wallet-sdk, so I had to add this line. Otherwise would get a "can't use import statement" error

extension/src/background/messageListener/__tests__/popupMessageListener.test.js
```diff
@@ -0,0 +1,65 @@
+import { SERVICE_TYPES } from "@shared/constants/services";
```
piyalbasu, line 1: Amazing! Thank you for adding these tests!

jest.config.js
piyalbasu: Hm, interesting. I can take a look into this, but def not a blocker. This solution looks fine to me!

piyalbasu (approved):

PR #540, opened 2022-06-24
use first non-xlm asset as default dest asset for swaps

acharb:


extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -219,9 +218,15 @@ export const SendAmount = ({
   // for swaps we're loading the destinationAsset here
   useEffect(() => {
     if (isSwap && !destinationAsset) {
-      dispatch(saveDestinationAsset(StellarSdk.Asset.native().toString()));
+      // default to first non-native asset if exists
+      const nonXlmAssets = Object.keys(accountBalances.balances || {}).filter(
+        (b) => b !== "native",
+      );
+      dispatch(
+        saveDestinationAsset(nonXlmAssets[0] ? nonXlmAssets[0] : "native"),
```
piyalbasu, line 226: Very NIT: we're not very consistent with this in the codebase, but we should probably use `StellarSdk.Asset.native().toString()` instead of `"native"`

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendAmount/index.tsx
acharb: ah ya  good catch 
https://github.com/stellar/freighter/pull/540/commits/6c7765138564c005fa4e1591ca473ff8a9c425a7

PR #544, opened 2022-06-28
store hardwareWallet in redux, background methods changes for when hardwareWallet active

acharb:
this PR is for 2 main things:
- store the new ledger account in redux, and add to the account list dropdown
- fixes `importAccount` and `addAccount` in `background popupMessageListener` for when checking password while the hardware wallet account is active (added tests for that case)

`+` also using WalletType instead of string everywhere

Screen Shot 2022-06-27 at 6 04 21 PM


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -73,15 +74,30 @@ export const popupMessageListener = (request: Request) => {
...
+  }) => keyManager.loadKey(keyID, password);
+
+  // this returns the first non hardware wallet (Hw) keyID, if it exists.
+  // Used for things like checking a password when a Hw is active.
+  const _getNonHwKeyID = () => {
+    const keyIdList = getKeyIdList();
+    const nonHwKeyIds = keyIdList.filter(
+      (k: string) => k.indexOf(HW_PREFIX) === -1,
+    );
+    return nonHwKeyIds[0] || "";
```
piyalbasu, line 92: I think this is good as it captures any weird edge case I'm not thinking of - but is there any reason to believe the first keyId in this list (the one the user onboards with) wouldn't always be "Non-HW" 🤔

extension/src/popup/components/account/AccountList/index.tsx
```diff
@@ -1,22 +1,39 @@
...
+export const OptionTag = ({
+  hardwareWalletType = WalletType.NONE,
+  imported,
+}: {
+  hardwareWalletType?: WalletType;
+  imported: boolean;
+}) => {
+  if (hardwareWalletType) {
+    return (
+      <span className="AccountList__option-tag">
```
piyalbasu, line 19: very NIT: possible to not repeat the element `<span className="AccountList__option-tag">`?

Maybe something like 
```
if (!hardwareWalletType && !imported) return null;
 return <span className="AccountList__option-tag">&bull; {hardwareWalletType || "Imported"} </span>
```

piyalbasu (approved):

extension/src/background/messageListener/popupMessageListener.ts
acharb: that's a good point I hadn't thought of that 🤔 

I think I'm in favor of this way, since it's more conservative. I also can't think of a way the first key won't always be non-hw, but in case that changes down the road, or if something funky happens with the list this method will still hold true

extension/src/popup/components/account/AccountList/index.tsx
acharb: ah nice I spent a couple minutes trying to make this a one-ish liner but ended up with a gross "if hw else if imported else null", but I like your way 👍 
https://github.com/stellar/freighter/pull/544/commits/257bb767e7b0886c42d6fa3428855211447280cc

PR #545, opened 2022-06-28
Bugfix/allow null network

piyalbasu:


@stellar/freighter-api/src/signTransaction.ts
```diff
@@ -2,6 +2,6 @@ import { submitTransaction } from "@shared/api/external";
 
 export const signTransaction = (
   transactionXdr: string,
-  network?: "PUBLIC" | "TESTNET",
+  network?: "PUBLIC" | "TESTNET" | null,
   accountToSign?: string
 ) => submitTransaction(transactionXdr, network, accountToSign);
```
piyalbasu, line 7: this should probably accept 1 object instead of 3 params, but in order to keep things backwards compatible, we would need to check the input for both an object and 3 params and then coerce it into the right structure. 

I decided not to do that for now because I wasn't sure that adding that abstraction was even worth the possible bugs/issues at this point. If we end up adding another param to this, I'll definitely switch it to accepting an object instead

acharb (approved):
lgtm!

PR #554, opened 2022-07-02
send transactions with ledger

acharb:
Last thing for this ticket will be some smaller items (ie animating the LedgerConnect component and handling non-default bip paths), will do in a subsequent PR since this PR was already getting pretty big

https://user-images.githubusercontent.com/30449853/177612097-7181c0a5-d8d8-4f0b-aecb-06a206d0e69c.mov

https://user-images.githubusercontent.com/30449853/177612215-e51e4d33-4423-44a7-bc78-aaa9e50b6544.mov


extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -245,6 +301,18 @@ const transactionSubmissionSlice = createSlice({
     saveAllowedSlippage: (state, action) => {
       state.transactionData.allowedSlippage = action.payload;
     },
+    startHwConnect: (state) => {
+      state.hardwareWalletData.status = HwOverlayStatus.IN_PROGRESS;
+      state.hardwareWalletData.transactionXDR = "";
```
acharb, line 306: clearing the transactionXDR just to be safe, since that's how the component will know the intent is to import an account and not sign

@shared/api/internal.ts
```diff
@@ -437,13 +437,17 @@ export const signFreighterTransaction = async ({
 
 export const submitFreighterTransaction = async ({
   signedXDR,
-  networkUrl,
+  networkDetails,
 }: {
   signedXDR: string;
```
acharb, line 442: ran into some funky typescript stuff, it looks like we've been passing the transaction object to this method to submit, instead of the xdr string. Not sure how Typescript allowed this before 🤷 , but changed it to pass the string and build the object here

extension/src/popup/components/hardwareConnect/LedgerConnect/index.tsx
```diff
@@ -37,37 +47,68 @@ const parseLedgerError = (err: any): string => {
...
+  const closeOverlay = () => dispatch(closeHwOverlay());
+
   const handleConnect = async () => {
     setIsConnecting(true);
+    setConnectError("");
     try {
       const transport = await TransportWebUSB.request();
       const ledgerApi = new LedgerApi(transport);
       const response = await ledgerApi.getPublicKey(defaultStellarBipPath);
+      setIsConnected(true);
```
piyalbasu, line 71: Just spitballing an idea here: It's a bit confusing to have the app status be able to be both `isConnecting === true` and `isConnected === true` at the same time.

Would it maybe be more clear if there was a `connectionStatus` that could have the value of an enum?

for ex: `connectionStatus === CONNECTION_STATUS.CONNECTING` or `connectionStatus === CONNECTION_STATUS.CONNECTED`

It might make some of the UI logic below more messy, so if that's the case, I think you can ignore. But it'd be nice if these 2 var's were related

@shared/api/internal.ts
piyalbasu: I had originally commented "why not just change the type?" but looks like it makes the ledger flow a little easier to pass in an xdr string here instead 👍

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -184,23 +193,26 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
   const handleSend = async () => {
-    const server = new StellarSdk.Server(networkDetails.networkUrl);
-
     try {
+      const server = new StellarSdk.Server(networkDetails.networkUrl);
       const transactionXDR = await server
         .loadAccount(publicKey)
-        .then((sourceAccount: Types.Account) => {
-          const transaction = new StellarSdk.TransactionBuilder(sourceAccount, {
+        .then((sourceAccount: Types.Account) =>
```
piyalbasu, line 200: ah, I can't believe I didn't notice this before. Would you mind using the `await` pattern here rather than `then`? We should be consistent in the codebase

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -38,23 +40,62 @@ export const signFreighterTransaction = createAsyncThunk<
...
+      const tx = StellarSdk.TransactionBuilder.fromXDR(
+        transactionXDR,
+        networkPassphrase,
+      );
+
+      const transport = await TransportWebUSB.create();
+      const ledgerApi = new LedgerApi(transport);
+      // TODO - move to redux
+      const result = await ledgerApi.signTransaction(
+        "44'/148'/0'",
```
piyalbasu, line 81: might be nice to have a comment explaining the significance of this value

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
acharb: ah ya  good catch https://github.com/stellar/freighter/pull/554/commits/1e24f7f47c5066c4232d5cc4b36c7e34b8d67eee

extension/src/popup/ducks/transactionSubmission.ts
acharb: it's the default bip path for ledger. Forgot to change this one to the constant, I'll be moving in my next PR but I'll change to the const now so I don't forget 👍 
https://github.com/stellar/freighter/pull/554/commits/7aec9cdaf66534f00ef5374e36bcc38fbad404af

extension/src/popup/components/hardwareConnect/LedgerConnect/index.tsx
acharb: ya  that's a good point 🤔 , maybe a good solution would moving the isConnected status to redux. Or else not ya  I agree an enum would also be more clear.

I'll explore that in my next PR to make it more clear

PR #555, opened 2022-07-08
support non default-bip path, store hw data in localstorage, add animation

acharb:
this should be the last of the ledger support work 🎸 

https://user-images.githubusercontent.com/30449853/178058800-e02f1e7f-e7ff-433c-b2d1-6e82509068cb.mov


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -92,49 +93,65 @@ export const popupMessageListener = (request: Request) => {
...
 
     const keyId = `${HW_PREFIX}${publicKey}`;
     const keyIdListArr = getKeyIdList();
     const accountName = `${hardwareWalletType} ${
       keyIdListArr.filter((k: string) => k.indexOf(HW_PREFIX) !== -1).length + 1
     }`;
-    keyIdListArr.push(keyId);
 
-    localStorage.setItem(KEY_ID_LIST, JSON.stringify(keyIdListArr));
+    if (keyIdListArr.indexOf(keyId) === -1) {
```
acharb, line 120: also making sure the same ledger account cant be imported twice. I noticed we can still for regular accounts though, I'll create a ticket for that

extension/src/popup/components/hardwareConnect/LedgerConnect/index.tsx
```diff
@@ -47,35 +50,55 @@ const parseLedgerError = (err: any): string => {
   return LEDGER_ERROR.OTHER;
 };
 
-export const LedgerConnect = () => {
+export const LedgerConnect = ({ newBipPath }: { newBipPath?: string }) => {
   const dispatch: AppDispatch = useDispatch();
-  const [isConnecting, setIsConnecting] = useState(false);
```
acharb, line 52: changed the names to "detecting" and "ledgerConnectSuccessful", I think that makes it a little less confusing instead of 2 "connect" var names

extension/src/background/messageListener/popupMessageListener.ts
acharb: https://github.com/stellar/freighter/issues/556

piyalbasu (approved):

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: Ah yeah, that's def an issue. Thanks for creating the ticket!

PR #572, opened 2022-07-14
moving ledger connect to own move, some refactoring to confirmPassword in background

acharb:


extension/src/background/messageListener/__tests__/popupMessageListener.test.js
```diff
@@ -1,6 +1,7 @@
 import { SERVICE_TYPES } from "@shared/constants/services";
 import { popupMessageListener } from "background/messageListener/popupMessageListener";
 import { store } from "background/store";
+import { sessionSlice } from "background/ducks/session";
```
acharb, line 4: I added some test cases here to be doubly sure that background functions are working as expected

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -508,42 +508,57 @@ export const popupMessageListener = (request: Request) => {
     }
   };
 
-  const confirmPasswordHW = async () => {
```
acharb, line 511: the `confirmPasswordHW` logic actually wasn't working (wasn't calling `login` in `confirmPasswordHW` oops). After looking over the two `confirmPassword` methods I think it's better to keep it as one to be more DRY and avoid future errors like this, so I removed the extra `confirmPasswordHW` function, and just refactored the original `confirmPassword` method to be able to handle hw or not

extension/src/popup/components/hardwareConnect/LedgerSign/index.tsx
```diff
@@ -0,0 +1,144 @@
+import React, { useState, useEffect, useRef } from "react";
```
acharb, line 1: we now have a LedgerSign component (that slides up when signing a transaction), and a LedgerConnect one (it's own view for when initially connecting)

piyalbasu (approved):

PR #573, opened 2022-07-15
sign external app transactions with hardware wallet

acharb:
https://user-images.githubusercontent.com/30449853/179158336-3c4f822b-61fe-4c78-9ee4-686e9351e725.mov


extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -234,126 +259,130 @@ export const SignTransaction = () => {
...
-          <Card variant={Card.variant.highlight}>
-            <PunycodedDomain domain={domain} isRow />
-            <div className="SignTransaction__subject">
-              {t("is requesting approval to a")} {isFeeBump ? "fee bump " : ""}
-              {t("transaction")}:
-            </div>
-            <div className="SignTransaction__approval">
-              <div className="SignTransaction__approval__title">
-                {t("Approve using")}:
+    <>
```
acharb, line 262: the only change I made in the jsx is adding the `{hwStatus === HwOverlayStatus.IN_PROGRESS && <LedgerSign />}`, the diff is a little deceiving

extension/src/popup/components/hardwareConnect/LedgerSign/index.tsx
```diff
@@ -78,12 +79,18 @@ export const LedgerSign = () => {
...
+          dispatch(
+            submitFreighterTransaction({
+              signedXDR: res.payload,
+              networkDetails,
+            }),
+          );
+        } else {
+          // right now there are only two cases after signing,
+          // submitting to network or handling in background script
+          await handleSignedHwTransaction({ transactionXDR: res.payload });
```
piyalbasu, line 92: this is so so pedantic, but I'm wondering if it might be nice to call this param `signedXDR` instead of just `transactionXDR`, similar to the `submitFreighterTransaction` param above.

It would also be nice in `handleSignedHwTransaction` to make it very clear that Ledger has already signed the XDR that is being passed

piyalbasu (approved):

extension/src/popup/components/hardwareConnect/LedgerSign/index.tsx
acharb: ooh good catch I agree. I think I'll change to `signedTransaction` since we're using that in the background script popupMessageListener already and to keep all the `handleSignedHwTransaction` methods using the same var name

acharb: https://github.com/stellar/freighter/pull/573/commits/129cfb6529859ded178012074482d3043e83960d

PR #585, opened 2022-07-27
show swaps in account history

acharb:
Screen Shot 2022-07-27 at 9 49 54 AM
Screen Shot 2022-07-27 at 9 50 01 AM

### DIFF
--- a/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
+++ b/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
@@ -14,7 +14,10 @@ import { HorizonOperation } from "@shared/api/types";
 import { TransactionDetailProps } from "../TransactionDetail";
 import "./styles.scss";
 
-export type HistoryItemOperation = HorizonOperation & { isPayment?: boolean };
+export type HistoryItemOperation = HorizonOperation & {
+  isPayment?: boolean;
+  isSwap?: boolean;
+};
 
 interface HistoryItemProps {
   operation: HistoryItemOperation;
@@ -31,6 +34,7 @@ export const HistoryItem = ({
   setDetailViewProps,
   setIsDetailViewShowing,
 }: HistoryItemProps) => {
+  const { t } = useTranslation();
   const {
     amount,
     asset_code: assetCode,
@@ -40,7 +44,12 @@ export const HistoryItem = ({
     type,
     transaction_attr: { operation_count: operationCount },
     isPayment = false,
+    isSwap = false,
   } = operation;
+  let sourceAssetCode;
+  if ("source_asset_code" in operation) {
+    sourceAssetCode = operation.source_asset_code;
+  }
   const operationType = camelCase(type) as keyof typeof OPERATION_TYPES;
   const operationString = `${OPERATION_TYPES[operationType]}${
     operationCount > 1 ? ` + ${operationCount - 1} ops` : ""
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -50,64 +59,90 @@ export const HistoryItem = ({
...
 
   let isRecipient = false;
   let paymentDifference = "";
+  let rowText = "";
+  let dateText = "";
   let IconComponent = <Icon.Shuffle className="HistoryItem__icon--default" />;
   let PaymentComponent = null as React.ReactElement | null;
 
-  if (isPayment) {
+  let transactionDetailProps: TransactionDetailProps = {
```
acharb, line 72: did a bit of refactoring here. Because now there's 3 states (swap, payment, or neither), I think cleaner to set each var in an if-elseif-else rather than a ternary for each one when used

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -40,7 +44,12 @@ export const HistoryItem = ({
     type,
     transaction_attr: { operation_count: operationCount },
     isPayment = false,
+    isSwap = false,
   } = operation;
+  let sourceAssetCode;
+  if ("source_asset_code" in operation) {
+    sourceAssetCode = operation.source_asset_code;
+  }
```
piyalbasu, line 52: hm, is there a danger in just setting `const sourceAssetCode = operation.source_asset_code;`?

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -30,12 +30,16 @@ const getIsPayment = (type: Horizon.OperationResponseType) =>
     Horizon.OperationResponseType.pathPaymentStrictSend,
   ].includes(type);
 
+const getIsSwap = (operation: HistoryItemOperation) =>
+  operation.type_i === 13 && operation.source_account === operation.to;
+
+enum SELECTOR_OPTIONS {
```
acharb, line 36: unrelated changed, I moved this enum outside the AccountHistory component. I noticed it was causing a re-render loop from the useEffect below (only noticed bc I added a log for debugging)

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
acharb: kind of annoying of typescript imo but 🤷

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -30,12 +30,16 @@ const getIsPayment = (type: Horizon.OperationResponseType) =>
...
+  ALL = "ALL",
+  SENT = "SENT",
+  RECEIVED = "RECEIVED",
+}
```
piyalbasu, line 40: no need to change this, but wanted to point out why i had moved `SELECTOR_OPTIONS` inside this component:

I needed translations for `SELECTOR_OPTIONS`, but the job that builds translations can't interpret code, it can only analyze it statically for strings inside of `t()`. So, a common workaround is to include a comment where you manually "call" the `t()` function so the builder knows to translate those strings:
``  
/*
      t("ALL");
      t("SENT");
      t("RECEIVED");
    */`
```

This comment needs to be inside of a React component for the builder to analyze it. I had moved the enum inside to be next to the comment so it was clear that they were related

Comment (acharb) [reply extension/src/popup/views/AccountHistory/index.tsx:40]: ahh gotchya makes sense 👍

File: extension/src/popup/components/accountHistory/HistoryItem/index.tsx [hunk 5]
```diff
@@ -50,64 +59,90 @@ export const HistoryItem = ({
...
+    setIsDetailViewShowing,
+  };
+
+  if (isSwap) {
+    PaymentComponent = (
+      <>
+        {new BigNumber(amount).toFixed(2, 1)} {destAssetCode}
+      </>
+    );
+    rowText = t(`${srcAssetCode} for ${destAssetCode}`);
```
piyalbasu, line 89: ah, this type of string interpolation won't work with the `t()` function unfortunately.
It'll need to be: 
```
headerTitle: t(`Swapped {{srcAssetCode}} for {{destAssetCode}}`, {
        srcAssetCode,
        destAssetCode,
      }),
```

https://www.i18next.com/translation-function/interpolation#basic

piyalbasu (approved):

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -50,64 +59,90 @@ export const HistoryItem = ({
...
+    dateText = t(`Swap \u2022 ${date}`);
+    transactionDetailProps = {
+      ...transactionDetailProps,
+      headerTitle: t(`Swapped ${srcAssetCode} for ${destAssetCode}`),
```
piyalbasu, line 93: ah, this type of string interpolation won't work with the `t()` function unfortunately.
It'll need to be: 
```
headerTitle: t(`Swapped {{srcAssetCode}} for {{destAssetCode}}`, {
        srcAssetCode,
        destAssetCode,
      }),
```

https://www.i18next.com/translation-function/interpolation#basic

acharb: ahh nice catch, I was wondering why the build translations wasn't catching them 😄 
https://github.com/stellar/freighter/pull/585/commits/1abcaeadf6126eb571b10461b047406fc383e893

PR #587, opened 2022-07-29
WAL-319, WAL-342: dont allow mnemonic re-view, bottom nav hover state

acharb:


extension/src/popup/views/MnemonicPhrase.tsx
```diff
@@ -10,9 +13,15 @@ import { Header } from "popup/components/Header";
...
 import { ConfirmMnemonicPhrase } from "popup/components/mnemonicPhrase/ConfirmMnemonicPhrase";
 import { DisplayMnemonicPhrase } from "popup/components/mnemonicPhrase/DisplayMnemonicPhrase";
+import { applicationStateSelector } from "popup/ducks/accountServices";
 
 export const MnemonicPhrase = () => {
   const mnemonicPhrase = useMnemonicPhrase();
+  const applicationState = useSelector(applicationStateSelector);
+
+  if (applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED) {
+    return <Redirect to={ROUTES.pinExtension} />;
```
acharb, line 23: chose to go to the pinExtension view. To my knowledge the only way this would happen is if someone manually tries entering the mnemonic url after it's already been confirmed, so it's def an edge case, and isn't terribly important where it goes to as long as it's not the mnemonic view. But open to suggestions on a different view to redirect to

piyalbasu: This makes sense to me!

piyalbasu (approved):

PR #588, opened 2022-08-02
handle LP shares

acharb:
Screen Shot 2022-08-02 at 3 13 37 PM

### DIFF
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -1,4 +1,3 @@
-import omitBy from "lodash/omitBy";
 import StellarSdk from "stellar-sdk";
 import { DataProvider } from "@stellar/wallet-sdk";
 import {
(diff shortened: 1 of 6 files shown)


@shared/api/internal.ts
```diff
@@ -280,14 +279,32 @@ export const getAccountBalances = async ({
...
 
   try {
-    ({ balances, subentryCount } = await dataProvider.fetchAccountDetails());
-    // let's filter out liquidity pool shares until freighter and wallet-sdk support
-    balances = omitBy(balances, (b: any) => b.liquidity_pool_id) as Balances;
+    const resp = await dataProvider.fetchAccountDetails();
+    balances = resp.balances;
+    subentryCount = resp.subentryCount;
+
+    for (let i = 0; i < Object.keys(resp.balances).length; i++) {
```
acharb, line 291: til (after much befuddlement) `.forEach` doesn't wait on the `await` calls inside its loop 🤷

@shared/api/internal.ts
```diff
@@ -280,14 +279,32 @@ export const getAccountBalances = async ({
     networkPassphrase,
   });
 
-  let balances = null;
+  let balances: any = null;
```
acharb, line 282: I played around with adding another type to BalanceMaps that would account for the LiquidityPool data. Something like
```
export interface BalanceMap {
  [key: string]: AssetBalance | NativeBalance | LPBalance
...
``` 
but the breaking change gets pretty ugly since LP data wouldn't share a lot of fields with AssetBalance and NativeBalance (eg. token), so it would need to be typecasted every time it's used. 

I think we'll need to take a holistic view of the wallet-sdk balance types and see how we can best incorporate a new LP share balance type (alternatively we could just rely on Horizon's BalanceLine type). But for now we're just displaying the shares so I think it's ok to set that aside as tech debt. Happy to discuss though

piyalbasu: Ah yeah, I was having some related issue with my these types with my work. Totally fine with using `any`. You'll see in my PR that doing it the "right way" is kinda ugly

@shared/api/internal.ts
piyalbasu: Argh, yeah. Sorry, one of those annoying gotchas. Glad you found the solution!

@shared/api/internal.ts
```diff
@@ -280,14 +279,32 @@ export const getAccountBalances = async ({
...
-    ({ balances, subentryCount } = await dataProvider.fetchAccountDetails());
-    // let's filter out liquidity pool shares until freighter and wallet-sdk support
-    balances = omitBy(balances, (b: any) => b.liquidity_pool_id) as Balances;
+    const resp = await dataProvider.fetchAccountDetails();
+    balances = resp.balances;
+    subentryCount = resp.subentryCount;
+
+    for (let i = 0; i < Object.keys(resp.balances).length; i++) {
+      const k = Object.keys(resp.balances)[i];
+      const v: any = resp.balances[k];
```
piyalbasu, line 293: this is totally fine - but just thought I'd mention you could use `Object.entries` to get a tuple of the key and the value so you wouldn't have to declare `v` manually here

piyalbasu (approved):

@shared/api/internal.ts
acharb: ya  I couldn't figure out a way to cast `v` to type `any` in the for loop 🤔 . I wish something like `for (let [k, v: any] of Object.entries(resp.balances))` works but typescript doesn't like it 😢 

I'll merge but if there's a way to do it better I can change in next PR

PR #590, opened 2022-08-03
add asset balance UI and refactor History to share code

piyalbasu:
https://user-images.githubusercontent.com/6789586/182660533-db51ae86-37a8-47f1-b721-883a55e0c207.mov

### DIFF
--- a/extension/src/popup/components/SlideupModal/index.tsx
+++ b/extension/src/popup/components/SlideupModal/index.tsx
@@ -1,0 +1,41 @@
+import React, { useEffect, useRef, useState } from "react";
+
+import { LoadingBackground } from "popup/basics/LoadingBackground";
+
+import "./styles.scss";
+
+interface SlideupModalProps {
+  children: React.ReactElement;
+  isModalOpen: boolean;
+  setIsModalOpen: (isModalOpen: boolean) => void;
+}
+
+export const SlideupModal = ({
+  children,
+  isModalOpen,
+  setIsModalOpen,
+}: SlideupModalProps) => {
+  const slideupModalRef = useRef<HTMLDivElement>(null);
+  const [slideupModalHeight, setSlideupModalHeight] = useState(-500);
+
+  useEffect(() => {
+    setSlideupModalHeight(slideupModalRef.current?.clientHeight || 0);
+  }, [slideupModalRef]);
+  return (
+    <>
+      <div
+        className="SlideupModal"
+        ref={slideupModalRef}
+        style={{
+          bottom: isModalOpen ? "0px" : `-${slideupModalHeight}px`,
+        }}
+      >
+        {children}
+      </div>
+      <LoadingBackground
+        onClick={() => setIsModalOpen(false)}
+        isActive={isModalOpen}
+      />
+    </>
+  );
+};
--- a/extension/src/popup/components/account/AccountAssets/index.tsx
+++ b/extension/src/popup/components/account/AccountAssets/index.tsx
@@ -15,6 +15,8 @@ import ImageMissingIcon from "popup/assets/image-missing.svg";
 
 import "./styles.scss";
 
+const getIsXlm = (code: string) => code === "XLM";
+
 export const AssetIcon = ({
   assetIcons,
   code,
@@ -34,7 +36,7 @@ export const AssetIcon = ({
     Method 2. We get an icon path directly from an API (like in the trustline flow) and just pass it to this component to render
   */
 
-  const isXlm = code === "XLM";
+  const isXlm = getIsXlm(code);
 
   // in Method 1, while we wait for the icon path to load, `assetIcons` will be empty until the promise resolves
   // This does not apply for XLM as there is no lookup as that logo lives in this codebase
@@ -95,13 +97,17 @@ export const AssetIcon = ({
   );
 };
 
+interface AccountAssetsProps {
+  assetIcons: AssetIcons;
+  sortedBalances: Array<any>;
+  setSelectedAsset?: (selectedAsset: string) => void;
+}
+
 export const AccountAssets = ({
   assetIcons: inputAssetIcons,
   sortedBalances,
-}: {
-  assetIcons: AssetIcons;
-  sortedBalances: Array<any>;
-}) => {
+  setSelectedAsset,
+}: AccountAssetsProps) => {
   const [assetIcons, setAssetIcons] = useState(inputAssetIcons);
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const [hasIconFetchRetried, setHasIconFetchRetried] = useState(false);
@@ -133,6 +139,12 @@ export const AccountAssets = ({
     }
   };
 
+  const handleClick = (code: string) => {
+    if (setSelectedAsset) {
+      setSelectedAsset(getIsXlm(code) ? "native" : code);
+    }
+  };
+
   const getLPShareCode = (reserves: Horizon.Reserve[]) => {
     if (!reserves[0] || !reserves[1]) {
       return "";
@@ -146,7 +158,7 @@ export const AccountAssets = ({
     <>
       {sortedBalances.map((rb: any) => {
         let issuer;
-        let code;
+        let code = "";
         let amountUnit;
         if (rb.liquidityPoolId) {
           issuer = "lp";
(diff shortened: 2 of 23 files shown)


extension/src/popup/components/accountHistory/HistoryList/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+  assetDetail?: boolean;
+}
+
+export const HistoryList = ({ children, assetDetail }: HistoryListProps) => (
+  <SimpleBar
+    className={`HistoryList ${assetDetail ? "HistoryList--assetDetail" : ""}`}
+  >
+    <div className="HistoryList__items">{children}</div>
+  </SimpleBar>
+);
```
piyalbasu, line 18: reusing the History scrolling component from Account History

extension/src/popup/components/accountHistory/NetworkInfo/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import "./styles.scss";
+
+interface NetworkInfoProps {
+  assetIssuer: string;
+  assetCode: string;
+  assetType: string;
+  assetDomain: string;
+}
+
+export const NetworkInfo = ({
```
piyalbasu, line 18: Reusing the logic to get an operation's Network from Transaction Detail

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +22,94 @@ export const sortBalances = (balances: Balances) => {
...
+
+export const getIsPayment = (type: Horizon.OperationResponseType) =>
+  [
+    Horizon.OperationResponseType.payment,
+    Horizon.OperationResponseType.pathPayment,
+    Horizon.OperationResponseType.pathPaymentStrictSend,
+  ].includes(type);
+
+export const getIsSwap = (operation: HorizonOperation) =>
+  operation.type_i === 13 && operation.source_account === operation.to;
```
piyalbasu, line 34: reusing getIsPayment and getIsSwap from Account History

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +22,94 @@ export const sortBalances = (balances: Balances) => {
...
+    const key = bal.token.type === "native" ? bal.token.type : bal.token.code;
+    assetOperationMap[key] = [];
+  });
+
+  operations.forEach((op) => {
+    if (getIsPayment(op.type)) {
+      Object.keys(assetOperationMap).forEach((asset) => {
+        if (op.asset_code === asset || op.asset_type === asset) {
+          assetOperationMap[asset].push(op);
+        } else if ("source_asset_type" in op || "source_asset_code" in op) {
```
piyalbasu, line 60: ugliness because of `Balance` types

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +22,94 @@ export const sortBalances = (balances: Balances) => {
...
+export const getStellarExpertUrl = (isTestnet: boolean) =>
+  `https://stellar.expert/explorer/${isTestnet ? "testnet" : "public"}`;
+
+interface GetAvailableBalance {
+  accountBalances: AccountBalancesInterface;
+  selectedAsset: string;
+  recommendedFee?: string;
+}
+
+export const getAvailableBalance = ({
```
piyalbasu, line 84: reusing this from the Send Payment flow

extension/src/popup/helpers/useAssetDomain.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useSelector } from "react-redux";
+import StellarSdk from "stellar-sdk";
+
+import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
+
+interface UseAssetDomain {
+  assetIssuer?: string;
+}
+
+export const useAssetDomain = ({ assetIssuer }: UseAssetDomain) => {
```
piyalbasu, line 11: reusing from Account History

extension/src/popup/components/SlideupModal/index.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useEffect, useRef, useState } from "react";
```
piyalbasu, line 1: reusing from Sign Transaction

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -83,13 +85,17 @@ export const AssetIcon = ({
   );
 };
 
+interface AccountAssetsProps {
+  assetIcons: AssetIcons;
+  sortedBalances: Array<any>;
+  setSelectedAsset?: (selectedAsset: string) => void;
```
acharb, line 103: thoughts on this name change? from the AccountAssets POV the setSelectedAsset makes it seem like it's a `useState` var w/in AccountAssets
```suggestion
  handleAssetSelect?: (selectedAsset: string) => void;
```

extension/src/popup/components/accountHistory/NetworkInfo/index.tsx
acharb, line 18: I got confused for a minute thinking `NetworkInfo` was pulling data on Mainnet vs testnet before looking at the code. Thoughts on changing the name to something like `AssetNetworkInfo`, or something similar?

extension/src/popup/components/accountHistory/NetworkInfo/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      setNetworkIconUrl(iconUrl);
+    };
+
+    if (assetIssuer) {
+      fetchIconUrl();
+    }
+  }, [assetCode, assetIssuer, networkDetails]);
+  return (
+    <div className="TransactionDetail__header__network">
```
acharb, line 49: and also the one below. Looks like it's catching the other style file but just to be safe 😄 
```suggestion
    <div className="NetworkInfo__header__network">
```

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +22,94 @@ export const sortBalances = (balances: Balances) => {
...
+export interface AssetOperations {
+  [key: string]: Array<HorizonOperation>;
+}
+
+export const sortOperationsByAsset = ({
+  balances,
+  operations,
+}: SortOperationsByAsset) => {
+  const assetOperationMap = {} as AssetOperations;
+  balances.forEach((bal) => {
```
acharb, line 50: I think this might break w/ the LP shares PR I just merged 🤔  since it woun't have the `token` field :/ 

I think we can handle by if the balance doesn't have a `token` field, we can ignore it. Ie ignoring LP shares for here

acharb: oh actually looks like the map should include it. So adding the key as the LP identifier to the AssetOperations map, and then adding the op to it

tomerweller:
This. Is. AWESOME.

extension/src/popup/components/account/AccountAssets/index.tsx
piyalbasu: So this actually is a `useState` var, just being passed from the parent. Does that make it less confusing? Totally fine if it's still confusing, but I have used this naming convention in a few other places (for ex: https://github.com/stellar/freighter/blob/master/extension/src/popup/views/Account/index.tsx#L138)

I'm happy to create a ticket to change this naming convention if you feel strongly about it!

extension/src/popup/components/accountHistory/NetworkInfo/index.tsx
piyalbasu: Argh good catch

extension/src/popup/components/account/AccountAssets/index.tsx
acharb: ahh ok,  now that I have that way of thinking in my head of it's a useState from a higher component makes sense to leave 👍 thanks for explaining

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +24,105 @@ export const sortBalances = (balances: Balances) => {
...
+  operations.forEach((op) => {
+    if (getIsPayment(op.type)) {
+      Object.keys(assetOperationMap).forEach((assetKey) => {
+        const asset = getAssetFromCanonical(assetKey);
+        const assetCode = asset.code === "XLM" ? "native" : asset.code;
+        const assetIssuer = asset.issuer;
+
+        if (
+          (op.asset_code === assetCode && op.asset_issuer === assetIssuer) ||
+          op.asset_type === assetCode
```
acharb, line 69: this case would only be for XLM right? can we hardcode that here?
```suggestion
          op.asset_type === "native"
```

extension/src/popup/helpers/account.ts
```diff
@@ -15,3 +24,105 @@ export const sortBalances = (balances: Balances) => {
...
+        ) {
+          assetOperationMap[asset].push(op);
+        } else if ("source_asset_type" in op || "source_asset_code" in op) {
+          if (
+            op.source_asset_type === assetCode ||
```
acharb, line 74: and same for here
```suggestion
            op.source_asset_type === "native" ||
```

acharb (approved):
lgtm!

I had 1 question to make the readibility easier, but not blocking

extension/src/popup/helpers/account.ts
piyalbasu: Ah this is clearer, but unfortunately does not work here. Since it's essentially a nested for loop, we need to make sure that not only does `asset_type === "native"` but also that the `asset` we are looking at it is "native"

In the truthy scenario, the operation loop's `asset_type` is the same as the asset loop's `assetCode`

PR #595, opened 2022-08-04
WAL-395, WAL-396: add loader, default self payment to sent

acharb:


extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -81,8 +84,7 @@ export const AccountHistory = () => {
         if (isPayment && !isSwap) {
           if (operation.source_account === publicKey) {
             segments[SELECTOR_OPTIONS.SENT].push(historyOperation);
-          }
-          if (operation.to === publicKey) {
+          } else if (operation.to === publicKey) {
```
acharb, line 87: @piyalbasu is there any reason we wouldn't want if-else here? I can't think of one, but just wanted to double check

piyalbasu: I think my original thinking was just that self payments should be under both SENT and RECEIVED. But since we now have isSwap and are defaulting to SENT, this makes more sense.

piyalbasu (approved):

PR #597, opened 2022-08-05
Prevent flash of "Stellar Network" for asset domain on non-XLM assets

piyalbasu:
https://user-images.githubusercontent.com/6789586/183160427-f7598350-8fc9-484f-bba9-0c0ea50f484c.mov


acharb (approved):
lgtm!

PR #598, opened 2022-08-05
categorize create external account ops as Sent

piyalbasu:
https://user-images.githubusercontent.com/6789586/183221527-541a9308-8d4a-4878-b23a-26e96f16a6e9.mov

The first op (`Create Account`) is the account being created.
The second `XLM` is me sending 5 XLM to another account to create it

### DIFF
--- a/@shared/api/types.ts
+++ b/@shared/api/types.ts
@@ -71,12 +71,7 @@ export interface AssetIcons {
 export type Balances = Types.BalanceMap | null;
 
 /* eslint-disable camelcase */
-export type HorizonOperation = (
-  | Horizon.PaymentOperationResponse
-  | Horizon.PathPaymentOperationResponse
-) & {
-  transaction_attr: Horizon.TransactionResponse;
-};
+export type HorizonOperation = any;
 /* eslint-enable camelcase */
 
 export interface AccountBalancesInterface {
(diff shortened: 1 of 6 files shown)


@shared/api/types.ts
```diff
@@ -71,12 +71,7 @@ export interface AssetIcons {
 export type Balances = Types.BalanceMap | null;
 
 /* eslint-disable camelcase */
-export type HorizonOperation = (
-  | Horizon.PaymentOperationResponse
-  | Horizon.PathPaymentOperationResponse
-) & {
-  transaction_attr: Horizon.TransactionResponse;
-};
+export type HorizonOperation = any;
```
piyalbasu, line 74: @acharb Ugh, I'm realizing I did a pretty bad job here. This doesn't take into account other Response types (like `CreateAccountOperationResponse`). When I add other types, TS doesn't do a great job of inferring the correct type when we use our helper methods like `getIsPayment`. I need to rethink this 🤔 

Associated ticket:
https://stellarorg.atlassian.net/browse/WAL-409

piyalbasu:
https://user-images.githubusercontent.com/6789586/183223234-0e303e65-6dc6-49d4-8c75-34eead13e22f.mov

acharb (approved):
lgtm!

@shared/api/types.ts
acharb: I think this is fine for now, agreed we should probably take another look at this, along with the Balances types

PR #602, opened 2022-08-10
WAL-394, WAL-360: dont round on dest asset amounts, show dest asset balances for swaps

acharb:
https://user-images.githubusercontent.com/30449853/184042837-bc5891e0-ae32-4f95-a600-225efc77b8f1.mov

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
@@ -6,6 +6,7 @@ import { ROUTES } from "popup/constants/routes";
 import { navigateTo } from "popup/helpers/navigate";
 import { AssetIcon } from "popup/components/account/AccountAssets";
 import { transactionSubmissionSelector } from "popup/ducks/transactionSubmission";
+import { useIsSwap } from "popup/helpers/useIsSwap";
 
 import "./styles.scss";
 
@@ -13,6 +14,7 @@ export enum ASSET_SELECT {
   QUERY_PARAM = "assetSelect",
   SOURCE = "source",
   DEST = "dest",
+  SWAP_QUERY_PARAM = "isSwap",
 }
 
 export function AssetSelect({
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/styles.scss
```diff
@@ -33,6 +33,8 @@
 
     &__right {
       color: var(--pal-text-primary);
+      overflow-y: hidden;
```
acharb, line 36: if the value has a lot of digits, then worst case it'll scroll

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -29,7 +31,13 @@ export const ManageAssets = () => {
     default:
   }
 
-  const { balances } = accountBalances;
```
acharb, line 32: I realized we're loading the source trustlines always even for path payment destination (oops), but fixed now :)

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
```diff
@@ -6,13 +6,15 @@ import { ROUTES } from "popup/constants/routes";
...
 import { transactionSubmissionSelector } from "popup/ducks/transactionSubmission";
+import { useIsSwap } from "popup/helpers/useIsSwap";
 
 import "./styles.scss";
 
 export enum ASSET_SELECT {
   QUERY_PARAM = "assetSelect",
   SOURCE = "source",
   DEST = "dest",
+  SWAP_QUERY_PARAM = "isSwap",
```
piyalbasu, line 17: I think this is fine for this PR, but I'm wondering if it makes sense to move all this asset select stuff either into Redux or (perhaps more appropriate for this kind of ephemeral data) React Context.

Relying too much on query params always make me a little nervous as route names, etc. can change, so sometimes it's nicer to rely on something more deterministic. Wdyt? It could be done in a future bug ticket.

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendAmount/AssetSelect/index.tsx
acharb: That's a good point 🤔 it is getting a little too complex for using query params only. I can change to redux in this PR

acharb: on second thought is a bigger change than I initially thought. I'll do this in a subsequent PR so it's not lost in this PR's changes 😄

acharb: tracking here https://stellarorg.atlassian.net/browse/WAL-446

PR #604, opened 2022-08-12
WAL-446: lift asset select code to redux

acharb:
[pointed out by piyal](https://github.com/stellar/freighter/pull/602#discussion_r943747568) that the asset select code is getting a little hairy for using routes alone. Lifting to redux makes it a lot more understandable / safer

### DIFF
--- a/extension/src/popup/components/manageAssets/SelectAssetRows/index.tsx
+++ b/extension/src/popup/components/manageAssets/SelectAssetRows/index.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { useHistory, useLocation } from "react-router-dom";
+import { useHistory } from "react-router-dom";
 import SimpleBar from "simplebar-react";
 import "simplebar-react/dist/simplebar.min.css";
 import { Types } from "@stellar/wallet-sdk";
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/manageAssets/SelectAssetRows/index.tsx
```diff
@@ -23,20 +23,15 @@ import "./styles.scss";
 interface SelectAssetRowsProps {
   assetRows: ManageAssetCurrency[];
   maxHeight: number;
-  selectingAssetType: string;
```
acharb, line 26: removing the selectingAssetType prop propagated through all the components anymore, much simpler 🕺

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -240,6 +246,10 @@ interface InitialState {
   accountBalances: AccountBalancesInterface;
   destinationBalances: AccountBalancesInterface;
   assetIcons: AssetIcons;
+  assetSelect: {
+    type: AssetSelectType;
+    source: boolean;
```
piyalbasu, line 251: can we call this `isSource` (or something similar) since it's a bool?

piyalbasu (approved):

extension/src/popup/ducks/transactionSubmission.ts
acharb: ooh ya  good call https://github.com/stellar/freighter/pull/604/commits/cbd168ded8a0de121d0690ff88f178531c09eb09

PR #612, opened 2022-08-24
Feature/wal 434 edit custom network

piyalbasu:
https://user-images.githubusercontent.com/6789586/186473075-21d89617-b69e-4578-8010-5bc8a676168c.mov


@shared/api/types.ts
```diff
@@ -101,5 +102,3 @@ declare global {
 }
 
 export type CURRENCY = { code: string; issuer: string; image: string };
-
-export type CustomNetwork = NetworkDetails & { isSwitchSelected: boolean };
```
piyalbasu, line 105: before I had `background` adding a custom network and then switching to that network if `isSwitchSelected`. Now I just have the UI call 2 services, `addCustomNetwork` and `changeNetwork` if that's what the user wants

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -429,43 +426,63 @@ export const popupMessageListener = (request: Request) => {
...
     const { networkName } = request;
 
-    const savedNetworks = JSON.parse(
-      localStorage.getItem(NETWORKS_LIST_ID) ||
-        JSON.stringify(DEFAULT_NETWORKS),
-    ) as NetworkDetails[];
+    const savedNetworks = getSavedNetworks();
+    const networkIndex = savedNetworks.findIndex(
+      ({ networkName: savedNetworkName }) => savedNetworkName === networkName,
+    );
```
piyalbasu, line 459: find a network by networkName and remove it from the arr

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -429,43 +426,63 @@ export const popupMessageListener = (request: Request) => {
...
+
+    localStorage.setItem(NETWORKS_LIST_ID, JSON.stringify(savedNetworks));
+
+    return {
+      networksList: savedNetworks,
+    };
+  };
+
+  const editCustomNetwork = () => {
+    const { networkDetails, networkIndex } = request;
```
piyalbasu, line 471: when editing, any field can change, so we can't identify the old version by any of the field. We just replace it by the index of where it falls in the networksList array instead

extension/src/helpers/stellar.ts
```diff
@@ -1,6 +1,6 @@
 import BigNumber from "bignumber.js";
 import StellarSdk from "stellar-sdk";
-import { isEqual } from "lodash";
+import isEqual from "lodash/isEqual";
```
piyalbasu, line 3: So this is the issue I mentioned to @acharb yesterday: I was unable to `yarn build:extension` locally. It turns out `background`, when not minified, was over 4mb (!). Firefox has a max file size of 4mb, so I had set Webpack to split files if they are over this limit. Unfortunately, we can't do that for `background`. It needs to be one file for a browser to load it in an extension

`lodash` is a huuuuge library. If you import lodash methods this way, it actually just loads the code you need rather than the whole lodash lib. I made this change everywhere I saw a lodash import

During the Manifest upgrade, I'll also be looking at other ways to reduce our bundle size because it's definitely too big right now and we'll continue to run into similar issues in the future

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -80,7 +80,7 @@ export const AssetDetail = ({
     selectedAsset,
   });
 
-  const stellarExpertUrl = getStellarExpertUrl(networkDetails.network);
+  const stellarExpertUrl = getStellarExpertUrl(networkDetails);
```
piyalbasu, line 83: `getStellarExpertUrl` now needs all of `networkDetails` to determine what network a user is on

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -71,15 +72,14 @@ const ResultsHeader = () => {
...
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const [assetRows, setAssetRows] = useState([] as ManageAssetCurrency[]);
   const [maxHeight, setMaxHeight] = useState(0);
   const [isSearching, setIsSearching] = useState(false);
   const [hasNoResults, setHasNoResults] = useState(false);
   const ResultsRef = useRef<HTMLDivElement>(null);
 
-  // TODO IN THIS PR
-  const isTestnet = network === "TESTNET";
+  const isNetworkTestnet = isTestnet(networkDetails);
```
piyalbasu, line 82: similarly `isTestnet` and `isMainnet` needs `networkDetails` to determine a user's network

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -99,7 +99,7 @@ export const SearchAsset = ({ setErrorAsset }: SearchAssetProps) => {
       try {
         res = await fetch(
           `https://api.stellar.expert/explorer/${
-            isTestnet ? "testnet" : "public"
+            isNetworkTestnet ? "testnet" : "public"
```
piyalbasu, line 102: Open question that needs to be addressed in a future PR: what stellar.expert API should we use for people on a custom network?

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -23,60 +38,202 @@ interface FormValues {
...
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const networksList = useSelector(settingsNetworksListSelector);
+  const settingsError = useSelector(settingsErrorSelector);
+  const [isNetworkInUse, setIsNetworkInUse] = useState(false);
+  const [isConfirmingRemoval, setIsConfirmingRemoval] = useState(false);
+  const history = useHistory();
+  const { search } = useLocation();
+
+  const networkIndex = new URLSearchParams(search).get("networkIndex");
+  const networkDetailsToEdit = networksList[Number(networkIndex)];
```
piyalbasu, line 62: If we're editing a Network's config, we grab the Network's index in the NetworksList from a query param

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -23,60 +38,202 @@ interface FormValues {
...
+        },
+      }),
+    );
+
+    const addCustomNetworkFulfilled = addCustomNetwork.fulfilled.match(
+      addCustomNetworkRes,
+    );
+    let changeNetworkFulfilled = true;
+
+    if (values.isSwitchSelected) {
```
piyalbasu, line 128: if the user wants to automatically switch to the new network after adding, we just call `changeNetwork` next

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -108,7 +265,22 @@ export const NetworkForm = ({ isEditing }: NetworkFormProps) => {
               placeholder={t("Enter passphrase")}
             />
             {isEditing ? (
-              "Remove"
+              <div className="NetworkForm__remove-wrapper">
+                {!isMainnetOrTestnet && (
```
piyalbasu, line 269: Disabling the `remove` button if someone tries to edit the Mainnet or Testnet config

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -127,25 +299,35 @@ export const NetworkForm = ({ isEditing }: NetworkFormProps) => {
...
+              <div className="NetworkForm__editing-buttons">
+                <Button
+                  onClick={() => history.goBack()}
+                  type="button"
+                  variant={Button.variant.tertiary}
+                  fullWidth
+                >
+                  {t("Cancel")}
+                </Button>
+                <Button
```
piyalbasu, line 311: Similarly, don't let the user edit Mainnent/Testnet configs

extension/src/popup/components/manageNetwork/NetworkIcon/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+import "./styles.scss";
+
+/*
+  Custom network icon colors are determined by where they fall in the list. We cycle through 4 colors and repeat if we get to the end. 
+*/
```
piyalbasu, line 7: I don't love the idea of having a different color for every Custom network. I also don't really like the idea of the color being determined by where it falls in the list.

But I really really don't like the idea of saving a Custom network's color in the "database". So I created this slightly over-engineered component to figure out a Network icon's color based on where it falls in the list of networks

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -23,60 +38,203 @@ interface FormValues {
...
+  const networksList = useSelector(settingsNetworksListSelector);
+  const settingsError = useSelector(settingsErrorSelector);
+  const [isNetworkInUse, setIsNetworkInUse] = useState(false);
+  const [isConfirmingRemoval, setIsConfirmingRemoval] = useState(false);
+  const history = useHistory();
+  const { search } = useLocation();
+
+  const networkIndex = Number(
+    new URLSearchParams(search).get(NETWORK_INDEX_SEARCH_PARAM),
+  );
```
piyalbasu, line 63: If we're editing a Network's config, we grab the Network's index in the NetworksList from a query param

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -127,25 +300,35 @@ export const NetworkForm = ({ isEditing }: NetworkFormProps) => {
...
-                fullWidth
-                isLoading={isSubmitting}
-                type="submit"
-                variant={Button.variant.tertiary}
-              >
-                {t("Save")}
-              </Button>
+              <div className="NetworkForm__add-button">
+                <Button
+                  disabled={!(isValid && dirty) || isMainnetOrTestnet}
```
acharb, line 324: doesn't this line block adding a new network if you're currently on Mainnet or testnet? I just tried it locally and I wasn't able to add a new network

acharb (commented):
looking great 👍  I had 1 comment / potential blocker

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
piyalbasu: Ugh, you're right - this was the last thing I added and I didn't test it enough

piyalbasu: Just addressed this. There was a def a bug. Also, `isMainnetOrTestnet` actually referred to whether you were trying to edit those configs, not if you were currently using it. I renamed that var to make it more clear

acharb (approved):
lgtm!

PR #619, opened 2022-08-29
scam asset pt 2: add new preferences flag, show warning overlay when adding trustline to scam asset

acharb:
https://user-images.githubusercontent.com/30449853/187309229-762f62ef-4842-4541-8333-bc38815f61c1.mov

### DIFF
--- a/@shared/api/types.ts
+++ b/@shared/api/types.ts
@@ -31,6 +31,7 @@ export interface Response {
   isTestnet: boolean;
   isMemoValidationEnabled: boolean;
   isSafetyValidationEnabled: boolean;
+  isValidatingSafeAssetsEnabled: boolean;
   networkDetails: NetworkDetails;
   networksList: NetworkDetails[];
   allAccounts: Array<Account>;
@@ -71,6 +72,7 @@ export interface Settings {
   networkDetails: NetworkDetails;
   isMemoValidationEnabled: boolean;
   isSafetyValidationEnabled: boolean;
+  isValidatingSafeAssetsEnabled: boolean;
   networksList: NetworkDetails[];
   error: string;
 }
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -1,9 +1,14 @@
-import React, { useState } from "react";
+import React, { useState, useRef, useEffect } from "react";
+import { useDispatch, useSelector } from "react-redux";
 import { createPortal } from "react-dom";
-import { Icon } from "@stellar/design-system";
+import { InfoBlock, Icon } from "@stellar/design-system";
 import { useTranslation } from "react-i18next";
+import { POPUP_HEIGHT } from "constants/dimensions";
 
 import { Button } from "popup/basics/buttons/Button";
+import { closeBlockedDomainWarning } from "popup/ducks/transactionSubmission";
+import { settingsSelector } from "popup/ducks/settings";
+import { ManageAssetRow } from "popup/components/manageAssets/ManageAssetRows";
 
 import "./styles.scss";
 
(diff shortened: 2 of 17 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -218,6 +246,39 @@ export const ManageAssetRows = ({
...
+};
+
+interface AssetRowData {
+  code: string;
+  issuer: string;
+  image: string;
+  domain: string;
+}
+
+export const ManageAssetRow = ({
```
acharb, line 264: moved to its own component so can re-use

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -214,7 +214,7 @@ export const getBlockedDomains = createAsyncThunk<
   }
 });
 
-export enum HwOverlayStatus {
+export enum ShowOverlayStatus {
```
acharb, line 217: changed this to something more general that we can re-use anytime we need a screen overlay, eg. ledger signing or this warning

acharb: if there's a better word than "overlay" that describes what this is doing happy to change the name 😄

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -64,7 +64,7 @@ export const AddAsset = ({ setErrorAsset }: AddAssetProps) => {
       setAssetRows(
         assetDomainToml.CURRENCIES.map((currency) => ({
           ...currency,
-          domain: assetDomainToml.DOCUMENTATION?.ORG_URL || "",
```
acharb, line 67: https://stellarfoundation.slack.com/archives/C016CPSLVLG/p1661363477552569

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -199,3 +204,96 @@ export const BackupPhraseWarningMessage = () => {
...
+              fullWidth
+              variant={Button.variant.tertiary}
+              type="button"
+              onClick={closeOverlay}
+            >
+              {t("Got it")}
+            </Button>
+            {!isValidatingSafeAssetsEnabled && (
+              <Button fullWidth onClick={closeOverlay} type="button">
+                {/* ALEC TODO - handle add anyway onclick */}
```
acharb, line 290: will do in subsequent PR

@shared/api/types.ts
```diff
@@ -31,6 +31,7 @@ export interface Response {
   isTestnet: boolean;
   isMemoValidationEnabled: boolean;
   isSafetyValidationEnabled: boolean;
```
acharb, line 33: I'm thinking we should change `isSafetyValidationEnabled` to something like `isValidatingSafeAccountsEnabled`. I don't _think_ there will be an issue changing this localStorage key. We could default to true for the new key

I can create a ticket for if agreed

piyalbasu: Yeah, this is a good call. This was an oversight on my part. I should have made this key name more specific to begin with 👍

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -199,3 +204,96 @@ export const BackupPhraseWarningMessage = () => {
...
+      dispatch(closeBlockedDomainWarning());
+    }, 300);
+  };
+
+  // animate entry
+  useEffect(() => {
+    if (warningRef.current) {
+      warningRef.current.style.bottom = "0";
+    }
+  }, [warningRef]);
```
piyalbasu, line 238: Is it possible to use `components/SlideupModal` here? It'll take care of this animation stuff for you

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -218,6 +246,39 @@ export const ManageAssetRows = ({
         </div>
         {children}
       </SimpleBar>
+      <LoadingBackground
+        onClick={() => {}}
+        isActive={blockedDomains.status === ShowOverlayStatus.IN_PROGRESS}
+      />
```
piyalbasu, line 252: you won't need to invoke this LoadingBackground with `SlideupModal` either 😃

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -166,9 +177,27 @@ export const ManageAssetRows = ({
     }
   }, [submitStatus, assetSubmitting, setErrorAsset, dispatch]);
 
+  const isBlockedDomain = (domain: string) => {
+    const found = blockedDomains.domains.filter((b) => b.domain === domain);
+    return found.length > 0;
+  };
+
+  const handleBlockedDomain = (assetData: AssetRowData) => {
+    dispatch(showBlockedDomainWarning());
```
piyalbasu, line 186: related to using `slideupModal`, but wondering if you even need to run this through redux. Seems like all the state changes are happening right in this component

@shared/api/types.ts
acharb: https://stellarorg.atlassian.net/browse/WAL-477?atlOrigin=eyJpIjoiZTE1OGVhMTVhYzFmNGIxOTk4NDczOWJkNTM4NTU0MGMiLCJwIjoiaiJ9

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: hmm good point technically it doesn't need to be in redux 🤔 . If alright with you I'd like to finish the scam asset work (should be 1 more major PR), and if it is simpler not in redux after that work then I can remove. They'll be 1 more warning screen, and possibly other data I'm not thinking of now that could be put into redux and it would be nice to keep that data together I think

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: ah I forgot about SlideupModal 🤔  we can, it's a little different because I'm having the warning view  appear and sliding up at the same time. It looks like some refactoring to be able to use the SlideupModal, since it assumes it's always there below but slides up when needed (just tried seeing if it'd be a quick plug and play and I'm having some css trouble with it). If alright with you I'd also like to chalk it up as tech debt, we could change LedgerSign to use as well

acharb: https://stellarorg.atlassian.net/browse/WAL-478

piyalbasu: Yup, sounds good. If it's not a quick 1:1 swap, we can hold off. Thanks for making the ticket!

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
piyalbasu: Good plan. No need to optimize if there's still more screens to add 👍

piyalbasu (approved):

PR #631, opened 2022-09-16
handle add asset from warning screen

acharb:
part of the [scam asset work](https://stellarorg.atlassian.net/browse/WAL-421?atlOrigin=eyJpIjoiOTEyNDdlOTZkNWFmNDc5N2I1MmI5ZDM0NWM0NjIyYmEiLCJwIjoiaiJ9)

https://user-images.githubusercontent.com/30449853/190730951-8a536966-a7bc-4f1c-9154-5f8f2ad5ea2e.mov

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -4,11 +4,27 @@ import { createPortal } from "react-dom";
 import { InfoBlock, Icon } from "@stellar/design-system";
 import { useTranslation } from "react-i18next";
 import { POPUP_HEIGHT } from "constants/dimensions";
+import StellarSdk, { Account } from "stellar-sdk";
 
+import { xlmToStroop, getCanonicalFromAsset } from "helpers/stellar";
+import { AppDispatch } from "popup/App";
 import { Button } from "popup/basics/buttons/Button";
-import { closeBlockedDomainWarning } from "popup/ducks/transactionSubmission";
-import { settingsSelector } from "popup/ducks/settings";
+import {
+  closeBlockedDomainWarning,
+  signFreighterTransaction,
+  submitFreighterTransaction,
+} from "popup/ducks/transactionSubmission";
+import {
+  settingsSelector,
+  settingsNetworkDetailsSelector,
+} from "popup/ducks/settings";
 import { ManageAssetRow } from "popup/components/manageAssets/ManageAssetRows";
+import { useNetworkFees } from "popup/helpers/useNetworkFees";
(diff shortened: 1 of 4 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -237,6 +259,48 @@ export const ScamAssetWarning = ({
     }
   }, [warningRef]);
 
+  const handleSubmit = async () => {
```
acharb, line 262: spent a bit of time trying to find a good way of abstracting this away since similar code is used in ManageAssetRows, I couldn't find a clean way of doing since both are tied to UIs differently and use a bunch of vars from the outer component scope

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -173,7 +173,7 @@ export const ManageAssetRows = ({
       navigateTo(ROUTES.trustlineError);
     } else if (submitStatus === ActionStatus.SUCCESS) {
       dispatch(resetSubmission());
-      navigateTo(ROUTES.trustlineError);
+      navigateTo(ROUTES.account);
```
acharb, line 176: idk why this was trustlineError before 🤔  it should go back to home on success

piyalbasu: ooh yeah - good catch!

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -237,6 +259,48 @@ export const ScamAssetWarning = ({
...
+    if (signFreighterTransaction.fulfilled.match(res)) {
+      const submitResp = await dispatch(
+        submitFreighterTransaction({
+          signedXDR: res.payload.signedTransaction,
+          networkDetails,
+        }),
+      );
+      if (submitFreighterTransaction.fulfilled.match(submitResp)) {
+        navigateTo(ROUTES.account);
+        emitMetric(METRIC_NAMES.manageAssetAddAsset, { code, issuer });
```
piyalbasu, line 296: pretty minor - but wondering if we should use a new METRIC_NAMES here to differentiate from adding a trustline that wasn't flagged

piyalbasu (approved):

extension/src/popup/components/WarningMessages/index.tsx
acharb: ah ya  I like that idea, for better tracking 
https://github.com/stellar/freighter/pull/631/commits/e7f9f422dc4dbc8eede89d940274cc2f06430f01

PR #634, opened 2022-09-22
check for suspicious characteristics when adding new asset w/ warning view

acharb:
will need 1 more PR to finish off the scam asset work, for flagging scam assets across views

https://user-images.githubusercontent.com/30449853/191856023-9e264653-0f3a-4701-ad70-8fdba7e44cba.mov

note: hardcoded so that the view would come up for this video, only comes up if at least one characteristic shown


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -365,3 +374,188 @@ export const ScamAssetWarning = ({
     </div>
   );
 };
+
+export const NewAssetWarning = ({
```
acharb, line 378: this is very similar to the `ScamAssetWarning` view above

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -182,20 +199,70 @@ export const ManageAssetRows = ({
...
+      isRevocable = resp.records[0]
+        ? resp.records[0].flags.auth_revocable
+        : false;
+    } catch (e) {
+      console.error(e);
+    }
+
+    // check if new asset
+    // TODO: stellar expert asset rating endpoint giving 403, resolve with OL
+    const isNewAsset = false;
```
acharb, line 220: I reached out to OL, the asset `/rating` endpoint is giving me a 403 (eg https://api.stellar.expert/explorer/testnet/asset/A-GD4PLJJJK4PN7BETZLVQBXMU6JQJADKHSAELZZVFBPLNRIXRQSM433II/rating)
but only from a react app, not from the browser url so I think something with CORS going on

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -367,12 +365,6 @@ const transactionSubmissionSlice = createSlice({
     saveAssetSelectSource: (state, action) => {
       state.assetSelect.isSource = action.payload;
     },
-    showBlockedDomainWarning: (state) => {
```
acharb, line 370: removed this stuff from redux to keep it w/in the component since it's only used there 😄

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -365,3 +374,188 @@ export const ScamAssetWarning = ({
...
+    if (warningRef.current) {
+      setTimeout(() => {
+        warningRef.current!.style.bottom = "0";
+      }, 10);
+    }
+  }, [warningRef]);
+
+  const closeOverlay = () => {
+    if (warningRef.current) {
+      warningRef.current.style.bottom = `-${POPUP_HEIGHT}px`;
```
piyalbasu, line 416: so so NIT: we're using `warningRef.current!.` up above, we should probably be consistent here

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -365,3 +374,188 @@ export const ScamAssetWarning = ({
...
+                <div className="NewAssetWarning__flag__icon">
+                  <img src={IconWarning} alt="invalid domain" />
+                </div>
+                <div className="NewAssetWarning__flag__content">
+                  <div className="NewAssetWarning__flag__header">
+                    {t("Invalid Format Asset")}
+                  </div>
+                  <div className="NewAssetWarning__flag__description">
+                    {t(
+                      "Asset home domain doesn't exist, TOML file format is invalid, or asset doesn't match currency description",
```
piyalbasu, line 532:
```suggestion
                      "Asset home domain doesn‘t exist, TOML file format is invalid, or asset doesn't match currency description",
```

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -182,20 +199,70 @@ export const ManageAssetRows = ({
...
+    issuer: string,
+    domain: string,
+  ): Promise<NewAssetFlags> => {
+    // check revocable
+    let isRevocable = false;
+    try {
+      const resp = await server.assets().forCode(code).forIssuer(issuer).call();
+      isRevocable = resp.records[0]
+        ? resp.records[0].flags.auth_revocable
+        : false;
```
piyalbasu, line 213: there's actually a small edge case. If for some reason the shape of `resp.records[0]` changed and there wasn't a `flags` object anymore, it would cause an error

It would be a little safer to do:

```
isRevocable  = resp?.records[0]?.flags?.auth_revocable
        || false
```

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -229,9 +296,26 @@ export const ManageAssetRows = ({
...
+                        setNewAssetFlags(resp);
+                        setSuspiciousAssetData({ domain, image, code, issuer });
+                      } else if (
+                        isBlockedDomain(domain) &&
+                        !isTrustlineActive
+                      ) {
+                        setShowBlockedDomainWarning(true);
+                        setSuspiciousAssetData({ domain, image, code, issuer });
                       } else {
                         changeTrustline(code, issuer, !isTrustlineActive);
```
piyalbasu, line 320: it might be nice to move this into a function up above since the logic is getting a little more involved. It's a little nicer to not have bigger pieces of logic mixed in with the JSX

piyalbasu (approved):

extension/src/popup/components/WarningMessages/index.tsx
acharb: thx for catching these i always forget lol 
https://github.com/stellar/freighter/pull/634/commits/a56954289da8e76c17a1d984d6d54ece5909c2a7

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: nice idea 👍 https://github.com/stellar/freighter/pull/634/commits/a56954289da8e76c17a1d984d6d54ece5909c2a7

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
acharb: 👍  https://github.com/stellar/freighter/pull/634/commits/a56954289da8e76c17a1d984d6d54ece5909c2a7

extension/src/popup/components/WarningMessages/index.tsx
acharb: the one's with `!` are actually only needed because of ts linting since the `warning.current.` line of code isn't directly after the `if (warning.current)`. Personally I like it better w/o for unneeded verbosity reasons. Will merge now but lmk if you feel strongly of including them

PR #635, opened 2022-09-23
add warning icon to scam assets

acharb:
Screen Shot 2022-09-23 at 11 45 33 AM
Screen Shot 2022-09-23 at 11 45 15 AM
Screen Shot 2022-09-23 at 11 45 08 AM
Screen Shot 2022-09-23 at 11 45 25 AM
Screen Shot 2022-09-23 at 1 46 12 PM


@shared/api/types.ts
```diff
@@ -45,12 +45,12 @@ export interface Response {
...
   bipPath: string;
-  blockedDomains: Array<BlockedDomain>;
+  blockedDomains: BlockedDomains;
+  assetDomain: string;
 }
 
-export interface BlockedDomain {
-  domain: string;
-  paging_token: string;
+export interface BlockedDomains {
```
acharb, line 52: I changed BlockedDomains to be a map w/ the domain as the key. Makes it much easier to use than an array of objects

@shared/api/internal.ts
```diff
@@ -412,6 +413,34 @@ export const retryAssetIcon = async ({
   return newAssetIcons;
 };
 
+export const getAssetDomains = async ({
```
acharb, line 416: I looked into combining this with `getIconUrlFromIssuer`, I think it's cleaner separate rather than trying to squeeze them into the same method. My worry initially was performance, but I don't notice when making the call twice, and especially since they both using caching anyways

@shared/api/helpers/getDomainFromIssuer.ts
```diff
@@ -0,0 +1,51 @@
+import StellarSdk from "stellar-sdk";
+import { sendMessageToBackground } from "./extensionMessaging";
+import { SERVICE_TYPES } from "../../constants/services";
+import { NetworkDetails } from "../../constants/stellar";
+
+export const getDomainFromIssuer = async ({
```
acharb, line 6: I looked into combining this with `getIconUrlFromIssuer`, I think it's cleaner / less convoluted separate rather than trying to squeeze them into the same method. My worry initially was performance, but I don't notice when making the call twice, and especially since they both using caching anyways

piyalbasu: Awesome! Yeah that's better. You're right about separating concerns

extension/src/popup/components/account/AssetDetail/styles.scss
```diff
@@ -1,6 +1,7 @@
 .AssetDetail {
   height: var(--popup--height);
-  overflow: hidden;
+  overflow-y: scroll;
```
piyalbasu, line 3: I think it would be better to use `simpleBar` instead of `overflow-y: scroll` so this matches the other scrolling areas.

Here's an example of how to implement: https://github.com/stellar/freighter/tree/master/extension/src/popup/components/accountHistory/HistoryList

acharb: good call 👍  https://github.com/stellar/freighter/pull/635/commits/391e0322cd82bfea472607f4f769e8aae8036c0b

also I just realized I think we've been hiding the bottom content of the asset detail view w/o a scroll lol

piyalbasu (approved):

PR #654, opened 2022-10-20
cleaning up some UI discrepancies

piyalbasu:
Screen Shot 2022-10-20 at 4 24 21 PM

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -376,7 +376,7 @@ export const ScamAssetWarning = ({
                       isValidatingSafeAssetsEnabled ? t("disable") : t("enable")
                     }`}{" "}
                     {t("this alert by going to")}{" "}
-                    <strong>{t("Settings > Security")}</strong>
+                    <strong>{t("Settings > Preferences")}</strong>
                   </p>
                 </div>
               </InfoBlock>
@@ -389,7 +389,7 @@ export const ScamAssetWarning = ({
               type="button"
               onClick={closeOverlay}
             >
-              {t("Got it")}
+              {isValidatingSafeAssetsEnabled ? t("Got it") : t("Cancel")}
             </Button>
             {isSendWarning && (
               <Button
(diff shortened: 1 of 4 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -376,7 +376,7 @@ export const ScamAssetWarning = ({
                       isValidatingSafeAssetsEnabled ? t("disable") : t("enable")
                     }`}{" "}
                     {t("this alert by going to")}{" "}
-                    <strong>{t("Settings > Security")}</strong>
+                    <strong>{t("Settings > Preferences")}</strong>
```
piyalbasu, line 379: The toggle is actually found under `Preferences`

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -389,7 +389,7 @@ export const ScamAssetWarning = ({
               type="button"
               onClick={closeOverlay}
             >
-              {t("Got it")}
+              {isValidatingSafeAssetsEnabled ? t("Got it") : t("Cancel")}
```
piyalbasu, line 392: The button's copy changes when we have validatingSafeAssetsEnabled toggled on

extension/src/popup/components/WarningMessages/styles.scss
```diff
@@ -154,14 +154,8 @@
...
-      background: var(--pal-error-background);
-      border-color: var(--pal-error-border);
-      --Button-color-background: var(--pal-error-background);
-      --Button-color-border: var(--pal-error-border);
-    }
-    .BasicButton--primary:hover {
-      background-color: var(--pal-error);
-      border-color: var(--pal-error);
+      background: #e5484d;
+      border-color: #e5484d;
```
piyalbasu, line 158: The new "red" button uses a different color according to the figma

quietbits (approved):

extension/src/popup/components/WarningMessages/styles.scss
acharb: ah nice catch, we'll still need the hover state though. Otherwise the color goes blue on hover

Screen Shot 2022-10-24 at 11 55 40 AM

not sure what the hover background color should be, maybe this?
```
    .BasicButton--primary:hover {
      background: #e5484d;
      border-color: #e5484d;
      background-color: var(--pal-error);
    }
```

PR #657, opened 2022-10-25
qa fixes

acharb:
- "BEFORE YOU ADD THIS ASSET" modal height should only be needed height
- show blocked domain modal first

https://user-images.githubusercontent.com/30449853/197889720-2281a9b4-ac82-4d94-9d1a-90d11b482f29.mov

https://user-images.githubusercontent.com/30449853/197889736-a7796542-4068-4639-9d93-b2954e4155b1.mov

### DIFF
--- a/extension/src/popup/components/WarningMessages/styles.scss
+++ b/extension/src/popup/components/WarningMessages/styles.scss
@@ -156,6 +156,13 @@
     .BasicButton--primary {
       background: #e5484d;
       border-color: #e5484d;
+      --Button-color-background: #e5484d;
+      --Button-color-border: #e5484d;
+    }
+
+    .BasicButton--primary:hover {
+      border-color: #e5484d;
+      background-color: var(--pal-error);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/WarningMessages/styles.scss
```diff
@@ -156,6 +156,13 @@
     .BasicButton--primary {
       background: #e5484d;
       border-color: #e5484d;
```
acharb, line 158: @piyalbasu  also looks like this got removed in https://github.com/stellar/freighter/pull/654/files 

but we need the hover state for the button otherwise turns blue. I went with `var(--pal-error);` hover color lmk if something else is better tho

piyalbasu: Ah yeah, you're right about the `hover`.

Why is 
 ```
 --Button-color-background: #e5484d;
      --Button-color-border: #e5484d;
 ``` 
 needed though?

piyalbasu: Ah you're right about `hover`

Let's use `--Button-color-background` and `--Button-color-border` since we're defining them

piyalbasu (approved):

PR #661, opened 2022-10-27
Feature/wal 456 start chrome storage

piyalbasu:


extension/public/static/manifest.json
```diff
@@ -41,5 +35,6 @@
     "48": "images/icon48.png",
     "128": "images/icon128.png"
   },
+  "permissions": ["storage"],
```
piyalbasu, line 38: enable browser storage in extension

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -910,35 +921,19 @@ export const popupMessageListener = (request: Request) => {
...
+        [NETWORK_ID]: defaultNetworkDetails,
+      });
     }
 
     localStorage.setItem(
       IS_EXPERIMENTAL_MODE_ID,
       JSON.stringify(isExperimentalModeEnabled),
     );
-
-    if (isExperimentalModeEnabled !== currentIsExperimentalModeEnabled) {
```
piyalbasu, line 920: As a result of a bade merge, this block was in here twice, so removing this 🤦

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -880,6 +884,13 @@ export const popupMessageListener = (request: Request) => {
 
     const currentIsExperimentalModeEnabled = getIsExperimentalModeEnabled();
 
+    await dataStorage.setItem({
+      [DATA_SHARING_ID]: isDataSharingAllowed,
+      [IS_VALIDATING_MEMO_ID]: isMemoValidationEnabled,
+      [IS_VALIDATING_SAFETY_ID]: isSafetyValidationEnabled,
+      [IS_VALIDATING_SAFE_ASSETS_ID]: isValidatingSafeAssetsEnabled,
+    });
```
piyalbasu, line 892: this is a preview of what the new "save" API will look like.

For now, recording new settings in both browser storage and local storage. We'll remove the latter as we continue ton refactor

extension/src/background/helpers/dataStorage.ts
```diff
@@ -0,0 +1,36 @@
...
+    // If the key is not found in storage, it will return the the value passed along in object as a default (in this case, "aValue")
+
+    const storageResult = await browser.storage.local.get(key);
+
+    return storageResult;
+  },
+  setItem: async (setItemParams: SetItemParams) => {
+    await browser.storage.local.set(setItemParams);
+  },
+};
```
piyalbasu, line 19: this is the new storage API wrapper. We'll eventually replace all instances of `localStorage` with `dataStorage`

extension/src/background/helpers/dataStorage.ts
```diff
@@ -0,0 +1,36 @@
...
+      const parsedValue = JSON.parse(v);
+      value = parsedValue;
+    } catch (e) {
+      // do not transform v
+    }
+
+    storage[k] = value;
+  });
+
+  await dataStorage.setItem(storage);
```
piyalbasu, line 35: this function migrates everything from localStorage to dataStorage. We'll run this every so often to keep the 2 data stores in sync (for the time being)

extension/src/background/index.ts
```diff
@@ -23,4 +24,6 @@ export const initMessageListener = () => {
 
     return res;
   });
+
+  browser.runtime.onInstalled.addListener(migrateLocalStorageToBrowserStorage);
```
piyalbasu, line 28: run this the migration every time we install. This will also run when the browser auto upgrades the extension when there is a new version available

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -951,9 +946,14 @@ export const popupMessageListener = (request: Request) => {
     };
   };
 
-  const loadSettings = () => {
-    const dataSharingValue = localStorage.getItem(DATA_SHARING_ID) || "true";
-    const isDataSharingAllowed = JSON.parse(dataSharingValue);
+  const loadSettings = async () => {
+    await migrateLocalStorageToBrowserStorage();
```
piyalbasu, line 950: `loadSettings` runs very often (every time you open the extension) so this seems like a good place to keep the 2 data stores in sync

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -951,9 +946,14 @@ export const popupMessageListener = (request: Request) => {
...
+
+    const {
+      [DATA_SHARING_ID]: isDataSharingAllowed,
+    } = await dataStorage.getItem({
+      [DATA_SHARING_ID]: true,
+    });
```
piyalbasu, line 956: this is a test case for storing something in the new data store!

`isDataSharingAllowed` will be stored and fetched from dataStorage.

extension/src/background/messageListener/popupMessageListener.ts
acharb: 😬  oof nice catch

extension/src/background/messageListener/popupMessageListener.ts
acharb: love how it can be done w/ one method call 💪 

can you add a todo line here since some of these lines are technically "repeat" for a short amount of time, so easy to remember later?

extension/src/background/helpers/dataStorage.ts
```diff
@@ -0,0 +1,36 @@
+import { browser } from "webextension-polyfill-ts";
```
acharb, line 1: dumb questions alert 🙂 : 
- is the idea for using webextension-polyfill for the background storage that it's using [chrome storage](https://developer.chrome.com/docs/extensions/reference/storage/) for chrome and firefox storage (whatever that is) for firefox? I tried glancing through the [repo](https://github.com/mozilla/webextension-polyfill) but didn't see where storage was defined
- we won't have to change localstorage in the extension right? the manifest v3 changes are only for changes of background script -> service worker right?

acharb (approved):
lgtm! had a Q but non-blocking

extension/src/background/helpers/dataStorage.ts
piyalbasu: Not dumb questions!

You're exactly right on both questions 😄

PR #663, opened 2022-11-01
analytics updates

acharb:
analytics updates [ticket](https://stellarorg.atlassian.net/browse/WAL-382?atlOrigin=eyJpIjoiZjUzZDZiNDY0YTFjNGI4OTgwZTRjZmUxMjkwNTA0M2YiLCJwIjoiaiJ9)

### DIFF
--- a/extension/src/helpers/metrics.ts
+++ b/extension/src/helpers/metrics.ts
@@ -3,7 +3,9 @@ import { Middleware, AnyAction } from "redux";
 import { createAsyncThunk, createAction } from "@reduxjs/toolkit";
 
 import { store } from "popup/App";
+import { METRICS_DATA } from "constants/localStorageTypes";
 import { settingsDataSharingSelector } from "popup/ducks/settings";
+import { AccountType } from "@shared/api/types";
 
 type metricHandler<AppState> = (state: AppState, action: AnyAction) => void;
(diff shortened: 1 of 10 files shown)


extension/src/popup/constants/metricsNames.ts
```diff
@@ -84,7 +84,10 @@ export const METRIC_NAMES = {
 
   accountScreenAddAccount: "account screen: created new account",
   accountScreenCopyPublickKey: "account screen: copied public key",
-  accountScreenImportAccount: "account screen: created new account",
+  accountScreenImportAccount: "account screen: imported new account",
```
acharb, line 87: pretty sure this should have been "imported" instead of "created" before, but we need the new metric now anyways

extension/src/helpers/metrics.ts
```diff
@@ -109,12 +125,21 @@ const getUserId = () => {
...
+    localStorage.getItem(METRICS_DATA) || "{}",
+  );
+
   cache.push({
     /* eslint-disable camelcase */
     event_type: name,
     event_properties: body,
     user_id: getUserId(),
     device_id: window.navigator.userAgent,
+    freighter_account_funded: metricsData.freighterFunded,
```
acharb, line 139: since we don't have a "dev" way to test this once this is deployed I can check amplitude to see if these fields are showing up right for new User Properties

PR #671, opened 2022-11-07
use abstracted function for localStorage calls in background

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-457?atlOrigin=eyJpIjoiYmQzZGVlNGVhZjczNGIzMWIxNjNlOTg5MWE5MTc4NmIiLCJwIjoiaiJ9)

so that all the localStorage calls can be triggered by same function


extension/src/background/helpers/dataStorage.ts
```diff
@@ -34,3 +34,11 @@ export const migrateLocalStorageToBrowserStorage = async () => {
 
   await dataStorage.setItem(storage);
 };
+
+// TODO - temporary wrapper around localStorage until we replace
+// localStorage all together
+export const freighterLocalStorage = {
```
piyalbasu, line 40: so the idea is that when we switch over to browser.storage, we can just change what's happening inside of this function. Because of that, I think it might be better if we call this something more generic since ideally this function won't be using localStorage anymore.

Maybe you can call it `backgroundDataStorage` (or something like that)?

extension/src/background/helpers/dataStorage.ts
```diff
@@ -34,3 +34,11 @@ export const migrateLocalStorageToBrowserStorage = async () => {
...
+  getItem: (keyId: string) => localStorage.getItem(keyId),
```
piyalbasu, line 41: Kind of annoying, but can you make these methods (getItem, setItem, clear) async? The browser.storage API is async, so at some point this will need to be async. It'd be nice if we started calling `await backgroundDataStorage` now so the call signature wouldn't have to change in the future

extension/src/background/helpers/dataStorage.ts
```diff
@@ -34,3 +34,11 @@ export const migrateLocalStorageToBrowserStorage = async () => {
...
+  setItem: (keyId: string, value: string) => localStorage.setItem(keyId, value),
```
piyalbasu, line 42: Could you also please make this method store in localStorage as well as in `dataStorage` above?

For ex:
```
  setItem: (keyId: string, value: string) => { localStorage.setItem(keyId, JSON.stringifu(value)); dataStorage.setItem({ [keyId]: value }) },

```

PR #673, opened 2022-11-11
fix issues with forgetting to await

piyalbasu:


extension/src/popup/ducks/accountServices.ts
```diff
@@ -173,7 +179,7 @@ export const recoverAccount = createAsyncThunk<
     });
   }
 
-  if (!res.publicKey) {
+  if (!res.publicKey || res.error) {
```
acharb, line 182: should we include the res.error here in the response?

acharb (approved):
ah thanks for catching these, just thad 1 thought otherwise lgtm!

PR #683, opened 2022-11-21
update key derivation num after recovering account

acharb:
bug fix from https://github.com/stellar/freighter/pull/679

need to update the key derivation number if found mnemonic accounts w/ a balance, otherwise creating a new account re-creates the already existing account

### DIFF
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -648,6 +648,8 @@ export const popupMessageListener = (request: Request) => {
             mnemonicPhrase: recoverMnemonic,
             imported: true,
           });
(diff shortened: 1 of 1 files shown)


piyalbasu (approved):

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -648,6 +648,8 @@ export const popupMessageListener = (request: Request) => {
             mnemonicPhrase: recoverMnemonic,
             imported: true,
           });
+          // eslint-disable-next-line no-await-in-loop
+          await dataStorageAccess.setItem(KEY_DERIVATION_NUMBER_ID, String(i));
```
piyalbasu, line 652: Good catch!

PR #684, opened 2022-11-23
add integration tests

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-553?atlOrigin=eyJpIjoiODhmZmI4MWU5OThhNDZlNTlhOWJkMDExMzlmZTk5NjAiLCJwIjoiaiJ9)

Pros of doing the integration tests in the browser is we can test directly the browser storage changes. As opposed to mocking it in the background and potentially giving false positives.

Con is it is more difficult to automate if we decide to do that, but can be addressed later.

notes: 
- the integration tests are only accessibly when running the app locally, and when the extension is built with `experimental` mode to prevent accidental data clears.
- one share/api/internal method not tested: `submitFreighterTransaction`, because can't run in an idempotent way (w/o having to make calls to the network, and it's a straightforward func anyways)

Directions to run added to readme

https://user-images.githubusercontent.com/30449853/203531526-d12a91fe-0a2b-4174-b4f1-e25d3f5ae962.mov


acharb:
@piyalbasu lmk what you think, open to suggestions if you have ideas to make most useful

piyalbasu (approved):
This is a really great start!!

In future iterations, we can start adding more specific use cases (for ex: `updateAccountName` returning the updated name as part of the `allAccounts` arr)

PR #692, opened 2022-12-10
upgrade web-extension-polyfill and upgrade manifest version

piyalbasu:
⚠️ This PR is just trying to get the code to build with the new libs and manifest version. There are still numerous issues with data not being transmitted from background to popup properly, so the app doesn't work at this state. Those will be fixed in the next PR ⚠️

### DIFF
--- a/@shared/api/helpers/extensionMessaging.ts
+++ b/@shared/api/helpers/extensionMessaging.ts
@@ -1,11 +1,10 @@
-import { browser } from "webextension-polyfill-ts";
+import browser from "webextension-polyfill";
 import {
   DEV_SERVER,
   EXTERNAL_MSG_RESPONSE,
   EXTERNAL_MSG_REQUEST,
 } from "../../constants/services";
 import { Response } from "../types";
-import { NoExtensionInstalledError } from "../../constants/errors";
 
 export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
   /* 
@@ -19,11 +18,7 @@ export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
     { source: EXTERNAL_MSG_REQUEST, messageId: MESSAGE_ID, ...msg },
     window.location.origin,
   );
-  return new Promise((resolve, reject) => {
-    if (!window.freighter) {
-      reject(new NoExtensionInstalledError());
-    }
-
+  return new Promise((resolve) => {
     const messageListener = (event: { source: any; data: Response }) => {
       // We only accept messages from ourselves
       if (event.source !== window) return;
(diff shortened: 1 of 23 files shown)


@shared/api/helpers/extensionMessaging.ts
```diff
@@ -1,11 +1,10 @@
-import { browser } from "webextension-polyfill-ts";
+import browser from "webextension-polyfill";
```
piyalbasu, line 1: `webextension-polyfill-ts` is deprecated. The latest has been rolled into `webextension-polyfill`. I was having trouble building without upgrading to the latest

@shared/api/helpers/extensionMessaging.ts
```diff
@@ -19,11 +18,7 @@ export const sendMessageToContentScript = (msg: {}): Promise<Response> => {
     { source: EXTERNAL_MSG_REQUEST, messageId: MESSAGE_ID, ...msg },
     window.location.origin,
   );
-  return new Promise((resolve, reject) => {
-    if (!window.freighter) {
```
piyalbasu, line 23: We no longer have the ability to set `window.freighter` on dapps with manifest version 3

extension/src/background/helpers/dataStorage.ts
```diff
@@ -1,52 +1,32 @@
...
-    await dataStorage.getItem({ [keyId]: "" });
-    return localStorage.getItem(keyId);
-  },
+  getItem: (keyId: string) => dataStorage.getItem(keyId),
   setItem: async (keyId: string, value: string) => {
     await dataStorage.setItem({ [keyId]: value });
-    localStorage.setItem(keyId, value);
   },
-  clear: () => localStorage.clear(),
+  clear: () => dataStorage.clear(),
```
piyalbasu, line 31: removing references to `localStorage` as a service worker does not support it

extension/public/contentScript.ts
```diff
@@ -1,5 +1,3 @@
-import { setIsConnected } from "contentScript/helpers/setIsConnected";
 import { redirectMessagesToBackground } from "contentScript/helpers/redirectMessagesToBackground";
 
-setIsConnected();
```
piyalbasu, line 4: this was the helper that would set `window.freighter = true` on dapps

acharb (approved):
lgtm!

had 1 question non-blocking i don't think

@shared/api/helpers/extensionMessaging.ts
```diff
@@ -45,7 +40,7 @@ export const sendMessageToBackground = async (msg: {}): Promise<Response> => {
     // treat this as an external call because we're making the call from the browser, not the popup
     res = await sendMessageToContentScript(msg);
   } else {
-    res = await browser.runtime.sendMessage(msg);
+    res = await browser?.runtime.sendMessage(msg);
```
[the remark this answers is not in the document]
piyalbasu: Good call - at one point I was having issues getting the codebase to build without this (since it's building in a node environment), but this no longer seems to be an issue.

You're right, there shouldn't be a time when the app is running and browser does not exist

PR #700, opened 2022-12-23
remove JSON parsing and stringifying

piyalbasu:


extension/src/background/helpers/cachedFetch.ts
```diff
@@ -10,24 +10,28 @@ export const cachedFetch = async (url: string, storageKey: string) => {
...
-    (await dataStorageAccess.getItem(storageKey)) || "{}",
-  );
+  let directoryLookup = (await dataStorageAccess.getItem(storageKey)) || "{}";
+
+  if (typeof directoryLookup === "string") {
+    try {
+      const directoryLookupJSON = JSON.parse(directoryLookup);
+      directoryLookup = directoryLookupJSON;
+    } catch (e) {
+      console.error(`cachedFetch JSON parse error: ${e}`);
```
piyalbasu, line 20: I noticed that `cachedAssetIconsId` was still being stored as a string, I put in this check (similar to one @acharb had proposed) to try to JSON parse if needed to make sure the output here is always an object

quietbits (approved):

PR #701, opened 2023-01-03
connecting sep24 UI to test anchor

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-578?atlOrigin=eyJpIjoiZTcyZWNjYjAwZWMxNDgwZTgxMzI1NjY0NDhjMzcyNGUiLCJwIjoiaiJ9)

limitation is that where the polling happens will have to change. Right now it starts polling from the MoneyGram component, but that view closes when a new tab opens. I was thinking it should start from the `/account` view, since that's where the ["sep-24 notification" is shown](https://www.figma.com/file/BR8lyECrgUHybq7XuQSEED?node-id=4800:12150#293081588)

https://user-images.githubusercontent.com/30449853/210432776-62755a38-6c7d-4685-a932-a49f5d7de969.mov

### DIFF
--- a/extension/src/popup/views/BuyAsset/MoneyGram/index.tsx
+++ b/extension/src/popup/views/BuyAsset/MoneyGram/index.tsx
@@ -1,70 +1,286 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
+import { useDispatch, useSelector } from "react-redux";
 import { useTranslation } from "react-i18next";
 import { TextLink } from "@stellar/design-system";
+import StellarSdk, { Account, StellarTomlResolver } from "stellar-sdk";
 
+import { xlmToStroop } from "helpers/stellar";
+import { emitMetric } from "helpers/metrics";
+import {
+  signFreighterTransaction,
+  submitFreighterTransaction,
+  startHwSign,
+  ShowOverlayStatus,
+  transactionSubmissionSelector,
+} from "popup/ducks/transactionSubmission";
+import { LedgerSign } from "popup/components/hardwareConnect/LedgerSign";
+import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
+import {
+  publicKeySelector,
+  hardwareWalletTypeSelector,
+} from "popup/ducks/accountServices";
+import { AppDispatch } from "popup/App";
 import { Button } from "popup/basics/buttons/Button";
 import { SubviewHeader } from "popup/components/SubviewHeader";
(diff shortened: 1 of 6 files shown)


extension/src/popup/views/BuyAsset/MoneyGram/index.tsx
```diff
@@ -1,70 +1,286 @@
...
+
+  useEffect(() => {
+    (async () => {
+      const signedXdr = freighterSignedXDR || hwSignedXDR;
+      if (signedXdr) {
+        const token = await getAuthToken(signedXdr);
+        startSep24Deposit(token);
+      }
+    })();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
acharb, line 64: not super happy about needing this. But otherwise all the helper functions below will have to either be defined in the useEffect or each get their own useCallback which gets pretty ugly, so stuck with the eslint comment

piyalbasu: So this actually meshes with my overall suggestion for this PR, which is to break out some of the logic out of this component into helpers to separate some of the UI concerns from the SEP-24 API concerns. This will make the code a little more modular while also suppressing the exhaustive-deps warning since the functions will be outside of the component and thus won't re-render:

1) `getAuthToken` can just be a standalone function outside of the component as it is stateless

2) I think `startSep24Deposit` should be a redux action and live inside the `transactionSubmission` duck. We'll need to move where polling starts (likely in `/account`) as you suggested, but that means the deposit's `id` and `token` will need to be saved in storage to be fetched later. `startSep24Deposit` can store the `id` and `token` in Background and update the redux store to let UI know that we have a deposit in process. 

Then, when the user reopens the app, it will load storage and check redux to see if it has an `id`/`token` pair in process, and `startPolling` based on that in the `Account` component. Then it removes the `id`/`token` pair on completion

piyalbasu: This can all be done in a follow up PR btw! Otherwise, this code mostly looks good 👍

piyalbasu (approved):

extension/src/popup/views/BuyAsset/MoneyGram/index.tsx
acharb: yup great suggestions. I think that polling mechanism works. I think it might get hairy if there's multiple deposits in process? 🤔 I'll explore/do in the next PR

PR #702, opened 2023-01-13
upgrade stellar-sdk to handle service workers

piyalbasu:
Upgrading stellar-sdk re-enables tx signing from dapps

### DIFF
--- a/@shared/api/helpers/getIconUrlFromIssuer.ts
+++ b/@shared/api/helpers/getIconUrlFromIssuer.ts
@@ -78,28 +78,20 @@ export const getIconUrlFromIssuer = async ({
 
   if (toml.CURRENCIES) {
     /* If we find some currencies listed, check to see if they have the currency we're looking for listed */
-    toml.CURRENCIES.every(
-      async ({
-        code: currencyCode,
-        image,
-      }: {
-        code: string;
-        image: string;
-      }) => {
-        if (currencyCode === code && image) {
-          /* We found the currency listing in the toml. 3. Get the image url from it */
-          iconUrl = image;
-          /* And also save into the cache to prevent having to do this process again */
-          await sendMessageToBackground({
-            assetCanonical: `${code}:${key}`,
-            iconUrl,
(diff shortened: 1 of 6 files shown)


@shared/api/helpers/getIconUrlFromIssuer.ts
```diff
@@ -78,28 +78,20 @@ export const getIconUrlFromIssuer = async ({
...
-            assetCanonical: `${code}:${key}`,
-            iconUrl,
-            type: SERVICE_TYPES.CACHE_ASSET_ICON,
-          });
-          return false;
-        }
-        return true;
-      },
-    );
+    toml.CURRENCIES.every(async ({ code: currencyCode, image }) => {
```
piyalbasu, line 81: making this async

leighmcculloch: This code diff doesn't look like it changes the async qualities of this code. It's already async, this just looks like different formatting. Is that correct?

leighmcculloch (approved):
Question above, otherwise LGTM.

@shared/api/helpers/getIconUrlFromIssuer.ts
leighmcculloch: Oh I see, there is a difference. I have no idea what that difference means though 😅.

acharb: @piyalbasu  ya I'm also confused here 🤔 . Is the only difference adding the return type in the async func? But isn't it returning the same thing?

piyalbasu: Apologies all - this is a few levels of bad engineering on my part. I chipped away at this PR over the course of weeks and lost track of what I was doing and made an erroneous comment 😞 And then, doubly bad, I let this PR sit here and did not address the comments. Very sorry about that - textbook bad PR hygiene.

The change is *not* making this function async, as you both noted. The change is removing the type because it is now inferring the type from toml.CURRENCIES which has `code` and `image` as optional params instead of required params like the type I removed described

PR #704, opened 2023-01-18
[WAL-233] Limit the number of chars in the amount field

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/browse/WAL-233

Adds `TX_SEND_MAX` constant and uses it to validate the max send amount and show the appropriate error message.
Lowers the `maxDigits` to 12, down from 16.

### DIFF
--- a/extension/src/popup/constants/transaction.ts
+++ b/extension/src/popup/constants/transaction.ts
@@ -1,0 +1,1 @@
+export const TX_SEND_MAX = "922337203685.4775807" as const;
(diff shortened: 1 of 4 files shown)


aristidesstaffieri:
@piyalbasu I didn't find docs on the max amount (922,337,203,685.4775807), was wondering if that would be the tx max send amount or some other limit?

extension/src/popup/constants/transaction.ts
```diff
@@ -0,0 +1 @@
+export const TX_SEND_MAX = "922337203685.4775807" as const;
```
piyalbasu, line 1: jc - why was `as const` necessary here?

piyalbasu: Ah gotcha - cool, you can leave as is 👍

piyalbasu (approved):

PR #705, opened 2023-01-18
[BUG] Don't lose cursor placement after sanitization/formatting

aristidesstaffieri:
This is likely pretty edge case but right now, in `SendAmount`, if you happen to need to change digits that are not the last character in the amount string, when `formatAmount` runs on change the user will experience a cursor reset. This is typically opaque because almost always when you are typing the amount your cursor is already on the last character where it gets reset after updating the amount value.

Example - 

https://user-images.githubusercontent.com/6886006/213299353-8b0ae370-7dc1-4698-b0ad-d5ddc3387c6f.mov

It might be hard to tell, but when I type the decimals in first and then go to type digits, my cursor is reset after every additional `,` added by the formatter in `formatAmount`.

To avoid this, we can track the cursor location by accounting for characters that have been removed by sanitization or added by formatting in order to move the cursor the correct place.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -26,8 +26,10 @@ import { useNetworkFees } from "popup/helpers/useNetworkFees";
 import { useIsSwap } from "popup/helpers/useIsSwap";
 import { LP_IDENTIFIER } from "popup/helpers/account";
 import { emitMetric } from "helpers/metrics";
+import { useRunAfterUpdate } from "popup/helpers/useRunAfterUpdate";
 import { SubviewHeader } from "popup/components/SubviewHeader";
 import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
+import { cleanAmount, formatAmount } from "popup/helpers/formatters";
 import {
   transactionSubmissionSelector,
   saveAmount,
@@ -103,6 +105,7 @@ export const SendAmount = ({
   const { t } = useTranslation();
   const dispatch: AppDispatch = useDispatch();
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const runAfterUpdate = useRunAfterUpdate();
 
   const {
     accountBalances,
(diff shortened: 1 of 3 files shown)


aristidesstaffieri:
Actually the way I set the cursor isn't quite right yet, I'm going to put a WIP tag on this while I sort that out.

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -301,7 +303,29 @@ export const SendAmount = ({
   // remove non digits and decimal
   const cleanAmount = (s: string) => s.replace(/[^0-9.]/g, "");
 
-  const formatAmount = (val: string) => {
+  const preserveCursor = (
```
aristidesstaffieri, line 306: I've tried to house the hairy logic into this one fn called preserveCursor.
Basically, this is my logic for tracking where the cursor should be after updates/clean/sanitize.

1. Determine wether we need to sanitize digits and/or decimals.
2. For digits only, compare the previous value with the newly formatted value and move the cursor according to the difference in number of commas, and difference in number of characters in the raw value vs the cleaned/sanitized(even though we filter out anything but numbers, invalid chars still move the cursor).
3. If digits & decimals, do previous step on chars before the dot and also account for characters after the dot that are cleaned out but have moved the cursor.

aristidesstaffieri:
This should be code complete but I'm still looking into why the translations action in the build fails, seems to run ok locally so not totally sure yet.

extension/src/popup/components/sendPayment/SendAmount/index.tsx
piyalbasu: Nice - this is a really good usability improvement. Maybe we should move `preserveCursor` and `formatAmount` to a helpers file as we'd probably want to reuse this for future currency inputs and it'd be nice to separate some of this complex logic from the tsx markup. Wdyt?

piyalbasu: Would also be great to add your explanation of the logic from this PR comment into a comment in the code as I found it very helpful to understand what was happening

aristidesstaffieri: Yeah totally, makes sense. I'll move these out into helpers and was going to take a stab at writing tests for them since they're kind of hairy.

acharb (approved):
lgtm! thanks for taking this ticket I definitely was looking forward for this getting fixed 😄

acharb:
> This should be code complete but I'm still looking into why the translations action in the build fails, seems to run ok locally so not totally sure yet.

that is weird it's working for me locally as well. It looks like the build translations command is failing during the build? 🤨 I can help debug tomorrow if you haven't figured it out by then

acharb:
ah looks like the branch needs to be updated with 2.9.2, and then `formatAmount` method needs that second argument which is what's breaking the build

Just tried it in a test PR [here](https://github.com/stellar/freighter/pull/708) and the build works then 👍 

Normally the PR won't run the build in github unless you've got the latest from the base branch (I thought), but in this case it did so the error was kind of misleading 🤷

aristidesstaffieri:
> ah looks like the branch needs to be updated with 2.9.2, and then `formatAmount` method needs that second argument which is what's breaking the build
> 
> Just tried it in a test PR [here](https://github.com/stellar/freighter/pull/708) and the build works then 👍
> 
> Normally the PR won't run the build in github unless you've got the latest from the base branch (I thought), but in this case it did so the error was kind of misleading 🤷

Oh interesting, thank you!!

extension/src/popup/components/sendPayment/SendAmount/index.tsx
aristidesstaffieri: This is done now, going to close this thread.

piyalbasu (approved):

PR #706, opened 2023-01-20
[WAL-68] Suggest removing trustline after sending entire balance

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/browse/WAL-68

When a user successfully submits a tx that results in that assets balance being empty for that user, they will now see an `InfoBlock` suggesting that they remove the trustline to that asset, with an inline button to take the action to remove it.

Notable Changes - 
In order to set the `errorAsset` for `TrustlineError` outside of the sub routes defined in `ManageAssets`, we now pull the error asset from the tx error state by parsing out the xdr we get back in the error response. This decouples the errorAsset from the call site, allowing us to just navigate to the trustline errors page at any point without having to explicitly set the `errorAsset`.


extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -67,13 +84,95 @@ export const SubmitSuccess = ({ viewDetails }: { viewDetails: () => void }) => {
...
+        dispatch(
+          getAccountBalances({
+            publicKey,
+            networkDetails,
+          }),
+        );
+        trackChangeTrustline();
+        dispatch(resetSubmission());
+        navigateTo(ROUTES.account);
+      }
```
acharb, line 160: we shouldn't need to getAccountBalances and resetSubmission, since when we load /account page it's done there. I noticed we're doing that in the ManageAssetRows/index file though, I don't think it's needed there as well

```suggestion
      if (submitFreighterTransaction.fulfilled.match(submitResp)) {
        trackChangeTrustline();
        navigateTo(ROUTES.account);
      }
```

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
```diff
@@ -88,13 +88,43 @@
...
+
+      button {
+        font-size: 14px;
+        font-weight: bold;
+        border: 0;
+        padding: 0;
+        text-align: initial;
+        margin-top: 10px;
+        background-color: transparent;
+        text-decoration: underline;
```
acharb, line 114: can we add `cursor: pointer` as well?

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -67,13 +84,95 @@ export const SubmitSuccess = ({ viewDetails }: { viewDetails: () => void }) => {
...
+
+  const suggestRemoveTrustline =
+    accountBalances.balances &&
+    accountBalances.balances[asset].available.isZero();
 
   return (
     <div className="SubmitResult">
       <div className="SubmitResult__header">
-        {t("Successfuly")} {isSwap ? t("swapped") : t("sent")}
+        {t("Successfully")} {isSwap ? t("swapped") : t("sent")}
```
acharb, line 175: nice catch 😄

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -13,12 +14,35 @@ import { TrustlineError } from "popup/components/manageAssets/TrustlineError";
...
+  React.useEffect(() => {
+    async function parseXdr() {
+      const { networkPassphrase } = await getNetworkDetails();
+      const xdrEnvelope = error?.response?.extras.envelope_xdr;
+      if (xdrEnvelope) {
+        const parsedTx = StellarSdk.TransactionBuilder.fromXDR(
+          xdrEnvelope,
+          networkPassphrase,
+        );
+        const { code, issuer } = parsedTx._operations[0].line;
```
acharb, line 37: really like this parsing the error tx xdr change/solution!

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -48,14 +48,32 @@ export const signFreighterTransaction = createAsyncThunk<
 
 export const submitFreighterTransaction = createAsyncThunk<
   Horizon.TransactionResponse,
-  { signedXDR: string; networkDetails: NetworkDetails },
+  {
+    publicKey: string;
+    signedXDR: string;
+    networkDetails: NetworkDetails;
+    refreshBalances?: boolean;
```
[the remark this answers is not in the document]
aristidesstaffieri: well I considered separating the actions into 2 actions, one where you can submit the tx standalone and one where you can submit the tx and refresh balances in one action. The reason is that I need a way to do both in one action lifecycle in order to avoid the "flash of screen" problem that happens if I first use the `submitFreighterTransaction` then the `getAccountBalances` since they have different lifecycles for each action(they each have their own pending/fulfilled/rejected). Typically I would do this with a layer that can compose actions like in redux-saga for example so maybe there is redux toolkit pattern for this that I haven't found yet?

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -13,12 +14,35 @@ import { TrustlineError } from "popup/components/manageAssets/TrustlineError";
...
+  const {
+    accountBalances,
+    destinationBalances,
+    assetSelect,
+    error,
+  } = useSelector(transactionSubmissionSelector);
   const [errorAsset, setErrorAsset] = useState("");
 
+  React.useEffect(() => {
+    async function parseXdr() {
```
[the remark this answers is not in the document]
aristidesstaffieri: yup, sure thing!

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
aristidesstaffieri: makes sense, thanks!

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/styles.scss
aristidesstaffieri: oh yup, nice catch.

extension/src/popup/ducks/transactionSubmission.ts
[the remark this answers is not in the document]
acharb: ah gotchya, I haven't used redux-saga before (looks cool) but, according to [RTK docs](https://redux.js.org/usage/side-effects-approaches) here looks like side effects are advised to be handled in middleware.

But in this case I'm not sure the side effect is needed? 🤔  Since `refreshBalances=true` is only being used in` TransactionDetails.index`, but the account balance isn't shown in the UI anywhere. Only when they go back to the /account view it's shown, but we're already dispatching getAccountBalances there.

wdyt? I may be missing something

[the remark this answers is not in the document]
aristidesstaffieri: Oh yeah, so the subtlety here is that we need to know the updated balance of the asset from the tx in order to know wether we need to show the InfoBlock to suggest removing the trustline when a user's balance for an asset is empty due to the tx.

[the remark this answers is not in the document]
acharb: ahh right I missed that. Ok that makes sense then thx for explaining

acharb (approved):
lgtm!

piyalbasu (approved):

PR #707, opened 2023-01-23
poll anchor and handle sep24 user todos

acharb:
handle the logic of storing sep24 deposit state in background, and polling anchor for updates + displaying todos for the user

https://user-images.githubusercontent.com/30449853/214017947-0d5160fa-e0dd-4a7c-b0db-b6e732c723ec.mov

❗️ note: there's a race condition between the sep24 state and the state the anchor holds for the user. 

eg: user starts deposit, fills out form and submits. And there's a delay (like 5-10s for test anchor) of test anchor still hasn't processed  the deposit and leaves it in INCOMPLETE. If the user polls the anchor during that delay, the next step for the user is start a new deposit - thus creating a second deposit even though the first is still active/valid. 

(another similar case: if the user deletes their extension data before finishing the deposit process)

I think this is something we'll have to think through for real anchors (and will depend on their implementation) to make sure users don't accidentally deposit multiple times. Possibly we should save multiple deposits (maybe adding a max?) for each account. Something to think about during the real anchor integration.


extension/src/background/helpers/dataStorage.ts
```diff
@@ -16,6 +16,9 @@ export const dataStorage = {
   setItem: async (setItemParams: SetItemParams) => {
     await browser.storage.local.set(setItemParams);
   },
+  removeItem: async (key: string) => {
+    await browser.storage.local.remove(key);
```
acharb, line 20: @piyalbasu is this right? tbh I'm not sure where the methods for browser.storage.local could be found

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -431,6 +488,10 @@ const transactionSubmissionSlice = createSlice({
     saveBuyAsset: (state, action) => {
       state.buyAsset = action.payload;
     },
+    // TODO - separate sign status from submit status so shouldnt need this method
```
acharb, line 491: https://stellarorg.atlassian.net/browse/WAL-612?atlOrigin=eyJpIjoiN2IwMzg4Mzk2NWViNDU2YjkzZjE3NDYwZWMxMmM3NDIiLCJwIjoiaiJ9

extension/src/popup/helpers/sep24.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (endStatuses.includes(currentStatus)) {
+      return currentStatus;
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+  }
+};
+
+export const addTrustline = async ({
```
acharb, line 119: added a tech debt ticket to consolidate the add trustline logic since we're doing it in separate places atm
https://stellarorg.atlassian.net/browse/WAL-613?atlOrigin=eyJpIjoiMjRhMzNlN2Y2MjVjNGZiYTljN2Q2ODg0ZTZjY2M1YjIiLCJwIjoiaiJ9

extension/src/popup/styles/global.scss
```diff
@@ -3,6 +3,7 @@
   --z-index--loader: -1;
   --z-index--scam-warning: 20;
   --z-index--ledger: 30;
+  --z-index-account-list: 40;
```
acharb, line 6: this was needed now that the ledgerSign component is used on the /account page. So had to make that parent the same z-index level as ledger, and make the account-list z-index above it

extension/src/popup/Router.tsx
```diff
@@ -244,9 +244,9 @@ export const Router = () => {
     <HashRouter>
       <RouteListener />
       <Switch>
-        <PublicKeyRoute exact path={ROUTES.account}>
+        <PrivateKeyRoute exact path={ROUTES.account}>
```
acharb, line 247: side effect of the /account view now needing to sign txns for the user for sep10 token, so the /account page now needs the private key active

really only changes when user switches accounts. I didn't find it very annoying while testing

extension/src/background/helpers/dataStorage.ts
piyalbasu: This is correct. The docs on this are not easy to find.

Here is the doc for chrome.storage: https://developer.chrome.com/docs/extensions/reference/storage/

And browser.storage is based off of that (with, weirdly, a few small differences): https://github.com/mozilla/webextension-polyfill

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1110,6 +1111,58 @@ export const popupMessageListener = (request: Request) => {
...
+    );
+    return { sep24Data: res };
+  };
+
+  const storeSep24Status = async () => {
+    const { status } = request;
+    const publicKey = publicKeySelector(store.getState());
+    const res = JSON.parse(
+      (await dataStorageAccess.getItem(`${SEP24_DATA}:${publicKey}`)) || "{}",
+    );
```
piyalbasu, line 1153: just a heads up, I believe you will need to remove the `JSON.parse` wrapper once you merge in 3.0.0. Sorry in advance for any headaches merging causes 😞

extension/src/background/helpers/dataStorage.ts
```diff
@@ -46,5 +49,9 @@ export const dataStorageAccess = {
     await dataStorage.setItem({ [keyId]: value });
     localStorage.setItem(keyId, value);
   },
+  remoteItem: async (keyId: string) => {
+    await dataStorage.removeItem(keyId);
+    localStorage.removeItem(keyId);
+  },
```
piyalbasu, line 55: Just a heads up, I think you're going to run into some issues when merging with Release/3.0.0

For ex, Release/3.0.0 has no localStorage calls at all now

Sorry for all the headaches rebasing to 3.0.0 is going to cause 😞 I can help you sort through the madness

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1110,6 +1111,58 @@ export const popupMessageListener = (request: Request) => {
...
+    return { sep24Data: res };
+  };
+
+  const storeSep24Status = async () => {
+    const { status } = request;
+    const publicKey = publicKeySelector(store.getState());
+    const res = JSON.parse(
+      (await dataStorageAccess.getItem(`${SEP24_DATA}:${publicKey}`)) || "{}",
```
piyalbasu, line 1152: related to my comment above, after merging in 3.0.0, you'll be able to (and probably should) just set/get things in storage as their actual data type rather than JSON.stringifying before and JSON.parsing after

extension/src/popup/Router.tsx
piyalbasu: Ah good catch - let's keep an eye on if this becomes a problem

extension/src/popup/components/sep24/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+  let onClick = () => {};
+
+  switch (todo) {
+    case Sep24Status.INCOMPLETE:
+      message = (
+        <>
+          <div className="Sep24Todo__title">
+            You have an unfinished deposit form
+          </div>
+          <div>click here to finish</div>
```
piyalbasu, line 65: super NIT and you're probably already thinking of this: Might be nice to make this a reusable component

extension/src/popup/helpers/sep24.ts
```diff
@@ -0,0 +1,180 @@
...
+}) => {
+  let currentStatus = status as Sep24Status;
+  const endStatuses = [
+    Sep24Status.INCOMPLETE,
+    Sep24Status.PENDING_TRUST,
+    Sep24Status.ERROR,
+  ];
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
```
piyalbasu, line 89: assuming `true` is placeholder for now?

piyalbasu (approved):

extension/src/popup/components/sep24/index.tsx
acharb: yup I think that is better 👍 
https://github.com/stellar/freighter/pull/707/commits/21e047ed8b71569f98b35e4443c65df645ec0669

extension/src/popup/helpers/sep24.ts
acharb: actually I thought of keeping it as is. The `startSep24Polling` should only be called if there's sep24 data in the storage, meaning there's a sep24 todo that needs to be resolved with the anchor. So the only polling ending case is the anchor giving use one of the "endStatuses" (or COMPLETED).

To be safe maybe we only ping a max number of times? Like 10? wdyt?

acharb:
> Just a heads up, I think you're going to run into some issues when merging with Release/3.0.0
> 
> For ex, Release/3.0.0 has no localStorage calls at all now
> 
> Sorry for all the headaches rebasing to 3.0.0 is going to cause 😞 I can help you sort through the madness

conflicts actually don't look that bad. However after merging in 3.0.0 the build break for some reason, not sure why yet. I'm gonna merge in this PR, and then start a fresh one for updating with 3.0.0

extension/src/background/helpers/dataStorage.ts
acharb: ah gotchya. I saw those docs and totally whiffed on seeing the `remove` method. See it now though thx!

acharb:
https://github.com/stellar/freighter/pull/712

PR #712, opened 2023-01-30
update with release 3.0.0

acharb:
update the sep24 code with release/3.0.0


extension/src/background/helpers/dataStorage.ts
```diff
@@ -1,57 +1,35 @@
-import { browser } from "webextension-polyfill-ts";
```
acharb, line 1: only merge conflicts resolved were in this file

acharb:
cc @piyalbasu

PR #716, opened 2023-01-31
use fetch in background script when calling stellar

acharb:
part of [3.0.0 QA](https://stellarorg.atlassian.net/wiki/spaces/PM/pages/4501078017/release+3.0.0+QA) - fixing `recovering account didn't add first couple of mnemonic accounts` 

looks like there's a bug with axios in service workers: https://stackoverflow.com/a/69384484

It looks like there could be some workarounds, but in the mean time this is an easy change.

### DIFF
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -619,7 +619,7 @@ export const popupMessageListener = (request: Request) => {
 
           // eslint-disable-next-line no-await-in-loop
           const resp = await fetch(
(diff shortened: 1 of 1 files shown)


acharb:
_I think_ the only other spot this may effect is here: https://github.com/stellar/freighter/blob/e75ec3358c6fd91fb8169e108db4a7cc1d02e53c/extension/src/background/messageListener/freighterApiMessageListener.ts#L162 (only other place we use `server` in the background)

I can't test right now because I can't get the isConnected working atm, but I think we should check here too

piyalbasu (approved):

PR #717, opened 2023-01-31
[BUG] clears auth error on unmount of add account, fixes typo

aristidesstaffieri:
Fixes two minor things:

1. Adds an action on `accountServices` to clear the error state, then uses it on dismount of AddAccount to clear the error in the reducer.
2.  Fixes typo on mnemonic phrase confirmation.

### DIFF
--- a/extension/src/popup/views/AddAccount/AddAccount/index.tsx
+++ b/extension/src/popup/views/AddAccount/AddAccount/index.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Input } from "@stellar/design-system";
 import { Field, Form, Formik, FieldProps } from "formik";
@@ -16,7 +16,11 @@ import { PopupWrapper } from "popup/basics/PopupWrapper";
 
 import { SubviewHeader } from "popup/components/SubviewHeader";
 
-import { addAccount, authErrorSelector } from "popup/ducks/accountServices";
+import {
+  addAccount,
+  authErrorSelector,
+  clearApiError,
+} from "popup/ducks/accountServices";
 
 import "./styles.scss";
 
@@ -45,6 +49,10 @@ export const AddAccount = () => {
       navigateTo(ROUTES.account);
     }
   };
+
+  useEffect(() => () => (dispatch(clearApiError()) as unknown) as void, [
+    dispatch,
+  ]);
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     },
   ],
   rules: {
+    "arrow-body-style": "off",
```
aristidesstaffieri, line 17: Looks like this rule prevents the syntax needed in order to return from `useEffect` which is how to trigger an action on dismount.

extension/src/popup/views/AddAccount/AddAccount/index.tsx
```diff
@@ -46,6 +50,12 @@ export const AddAccount = () => {
     }
   };
 
+  React.useEffect(() => {
```
piyalbasu, line 53: Just to match style - I'd `import {useEffect} from "react";`

and just call `useEffect` here

extension/src/popup/views/AddAccount/AddAccount/index.tsx
```diff
@@ -46,6 +50,12 @@ export const AddAccount = () => {
...
+    return function cleanup() {
+      dispatch(clearApiError());
```
piyalbasu, line 55: Any reason to not just `return () => {`?

It would prevent you from having to change the eslint rule

aristidesstaffieri: hmm, it didn't seem to. I believe it's complaining about the use of the curly brackets in the return instead of an implicit return but if I do that then I get a pretty cryptic type error that I haven't understood yet.

extension/src/popup/views/AddAccount/AddAccount/index.tsx
aristidesstaffieri: yup, will do

extension/src/popup/views/AddAccount/AddAccount/index.tsx
aristidesstaffieri: I did find one way to avoid removing the rule which is by doing this -

```
useEffect(() => {
    return () => dispatch(clearApiError()) as unknown as void;
  }, [dispatch]);
```

Don't love it but the type of the return sig of useEffect doesn't seem to be able to infer the return correctly there.

aristidesstaffieri: This works but I do have to cast the return value since the hook can't infer it here
`useEffect(() => () => dispatch(clearApiError()) as unknown as void, [dispatch]);`

.eslintrc.js
aristidesstaffieri: removed this in favor of the implicit return syntax

piyalbasu (approved):

PR #718, opened 2023-02-01
[BUG] Don't display negative balance when balance ≈ min balance

aristidesstaffieri:
Right now if you send the max allowed XLM, leaving you with only the min balance required according to your account's subentries, then go to the swap/send tab and  depending on the current fee you could see your available xlm balance be -0.00001.

In order to fix this we can check for an available balance that is less that the min balance and if so, display a 0 available balance as that is effectively what they have.

I've also added a check to avoid doing a call to `getBestPath` in the case where the source amount is 0 since that will always  fail and you cannot complete a swap or send in that scenario anyways.

Screenshot 2023-02-01 at 9 30 11 AM

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -45,6 +45,8 @@ import { ScamAssetWarning } from "popup/components/WarningMessages";
 
 import "../styles.scss";
 
+const BASE_RESERVE = 0.5 as const;
+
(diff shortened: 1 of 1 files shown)


aristidesstaffieri:
As a side note - I wonder if it would be useful to have a link or help text somewhere in swap/send explaining or leading users to a doc so they can understand why their available balance does not match the balance they see on the accounts tab?

piyalbasu (approved):

PR #719, opened 2023-02-01
More consistent loading on Accounts

aristidesstaffieri:
Right now, If you navigate to the Account(Home) tab you might notice that there is a very quick "flash of screen".

This happens because we use the `accountBalanceStatus` to conditionally render the page body but nothing resets it after navigating away from this page so when you land on the page you are in `accountBalanceStatus.success` which does not skip the render and quickly flashes the bare screen before the effect kicks off the api call and resets the loading/data fetching cycle.

I added an action to `resetAccountBalanceStatus` and dispatch it on dismount of Account in order to avoid this.
I also tweaked the height on the loader div for SendAmount in order to expose the tabs underneath, before they were covered by the transparent bg of the loader div which prevented users from being able to nav while loading is taking place. This likely never happens but if the call took a long time for some reason or hung then the user could still nav away from the tab and do other stuff.

### DIFF
--- a/extension/src/popup/ducks/transactionSubmission.ts
+++ b/extension/src/popup/ducks/transactionSubmission.ts
@@ -394,6 +394,9 @@ const transactionSubmissionSlice = createSlice({
   initialState,
   reducers: {
     resetSubmission: () => initialState,
+    resetAccountBalanceStatus: (state) => {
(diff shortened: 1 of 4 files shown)


aristidesstaffieri:
Netlify build should pass here once we merge in https://github.com/stellar/freighter/pull/720

piyalbasu (approved):

PR #720, opened 2023-02-02
Fix docs:build

aristidesstaffieri:
It looks like our docs build was failing because of a bug that's been in terser-webpack-plugin for a while, which is consumed by docusaurus in newer versions.

Details - 
https://github.com/facebook/docusaurus/issues/3657
https://github.com/facebook/docusaurus/issues/3515

Solution - 
Pin terser-webpack-plugin to the version before this was introduced until it is resolved in future versions by setting a specific version in the top level yarn resolution.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,0 +1,21589 @@
+{
+  "name": "docs",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "docs",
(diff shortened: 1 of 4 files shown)


aristidesstaffieri:
Hmm, still lands in a failing state but no obvious error in the logs anymore. Still looking into this on the Netlify side, seem ok locally.

piyalbasu:
@aristidesstaffieri Ah, I see the issue - it's using an old version of Node (12) rather than the one it expects. I'll show you how to upgrade

PR #722, opened 2023-02-03
check for window.freighter before attempting new api method

piyalbasu:
This will allow dapps to upgrade well ahead of the rollout of 3.0.0 and not lose any functionality

### DIFF
--- a/@stellar/freighter-api/src/isConnected.ts
+++ b/@stellar/freighter-api/src/isConnected.ts
@@ -1,3 +1,9 @@
 import { requestConnectionStatus } from "@shared/api/external";
 
-export const isConnected = () => requestConnectionStatus();
+export const isConnected = () => {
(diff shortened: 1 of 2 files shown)


@stellar/freighter-api/src/isConnected.ts
```diff
@@ -1,3 +1,9 @@
 import { requestConnectionStatus } from "@shared/api/external";
 
-export const isConnected = () => requestConnectionStatus();
+export const isConnected = () => {
+  if (window?.freighter) {
+    return window.freighter;
+  }
```
piyalbasu, line 6: miraculously, all the other api methods still work fine with the 3.0.0 switch, so just adding a little check here to make this method work with both versions

aristidesstaffieri (approved):

Mateusz9611 (approved):

PR #724, opened 2023-02-09
increasing node version because docusaurus requires this specific ver…

piyalbasu:
…sion


.github/workflows/submitProduction.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Build extension
         uses: actions/setup-node@v1
         with:
-          node-version: "12"
+          node-version: "16.14"
```
piyalbasu, line 44: interestingly - upgrading to the latest stable version of node (18) caused build errors with the old version of Webpack we're using

PR #728, opened 2023-02-15
Add Soroban Token support in accounts

aristidesstaffieri:
Screenshot 2023-02-21 at 9 32 29 PM

Adds SorobanContext to provide app with a Soroban server and transaction builder.
Adds a Soroban redux slice and actions to track token balances and request lifecycles.
Adds Soroban tokens in accounts, after classic assets but before LP shares.

This uses a placeholder icon for Soroban assets, it displays an `S` in a similar way to LP shares displaying `LP`.

Soroban transactions only support 1 op per tx right now so we have to send 1 tx per value that we want from a contract. There is ongoing work to allow multiple ops per tx, once we have that we can refactor to only send 1 tx with several ops.

NOTE - I used the token at `4a7a254e803102a2da255e634f26b9b9fe5655ad89bf578dd04f8fcbdecf0d95` on futurenet to test this. I also deployed this token with a hardcoded symbol and I did that incorrectly so if you use this token you will see the symbol/code for the FTT token be a dot instead of FTT. This is due to my mistake in deploying that token and the logic in the Freighter should be correct to handle the ascii value for a symbol when deployed properly. For reassurance, you can see that the name property is displayed correctly which is of the same type as the symbol.


@shared/constants/stellar.ts
```diff
@@ -47,3 +47,7 @@ export const DEFAULT_NETWORKS: Array<NetworkDetails> = [
   MAINNET_NETWORK_DETAILS,
   TESTNET_NETWORK_DETAILS,
 ];
+
+export const SOROBAN_RPC_URLS = {
```
aristidesstaffieri, line 51: There isn't a public Soroban RPC at the moment so not sure what we can set for this in the time being, I use a local quickstart node during development,

extension/src/popup/SorobanContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const SorobanContext = React.createContext({} as SorobanTxBuilder);
+
+export class SorobanTxBuilder {
+  public source: SorobanClient.Account;
+  private fee: string;
+  private pubKey: string;
+  private networkPassphrase: string;
+  public server: SorobanClient.Server;
+
+  constructor(pubKey: string, fee: string, networkPassphrase: string) {
```
piyalbasu, line 18: this type of structure makes sense to me if he were to have multiple instances of this class with different public keys, fees, and network phrases.

Since the inputs and outputs will always be the same (we're only ever dealing with one public key, for ex), does this make sense for this be a function rather than a class?

I also bring it up because we just don't tend to use classes in this codebase, so there's also kind of a stylistic component to this comment, as well.

extension/src/popup/SorobanContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+    this.fee = fee;
+    this.pubKey = pubKey;
+    this.networkPassphrase = networkPassphrase;
+
+    this.server = new SorobanClient.Server(SOROBAN_RPC_URLS.futureNet, {
+      allowHttp: SOROBAN_RPC_URLS.futureNet.startsWith("http://"),
+    });
```
piyalbasu, line 25: we should also make sure this works for local networks. The cheesy solution this might be something like:

```
if (networkPassphrase === "Test SDF Future Network ; October 2022" && networkUrl === "https://horizon-futurenet.stellar.org/" ) {
  serverUrl = SOROBAN_RPC_URLS.futureNet
} else {
  serverUrl = networkUrl; // in this case, localhost:8000/soroban/rpc
}

piyalbasu:
also watch out for that build.zip in this pr :)

extension/src/popup/SorobanContext.tsx
aristidesstaffieri: yeah I think that makes sense, I thought it would do more when I first wrote it as a class but I see no reason why it can't be a function or two, will refactor.

extension/src/popup/SorobanContext.tsx
aristidesstaffieri: Ah yeah good point, I can refactor to explicitly pass the `allowHttp` flag when I turn this into a function.

aristidesstaffieri:
> also watch out for that build.zip in this pr :)

whoops, thanks for the catch! Removing and I'll add it to .gitignore

extension/src/popup/SorobanContext.tsx
aristidesstaffieri: refactored this in commit 16200e44ab98a1c820f7e795091a04b53b76c325

extension/src/popup/SorobanContext.tsx
aristidesstaffieri: Added in 16200e44ab98a1c820f7e795091a04b53b76c325

piyalbasu (approved):

PR #744, opened 2023-03-01
WAL 643 soroban payment

aristidesstaffieri:


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -917,6 +917,24 @@ export const popupMessageListener = (request: Request) => {
...
+    const transaction2 = SorobanSdk.TransactionBuilder.fromXDR(
+      transaction,
+      network,
+    );
+
+    const privateKey = privateKeySelector(store.getState());
+    if (privateKey.length) {
+      const sourceKeys = SorobanSdk.Keypair.fromSecret(privateKey);
+      transaction2.sign(sourceKeys);
+      return { signedTransaction: transaction2.toXDR() };
```
aristidesstaffieri, line 932: @piyalbasu as expected, encoding the signed tx to XDR was the last thing needed to make this work. So now you can successfully send some amount of a Soroban token! 🎉  (you should have a few FTT from my tests).

extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
```diff
@@ -89,6 +92,24 @@ export const ChooseAsset = ({ balances }: ChooseAssetProps) => {
             image: "",
             domain: "",
           });
+
+          if (isExperimentalModeEnabled && sorobanBalances.length) {
+            sorobanBalances.forEach(({ symbol, contractId, name }) => {
+              // TODO:
```
aristidesstaffieri, line 98: Interesting interaction, made me think a bit more about how we might want to handle values coming from a smart contract that could now break the interface but haven't found the right way to escape this yet.

aristidesstaffieri:
@piyalbasu I ended up rebasing this onto master after the 2.11.0 release, apologies for the need to `git reset --hard origin/feature/WAL-643-soroban-payment`.

.gitignore
```diff
@@ -20,6 +20,7 @@ node_modules
 .env.development.local
 .env.test.local
 .env.production.local
+.soroban
```
piyalbasu, line 23: Can you refresh my memory: what was this is needed for?

@shared/api/internal.ts
```diff
@@ -545,6 +566,48 @@ export const submitFreighterTransaction = async ({
...
+  const server = new SorobanClient.Server(SOROBAN_RPC_URLS.futureNet, {
+    allowHttp: true,
+  });
+
+  // TODO: fixed in Sorobanclient, not yet released
+  let response = (await server.sendTransaction(tx)) as any;
+
+  try {
+    // Poll this until the status is not "pending"
+    while (response.status === "pending") {
```
piyalbasu, line 596: `pending` should probably be part of an enum. That's generally the pattern we use for dealing with statuses

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -917,6 +917,24 @@ export const popupMessageListener = (request: Request) => {
     return { error: "Session timed out" };
   };
 
+  const signFreighterSorobanTransaction = () => {
+    const { transaction, network } = request;
+
+    const transaction2 = SorobanSdk.TransactionBuilder.fromXDR(
```
piyalbasu, line 923: Ah, I came up with these var names quickly while debugging and I did kind of a bad job here tbh. Would you mind following the pattern from ln 904 above?

It would be more descriptive and easier to follow if it were:

```
    const { transactionXDR, network } = request;
    
    const transaction = SorobanSdk.TransactionBuilder.fromXDR(
```

Thanks and sorry!! You'll need to, unfortunately, rename the params in all the function invocations

piyalbasu (approved):

.gitignore
aristidesstaffieri: Hmm maybe it's not needed, not even sure how I actually generated a `.soroban` dir so let me get rid of that.

aristidesstaffieri: removed in d172235d8ea857ddb35022b80367ea7dbba8ff71

@shared/api/internal.ts
aristidesstaffieri: Added the enum in d172235d8ea857ddb35022b80367ea7dbba8ff71

PR #747, opened 2023-03-07
skip UnlockAccount in history stack

acharb:
fixes: https://stellarorg.atlassian.net/browse/WAL-399
and also came up: https://stellarfoundation.slack.com/archives/C03SJU2D5H9/p1678121113589049

if the view after UnlockAccount calls `history.goBack()` , it goes back to unlock-account causing an annoying UX for user. So instead don't add UnlockAccount to the history stack, since after leaving we won't need to go back to it

### DIFF
--- a/extension/src/popup/views/UnlockAccount/index.tsx
+++ b/extension/src/popup/views/UnlockAccount/index.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import get from "lodash/get";
 import { useDispatch, useSelector } from "react-redux";
-import { useLocation } from "react-router-dom";
+import { useLocation, useHistory } from "react-router-dom";
 import { Field, Form, Formik, FieldProps } from "formik";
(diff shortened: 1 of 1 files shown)


extension/src/popup/views/UnlockAccount/index.tsx
```diff
@@ -39,6 +40,8 @@ export const UnlockAccount = () => {
   const handleSubmit = async (values: FormValues) => {
     const { password } = values;
     await dispatch(confirmPassword(password));
+    // skip this location in history, we won't need to come back here after unlocking account
+    history.replace(destination);
```
acharb, line 44: [replace](https://v5.reactrouter.com/web/api/history) replaces the current location in history stack, so replacing with next destination (which actually doesn't add since it's already there, but it prevents a new addition to the stack)

piyalbasu (approved):

PR #749, opened 2023-03-14
use version from package.json as app version in settings

aristidesstaffieri:
Replaces hard coded version number with dynamic version number pulled from package.json


piyalbasu (approved):
wish I knew it was that easy - would've done it months ago LOL

aristidesstaffieri:
Even though we already cut the BETA for 2.11.1, I'm considering adding this to it since I believe we will have to cut another one [due to this bug](https://github.com/stellar/freighter/pull/750), and this change seems very low risk. @piyalbasu

PR #750, opened 2023-03-14
[BUG] Fix paths for native stops in path payments

aristidesstaffieri:
I'm not sure how this worked before, I looked back through the previous states of `getCanonicalFromAsset` and of the `getBestPath` fulfilled builder step and I haven't made sense of how this used to work. It looks to me like this would have always failed for native steps in a path payment when using the `getCanonicalFromAsset` helper to build the paths. I'm wondering if maybe the Horizon response changed?


aristidesstaffieri:
I'm still trying to figure out how to test some more complex paths, haven't been able to find the right asset combo to produce a path that isn't either empty or only goes through XLM.

piyalbasu (approved):

PR #757, opened 2023-03-17
[BUG] Bugfix/consistent soroban icons

aristidesstaffieri:
Uses the `isSorobanToken` flag on all active instances of AssetIcon for consistent Soroban token icons across the app.

### DIFF
--- a/extension/src/popup/components/account/AccountAssets/index.tsx
+++ b/extension/src/popup/components/account/AccountAssets/index.tsx
@@ -8,6 +8,7 @@ import { AssetIcons } from "@shared/api/types";
 import { retryAssetIcon } from "@shared/api/internal";
 
 import { getCanonicalFromAsset } from "helpers/stellar";
+import { isSorobanIssuer } from "popup/helpers/account";
 import StellarLogo from "popup/assets/stellar-logo.png";
 import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
 import { transactionSubmissionSelector } from "popup/ducks/transactionSubmission";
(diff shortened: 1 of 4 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -384,6 +384,7 @@ export const ManageAssetRow = ({
         assetIcons={code !== "XLM" ? { [canonicalAsset]: image } : {}}
         code={code}
         issuerKey={issuer}
+        isSorobanToken={!!(issuer && isSorobanIssuer(issuer))}
```
piyalbasu, line 387: hm, looks like in these instances, you're always defining `isSorobanToken` based on the `issuerKey` value.

Would it be cleaner to have this check inside `AssetIcon`? Maybe something like:

`isSorobanToken = !isSorobanToken ? isSorobanIssuer(issuerKey) : isSorobanToken`

I'm not sure if this is actually better, but it allows us to do the ` isSorobanIssuer(issuerKey)` logic in one place while also allowing a dev to override. I see that the only other place we're actually passing `isSorobanToken` is when we deal with balances in `AccountAssets` 🤔

aristidesstaffieri: yeah 100%, not even sure why I went the prop route so I'll refactor.

aristidesstaffieri: refactored in a6f66dba1025348d5e6c8fadf4006e0cd8a86376

piyalbasu (approved):

PR #772, opened 2023-04-03
remove popupwrapper scroll and move simplebar use to simplebarwrapper 

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-709)

removes the popupwrapper scroll that used `overflow-y`, and use simplewrapper instead.

also move simplebar into a wrapper


extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -236,38 +237,40 @@ export const SendTo = ({ previous }: { previous: ROUTES }) => {
...
-                        <span>
-                          {isFederationAddress(address)
-                            ? address
-                            : truncatedPublicKey(address)}
-                        </span>
-                      </button>
-                    </li>
-                  ))}
-                </ul>
+                <SimpleBarWrapper className="SendTo__simplebar">
```
acharb, line 240: adding simple bar here since it was using the popupwrapper scroll

aristidesstaffieri (approved):
lgtm

piyalbasu:
@acharb I believe this branch is a duplicate of https://github.com/stellar/freighter/pull/777

Let me know if this is not the case!

PR #778, opened 2023-04-04
Check for blocked accounts before send/swap

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/36?modal=detail&selectedIssue=WAL-465

Right now, when signing a transaction using freighter we decode the tx and check the destination address against known blocked accounts in order to surface the right warning using the tags associated to the blocked accounts from Horizon.

We can do this same check when sending/swapping assets.

Adds getBlockedAccounts internal API
Uses this API in transaction details when the action is in idle status, which means this will only show on the "review" screen before submitting a send/swap.

Warnings will be shown for tags: malicious, unsafe, memo_required.

Screenshot 2023-04-04 at 9 38 21 AM

Screenshot 2023-04-04 at 10 27 48 AM


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -120,6 +125,60 @@ const TwoAssetCard = ({
...
 
+const computeDestMinWithSlippage = (
+  slippage: string,
+  destMin: string,
+): BigNumber => {
+  const mult = 1 - parseFloat(slippage) / 100;
+  return new BigNumber(destMin).times(new BigNumber(mult));
+};
+
+const getOperation = (
```
aristidesstaffieri, line 136: I pulled these out of the component body and moved all closure variables to fn args in order to avoid the re-renders that were caused by the closure style before, since this is now needed in an effect to compute the blocked account checks.

extension/src/helpers/stellar.ts
```diff
@@ -56,19 +55,12 @@ export const getTransactionInfo = (search: string) => {
...
 
-export const getAssetFromCanonical = (canonical: string) => {
+export const getAssetFromCanonical = (canonical: string): Asset => {
   if (canonical === "native") {
     return StellarSdk.Asset.native();
   }
   if (canonical.includes(":")) {
     const [code, issuer] = canonical.split(":");
-
-    if (isSorobanIssuer(issuer)) {
```
piyalbasu, line 66: hm, I think you may need to rebase or something - I don't think you want to drop this

extension/src/popup/basics/Modal/index.tsx
```diff
@@ -26,9 +25,9 @@ export const ModalWrapper = ({ children }: ModalWrapperProps) => {
   }, []);
 
   return (
-    <SimpleBar className="ModalWrapper__scrollbar">
+    <SimpleBarWrapper className="ModalWrapper__scrollbar">
```
piyalbasu, line 28: rebasing will probably help get rid of all these SimpleBar changes

extension/src/helpers/stellar.ts
aristidesstaffieri: oh actually this is intentional. While working in this file I noticed that there isn't a difference in the canonical for a Soroban asset(still can be represented as a code/issuer) so I simplified this a bit. It might be more clear to look [at this commit in isolation](https://github.com/stellar/freighter/pull/778/commits/91fae58372661ae52f852470ff5e63cbc6933363)

Apologies, maybe I should have pulled this change out on its own since it doesn't really follow the theme of the PR.

aristidesstaffieri:
@piyalbasu I caught the branch up, so the diff should just be this change set now.

extension/src/helpers/stellar.ts
piyalbasu: No worries - just making sure!

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -172,47 +234,61 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
-        allowedSlippage,
-        destinationAmount,
-      );
-      return StellarSdk.Operation.pathPaymentStrictSend({
-        sendAsset: sourceAsset,
-        sendAmount: amount,
-        destination: isSwap ? publicKey : destination,
+  // checked tags for blocked accounts
+  useEffect(() => {
+    if (isValidatingSafeAssetsEnabled) {
```
piyalbasu, line 243: Ah, this is actually the incorrect setting to check. It should be `isValidatingSafetyValue`. (This naming is pretty confusing!)

`isValidatingSafeAssetsEnabled` is a toggle used to check  if a trustline is safe to be added

piyalbasu: These lines might be a good model to use here: https://github.com/stellar/freighter/blob/master/extension/src/background/messageListener/freighterApiMessageListener.ts#L127-L129

Also, we probably only want to do this check on Mainnet as the mainnet account is actually the one that's flagged

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -172,47 +234,61 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
+        allowedSlippage,
+        path,
+        isPathPayment,
+        isSwap,
+        destinationBalances.isFunded!,
+        publicKey,
+      );
+
+      blockedAccounts.forEach(({ address, tags }) => {
+        if (address === operation.destination) {
```
piyalbasu, line 259: Hm, I'm actually having an issue getting the warning to show. It looks like `operation.destination` is undefined 🤔 
Screenshot 2023-04-07 at 1 29 20 PM

piyalbasu:
Another thing I noticed - when we're attempting to sign a tx to a malicious/memo req destination (and we have the settings enabled, of course), we block Approving of the tx

Screenshot 2023-04-07 at 1 35 17 PM

I think we should do the same here - block the `Send` button

https://github.com/stellar/freighter/blob/master/extension/src/popup/views/SignTransaction/index.tsx#L244

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: thanks, I corrected this in 8dea16edd6ef4a38d6795e0ce204f909df6414bc

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: whoops, yeah this is just supposed to be `destination` here, not sure what I was thinking. Thanks for catching it, fixed in 38b4bb1df61ce09bb498335baae79544404581f8

aristidesstaffieri:
> Another thing I noticed - when we're attempting to sign a tx to a malicious/memo req destination (and we have the settings enabled, of course), we block Approving of the tx
> 
> Screenshot 2023-04-07 at 1 35 17 PM
> 
> I think we should do the same here - block the `Send` button
> 
> https://github.com/stellar/freighter/blob/master/extension/src/popup/views/SignTransaction/index.tsx#L244

makes sense, added this flag in ad90fbd5b9315c1d0d70891dd8efe44f11fcbc18

PR #783, opened 2023-04-10
Use decimals values to format/parse Soroban token values

aristidesstaffieri:
Ticket - https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/36?modal=detail&selectedIssue=WAL-737

Before this, we assume all assets follow the classic asset path of using 7 decimals for their amount precisions.
While this will be true for all wrapped assets and may become the canonical pattern for non wrapped assets, token deployments can choose a different decimal precision for whatever reason.

This change takes into the account the token decimal interface and uses that throughout the app to format/parse values to/from human readable values.

Example - 

We have deployed Freighter Test Token with `7` decimals, which mimics the relationship between a lumen and a stroop. Without floating point representations, we can pad the values with zeros in order to represent fractional values using an i128.

So to send `1.5 FTT`, you interact with the contract using the value `150000000` for an `xfer`. When you account for the number of decimals set in the contract, you can display this value as `1.5`.

The parser does the opposite of the formatter, it takes a human readable value from the input and pads it with the appropriate amount of zeros when necessary.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -48,12 +48,30 @@ import { ScamAssetWarning } from "popup/components/WarningMessages";
 import { TX_SEND_MAX } from "popup/constants/transaction";
 
 import "../styles.scss";
+import { TokenBalances } from "@shared/api/types";
 
 enum AMOUNT_ERROR {
   TOO_HIGH = "amount too high",
   DEC_MAX = "too many decimal digits",
   SEND_MAX = "amount higher than send max",
 }
+
+const getAssetDecimals = (
+  asset: string,
+  balances: TokenBalances,
+  isToken: boolean,
+) => {
+  if (isToken) {
+    const contractId = asset.split(":")[1];
+    const balance = balances.find(({ contractId: id }) => id === contractId);
+
+    if (balance) {
+      return Number(balance.decimals);
+    }
+  }
+
+  return 7;
+};
 
 const ConversionRate = ({
   source,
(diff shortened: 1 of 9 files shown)


extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -48,13 +48,31 @@ import { ScamAssetWarning } from "popup/components/WarningMessages";
...
+  if (isToken) {
+    const contractId = asset.split(":")[1];
+    const balance = balances.find(({ contractId: id }) => id === contractId);
+
+    if (balance) {
+      return Number(balance.decimals);
+    }
+  }
+
+  return 7;
```
piyalbasu, line 73: I think it'd be nice to save `7` as a const with a little comment explaining where it comes from

aristidesstaffieri: yeah good call, always hard to glean what these magic numbers are.

extension/src/popup/helpers/formatters.ts
```diff
@@ -44,7 +45,7 @@ export const formatAmount = (
   if (cleaned.indexOf(".") !== -1) {
     const parts = cleaned.split(".");
     parts[0] = decimal.format(Number(parts[0].slice(0, maxDigits))).toString();
-    parts[1] = parts[1].slice(0, 7);
+    parts[1] = parts[1].slice(0, decimals);
```
piyalbasu, line 48: you could import the default decimals value (7) here: `decimals || classicDefaultPrecision` (not sold on that var name)

extension/src/popup/components/sendPayment/SendAmount/index.tsx
piyalbasu: https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/assets#amount-precision

piyalbasu (approved):

PR #792, opened 2023-04-18
add jest test for views/Account

acharb:
[ticket](https://stellarorg.atlassian.net/browse/WAL-746)

tests for:

- Rendering
- loads accounts
- switching accounts works
- shows balances
- clicking on balance goes go account balance screen


extension/src/popup/App.tsx
```diff
@@ -22,11 +18,6 @@ import { Router } from "./Router";
 
 import "./styles/global.scss";
 
-// .isBigNumber() not catching correctly, so checking .isBigNumber
-// property as well
-const isSerializable = (value: any) =>
```
acharb, line 27: for some reason jest didn't like importing this from this file. So moved the func to a helper file

aristidesstaffieri (approved):

extension/src/popup/App.tsx
piyalbasu: Oh weird - I wonder if it had something to do with the other code in this file 🤔 Prob best to have it in a helper anyway

piyalbasu (approved):

PR #802, opened 2023-04-27
Session storage for background redux store with feature flag

aristidesstaffieri:
[Ticket](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/36?modal=detail&selectedIssue=WAL-751)

In an effort to merge the manifest v3 changes, I've added an option to set a static feature flag in order to use session storage to persist the background redux store.

This change is required after Manifest v3 because the background script is deprecated in favor of service workers, but service workers are ephemeral and go idle after 30 seconds of inactivity which means our redux store in the background is garbage collected.

The reason we need a feature flag and can't use this approach as the default is because [Firefox currently does not support the session storage APIs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/session). So we cannot migrate to manifest v3 until they either add support or we decide to add special casing for each browser. Right now, we can stay on manifest v2 until Chrome forces us to upgrade to v3.

Notable changes:
Both manifest versions are supported, we now have a `/public/manifest/{v2|v3}.json` and the common webpack config uses the v2 version as a default. The `SESSION_STORAGE_ENABLED` feature flag in dataStorage is set to false which means the redux store will be kept on the top level scope of the background script.

To switch to manifest v3 mode, flip `SESSION_STORAGE_ENABLED` to true and update `webpack.common.ts` to use `manifest/v3.json` for the manifest in the build output.

The dataStore now supports local storage or session storage(they use the same APIs), it's configured to use browser storage everywhere that it did before and uses sessions storage for the redux store hydration when it is enabled.

The SessionTimer now uses `chrome.alarms` in order to timeout session after a day.

### DIFF
--- a/extension/public/background.ts
+++ b/extension/public/background.ts
@@ -2,9 +2,17 @@ import {
   initContentScriptMessageListener,
   initExtensionMessageListener,
   initInstalledListener,
+  initAlarmListener,
 } from "background";
 
-initContentScriptMessageListener();
-initExtensionMessageListener();
+import { buildStore } from "background/store";
 
-initInstalledListener();
+async function main() {
+  const store = await buildStore();
+  initContentScriptMessageListener();
+  initExtensionMessageListener(store);
+  initInstalledListener();
+  initAlarmListener(store);
+}
+
+main();
--- a/extension/src/background/helpers/dataStorage.ts
+++ b/extension/src/background/helpers/dataStorage.ts
@@ -14,39 +14,58 @@ interface SetItemParams {
   [key: string]: any;
 }
 
-export const browserStorage = browser.storage?.local;
+// https://github.com/mozilla/webextension-polyfill/issues/424
+interface BrowserStorage extends browser.Storage.Static {
+  session: browser.Storage.LocalStorageArea;
+}
 
-export const dataStorage = {
+const storage = browser.storage as BrowserStorage;
+
+// browser storage uses local storage which stores values on disk and persists data across sessions
+// session storage uses session storage which stores data in memory and clears data after every "session"
+// only use session storage for secrets or sensitive values
+export const localStorage = storage?.local;
+export const sessionStorage = storage?.session;
+
+// Session Storage Feature Flag - turn on when storage.session is supported
+export const SESSION_STORAGE_ENABLED = false;
+
+export type StorageOption = typeof localStorage | typeof sessionStorage;
+
+export const dataStorage = (storageApi: StorageOption = localStorage) => ({
   getItem: async (key: string) => {
     // TODO: re-enable defaults by passing an object. The value of the key-value pair will be the default
 
-    const storageResult = await browser.storage.local.get(key);
+    const storageResult = await storageApi.get(key);
 
     return storageResult[key];
   },
   setItem: async (setItemParams: SetItemParams) => {
-    await browser.storage.local.set(setItemParams);
+    await storageApi.set(setItemParams);
   },
 
   clear: async () => {
-    await browser.storage.local.clear();
+    await storageApi.clear();
   },
-};
+});
 
-export const dataStorageAccess = {
-  getItem: (keyId: string) => dataStorage.getItem(keyId),
-  setItem: async (keyId: string, value: any) => {
-    await dataStorage.setItem({ [keyId]: value });
-  },
-  clear: () => dataStorage.clear(),
+export const dataStorageAccess = (storageApi: StorageOption = localStorage) => {
+  const store = dataStorage(storageApi);
+  return {
+    getItem: store.getItem,
+    setItem: async (keyId: string, value: any) => {
+      await store.setItem({ [keyId]: value });
+    },
+    clear: () => store.clear(),
+  };
 };
 
 // This migration adds a friendbotUrl to testnet and futurenet network details
 export const migrateFriendBotUrlNetworkDetails = async () => {
-  const networkList = await dataStorageAccess.getItem(NETWORKS_LIST_ID);
-  const networksList: NetworkDetails[] = networkList
-    ? JSON.parse(networkList)
-    : DEFAULT_NETWORKS;
+  const localStore = dataStorageAccess(localStorage);
+
+  const networksList: NetworkDetails[] =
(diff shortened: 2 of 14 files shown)


extension/public/background.ts
```diff
@@ -2,9 +2,17 @@ import {
...
+  initInitAlarmListener,
 } from "background";
 
-initContentScriptMessageListener();
-initExtensionMessageListener();
+import { buildStore } from "background/store";
 
-initInstalledListener();
+async function main() {
+  const store = await buildStore();
```
aristidesstaffieri, line 11: We need to build the store async and at the top level to make sure that in v2 mode, it is only built when the background script is ran and to use the same hydrated store across listeners in v3 mode.

extension/src/background/store.ts
```diff
@@ -2,9 +2,60 @@ import { combineReducers } from "redux";
...
+  sessionStorage,
+  browserStorage,
+} from "./helpers/dataStorage";
 
-export const store = configureStore({
+// Session storage can be used to persist redux store in Manifest v3 because service workers go idle after 30 seconds
+// Session storage is not currently supported in Firefox, which blocks our migration to using this by default
+const REDUX_STORE_KEY = "redux-store";
+const sessionStore = dataStorageAccess(
+  SESSION_STORAGE_ENABLED ? sessionStorage : browserStorage,
```
piyalbasu, line 16: correct me if I'm wrong - but it looks like `sessionStore` will never be saved to `browserStorage` as we're just using redux when `SESSION_STORAGE_ENABLED` is false (`loadState` and `saveStore` are only used when `SESSION_STORAGE_ENABLED` === true)

If that's the case, I wonder if we shouldn't even bother having `browserStorage` as a backup. Especially since we considered `browserStorage` to be a risky place to save private creds.

extension/src/background/store.ts
```diff
@@ -2,9 +2,60 @@ import { combineReducers } from "redux";
...
+  SESSION_STORAGE_ENABLED,
+  dataStorageAccess,
+  sessionStorage,
+  browserStorage,
+} from "./helpers/dataStorage";
 
-export const store = configureStore({
+// Session storage can be used to persist redux store in Manifest v3 because service workers go idle after 30 seconds
+// Session storage is not currently supported in Firefox, which blocks our migration to using this by default
+const REDUX_STORE_KEY = "redux-store";
```
piyalbasu, line 14: this is pretty NIT: but would you mind calling this something a bit more descriptive than `redux-store`? Maybe something that conveys that it's the user's session creds or something 🤔

extension/src/background/index.ts
```diff
@@ -50,3 +53,11 @@ export const initInstalledListener = () => {
   });
   browser?.runtime?.onInstalled.addListener(migrateFriendBotUrlNetworkDetails);
 };
+
+export const initInitAlarmListener = (store: Store) => {
```
piyalbasu, line 57: probably want to rename this to `initAlarmListener`

extension/src/background/helpers/dataStorage.ts
```diff
@@ -13,37 +13,60 @@ interface SetItemParams {
...
+interface BrowserStorage extends browser.Storage.Static {
+  session: browser.Storage.LocalStorageArea;
+}
+
+const storage = browser.storage as BrowserStorage;
+
+// browser storage uses local storage which stores values on disk and persists data across sessions
+// session storage uses session storage which stores data in memory and clears data after every "session"
+// only use session storage for secrets or sensitive values
+export const browserStorage = storage?.local;
```
piyalbasu, line 26: I think now that we're starting to deal with 2 types of storage (1 long lived, 1 ephemeral) in many parts of the code, we should start using consistent naming to make it clear when you're saving something to the long lived store vs the short lived store.

why don't we call this var `localStorage`. And then, going forward in the code, every time we invoke `dataStorage` with `localStorage`, we call that `const localStore = dataStorage(localStorage)`. Similary, every time we're dealing with the short lived (either redux or storage.session), we call it `sessionStore`

extension/src/background/helpers/dataStorage.ts
```diff
@@ -13,37 +13,60 @@ interface SetItemParams {
...
+    setItem: async (keyId: string, value: any) => {
+      await store.setItem({ [keyId]: value });
+    },
+    clear: () => store.clear(),
+  };
 };
 
 // This migration adds a friendbotUrl to testnet and futurenet network details
 export const migrateFriendBotUrlNetworkDetails = async () => {
+  const store = dataStorageAccess(browserStorage);
```
piyalbasu, line 66: for ex: we can call this `const localStore = dataStorage(localStorage)` to make it clear that these methods are saving to the long lived store

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -100,7 +99,8 @@ interface KeyPair {
   privateKey: string;
 }
 
-export const popupMessageListener = (request: Request) => {
+export const popupMessageListener = (request: Request, store: Store) => {
```
piyalbasu, line 102: related to the comments above: we can call the `store` param `sessionStore`

extension/src/background/store.ts
aristidesstaffieri: Hmm yeah I was trying to not risk using the session store when it's not supported but I think it's probably ok to do what you said since it won't actually be used in the case where session storage is disabled.

extension/src/background/store.ts
aristidesstaffieri: ah yeah fair, how about `session-store`?

extension/src/background/index.ts
aristidesstaffieri: 😄 thanks

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: makes sense, will update the var names.

aristidesstaffieri:
addressed all comments in e4838121cf26c2d7a87f46a8a93cfa67229a03ce

piyalbasu (approved):

PR #808, opened 2023-05-01
adds amount display formatter, uses on amount display

aristidesstaffieri:
[Ticket](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/36?modal=detail&selectedIssue=WAL-779)

This formats all instances of asset amounts in order to use decimal formatting(commas after every 3rd digit).


extension/src/popup/helpers/formatters.ts
```diff
@@ -82,3 +82,15 @@ export const formatAmount = (
...
+export const formatAmount = (val: string) => {
+  const decimal = new Intl.NumberFormat("en-US", { style: "decimal" });
+  const [wholeVal, remainderVal] = val.split(".");
+  const formattedWholeVal = decimal.format(Number(wholeVal)).toString();
+
+  if (remainderVal) {
+    return `${formattedWholeVal}.${remainderVal}`;
+  }
+
+  return formattedWholeVal;
```
piyalbasu, line 95: I think you can use this in `formatAmountPreserveCursor` above, right?

aristidesstaffieri: Felt risky to me to try to combine these since one targets user input and accounts for the cursor positioning while the other formats read only display values. I could take another look at combining these.

piyalbasu (approved):

PR #826, opened 2023-05-16
Bugfix/token id list

piyalbasu:


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1144,13 +1145,10 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
-      TOKEN_ID_LIST,
-      JSON.stringify({
-        ...tokenIdList,
-        [keyId]: accountTokenIdList,
-      }),
-    );
+    await localStore.setItem(TOKEN_ID_LIST, {
+      ...tokenIdList,
+      [keyId]: accountTokenIdList,
+    });
```
piyalbasu, line 1151: this is the real change - this shouldn't be stringify'd

aristidesstaffieri (approved):

PR #827, opened 2023-05-17
normalize any faulty localstorage data and better naming for storage

piyalbasu:
I ran into an edge case where I added a token id in 4.0.2 and the `migrateLocalStorageToBrowserStorage` function didn't run. When I switched to 5.0.0, my token data was in stringified json instead of just an object. This is an issue we keep running into in this migration, so I created a script that runs on install that finds any weird data and puts it into the right form for 5.0.0

This also led me to discover that it was a mistake to name the `browser.storage.local` var as `localStorage` because `localStorage` is a reserved var name when `window.localStorage` is available. So I had to rename the storage var's, as well. Not married to the new names. Open to suggestions there

### DIFF
--- a/extension/src/background/helpers/dataStorage.ts
+++ b/extension/src/background/helpers/dataStorage.ts
@@ -24,15 +24,19 @@ const storage = browser.storage as BrowserStorage;
 // browser storage uses local storage which stores values on disk and persists data across sessions
 // session storage uses session storage which stores data in memory and clears data after every "session"
 // only use session storage for secrets or sensitive values
-export const localStorage = storage?.local;
-export const sessionStorage = storage?.session;
+export const browserLocalStorage = storage?.local;
+export const browserSessionStorage = storage?.session;
 
 // Session Storage Feature Flag - turn on when storage.session is supported
 export const SESSION_STORAGE_ENABLED = false;
 
-export type StorageOption = typeof localStorage | typeof sessionStorage;
+export type StorageOption =
+  | typeof browserLocalStorage
+  | typeof browserSessionStorage;
 
-export const dataStorage = (storageApi: StorageOption = localStorage) => ({
+export const dataStorage = (
+  storageApi: StorageOption = browserLocalStorage,
+) => ({
   getItem: async (key: string) => {
     // TODO: re-enable defaults by passing an object. The value of the key-value pair will be the default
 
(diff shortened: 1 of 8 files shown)


extension/src/background/helpers/dataStorage.ts
```diff
@@ -24,15 +24,19 @@ const storage = browser.storage as BrowserStorage;
 // browser storage uses local storage which stores values on disk and persists data across sessions
 // session storage uses session storage which stores data in memory and clears data after every "session"
 // only use session storage for secrets or sensitive values
-export const localStorage = storage?.local;
```
piyalbasu, line 27: we need the name `localStorage` available so we can call `window.localStorage` later

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -976,8 +976,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
       isExperimentalModeEnabled,
     } = request;
 
-    const currentIsExperimentalModeEnabled =
-      await getIsExperimentalModeEnabled();
+    const currentIsExperimentalModeEnabled = await getIsExperimentalModeEnabled();
```
piyalbasu, line 979: i have no idea why my PR's keep doing and undoing this line break 🤷

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1183,8 +1180,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     [SERVICE_TYPES.HANDLE_SIGNED_HW_TRANSACTION]: handleSignedHwTransaction,
     [SERVICE_TYPES.REJECT_TRANSACTION]: rejectTransaction,
     [SERVICE_TYPES.SIGN_FREIGHTER_TRANSACTION]: signFreighterTransaction,
-    [SERVICE_TYPES.SIGN_FREIGHTER_SOROBAN_TRANSACTION]:
-      signFreighterSorobanTransaction,
+    [SERVICE_TYPES.SIGN_FREIGHTER_SOROBAN_TRANSACTION]: signFreighterSorobanTransaction,
```
piyalbasu, line 1183: same with this line break

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1133,9 +1132,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
 
   const addTokenId = async () => {
     const { tokenId } = request;
-    const tokenIdList = JSON.parse(
-      (await localStore.getItem(TOKEN_ID_LIST)) || "{}",
-    );
+    const tokenIdList = (await localStore.getItem(TOKEN_ID_LIST)) || {};
```
piyalbasu, line 1135: now that the tokenIdlist is in the right format, we don't need to worry about parsing it

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: ah yup 😄

aristidesstaffieri (approved):

PR #828, opened 2023-05-19
edit friendbot network data after all data has been migrated; fix boo…

piyalbasu:
…lean data defaults

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -64,16 +64,16 @@ export const getIsFuturenet = async () => {
 };
 
 export const getIsMemoValidationEnabled = async () =>
-  (await localStore.getItem(IS_VALIDATING_MEMO_ID)) || true;
+  (await localStore.getItem(IS_VALIDATING_MEMO_ID)) ?? true;
 
 export const getIsSafetyValidationEnabled = async () =>
-  (await localStore.getItem(IS_VALIDATING_SAFETY_ID)) || true;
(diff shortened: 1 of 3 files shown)


extension/src/background/helpers/account.ts
```diff
@@ -64,16 +64,16 @@ export const getIsFuturenet = async () => {
 };
 
 export const getIsMemoValidationEnabled = async () =>
-  (await localStore.getItem(IS_VALIDATING_MEMO_ID)) || true;
+  (await localStore.getItem(IS_VALIDATING_MEMO_ID)) ?? true;
```
piyalbasu, line 67: now that these values are bools and no longer string "true" or "false", when the value is set to `false`, it was defaulting to the backup case of `true`

extension/src/background/index.ts
```diff
@@ -54,8 +54,8 @@ export const initInstalledListener = () => {
       default:
     }
   });
-  browser?.runtime?.onInstalled.addListener(migrateFriendBotUrlNetworkDetails);
   browser?.runtime?.onInstalled.addListener(normalizeMigratedData);
+  browser?.runtime?.onInstalled.addListener(migrateFriendBotUrlNetworkDetails);
```
piyalbasu, line 58: unrelated, but we need to do the friendbot migration after normalize or else the browser.storage data that needs to be updated may not exist yet

acharb (approved):
nice catch, lgtm!

aristidesstaffieri (approved):

PR #842, opened 2023-05-31
fixes soroban transfers asfter upgrade and contract interface changes

aristidesstaffieri:
This doesn't resolve the problem in the operation history display but does re-enable sending payments.

Ongoing convo about history items - https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1685551006105059

### DIFF
--- a/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
@@ -93,14 +93,7 @@ export const ChooseAsset = ({ balances }: ChooseAssetProps) => {
             domain: "",
           });
 
-          // disable soroban payment until we can get the new SDK working
-          const disableSorobanPayment = true;
-
-          if (
-            isExperimentalModeEnabled &&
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
```diff
@@ -94,7 +94,7 @@ export const ChooseAsset = ({ balances }: ChooseAssetProps) => {
           });
 
           // disable soroban payment until we can get the new SDK working
-          const disableSorobanPayment = true;
+          const disableSorobanPayment = false;
```
piyalbasu, line 97: can you please just remove this feature flag altogether now? I just added it to quickly block token payments

piyalbasu (approved):

extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
aristidesstaffieri: oh right, removing

PR #848, opened 2023-06-05
retry tx submission on 504

piyalbasu:


aristidesstaffieri (approved):
lgtm. After reading the doc and thinking about our convo, I do wonder if at times of network congestion if this can keep happening for potentially a long time. Seems like a bump network fee solution would be goo din the long term.

Shaptic (approved):
I recommend dropping a reference to https://developers.stellar.org/docs/encyclopedia/error-handling in the docs so people know where to look for more strategies if you guys revisit this code, but LGTM as a basic resolution strategy alongside a max timebound!

PR #858, opened 2023-06-13
[BUGFIX] Safely sets title default when tab title is not present

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/browse/WAL-875

This is an edge case that happens when an external api call is made to Freighter but Freighter requires the user to login. In that case, the GrantAccess is rendered without a tab title and the helper does not always return a `tab`.


piyalbasu (approved):
Ah, nice

PR #869, opened 2023-06-20
feat: make compatible with node envs

chadoh:
You can see that a recent addition of `isConnected` to some app logic caused my builds to start failing: https://github.com/stellar/soroban-example-dapp/pull/117/commits

In general, Freighter needs to be careful to exit gracefully in Node, in order to support isomorphic/universal apps, and even just build scripts for NextJS and such. There's no reason to make app authors perform these checks themselves.

The messiest part of this was getting the types to be sensible. This part is also somewhat untested! However, since the exports from `@stellar/freighter-api/src/*` are all pretty thin wrappers around other logic, the hard-coded types I've added here to those files should alert us to problems if the functions they wrap change signatures.

### DIFF
--- a/@stellar/freighter-api/jest.config.js
+++ b/@stellar/freighter-api/jest.config.js
@@ -1,5 +1,0 @@
-const parentConfig = require("../../jest.config");
-
-module.exports = {
-  ...parentConfig
-};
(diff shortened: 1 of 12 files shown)


piyalbasu:
Thanks! This is a good idea. I'd like to have a more explicit check for browser vs node and I'd like to not have to check `isBrowser` manually in every method call. I'll follow up in the next version

piyalbasu (approved):

PR #879, opened 2023-06-30
add test ci workflow and move translations to git hook

piyalbasu:
This PR also removes the PR Preview jobs as they were quite flakey: they would create GH releases for each preview but only sometimes properly clean them up, thus polluting the GH releases with a lot of "orphaned" previews that I would manually clean up. Now that we have a Beta env, it's really not that useful to have them, so I've removed them


README.md
```diff
@@ -86,3 +86,23 @@ symlinks all the workspaces, so doing so will allow you to import files from the
 ### Dependencies
 
 Many dev dependencies (such as Typescript, linters, Webpack, etc.) have been moved to the root `package.json` to allow devs to upgrade these libraries all in one place.
+
+### Pushing to repo
+
+This repo will run a pre-push hook before pushing. This hook will run the cmd `yarn build:extension:translations` to check if any strings in the extension need to be added to the translations JSON. If there is no need to update the translations JSON, the push will go through. If there is a need to update, the changes will be automatically committed to your branch and the push will be aborted. You will need to run `git push` again.
+
+NOTE: If you're using nvm and run into an error where the git hook is using an incompatible version of node, create a file `~/.huskryc` on your system and added the following:
```
piyalbasu, line 94: this issue is only temporary as I have a branch that upgrades this repo to Node 18 ready to test 🎉

.husky/addTranslations.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+yarn build:extension:translations
+git diff --quiet || (echo "Adding translations commit. Please retry git push" && git add . && git commit -m "Added translations" && exit 1)
```
piyalbasu, line 4: on push, this runs the translate job, determines if any translations need to be added, and then aborts the push if translations are found. It will automatically commit and stage the translations, so all you need to do is `git push` again (or `git push --no-verify` if you don't want to run the whole translate job again and you've made no other changes)

aristidesstaffieri (approved):

PR #881, opened 2023-07-03
Feature/improve sign xdr

piyalbasu:
Screenshot 2023-07-03 at 4 43 33 PM


extension/src/popup/components/signTransaction/Operations/index.tsx
```diff
@@ -300,6 +313,21 @@ export const Operations = ({
     t("Authorization Required; Authorization Required; Authorization Immutable")
   */
 
+  const [contractId, setContractId] = useState("");
+  const [decimals, setDecimals] = useState(0);
+
+  useEffect(() => {
+    const fetchContractDecimals = async () => {
+      const contractDecimals = await getContractDecimals(
```
piyalbasu, line 321: we need to fetch `decimals` to properly show the amount a user is sending. Kind of annoying, but it's probably better to format the amount using the appropriate number of decimals

extension/src/popup/ducks/soroban.ts
```diff
@@ -8,11 +8,11 @@ import {
 } from "@shared/api/internal";
 import { ErrorMessage, ActionStatus, TokenBalances } from "@shared/api/types";
 import { accountIdentifier } from "@shared/api/helpers/soroban";
-import { SororbanContext } from "popup/SorobanContext";
+import { SorobanContextInterface } from "popup/SorobanContext";
 
 export const getTokenBalances = createAsyncThunk<
   TokenBalances,
-  { sorobanClient: SororbanContext },
```
piyalbasu, line 15: fixing a little typo here :)

extension/src/popup/helpers/soroban.ts
```diff
@@ -19,6 +20,11 @@ export enum SorobanTokenInterface {
   mint = "mint",
 }
 
+export const SOROBAN_OPERATION_TYPES = [
+  "invoke_host_function",
+  "invokeHostFunction",
```
piyalbasu, line 25: this is kinda weird that an op type can either be camelCase or snake_case. I'll raise this with the Platform team

extension/src/popup/helpers/soroban.ts
```diff
@@ -109,10 +115,33 @@ export const getTokenBalance = (
   );
 };
 
+export const getContractDecimals = async (
```
piyalbasu, line 118: similar to the code in `getTokenBalances`, but broken out here

extension/src/popup/helpers/soroban.ts
```diff
@@ -186,3 +201,33 @@ export const getAttrsFromSorobanOp = (
     ...opArgs,
   };
 };
+
+export const getAttrsFromSorobanTxOp = (operation: HorizonOperation) => {
```
piyalbasu, line 205: we now have 2 helper methods to decode a soroban tx as the tx built from an xdr is _slightly_ different than what Horizon provides. Maybe this can be combined once I understand this a little better

piyalbasu: Breaking out the shared rootInvocation stuff into its own function

extension/src/popup/helpers/soroban.ts
```diff
@@ -186,3 +201,33 @@ export const getAttrsFromSorobanOp = (
...
+  if (!isSorobanOp(operation)) {
+    return null;
+  }
+  const hostFn = operation.functions[0];
+  return getRootInvocationArgs(hostFn);
+};
+
+export const getAttrsFromSorobanHorizonOp = (
```
piyalbasu, line 213: renaming this to make it clear that it's for an operation provided from Horizon

extension/src/popup/views/UnlockAccount/index.tsx
```diff
@@ -41,8 +41,7 @@ export const UnlockAccount = () => {
     const { password } = values;
     await dispatch(confirmPassword(password));
     // skip this location in history, we won't need to come back here after unlocking account
-    history.replace(destination);
-    navigateTo(destination, queryParams);
+    history.replace(`${destination}${queryParams}`);
```
piyalbasu, line 44: fixing a bug where if you weren't logged in when you tried to Sign a transaction and were re-directed to the Unlock Account screen, authenticating was redirecting you to a broken page

@aristidesstaffieri Didn't we talk about this in Slack? I can't seem to find the thread or a Jira ticket :/

aristidesstaffieri: hmm ya know I don't remember this one specifically but this looks right to me now at a glance

extension/src/popup/helpers/soroban.ts
aristidesstaffieri: I don't have a suggested change here but one thought - we end up having these very similar helpers to get contract data, but the only difference between them is - the contract ID, the function name, the args to the function, and the decoders used on the response. I think there's probably some abstraction here, and even cooler I wonder how feasible it would be to write something that can take the contract function signature as input and give you everything else we need for this as output. This looks good to me for this though.

extension/src/popup/components/signTransaction/Operations/index.tsx
aristidesstaffieri: I think one future improvement could be to hold certain high level values related to Soroban tokens in a globally available store for uses like this. Since users have to land on the account tab first when using Freighter, that's a good opportunity to save these off for uses in other workflows in the wallet. Decimals, along with the rest of the token interface can never change so maybe even worth thinking about options for persisting these past a single session.

aristidesstaffieri (approved):
lgtm!

extension/src/popup/components/signTransaction/Operations/index.tsx
piyalbasu: Yeah, definitely! So my original idea was just to grab this from `tokenBalances` in redux, but realized that `getTokenBalances` doesn't run when you just launch `SignTransaction` from freighter-api, we just run that in the `Account` view when you open the extension. There's also the case where a user may want to sign a transaction for a transfer/mint for a token that they haven't loaded into Freighter yet.

So, they would need to `getTokenBalances` to get all their tokens, check to see if they have the one we're looking for, and then possibly do an additional fetch to get the token related to this transaction. Rather than do all that, I thought it'd be more efficient for v1 just to fetch the one decimal we were looking for

But definitely agree that we should come up with something more ergonomic for this

PR #887, opened 2023-07-10
refactors getTxArgs to be able to use args instead of auth args

aristidesstaffieri:
Bringing these learnings back from the example dapps.

It turns out that our arg parsing here kinda worked by coincidence. Since both mint & transfer on the token interface only use tx arg, auth args had all arguments but in any case that uses auth next this will only represent the args that were used in contract auth.

This updates our helpers to pull args from `hostFn.args` instead which is the raw list of `[contractId, methodSymbol, ...method arguments]`

### DIFF
--- a/extension/src/popup/helpers/soroban.ts
+++ b/extension/src/popup/helpers/soroban.ts
@@ -6,15 +6,6 @@ import { decodeScVal, valueToI128String } from "@shared/api/helpers/soroban";
 import { NetworkDetails } from "@shared/constants/stellar";
 import { SorobanContextInterface } from "popup/SorobanContext";
 
-interface RootInvocation {
-  _attributes: {
-    contractId: Buffer;
-    functionName: Buffer;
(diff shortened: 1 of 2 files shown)


extension/src/popup/helpers/soroban.ts
```diff
@@ -168,22 +159,19 @@ export const getOpArgs = (fnName: string, args: SorobanClient.xdr.ScVal[]) => {
...
+  if (!args.length) {
     return null;
   }
 
-  const {
-    _attributes: attrs,
-  } = (txAuth[0].rootInvocation() as unknown) as RootInvocation;
-
-  const fnName = attrs.functionName.toString();
+  const fnName = (methodSymbol.value() as string).toString();
```
aristidesstaffieri, line 173: this kinda sucks but the type isn't generic even though its from a union so I think you have to cast these until they become generic in the sdk.

piyalbasu (approved):

PR #912, opened 2023-07-25
API signTransaction accepts arbitrary blobs

aristidesstaffieri:
This extends signTransaction to be able to do sign arbitrary blobs encoded as b64.
It splits our singing workflows into either tx or blob and splits the review screen as well.

Adds new `signBlob` which is an equivalent to signTransaction, which signs arbitrary base64 data blobs.


extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -105,94 +105,109 @@ export const freighterApiMessageListener = (
...
-                collectedTags = collectedTags.filter(
-                  (tag) => tag !== TRANSACTION_WARNING.unsafe,
-                );
-                collectedTags = collectedTags.filter(
-                  (tag) => tag !== TRANSACTION_WARNING.malicious,
-                );
+    // try to build a tx xdr, if you cannot then assume the user wants to sign an arbitrary blob
+    let encodedBlob = ""
+    try {
+      const transaction = SDK.TransactionBuilder.fromXDR(
```
piyalbasu, line 120: Yeah, I like your idea of splitting this up into a new API method called `signBlob`. I think signing a blob is a different enough, and perhaps more of a super-user feature, that we can just have its own dedicated method for it rather than splitting it here on the try/catch

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -63,28 +62,299 @@ import {
...
+      tx={blobOrTx}
+     />
+  )
+};
+
+interface SignBlobBodyProps {
+  blob: BlobToSign
+}
+
+const SignBlobBody = ({ blob }: SignBlobBodyProps) => {
```
piyalbasu, line 93: what are the differences between this and SignTxBody? It seems like there's some repeated logic here. Is it possible to either combine these into one function or split out the reusable stuff? (for ex: selecting the correct account, checking password, showing warning messages, etc.)

aristidesstaffieri: yeah I think so, looking at it more I think we can probably define something like `isBlob` because the main difference is you don't show some elements for blobs and change some copy slightly.

aristidesstaffieri:
@piyalbasu one detail here is that blob must be b64 encoded right now, I can't think of any problems about this right away but wanted to bring it up.

extension/src/popup/views/SignTransaction/index.tsx
aristidesstaffieri: I refactored this in cddc615c530f802fc23ca3652c2fbb9a1ccc042b to move common state/dom up to the parent, there is still some repetition in the markup but it seemed better to me to have this bit of repetition for now vs trying to make one common render with more complicated logic to switch between blob or tx mode, I suspect that the render will only get more complicated as we expand on how we display common blobs. What do you think of this approach? @piyalbasu

extension/src/background/messageListener/freighterApiMessageListener.ts
aristidesstaffieri: Added in f3865d3d3d44002448a251d1a4055d4852bc624f

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -63,28 +62,348 @@ import {
...
+  dispatch: AppDispatch
+  setStartedHwSign: (hasStarted: boolean) => void
+  setIsPasswordRequired: (isRequired: boolean) => void
+  setIsDropdownOpen: (isRequired: boolean) => void
+  rejectAndClose: () => void
+  handleApprove: (signAndClose: () => Promise<void>) => () => Promise<void>
+}
+
+const SignBlobBody = ({ publicKey, allAccounts, isDropdownOpen, handleApprove, isConfirming, accountNotFound, rejectAndClose, currentAccount, setIsDropdownOpen, setIsPasswordRequired, isPasswordRequired, blob, isExperimentalModeEnabled, t, dispatch, hwStatus, isHardwareWallet, setStartedHwSign }: SignBlobBodyProps) => {
+  const { accountToSign, domain, isDomainListedAllowed } = blob
```
piyalbasu, line 208: looks like you need to run prettier on this file to clean up some of the formatting

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -63,28 +62,348 @@ import {
...
+  t: any
+  dispatch: AppDispatch
+  setStartedHwSign: (hasStarted: boolean) => void
+  setIsPasswordRequired: (isRequired: boolean) => void
+  setIsDropdownOpen: (isRequired: boolean) => void
+  rejectAndClose: () => void
+  handleApprove: (signAndClose: () => Promise<void>) => () => Promise<void>
+}
+
+const SignBlobBody = ({ publicKey, allAccounts, isDropdownOpen, handleApprove, isConfirming, accountNotFound, rejectAndClose, currentAccount, setIsDropdownOpen, setIsPasswordRequired, isPasswordRequired, blob, isExperimentalModeEnabled, t, dispatch, hwStatus, isHardwareWallet, setStartedHwSign }: SignBlobBodyProps) => {
```
piyalbasu, line 207: what do you think about moving SignBlobBody and SignTxBody to their own `components` files? Might make this file a little easier to digest

piyalbasu: Also, if you move these to their own files - you won't need to pass `t` and `dispatch`, you can just re-instantiate them with `useTranslation` and `useDispatch` respectively

extension/src/popup/views/SignTransaction/index.tsx
aristidesstaffieri: oh whoops, I expected it to run on commit 🤔 I'll run it manually

extension/src/popup/views/SignTransaction/index.tsx
aristidesstaffieri: yup, can do on both 👍

aristidesstaffieri: I split out the body components in their own files, moved t and dispatch into the body components and formatted the files manually, all in 5f96ea119d1879752620ec642e002d64e92bea1b

extension/src/background/messageListener/freighterApiMessageListener.ts
Shaptic: 100% agreed: signing arbitrary data is **very** dangerous

aristidesstaffieri: @Shaptic can you expand your thoughts on this? I'm trying to think through ways to make this safer. I think we can safely decode the b64 blob and display it in the signing window/make the window clearly show that this is not XDR, but can you think of anything else we can do to make this safer for end users?

Shaptic: I think a warning around "Do not sign random blobs unless you know what you're signing/doing, trust their source, etc." would be really helpful! As in "we have no idea what this blob represents so be aware of that"

aristidesstaffieri: @piyalbasu I added these kind of off the cuff, any thoughts or additions for this warning?
Screenshot 2023-07-31 at 2 40 58 PM

Screenshot 2023-07-31 at 2 41 10 PM

aristidesstaffieri: It also shows the blob you are signing underneath

piyalbasu: Nice - I think this is a great start. We can run this by Bruno to see if he wants to adjust this copy at all

piyalbasu (approved):

PR #924, opened 2023-08-02
fixes bad references to request/response keys in the listener layer

aristidesstaffieri:
I named these incorrectly when splitting the tx and blob paths, tested locally on both paths.

### DIFF
--- a/@shared/api/external.ts
+++ b/@shared/api/external.ts
@@ -21,7 +21,7 @@ export const requestPublicKey = async (): Promise<string> => {
 };
 
 export const submitTransaction = async (
-  blob: string,
+  transactionXdr: string,
   opts?:
     | string
     | {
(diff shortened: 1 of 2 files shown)


@shared/api/external.ts
```diff
@@ -81,7 +81,7 @@ export const submitBlob = async (
   const { network, networkPassphrase, accountToSign } = opts;
   try {
     response = await sendMessageToContentScript({
-      transactionXDR: blob,
+      transactionXdr: blob,
```
piyalbasu, line 84: can we call this var `transactionBlob` for clarity?

You'll need to change the var name in `freighterApiMessageListener` as well

piyalbasu: I also noticed that the function in `freighterApiMessageListener` isn't using the `network` and  `networkPassphrase` params. Can we remove?

aristidesstaffieri: ah yeah totally we can get rid of those, assuming that a blob has no relation to a specific network.

aristidesstaffieri: oh but I guess it wouldn't be transactionBlob, the data can be any b64 encoded data so how about just blob?

aristidesstaffieri: addressed these in 11e8d4cef1b23309fefbd14edc064323e1221e87

piyalbasu (approved):

PR #929, opened 2023-08-04
Feature/safari support

aristidesstaffieri:
Adds fastlane support in order to deploy to Safari from an action in CI.

Adds:
- new action submitSafari(separate from submitProduction because we need to use a mac image for xcode
- fastlane integration
- safari store env vars


aristidesstaffieri:
Should this be based from master? It's only build related so it seems ok to me but happy to target it for a release branch also.

piyalbasu:
> Should this be based from master? It's only build related so it seems ok to me but happy to target it for a release branch also.

Yup, I think master makes sense. Do you want to double check the formatting changes? The line breaks seem a little off on some of them

piyalbasu (approved):

aristidesstaffieri:
> > Should this be based from master? It's only build related so it seems ok to me but happy to target it for a release branch also.
> 
> Yup, I think master makes sense. Do you want to double check the formatting changes? The line breaks seem a little off on some of them

Hmm which line? I took it into the editor that parses it and it didn't complain, I don't visually see it yet 🤔

aristidesstaffieri:
@piyalbasu we have the env vars set now, so once we are good here should I re-run the action and submit to the store for the first time?

piyalbasu (approved):

PR #931, opened 2023-08-05
first pass at upgrading to node v18 and webpack 5

piyalbasu:
This still needs some testing, but running through most happy paths, this appears to be mostly working 🎉

Also removing some freighter-api tarballs I had mistakenly committed to the repo 😬


extension/src/popup/components/ErrorTracking/index.tsx
```diff
@@ -3,15 +3,15 @@ import * as Sentry from "@sentry/browser";
 import { Integrations } from "@sentry/tracing";
 
 import { settingsDataSharingSelector } from "popup/ducks/settings";
-import { version } from "../../../../package.json";
+import packageJson from "../../../../package.json";
```
piyalbasu, line 6: this started throwing a warning - it didn't like a named export from a json file

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -15,34 +16,28 @@ import { AddToken } from "popup/components/manageAssets/AddToken";
...
     destinationBalances,
     assetSelect,
     error,
   } = useSelector(transactionSubmissionSelector);
+  const { networkPassphrase } = useSelector(settingsNetworkDetailsSelector);
   const [errorAsset, setErrorAsset] = useState("");
 
-  React.useEffect(() => {
-    const parseXdr = async () => {
-      const { networkPassphrase } = await getNetworkDetails();
```
piyalbasu, line 31: This was throwing an error when running locally in the browser on localhost. `getNetworkDetails` relied on accessing chrome.storage to get the network info. This API, however, doesn't exist in the browser. So, in the `popup`, we should rely on `background` to grab stuff from `chrome.storage` and transmit it to `popup`.

This works fine in production because `popup` gets packaged into the extension and thus has access to chrome.storage. But we should structure our code in a way that works in both the extension and in the browser.

piyalbasu: Not sure if this worked before and maybe we didn't notice? 🤷

aristidesstaffieri: ah interesting, but feels weird that we wouldn't have seen this in dev right? Either way this lgtm.

aristidesstaffieri (approved):
lgtm, thanks! 🚀

PR #945, opened 2023-08-16
Bugfix/tx history restore bump

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/browse/WAL-1008

This adds handling for the bump or restore operations that are new in Soroban with preview 10.
Previously we assumed that any Soroban related transactions would be accounted for in our `OPERATION_TYPES` enum, and this was ok because we controlled what contracts we supported(just token basically).

P10 introduced state expiry which means that now any user may have restored or bumped state through a transaction which broke this model.

This is fixed in 2 ways -
1. We add the `OPERATION_TYPES` key for bump and restore ops.
2. We no longer assume anything about unknown op types and render them as generic host invoke functions.

Unrelated bug fix -
I found that the mint token history item was flawed because it assumed the the minter already had some of the token, and the token ID saved in their tokenIds list. It also didn't account for the mint to be to the current wallet, so it didn't properly show a change in balance.
This is fixed by attempting to fetch token details when the mint operation is for a token that we do not have in our tokenIds list, and accounting for a change in balance when the mint is to the current public key.

This also includes refactors/re-organization of Soroban related helpers/tooling.

moved helpers to the shared namespace.
added server helper for simTx.
added specific token helpers and refactored to use new simTx helper.
removed helpers that could be replaced by ones in soroban-client.

Unresolved Issues -
Working on this uncovered interesting UX fallout from now potentially dealing with tokens where the state or code can be expired. For either account details or account history, we could run into a scenario where the contract has an expired entry and then we fail to fetch token details through a transaction simualtion, we can do a few things -

If the contract is a SAC, we can get this info from Horizon and we should anyway because it's free.
Quote from Dima about knowing when a contract is a SAC - 
```
that's a good question. I'm not sure how SAC tokens should be added. they can be added either via classic Asset or via contract id. in the latter case you can easily figure out that it's SAC (just lookup executable in the contract instance entry), and then I believe you can get Asset spec from the token name (it's ASSET_NAME:ISSUER_KEY)
```
We could also index Soroban to always have up to date data about every token contract that we care about, quote from Tomer on that
```
Using SAC optimizations is an ok temporary fix. With that said I think we should take a step back and acknowledge that soroban-rpc is actually a poor wallet backend. and that’s ok - that’s not what it was built for. So obviously we can tackle building a wallet backend, though that might be a lot of work. We can see if we can utilize the work of someone else in the ecosystem. maybe stellar.expert (I suspect orbit will have something like this ready) or one of the incoming indexers. Or, we can potentially solve some of these issues by having more robust client-side state management
```

There is also the option to fetch the ledger entry directly, but I am still gathering details about it and don't yet understand how that would be different from the RPC getting it.


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -80,7 +84,8 @@ export const HistoryItem = ({
     sourceAssetCode = operation.source_asset_code;
   }
   const operationType = camelCase(type) as keyof typeof OPERATION_TYPES;
-  const operationString = `${OPERATION_TYPES[operationType]}${
+  const opTypeStr = OPERATION_TYPES[operationType] || t("Transaction");
```
aristidesstaffieri, line 87: this was basically the source of the bug

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -112,169 +109,229 @@ export const HistoryItem = ({
...
-        new BigNumber(attrs.amount),
-        Number(token.decimals),
-      );
-      PaymentComponent = (
-        <>
-          {formattedTokenAmount} {token.symbol}
-        </>
-      );
-      IconComponent = <Icon.ArrowUp className="HistoryItem__icon--sent" />;
+  React.useEffect(() => {
```
aristidesstaffieri, line 127: I needed to make this async to better handle the mint token case, explained in the PR description.

@shared/api/internal.ts
```diff
@@ -859,94 +863,37 @@ export const getBlockedAccounts = async () => {
...
-    _prev[curr] = decoder(result.xdr);
-
-    return _prev;
-  }, Promise.resolve({} as SorobanTokenRecord));
+  // for now we need to do 4 tx simulations to show 1 user balance. :(
+  // TODO: figure out how to fetch ledger keys to do this more efficiently
+  const decimals = await getDecimals(contractId, server, txBuilders.decimals);
+  const name = await getName(contractId, server, txBuilders.name);
+  const symbol = await getSymbol(contractId, server, txBuilders.symbol);
+  const balance = await getBalance(
```
piyalbasu, line 884: nice - I find this pattern a lot more straightforward!

@shared/helpers/soroban/server.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!results || results.length !== 1) {
+    throw new Error("Invalid response from simulateTransaction");
+  }
+  const result = results[0];
+  const scVal = xdr.ScVal.fromXDR(result.xdr, "base64");
+  let convertedScVal: any;
+  try {
+    // handle a case where scValToNative doesn't properly handle scvString
+    convertedScVal = scVal.str().toString();
+    return convertedScVal;
```
piyalbasu, line 26: Ah, I _believe_ this should be fixed now in the last soroban-client. So I don't think you need to handle this scvString case

@shared/helpers/soroban/token.ts
```diff
@@ -0,0 +1,93 @@
...
+) => {
+  const contract = new Contract(contractId);
+
+  const tx = builder
+    .addOperation(contract.call("symbol"))
+    .setTimeout(TimeoutInfinite)
+    .build();
+
+  const result = await simulateTx<string>(tx, server);
+  return result;
```
piyalbasu, line 92: no need to do this now since things could still change - but if this API holds, in the future we could probably refactor these helpers to share some logic as they're all kinda doing the same thing.

Also, @Shaptic - do we consider this a common enough use case (getting balance + decimals + name + symbol) that soroban-client should offer this?

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -112,169 +109,229 @@ export const HistoryItem = ({
...
-        to: account,
-      },
-      operationText: `-${new BigNumber(startingBalance)} XLM`,
-    };
-  } else if (isSorobanTx) {
-    const attrs = getAttrsFromSorobanHorizonOp(operation, networkDetails);
-    const token = tokenBalances.find(
-      (balance) => attrs && balance.contractId === attrs.contractId,
-    );
+  const [txDetails, setTxDetails] = React.useState(transactionDetailPropsBase);
```
piyalbasu, line 112: Super NIT: but in other files in the repo, we usually just `import { useState } from "react"` at the top of the file so we just call `useState` directly instead of calling `React.useState`. So minor, but would be good to keep the pattern consistent

piyalbasu: Same deal with `useEffect`

piyalbasu (approved):

piyalbasu (approved):

@shared/helpers/soroban/server.ts
aristidesstaffieri: oh yeah I did hear that also, I'll test it out to make sure.

@shared/helpers/soroban/token.ts
aristidesstaffieri: Additionally - there's also a good convo happening about how we should be fetching data for Soroban now that expired state can be involved. It seems like we'll have to at least augment our approach anyway pretty soon. Sem good ideas that came up includes -
1. Caching things like decimals/name/symbol when a user firsts gets a balance of a token.
2. Fetching the ledger entries directly(this is cheaper than the rpc call).
3. Getting these details from Horizon when token is a SAC.
4. Allowing users to restore their data when their token data is expired(implies we start storing expiration time along with other details).

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
aristidesstaffieri: ah whoops, yup will update.

@shared/helpers/soroban/token.ts
Shaptic: I don't think this belongs in soroban-client, per se, but maybe this be handled by using soroban-cli to generate the TypeScript bindings for the SAC? cc @chadoh @willemneal

extension/src/constants/transaction.ts
```diff
@@ -28,6 +28,7 @@ export enum OPERATION_TYPES {
   revokeTrustlineSponsorship = "Revoke Trustline Sponsorship",
   setOptions = "Set Options",
   setTrustLineFlags = "Set Trustline Flags",
+  bumpFootprintExpiration = "Bump Footprint Expiration",
```
Shaptic, line 31: missing `restoreFootprint`?

extension/src/popup/SorobanContext.tsx
```diff
@@ -9,9 +9,11 @@ import {
 
 import { settingsNetworkDetailsSelector } from "./ducks/settings";
 
+const BASE_FEE = "100";
```
Shaptic, line 12: you should use `SorobanClient.BASE_FEE`, I think

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -281,33 +281,20 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
-        ...[
-          accountIdentifier(publicKey), // from
-          accountIdentifier(destination), // to
-          numberToI128(parsedAmount.toNumber()), // amount
-        ],
-      );
+      const params = [
+        new SorobanClient.Address(publicKey).toScVal(), // from
+        new SorobanClient.Address(destination).toScVal(), // to
+        SorobanClient.nativeToScVal(parsedAmount.toNumber(), { type: "i128" }), // amount
```
Shaptic, line 287: `new SorobanClient.XdrLargeInt('i128', parsedAmount).toScVal()` may be faster/more direct, I think?

@shared/helpers/soroban/server.ts
aristidesstaffieri: done in 1ba0450fb6a466570f2ab0318b838f1072f0ae47

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
aristidesstaffieri: done in 1ba0450fb6a466570f2ab0318b838f1072f0ae47

extension/src/constants/transaction.ts
aristidesstaffieri: yes it is, thanks!

aristidesstaffieri: done in 05fa8fe00e388f8cbf24ba50b6f75548e8ba489d

extension/src/popup/SorobanContext.tsx
aristidesstaffieri: thanks, I keep meaning to update that. Done in 05fa8fe00e388f8cbf24ba50b6f75548e8ba489d

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: done in 05fa8fe00e388f8cbf24ba50b6f75548e8ba489d

aristidesstaffieri:
New Entries look like this, also improved mint & transfer a bit.

Screenshot 2023-08-18 at 11 30 30 AM

PR #949, opened 2023-08-24
Sign Auth Entry API & modal flow

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/36?modal=detail&selectedIssue=WAL-1016&assignee=63bc81730a1b5442166a2f79

This adds a new API `signAuthEntry` and a matching popup flow for signing auth entries.
Splits `signTx` and `signBlob`, so now we have `sign-transaction` and `sign-blob` paths instead of a shared one.
Adds `buildInvocationTree`(can be deleted once the sdk version lands) for crawling and unpacking xdr structures from a root invocation. Adds this to the `Transaction` component used to render transaction details during signing. Applied feedback to focus more on post sim `auth` than `op.func` in order to surface the invocation.


extension/src/popup/components/signAuthEntry/invocation.ts
```diff
@@ -0,0 +1,87 @@
+import { Asset, Address, scValToNative, xdr } from "soroban-client";
+
+interface RootOutput {
+  type: string;
+  args: Record<string, any>;
+  subInvocations: RootOutput[];
+}
+
+export function buildInvocationTree(root: xdr.SorobanAuthorizedInvocation) {
```
aristidesstaffieri, line 9: This can be deleted once the sdk version is out.

extension/src/popup/components/signBlob/index.tsx
```diff
@@ -10,6 +10,6 @@ interface BlobProps {
 export const Blob = (props: BlobProps) => (
   <Card variant={Card.variant.highlight}>
     <Heading4>Signing data:</Heading4>
-    <div className="Blob">{props.blob}</div>
+    <div className="Blob">{atob(props.blob)}</div>
```
aristidesstaffieri, line 13: not sure why we didn't decode the b64 blob before, just oversight.

aristidesstaffieri:
There is one more refactor here that I haven't made because this PR is already large but I could be convinced to do it here also.

We should split `SignTransaction` to different render branches for the `Transaction` vs `FeeBumpTransaction` cases. Right now, is we let the types come through properly we can uncover some edge cases of our current approach. Also we can refactor this component to better make use of the xdr class methods instead of digging into the xdr class internals in order to improve `signTransaction/Operations` to more generically cover all transactions cases.

I plan to do this as a separate PR to keep things comprehensible but let me know if you feel strongly that this should happen in this one.

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -60,11 +102,45 @@ export const SignTransaction = () => {
...
+  } = tx;
+
+  /* 
+  Reconstruct the tx from xdr as passing a tx through extension contexts 
+  loses custom prototypes associated with some values. This is fine for most cases 
+  where we just need a high level overview of the tx, like just a list of operations.  
+  But in this case, we will need the hostFn prototype associated with Soroban tx operations.
+  */
+
+  const SDK = isExperimentalModeEnabled ? SorobanSdk : StellarSdk;
```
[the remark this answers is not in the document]
Shaptic: Wait no, `StellarSdk` is [`js-stellar-sdk`](https://github.com/stellar/js-stellar-sdk) which connects to Horizon.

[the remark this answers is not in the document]
aristidesstaffieri: ah whoops, thanks for catching that. I thin the rest of it still makes sense but let me know if you think otherwise.

extension/src/popup/components/signTransaction/Operations/index.tsx
```diff
@@ -166,6 +168,37 @@ const KeyValueWithScValue = ({
...
+  const rootJson = buildInvocationTree(authEntry.rootInvocation());
+  return (
+    <div className="Operations__pair--smart-contract">
+      <div>
+        {operationKey}
+        {operationKey ? ":" : null}
+      </div>
+      <SimpleBarWrapper className="Operations__scValue">
+        <div>
+          <pre>{JSON.stringify(rootJson, null, 2)}</pre>
```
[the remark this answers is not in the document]
aristidesstaffieri: oh in the version I wrote I handled it inline but can switch to handling in after building the tree once I pull in the sdk one.

extension/src/helpers/urls.ts
```diff
@@ -11,6 +11,15 @@ export interface BlobToSign {
   accountToSign: string;
 }
 
+export interface EntryToSign {
+  isDomainListedAllowed: boolean;
+  domain: string;
+  tab: browser.Tabs.Tab | undefined;
```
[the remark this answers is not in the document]
aristidesstaffieri: for sure, thanks. Done in 63a77ea851e016ad3f52821450823fff7d6c7b75

@shared/api/external.ts
```diff
@@ -95,6 +95,32 @@ export const submitBlob = async (
...
+  const _opts = opts || {};
+  const accountToSign = _opts.accountToSign || "";
+  try {
+    response = await sendMessageToContentScript({
+      entryXdr,
+      accountToSign,
+      type: EXTERNAL_SERVICE_TYPES.SUBMIT_AUTH_ENTRY,
+    });
+  } catch (e) {
+    console.error(e);
```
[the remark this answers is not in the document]
aristidesstaffieri: I think it's more appropriate to throw here because that would be an implementation error when the other is a type or API error. Added in 63a77ea851e016ad3f52821450823fff7d6c7b75 but let me know if you disagree.

extension/src/popup/components/signAuthEntry/AuthEntry/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { buildInvocationTree } from "../invocation";
+import "./styles.scss";
+
+interface TransactionProps {
+  authEntryXdr: string;
+}
+
+export const AuthEntry = ({ authEntryXdr }: TransactionProps) => {
+  const { t } = useTranslation();
+  const authEntry = xdr.SorobanAuthorizationEntry.fromXDR(authEntryXdr, "hex");
```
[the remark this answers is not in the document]
aristidesstaffieri: ah thanks, I was just testing with a hex string I had. Updated to b64 in 63a77ea851e016ad3f52821450823fff7d6c7b75

@shared/api/external.ts
```diff
@@ -95,6 +97,33 @@ export const submitBlob = async (
...
+  }
+  const { signedAuthEntry, error } = response;
+
+  if (error) {
+    throw error;
+    throw error;
```
piyalbasu, line 122: repeated line

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -88,18 +89,20 @@ export const AccountHistory = () => {
   const stellarExpertUrl = getStellarExpertUrl(networkDetails);
 
   // differentiate between if data is still loading and if no account history results came back from Horizon
+  const shouldLoadToken =
+    isExperimentalModeEnabled || networkDetails.network === NETWORKS.FUTURENET;
```
piyalbasu, line 93: should this be `&&`? It'll be using `stellar-sdk` instead of `soroban-client` if `isExperimentalModeEnabled` is false

piyalbasu, line 93: isn't this more dependent on `isExperimentalModeEnabled`? If that flag is true, `soroban-client` will be used, but it will use `stellar-sdk` if it's false.

Also, since we're going live on Testnet in a few weeks, maybe we should just key this off of `isExperimentalModeEnabled`. I'm guessing the Futurenet check is to make sure the data exists (or something like that). Maybe we should gracefully fail if the data isn't found on network

aristidesstaffieri: i'm not sure that I understand, how does this toggle the sdk? AFAIk this toggles wether we fetch and add soroban tokens to the account history. The network check is needed because you can be in experimental mode while not being on Futurenet, so as currently implemented it incorrectly shows tokens that are not for the network when on Testnet.

extension/src/popup/views/SignBlob/index.tsx
```diff
@@ -1,85 +1,141 @@
...
+    setIsConfirming(false);
+  };
+
+  useEffect(() => {
+    if (startedHwSign && hwStatus === ShowOverlayStatus.IDLE) {
+      window.close();
+    }
+  }, [startedHwSign, hwStatus]);
+
+  useEffect(() => {
```
piyalbasu, line 107: I think something you did a good job of in the PR to add sign-blob was put this "auto-select account/set current account" logic in one place and pass it to different views. I think you should do the same here so we're not repeating this logic in all the 3 `sign-X` views

@shared/api/external.ts
aristidesstaffieri: whops, thanks for the catch. Fixed in 4b0d05a220d333be1420ba3a45be60475512e133

extension/src/popup/views/SignBlob/index.tsx
aristidesstaffieri: Hmm yeah I chose the repetition because they no longer share a common ancestor where it would make sense to hold common logic. I think the solution now would have to be a wrapper component that does the shared stuff and passes it to a generic child prop. Let me give that a shot then.

extension/src/popup/views/AccountHistory/index.tsx
piyalbasu: Oh actually - you're right. We're just grabbing the Horizon entry, so we can just use `stellar-sdk` here. I was thinking of other places in the app where we do: `const SDK = isExperimentalModeEnabled ? SorobanSdk : StellarSdk;`. You can ignore that first part 😛

aristidesstaffieri: but still I think also the second part we cannot do right now because the tokens as we store them today have no relation to any network(we should address this before testnet) so when on Testnet but with experimental mode toggled on, you incorrectly see tokens that are on Futurenet without the network check.

piyalbasu: I guess for the second part: I think instead of checking for Futurenet explicitly here, we should adjust the network url that `SorobanContext` is using.  On closer inspection, looks like `SorobanContext` is hardcoded to use Futurenet. So, in `getTokenBalances` on ln 162, it's always checking Fututenet. 

We should fix that in SorobanContext (but maybe we just do that in a later PR)

This is mostly so when Soroban is running on both Futurenet and Testnet, we have fewer places in the code where we have explicit checks for Futurenet that we have to manually remove

aristidesstaffieri: We must have been writing this at the same time. Yeah we should def be able to configure the SorobanContext for other networks before Testnet support is live. We still can't set the network in the SorobanContext to whatever the current network is here to fix this bug because a) the saved tokens have no relation to a network today, b) There is still only futurenet support so this is incorrect today.

aristidesstaffieri: There's def a bigger refactor here to support future networks but this is just meant to address the live bug where we show users the wrong tokens when in experimental mode and not on Futurenet. In order to fix this to work for other networks, we also need to refactor how we store tokens to include network details, then change all token related checks to instead map to tokens that are saved for that network.

extension/src/popup/views/SignBlob/index.tsx
piyalbasu: Might also work as a custom hook?

Really want to avoid the repetition bc the logic is a little hairy

extension/src/popup/views/AccountHistory/index.tsx
piyalbasu: Cool - yeah, let's just leave this is as-is then, and we can tackle the network stuff later on

aristidesstaffieri: but good call out as I hadn't really considered before this that there will be significant work to support tokens on other networks, and it will include the always hairy work of migrating storage schemas.

@shared/api/external.ts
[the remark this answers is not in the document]
Shaptic: def up to y'all cuz I'm unfamiliar w/ your error-handling conventions, I just more wanted to call out the distinction between exceptions and the response.error field

extension/src/popup/views/SignTransaction/index.tsx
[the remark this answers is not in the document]
aristidesstaffieri: hold up, I'm back to my original statement that we do not need to do this because we only use an sdk in this workflow to rebuild the tx from xdr with `TransactionBuilder` which is available and the same through either sdk so we should just use `SorobanSdk` here.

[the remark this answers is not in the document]
Shaptic: oh sure yeah, that makes sense. but if you guys need horizon as part of freighter beyond experimental/futurenet mode then you may need both (until https://github.com/stellar/js-stellar-sdk/issues/850 lands)

extension/src/popup/views/SignBlob/index.tsx
aristidesstaffieri: made a new hook that worked pretty nicely for this in 47ea4f923d0ef3128e286610f5b783a9688ce084

extension/src/popup/views/SignTransaction/index.tsx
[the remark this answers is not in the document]
aristidesstaffieri: ah nice I'll watch that one. I think we mostly did this in anticipation and don't really need to do it many of the places that we do.

extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -67,6 +68,25 @@ export const SignAuthEntry = () => {
...
+      <ModalWrapper>
+        <WarningMessage
+          variant={WarningMessageVariant.warning}
+          handleCloseClick={() => window.close()}
+          isActive
+          header={t("Unsupported signing method")}
+        >
+          <p>
+            {t(
+              "Hardware wallets are currently not supported for signing arbitrary blobs.",
```
Shaptic, line 82: The phrasing/ordering of words here _really_ throws me off for some reason :joy: suggestion:

```suggestion
              "Signing arbitrary data with a hardware wallet is currently not supported.",
```

@shared/constants/stellar.ts
```diff
@@ -23,7 +23,9 @@ export enum FRIENDBOT_URLS {
   FUTURENET = "https://friendbot-futurenet.stellar.org",
 }
 
-export const SOROBAN_RPC_URLS = {
+export const SOROBAN_RPC_URLS: { [index: string]: string } = {
+  PUBLIC: "https://rpc-futurenet.stellar.org/",
```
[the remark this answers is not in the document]
aristidesstaffieri: ah whoops, didn't mean to commit this actually. It got included by the translations action 😄 . Backing out.

extension/src/popup/views/SignAuthEntry/index.tsx
aristidesstaffieri: works for me, done in d1f4093f9203f7bba5a6395d1a19540fd02ef47b

@shared/constants/stellar.ts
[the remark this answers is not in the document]
aristidesstaffieri: fixed in e03372a9570ae51c0a4cf2f96a5c2428571b4b97

extension/src/popup/helpers/useSetupSigningFlow.ts
```diff
@@ -0,0 +1,138 @@
...
+} from "popup/ducks/accountServices";
+
+import {
+  ShowOverlayStatus,
+  startHwSign,
+  transactionSubmissionSelector,
+} from "popup/ducks/transactionSubmission";
+
+export function useSetupSigningFlow(
+  dispatch: AppDispatch,
```
piyalbasu, line 24: do you need to pass dispatch? or can you just instantiate a new one with `useDispatch`? Might make the API a little nicer to have one fewer param

piyalbasu (approved):

extension/src/popup/helpers/useSetupSigningFlow.ts
aristidesstaffieri: yeah 100%, moved in 06295ca863c4e3674866541069298a58a539e699

PR #950, opened 2023-08-28
fix inconsistencies with SDS v1

piyalbasu:
This mostly fixes issues with light/dark mode and list styling.

Before there was some unexpected behaviors when viewing Freighter in Light Mode.  Now, all styles hardcode to Dark Mode regardless of user preference. We'll incorporate Light Mode in a later release

This also overrides the standard SDS styling of using a `-` before all list items as this breaks Freighter's current li styling

### DIFF
--- a/extension/src/popup/components/PasswordRequirements/styles.scss
+++ b/extension/src/popup/components/PasswordRequirements/styles.scss
@@ -2,15 +2,15 @@
   padding: 1.25rem 0 2.5rem;
 
   &__list {
-    color: rgba(255, 255, 255, 0.6);
-    list-style-type: disc;
+    color: var(--color-gray-60) !important;
+    list-style-type: disc !important;
     list-style-position: inside;
     font-size: 1rem;
     padding: 0;
     margin: 0;
 
-    li::before {
-      content: none;
+    ::before {
+      content: none !important;
     }
   }
 }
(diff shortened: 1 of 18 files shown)


extension/src/popup/styles/global.scss
```diff
@@ -19,17 +19,13 @@
   --back--button-dimension: 1rem;
   --back--button-z-index: 2;
 
-  --background-color: var(--color-gray-00);
-  --background-color--overlay: var(--color-gray-10);
-  --secondary-text: var(--color-gray-70);
```
piyalbasu, line 24: interestingly - aliasing these colors doesn't work in scss. When you toggle between light/dark mode, it doesn't grab the respective value, it always just gets the light version. I couldn't quite figure out why this is, but it's probably okay because we wanted to move to a more declarative use of colors where components always just use the same color var names as listed in the figma rather than using a project-dependent alias

extension/src/popup/components/PasswordRequirements/styles.scss
```diff
@@ -2,15 +2,15 @@
...
+    list-style-type: disc !important;
     list-style-position: inside;
     font-size: 1rem;
     padding: 0;
     margin: 0;
 
-    li::before {
-      content: none;
+    ::before {
+      content: none !important;
```
piyalbasu, line 13: we need to manually add this for all lists to clear out SDS default styling. We can figure out how to address this later

aristidesstaffieri (approved):

PR #953, opened 2023-09-06
Network Settings - Soroban RPC URL

aristidesstaffieri:
Ticket: https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/61?modal=detail&selectedIssue=WAL-1077&assignee=63bc81730a1b5442166a2f79
Also fixes: https://stellarorg.atlassian.net/browse/WAL-1083

Renames `URL` field in network settings to `Horizon RPC URL`
Adds `Soroban RPC URL` field to network settings
Updates network settings in storage to support optional `sorobanRpcUrl`
Uses new `sorobanRpcUrl` field to replace hardcoded soroban rpc url throughout
Migrates Network details to add `sorobanRpcUrl`
Adds handling for optional `sorobanRpcUrl`


extension/src/background/helpers/dataStorage.ts
```diff
@@ -105,3 +106,23 @@ export const migrateFriendBotUrlNetworkDetails = async () => {
 
   await localStore.setItem(NETWORKS_LIST_ID, migratedNetworkList);
 };
+
+export const migrateSorobanRpcUrlNetworkDetails = async () => {
```
aristidesstaffieri, line 110: We'll have to do this again once we support Soroban in testnet/pubnet

extension/src/popup/SorobanContext.tsx
```diff
@@ -31,23 +31,36 @@ export const SorobanProvider = ({
...
+    server = new SorobanClient.Server(serverUrl, {
+      allowHttp: serverUrl.startsWith("http://"),
+    });
+
+    newTxBuilder = () =>
+      new SorobanClient.TransactionBuilder(source, {
+        fee,
+        networkPassphrase: networkDetails.networkPassphrase,
+      });
+  } else if (networkDetails.sorobanRpcUrl) {
```
piyalbasu, line 52: pretty NIT: but should this just be an `else` rather than `else if`? We prob don't want to stumble into a scenario where `server` and `newTxBuilder` are undefined, right?

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
```diff
@@ -180,6 +187,13 @@ export const NetworkForm = ({ isEditing }: NetworkFormProps) => {
     }
   };
 
+  const supportsSorobanRpc = (network: string) => {
+    if (network === NETWORK_NAMES.FUTURENET) {
+      return true
+    }
+    return false
+  }
```
piyalbasu, line 195: very NIT: but this would be a little tidier as `const supportsSorobanRpc = (network: string) => network === NETWORK_NAMES.FUTURENET`

extension/src/popup/helpers/soroban.ts
```diff
@@ -111,15 +111,18 @@ export const getContractDecimals = async (
   contractId: string,
 ) => {
   const contract = new SorobanClient.Contract(contractId);
-  const server = sorobanClient.server;
+
+  if (!sorobanClient.server || !sorobanClient.newTxBuilder) {
+    throw new Error("soroban rpc not supported")
+  }
```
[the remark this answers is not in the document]
aristidesstaffieri: well we can't throw it there because there is no way to instantiate a SorobanClient on pubnet/testnet yet so we had to move to an optional server/tx builder in order to allow custom ones as well as support some default networks but not all of them.

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
aristidesstaffieri: oh yeah for sure 😄

extension/src/popup/helpers/soroban.ts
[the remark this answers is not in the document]
aristidesstaffieri: this was sort of overlooked before because we incorrectly passed in a horizon instance when we were on pubnet/testnet.

extension/src/popup/components/manageNetwork/NetworkForm/index.tsx
aristidesstaffieri: done in 704bac55565f361a39826bbfd8d66bc062ced811

piyalbasu (approved):

esteblock:
Hello. What is the difference between normal horizon and this soroban rpc? ( I don't have access to those links)
Also, I think that the freighter-api should be upgraded to get the new sorobanRpc
@joaquinsoza

PR #958, opened 2023-09-13
Refactor/multi network custom token

aristidesstaffieri:
Adds migration to key token ID lists by network.
Refactors storage getters/setters to work with network to get/set token IDs.

Ticket: https://stellarorg.atlassian.net/browse/WAL-1100


extension/src/background/helpers/dataStorage.ts
```diff
@@ -126,3 +127,14 @@ export const migrateSorobanRpcUrlNetworkDetails = async () => {
...
 }
+
+// This migration migrates the storage for custom tokens IDs to be keyed by network
+export const migrateTokenIdList = async () => {
+  const localStore = dataStorageAccess(browserLocalStorage);
+
+  const tokenIdList = await localStore.getItem(TOKEN_ID_LIST)
+
+  // intially this was assumed to be Futurenet
+  const compositeKey = `${TOKEN_ID_LIST}__${Networks.FUTURENET}`
```
piyalbasu, line 138: this might be a little easier to deal with if you structured it as:

```
{
   FUTURENET: {
      [KEYID1]: ["C..."],
      [KEYID2]: ["C..."]
   },
   TESTNET: {
      [KEYID1]: ["C..."]
   }
}
```

I don't think there's anything necessarily wrong with the way you're doing it, but since you can use any kind of data type here, might be more ergonomic to use one where you don't need to encode/decode this compositeKey each time you want to use it

aristidesstaffieri: works for me, I refactored to this in ae537314293f7e7b93d4c612ba244371bf8f9857

extension/src/background/helpers/dataStorage.ts
```diff
@@ -126,3 +127,53 @@ export const migrateSorobanRpcUrlNetworkDetails = async () => {
...
+
+  const isNetwork = (key: string) => (Object.keys(Networks) as Array<keyof typeof Networks>).some((network) => key === network)
+
+  let newStorageSchema = tokenIdsByKey
+  if (tokenIdsByKey) {
+    // intially this was assumed to be Futurenet
+    const futurenetNetworks = Object.keys(tokenIdsByKey).reduce((prev, curr) => {
+      const next = prev
+      // if the key is not a network, assume old schema and move all keys(key ids) under futurenet key
+      if (!isNetwork(curr)) {
```
piyalbasu, line 163: I believe this would be truthy in the new schema if a user added a token under a custom network. I believe curr would equal `CUSTOM` in this case

piyalbasu: I like your idea of adding a schema version! Would it be easier if we introduce a schema version in storage and say `if (!schemaVersion || schemaVersion < 5.5.0)`: don't do any migration?

extension/src/background/index.ts
```diff
@@ -59,7 +60,9 @@ export const initInstalledListener = () => {
   browser?.runtime?.onInstalled.addListener(normalizeMigratedData);
   browser?.runtime?.onInstalled.addListener(migrateFriendBotUrlNetworkDetails);
   browser?.runtime?.onInstalled.addListener(migrateSorobanRpcUrlNetworkDetails);
-  browser?.runtime?.onInstalled.addListener(migrateTokenIdList); 
+  browser?.runtime?.onInstalled.addListener(migrateTokenIdList);
+  // migrate storage version last so that other migrations can compare versions
+  browser?.runtime?.onInstalled.addListener(migrateDataStorageVersion);
```
piyalbasu, line 65: I think to be safe, this should be at the top. I'm not 100% sure, but I imagine these would run sequentially

piyalbasu, line 65: Do these run sequentially? Should this be at the top? I'm not 100% sure but we should double check to be safe

piyalbasu (approved):

PR #972, opened 2023-09-20
Update stellar sdk

piyalbasu:


extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -146,7 +153,7 @@ export const freighterApiMessageListener = (
       (await getIsSafetyValidationEnabled()) && isMainnet;
 
     if (isValidatingMemo || isValidatingSafety) {
-      _operations.forEach((operation: { destination: string }) => {
+      _operations.forEach((operation: any) => {
```
aristidesstaffieri, line 156: Can this `any` be an `xdr.Operation`?

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -179,7 +186,7 @@ export const freighterApiMessageListener = (
     const server = stellarSdkServer(networkUrl);
 
     try {
-      await server.checkMemoRequired(transaction);
+      await server.checkMemoRequired(transaction as any);
```
aristidesstaffieri, line 189: curious that we have to cast these as any after the types come through. I don't have a suggestion yet but seems like this should be an `xdr.Transaction` maybe?

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -90,7 +90,7 @@ export const SignTransaction = () => {
   const transaction = SDK.TransactionBuilder.fromXDR(
     transactionXdr,
     networkPassphrase,
-  );
+  ) as any;
```
aristidesstaffieri, line 93: I think I've run into this before, we can probably drop this `any` in the future if we refactor this to deal with the different transaction types separately.

aristidesstaffieri (approved):
lgtm

PR #973, opened 2023-09-21
Fixes bad check in buildInvocationTree + authEntry sig

aristidesstaffieri:
Fixes 2 unrelated issues -

Our version of buildInvocationTree was wrong and missed an important change in the host function type checks, so now that the sdk has a version of this, we pull that in instead.
Fixes a bad decoding of the auth entry arg, which didn't prevent signing but didn't match client encoding expectations.


Shaptic (approved):
woo! :tada:

piyalbasu (approved):

PR #974, opened 2023-09-21
add Soroban support to testnet

piyalbasu:


extension/src/background/helpers/dataStorage.ts
```diff
@@ -147,14 +148,50 @@ export const migrateTokenIdList = async () => {
...
+    await localStore.setItem(NETWORKS_LIST_ID, migratedNetworkList);
+    await migrateDataStorageVersion("2.0.0");
+  }
+};
+
+export const versionedMigration = async () => {
+  // sequentially call migrations in order to enforce smooth schema upgrades
+
+  await migrateTokenIdList();
+  await migrateTestnetSorobanRpcUrlNetworkDetails();
```
piyalbasu, line 188: the idea here is to make sure people who are multiple migrations behind go through each migration iteratively. jumping ahead to a migration script they're not yet ready for could cause big issues.

I def still need to do some testing on this

extension/src/popup/ducks/settings.ts
```diff
@@ -335,6 +336,13 @@ export const settingsExperimentalModeSelector = createSelector(
   (settings) => settings.isExperimentalModeEnabled,
 );
 
+export const settingsSorobanSupportedSelector = createSelector(
+  settingsSelector,
+  (settings) =>
+    settings.isExperimentalModeEnabled ||
+    settings.networkDetails.network === NETWORKS.TESTNET,
```
piyalbasu, line 343: we should use this new selector (rather than isExperimentalModeEnabled) to determine if we should turn off/on Soroban features

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: oh interesting yeah I guess all of the listeners aren't really sequential.

aristidesstaffieri (approved):

extension/src/background/helpers/dataStorage.ts
piyalbasu: Yeah, I haven't really dug into how they work, but from testing, I've observed that they don't guarantee sequence. Assuming they all just work async 🤷

aristidesstaffieri: for sure, I guess that makes sense with an event loop. This makes sense to me at a glance, even if the events are async, if we only have one "migration" event listener that sequentially runs the migrations then that should enforce order... _I think_

PR #978, opened 2023-09-23
Release/5.6.1

piyalbasu:


extension/src/popup/helpers/parseTransaction.ts
```diff
@@ -12,7 +12,11 @@ export const decodeMemo = (memo: {}) => {
     return get(memo, "_value.low", "");
   }
   const decodeMethod = memoType === MEMO_TYPES.MEMO_HASH ? "hex" : "utf-8";
-  return buffer.Buffer.from(get(memo, "_value", "")).toString(decodeMethod);
```
piyalbasu, line 15: iff memo value was `null`, this was passing `null` (instead of "") to `Buffer.from`, which throws an error

PR #985, opened 2023-10-02
Show consumed resource estimation to the user and projected fees in Send Token flow

aristidesstaffieri:
Adds preflight info to token payments, includes resource cost & fee.
Ticket: https://stellarorg.atlassian.net/browse/WAL-744

Screenshot 2023-09-28 at 10 53 06 AM

Adds `transactionSimulation` to transactionSubmission store.
Moves payment tx sim to network settings.
Uses tx sim from store in tx details, displays resource info if available.


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -269,44 +269,19 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
 
-      const builder = await sorobanClient.newTxBuilder(
-        xlmToStroop(transactionFee).toFixed(),
-      );
-
-      const transaction = await transfer(assetAddress, params, memo, builder);
-
-      const preparedTransaction = await sorobanClient.server.prepareTransaction(
-        transaction,
+      const preparedTransaction = SorobanClient.assembleTransaction(
```
aristidesstaffieri, line 276: I think we can refactor this store so that the state is a union type where the tx stuff is no longer nullable in certain states(sim success/submit success) so that we don't have to cast these to non-nullable. I didn't think it'd be worth it in this PR but let me know if you think otherwise @piyalbasu

piyalbasu: Cool - yeah, I thin we can clean up the types in another PR 👍

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -269,44 +267,19 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
-
-      const params = [
-        new SorobanClient.Address(publicKey).toScVal(), // from
-        new SorobanClient.Address(destination).toScVal(), // to
-        new SorobanClient.XdrLargeInt("i128", parsedAmount.toNumber()).toI128(), // amount
-      ];
-
       if (!hasSorobanClient(sorobanClient)) {
         throw new Error("Soroban RPC not supported for this network");
       }
```
piyalbasu, line 272: do we still need this check? We're not actually using `sorobanClient` (lowercase) anymore since we've already built and simmed the tx earlier

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: we do end up using it in the following action [for refreshing balances](https://github.com/stellar/freighter/blob/61277e9d7d582ca2cc41daa6bc828ca4557207e3/extension/src/popup/ducks/transactionSubmission.ts#L147C13-L147C26). It's kind of annoying to have to pass in the client just for the refresh so maybe it would be nice to refactor those actions to build their own clients or something?

piyalbasu: Oh gotcha. Yeah, maybe we refactor those to use their own clients later on 👍

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -541,6 +514,27 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
+          </div>
+        )}
+        {transactionSimulation.response && (
+          <div className="TransactionDetails__row">
+            <div>{t("Minimum resource fee")} </div>
+            <div className="TransactionDetails__row__right">
+              {transactionSimulation.response.minResourceFee} XLM
+            </div>
+          </div>
+        )}
```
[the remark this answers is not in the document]
aristidesstaffieri: yup 😅 fixed in 7171907199f8efc4b3d72e7c4b4cb7e4edf4879d

PR #1004, opened 2023-10-13
Rm stale tokens

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/02ec7a60-c21e-4bf0-8001-33d37b59a749


aristidesstaffieri:
On the screen where you manage your assets, the buttons say "Add Soroban Token" and "Add another asset", should that second one be something like "Add Classic Asset or Token"?

aristidesstaffieri (approved):

piyalbasu:
> On the screen where you manage your assets, the buttons say "Add Soroban Token" and "Add another asset", should that second one be something like "Add Classic Asset or Token"?

Good call - so that second button only allows you to add a Classic asset, so it should probably say something like "Add Classic asset". I'll confirm with product if that's the nomenclature we want to use

PR #1015, opened 2023-10-23
rm serializable check to reduce noise

piyalbasu:


extension/src/helpers/stellar.ts
```diff
@@ -13,11 +12,6 @@ import {
 import { TransactionInfo } from "types/transactions";
 import { parsedSearchParam, getUrlHostname } from "./urls";
 
-// .isBigNumber() not catching correctly, so checking .isBigNumber
-// property as well
-export const isSerializable = (value: any) =>
```
piyalbasu, line 18: I actually tried to follow this pattern here, but it was getting pretty messy because of the sheer amount of non-serializable values inside of a Transaction:

It started to look like this, and I didn't even finish getting all the edge cases:
```
export const isSerializable = (value: any) => {
  // .isBigNumber() not catching correctly, so checking .isBigNumber
  // property as well
  if (value?.isBigNumber || BigNumber.isBigNumber(value) || isPlain(value)) {
    return true;
  }

  if (
    value._attributes ||
    value._armType ||
    typeof value === "function" ||
    value.name ||
    value._maxLength ||
    Object.prototype.toString.call(value) === "[object Uint8Array]" ||
    value._childType
  ) {
    return true;
  }
  return false;
};
```

I started to think it wasn't even worth it. But curious to hear other toughts!

aristidesstaffieri: LGTM. Is it worth it to make the change in this PR to not store the serializable value? Like store it(the tx object) in redux as xdr and fromXDR when needed out of redux?

aristidesstaffieri (approved):

Issue #903, opened 2023-07-19
Sync authentication state between different accounts


piyalbasu:
When switching between accounts in Freighter, users are prompted to re-enter their password. Is there a way to keep a user "logged in"?

piyalbasu:
Dupe of #1495

PR #1026, opened 2023-11-07
Feature/account migration pt 1

piyalbasu:
Most of this PR is populating this "Review accounts to migrate" page

Screenshot 2023-11-06 at 9 43 10 PM


@shared/api/internal.ts
```diff
@@ -310,6 +326,47 @@ export const confirmPassword = async (
...
+    account = await server.loadAccount(publicKey);
+    signerArr = await server.accounts().forSigner(publicKey).call();
+  } catch (e) {
+    console.error(e);
+  }
+
+  return {
+    account,
+    isSigner: signerArr.records.length > 1,
+  };
```
piyalbasu, line 353: we need some slightly different data than what we store in allAccounts, so we unfortunately need to refetch from Horizon

@shared/api/internal.ts
```diff
@@ -337,11 +394,13 @@ export const getAccountBalances = async ({
...
     subentryCount = resp.subentryCount;
 
+    // eslint-disable-next-line no-plusplus
     for (let i = 0; i < Object.keys(resp.balances).length; i++) {
       const k = Object.keys(resp.balances)[i];
       const v: any = resp.balances[k];
       if (v.liquidity_pool_id) {
         const server = stellarSdkServer(networkUrl);
+        // eslint-disable-next-line no-await-in-loop
         const lp = await server
```
piyalbasu, line 404: just adding these in bc they were throwing linting errors

extension/src/background/ducks/session.ts
```diff
@@ -57,6 +60,25 @@ export const sessionSlice = createSlice({
...
       };
     },
+    setPassword: (state, action: { payload: AppData }) => {
+      const { password } = action.payload;
+
+      return {
+        ...state,
+        password,
+      };
+    },
```
piyalbasu, line 70: in order to prevent the user from having to re-enter their password, we can store the password in the background in memory for use later in the flow. 

This actually makes this [issue](https://github.com/stellar/freighter/issues/903) possible, so I thought it made sense to introduce this feature here

extension/src/background/ducks/session.ts
```diff
@@ -115,10 +139,18 @@ export const hasPrivateKeySelector = createSelector(
...
-  (session) => session.privateKey,
+  (session) => session.privateKey || "",
+);
+export const passwordSelector = createSelector(
+  sessionSelector,
+  (session) => session.password,
+);
+export const migratedMnemonicPhraseSelector = createSelector(
+  sessionSelector,
+  (session) => session.migratedMnemonicPhrase,
```
piyalbasu, line 155: we need to hold the old mnemonic and the new mnemonic for a little while, so storing this in a different key

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1264,6 +1268,39 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     return { tokenIdList: updatedTokenIdList };
   };
 
+  const getMigratableAccounts = async () => {
+    const keyIdList = await getKeyIdList();
+
+    const mnemonicPhrase = mnemonicPhraseSelector(sessionStore.getState());
+    const allAccounts = allAccountsSelector(sessionStore.getState());
+    const wallet = fromMnemonic(mnemonicPhrase);
```
piyalbasu, line 1276: unfortunately, when users add new accounts, we don't properly differentiate between accounts that were generated from the mnemonic vs accounts that were imported via secret key. Because of this, we have to do this roundabout process of generating a bunch of accounts from the mnemonic and seeing if the accounts the users has stored belong to the mnemonic. Yuck.

I think this is a band-aid for right now and in the future we should properly tag mnemonic-derived accounts as they are added so we can easily identify them without having to do this

extension/src/popup/components/accountMigration/MnemonicPhrase/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    fetchMnemonicPhrase();
+  }, []);
+
+  return isConfirmed ? (
+    <ConfirmMnemonicPhrase words={shuffle(mnemonicPhrase.split(" "))} />
+  ) : (
+    <DisplayMnemonicPhrase
+      mnemonicPhrase={mnemonicPhrase}
+      setIsConfirmed={setIsConfirmed}
+    />
```
piyalbasu, line 30: reusing the existing "confirm your mnemonic" flow from onboarding

extension/src/popup/components/accountMigration/ReviewMigration/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+  <div className="ReviewMigration__account">
+    <div className="ReviewMigration__account__identicon-wrapper">
+      <IdenticonImg publicKey={account.publicKey} />
+    </div>
+    <div className="ReviewMigration__account__name">{account.name}</div>
+    <div className="ReviewMigration__account__public-key">
+      ({truncatedPublicKey(account.publicKey)})
+    </div>
+  </div>
+);
```
piyalbasu, line 52: These subcomponents will likely be broken out into their own files as they'll be needed again later in the flow

extension/src/popup/components/accountMigration/ReviewMigration/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+                    fieldSize="md"
+                    autoComplete="off"
+                    id="isMergeSelected-input"
+                    label={
+                      <div>
+                        <span className="ReviewMigration__highlight">
+                          {t("Optional")}:{" "}
+                        </span>
+                        {t(
+                          "Merge accounts after migrating (you funding lumens used to fund the current accounts will be sent to the new ones - you lose access to the current accounts.)",
```
aristidesstaffieri, line 224: typo? `you funding lumens` => `your funding lumens`

extension/src/background/messageListener/popupMessageListener.ts
aristidesstaffieri: ah yeah that's too bad, makes sense 👍

aristidesstaffieri (approved):

extension/src/popup/components/accountMigration/ReviewMigration/index.tsx
piyalbasu: ah good catch!

PR #1033, opened 2023-11-15
Update layout

quietbits:


extension/src/popup/basics/layout/Box/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface BoxProps {
+  display?: "grid" | "flex";
+  children: React.ReactElement | React.ReactElement[];
+  gridCellWidth?: string;
+  gapHorizontal?: string;
+  gapVertical?: string;
+  isFlexRow?: boolean;
+}
+
+export const Box: React.FC<BoxProps> = ({
```
quietbits, line 15: This component attempts to create a layout system that we can use to build the most common layouts. It's very limited right now, but we can keep growing it if the idea seems OK.

extension/src/popup/basics/layout/View/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+  Footer: React.FC<ViewFooterProps>;
+  Inset: React.FC<ViewInsetProps>;
+}
+
+interface ViewLayoutProps {
+  children: React.ReactNode;
+  isAppLayout?: boolean;
+}
+
+export const View: React.FC<ViewLayoutProps> & ViewComponent = ({
```
quietbits, line 240: `View` is the component we should use on every screen/page that is its own view. It shouldn't matter if it's a main view or subview. Only overlays and modals are not part of this.

extension/src/popup/components/account/AccountHeader/index.tsx
```diff
@@ -24,14 +25,14 @@ import IconCube from "popup/assets/icon-cube.svg";
...
-  accountDropDownRef: React.RefObject<HTMLDivElement>;
+  // accountDropDownRef: React.RefObject<HTMLDivElement>;
   allAccounts: Array<Account>;
   currentAccountName: string;
   publicKey: string;
 }
 
 export const AccountHeader = ({
-  accountDropDownRef,
+  // accountDropDownRef,
```
quietbits, line 35: What is this `ref` used for? I noticed we use refs in a few other places.

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -20,7 +20,7 @@ import { useAssetDomain } from "popup/helpers/useAssetDomain";
 import { navigateTo } from "popup/helpers/navigate";
 import { formatTokenAmount } from "popup/helpers/soroban";
 import { getAssetFromCanonical } from "helpers/stellar";
-import { SimpleBarWrapper } from "popup/basics/SimpleBarWrapper";
+// import { SimpleBarWrapper } from "popup/basics/SimpleBarWrapper";
```
quietbits, line 23: We don't need to keep the `SimpleBarWrapper` to use JS for content scrolling, right? I kept it in most places where it didn't affect the new layout, but it seems unnecessary for the new responsive layout. Removing all components and styles associated with it is a big update (many files will be affected), so we might handle that in a separate PR.

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -263,8 +267,10 @@ export const AssetDetail = ({
               {t("No transactions to show")}
             </div>
           )}
-        </SimpleBarWrapper>
-      </div>
+          {/* </SimpleBarWrapper> */}
+        </div>
+      </View.Content>
+      {/* TODO: fix the slideup modal */}
```
quietbits, line 273: We should create a "slide over" modal component. I haven't updated it yet, but it should look OK on smaller screen sizes. The width will stretch across the whole screen, which is not great for larger screen sizes.

extension/src/popup/components/accountHistory/HistoryList/styles.scss
```diff
@@ -1,10 +1,10 @@
 .HistoryList {
-  max-height: 25rem;
-  padding-right: var(--popup--side-padding);
-  width: calc(var(--popup--width) - var(--popup--side-padding));
+  // max-height: 25rem;
```
quietbits, line 2: I'll do a cleanup for comments once we're sure the new layout functions well. I wanted to keep some code for reference in case we need to fix some things.

extension/src/popup/styles/global.scss
```diff
@@ -105,15 +103,26 @@ a {
   }
 }
 
-* {
-  scrollbar-color: rgba(var(--color-shadow-rgb), 0.25) var(--color-gray-00);
+// TODO: do we have to do this? seems hacky
```
quietbits, line 106: Do we want to modify browser scrollbar?

extension/src/popup/components/account/AccountHeader/index.tsx
aristidesstaffieri: hmm I don't see this used anywhere, maybe leftover from a refactor that removed the need for it?

piyalbasu: Hm, we usually use refs to dynamically set modal/slide up heights and positioning. I believe at one point this ref was used to show the dropdown of accounts on the main screen. But looking at this code, it looks like we refactored this at some point and this is not needed anymore. I'll QA this to make sure we can safely remove

extension/src/popup/components/account/AssetDetail/index.tsx
piyalbasu: Is your question if we can get rid of the wrapper and just directly use `import SimpleBar from "simplebar-react";`? We could probably do that. 

But we would still need to use simplebar-react to have a JS scrollable div (especially when we're in a smaller screen/view). For context, we used native browser div scrolling for a while, but it was pretty ugly. This created a nicer UI

aristidesstaffieri:
A few general notes - 

It looks like the general loading screen does not maintain the sticky-ness of the footer, it's hard to catch on fast loads but more noticeable when you do things like change networks.

Account balance items have some layout shift when the logo takes longer to load than the rest of the line item. I'm not sure if this existed before and maybe if more noticeable now? I think either the whole row should load at once or we should retain the screen real estate for the logo while loading to avoid the shift.

It appears there is a "flash of screen" when you click on an item in the account balance screen for more details. Looking at the changes, I would guess this existed before but is for some reason easier to notice now.

aristidesstaffieri (approved):
Approving since Iveta is working on the footer during loading and that was all I found in QA.

aristidesstaffieri:
Layout shift looks fixed for the loading screen, lgtm!

PR #1035, opened 2023-11-17
Account Balance Mercury V1

aristidesstaffieri:
This is the first pass at using our backend to get account balances, the backend integrates with Mercury and falls back to using the current RPCs.

Prerequisites:
In order to test this you need a backend running. At this time, the dev deployment infra is still being worked on so this requires running a local instance of https://github.com/stellar/freighter-backend.
Reach out to Aristides for env vars.

**High Level Changes**

- Updates `getAccountBalance` to use the backend endpoint.
- Removes `ducks/soroban` and moves `removeTokenId` to the `ducks/transactionSubmission`.
- Moves away from the wallet-sdk provided `Types.BalanceMap` to our own extension of that, which adds a `TokenBalance` branch in balances, to aid our effort in combining the balance keys.
- Combines the `accountBalances` and `tokenBalances` into the `accountBalances` key in the store, in order to better support the new combined data source.
- Adds a new storage key `HAS_ACCOUNT_SUBSCRIPTION`(and migration to add it) which is toggle to let us know when to subscribe a new account.
- Adds subscribe account internal action, which uses the new storage key to optionally setup a new account sub when a user logs in.
- Removes refresh balances actions.
- Updates views to only use accountBalances.
- Updates tests to only use accountBalances.

**Next Steps**
This makes the migration for the data sources but does not refactor away some of the old code that used to handle the split paths. There are likely some more simplifications that can be gained after these changes.


@shared/api/internal.ts
```diff
@@ -310,6 +315,92 @@ export const confirmPassword = async (
...
+  issuer?: {
+    key: string;
+  };
+  decimals: number;
+  name: string;
+  symbol: string;
+  valueXdr: string;
+}[];
+
+export const getAccountIndexerBalances = async (
```
aristidesstaffieri, line 329: This takes the data from the indexer and formats it to match our current use from the RPCs, following the BalanceMap pattern.
I thought it would be best to do this here instead of in the backend because we only do this in order to be able to achieve the fallback strategy in a cleaner way and hopefully one day this can be removed. 
Open to suggestions on alternatives.

@shared/api/types.ts
```diff
@@ -146,7 +147,16 @@ export interface AssetDomains {
   [code: string]: string;
 }
 
-export type Balances = Types.BalanceMap | null;
+export interface TokenBalance extends AssetBalance {
```
aristidesstaffieri, line 150: This new type extends the existing pattern in order to combine the data keys in our store for both classic and soroban assets.

@shared/constants/mercury.ts
```diff
@@ -0,0 +1,3 @@
+export const INDEXER_URLS: { [index: string]: string } = {
+  TESTNET: "http://localhost:3002/api/v1",
```
aristidesstaffieri, line 2: We don't have env vars yet and this isn't really a secret but maybe this should go in an env var?

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -261,20 +245,128 @@ const storeBalanceMetricData = (publicKey: string, accountFunded: boolean) => {
...
-    return res;
+    await internalRemoveTokenId({ contractId, network });
   } catch (e) {
-    return thunkApi.rejectWithValue({ errorMessage: e });
+    console.error(e);
+    thunkApi.rejectWithValue({ errorMessage: e as string });
   }
 });
 
+export const getAccountBalancesWithFallback = createAsyncThunk<
```
aristidesstaffieri, line 266: There's a lot of logic here but it's the one place where we house the complexity of the fallback strategy, should this be broken up/split up in some way?

@shared/api/internal.ts
```diff
@@ -310,6 +315,92 @@ export const confirmPassword = async (
...
+        total: scValToNative(totalScVal),
+      };
+    });
+
+    const balances = formattedBalances.reduce((prev, curr) => {
+      if (curr.symbol === "XLM") {
+        prev["native"] = {
+          token: { type: "native", code: "XLM" },
+          total: new BigNumber(curr.total),
+          available: new BigNumber(curr.total), // TODO: how to get available for xlm?
```
aristidesstaffieri, line 360: I believe this should be `currentBalance - ((2 + accountBalances.subentryCount) * BASE_RESERVE)`?

@shared/constants/mercury.ts
```diff
@@ -0,0 +1 @@
+export const INDEXER_URL = "http://localhost:3002/api/v1";
```
aristidesstaffieri, line 1: Should this come from an env var?

aristidesstaffieri:
One thing I'm worried about is missing a situation where we need to subscribe a user, and incorrectly showing no assets/history. 
We do the subscription check & sub in the logIn action, with the assumption that this happens in all cases that we need to subscribe or check for a user sub - new user account, switch accounts, import account. 
Are there any other times where we would need to subscribe a user?

@shared/constants/mercury.ts
aristidesstaffieri: we do now have the dev instance stood up, and it is currently at `https://freighter-backend-fc55eaf.previews.kube001.services.stellar-ops.com/api/v1`

piyalbasu:
> One thing I'm worried about is missing a situation where we need to subscribe a user, and incorrectly showing no assets/history. We do the subscription check & sub in the logIn/setPublicKey actions, with the assumption that this happens in all cases that we need to subscribe or check for a user sub - new user account, switch accounts, import account. Are there any other times where we would need to subscribe a user?

I think this should cover it! Any time a user changes what public key they're working with, it goes through either of these 2 redux actions.

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -605,6 +609,7 @@ const transactionSubmissionSlice = createSlice({
     });
     builder.addCase(getAccountBalances.fulfilled, (state, action) => {
       state.accountBalances = action.payload;
+      state.tokensWithNoBalance = []; // TODO
```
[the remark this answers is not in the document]
piyalbasu: Ah, yeah - so this was intended to cast a wider net. This was implemented to allow users to remove a contract ID that they no longer wanted in Freighter's UI. It could be because a) the contract expired or b) the user mistyped  a contract id and it can't be found. It would def be great to be able to delineate between expired contracts that existed and contracts that never existed


(example of how `tokensWithNoBalance` is used to show a bad contract id that a user may want to remove)
Screenshot 2023-12-04 at 5 56 30 PM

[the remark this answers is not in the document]
aristidesstaffieri: nice yeah makes sense. I think the move to the backend makes recreating the behavior exactly kind of tough for a few reasons -
We will now have access to balances even when the underlying contract has expired, due to indexing.
Mistyping should result in a few cases depending on how bad the mis-type is - if the string is a contract ID but does not have the same interface as a token contract, this will fail at index time which should mean that we can reject it when they enter it in(not true for existing ones exactly), or if the contract string does not parse to a contract ID, the API will 400 the request to index it.

So I think that more or less leaves us with handling only the case of truly having a 0 balance in order to suggest removing, do you agree? Also, to complicate things further, this ignores the fallback to rpc cases which would work as you have already written.
@piyalbasu

aristidesstaffieri:
> > > One thing I'm worried about is missing a situation where we need to subscribe a user, and incorrectly showing no assets/history. We do the subscription check & sub in the logIn/setPublicKey actions, with the assumption that this happens in all cases that we need to subscribe or check for a user sub - new user account, switch accounts, import account. Are there any other times where we would need to subscribe a user?
> > 
> > 
> > I think this should cover it! Any time a user changes what public key they're working with, it goes through either of these 2 redux actions.
> > EDIT: I just a found an instance that currently isn't covered. When you switch networks (testnet to mainnet, for ex). You would need to check subscription status then, even though the `publicKey` hasn't changed
> 
> oh yeah good point, and actually this requires that I finish the work to pass network into the calls in the backend. I'll work on that today and should have a patch for this soon.

This should be fixed in ea35feede4ba0d28dfc360f490f036d4506785e4, and I did need to tweak the routes so to test that out you just need to pull the latest commit in https://github.com/stellar/freighter-backend/pull/8/commits

@shared/api/internal.ts
```diff
@@ -35,6 +35,7 @@ import { sendMessageToBackground } from "./helpers/extensionMessaging";
 import { getIconUrlFromIssuer } from "./helpers/getIconUrlFromIssuer";
 import { getDomainFromIssuer } from "./helpers/getDomainFromIssuer";
 import { stellarSdkServer } from "./helpers/stellarSdkServer";
+import BigNumber from "bignumber.js";
```
piyalbasu, line 38: this should actually throw a linting error. I'll fix the husky pre-commit hook for this

@shared/api/internal.ts
```diff
@@ -310,65 +311,44 @@ export const confirmPassword = async (
...
-  networkDetails: NetworkDetails;
-}): Promise<AccountBalancesInterface> => {
-  const { networkUrl, networkPassphrase } = networkDetails;
-
-  let balances: any = null;
-  let isFunded = null;
-  let subentryCount = 0;
-
+export const getAccountIndexerBalances = async (
+  pubKey: string,
```
piyalbasu, line 315: super super NIT: we tend to refer to public keys as `publicKey` in this repo. Would you mind using that var name here to keep it consistent? I think it makes it easier when var naming conventions are the same throughout the app

@shared/constants/mercury.ts
piyalbasu: yeah, maybe we should add an `env.development` file

@shared/api/internal.ts
aristidesstaffieri: oh yup, will do!

aristidesstaffieri: fixed in 12547c76cb7701443423ea42cb16c0f825ec746c

@shared/api/internal.ts
```diff
@@ -310,65 +311,44 @@ export const confirmPassword = async (
...
+      balances: formattedBalances,
+    };
+  } catch (error) {
+    console.error(error);
     return {
-      balances,
-      isFunded: false,
-      subentryCount,
+      balances: {} as AccountBalancesInterface["balances"],
+      isFunded: null,
```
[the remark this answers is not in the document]
aristidesstaffieri: didn't see any reason not to do this, so fixed it in 12547c76cb7701443423ea42cb16c0f825ec746c

@shared/constants/mercury.ts
aristidesstaffieri: @piyalbasu I added an env var for this in 46a4341c09059ee193aa1dd9169dc6a37452f908.

Just noting(mainly for myself later!) that we will need to add this as a secret in github before deployment for the action to use it.

aristidesstaffieri: jus deployed an update to the backend, so the new remote instance is at https://freighter-backend-88d8d43.previews.kube001.services.stellar-ops.com/

extension/src/popup/ducks/transactionSubmission.ts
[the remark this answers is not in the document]
aristidesstaffieri: updated to handle the missing balance case in 4de74c01ea630a69c161d4d90a4a77a96ae4f8d3. I tried to go for "ban-aid" solution which maybe isn't the cleanest but will hopefully be easiest to change once we move into handling this more explicitly with different states for expired entries, etc.

aristidesstaffieri:
newest dev instance is http://freighter-backend-a880c3e.previews.kube001.services.stellar-ops.com/

PR #1038, opened 2023-11-21
Feature/account migration submit tx

piyalbasu:


@shared/api/types.ts
```diff
@@ -18,6 +18,8 @@ export interface UserInfo {
   publicKey: string;
 }
 
+export type MigratableAccount = Account & { keyIdIndex: number };
```
piyalbasu, line 21: when we migrate an account, we need to know where the in the keyIdList this specific account falls so we can seamlessly replace the old account with the new one

extension/src/background/helpers/migration.ts
```diff
@@ -0,0 +1,108 @@
...
+    try {
+      builtChangeTrustTx.sign(recipientSourceKeys);
+    } catch (e) {
+      console.error(e);
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await server.submitTransaction(builtChangeTrustTx);
+
+    if (asset) {
```
aristidesstaffieri, line 75: small nit, is this check needed? Seems like the check on like 54/ instantiating an `Asset` would take care of this.

extension/src/background/helpers/migration.ts
```diff
@@ -0,0 +1,108 @@
...
+    const builtChangeTrustTx = changeTrustTx.setTimeout(180).build();
+
+    try {
+      builtChangeTrustTx.sign(recipientSourceKeys);
+    } catch (e) {
+      console.error(e);
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    await server.submitTransaction(builtChangeTrustTx);
```
aristidesstaffieri, line 73: Should there be a check for a failed change trust transaction before sending a balance?

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1301,6 +1398,192 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
+        privateKey: newWallet.getSecret(keyIdIndex),
+      };
+
+      // eslint-disable-next-line no-await-in-loop
+      const transaction = await new TransactionBuilder(sourceAccount, {
+        fee,
+        networkPassphrase: Networks.TESTNET,
+      });
+
+      if (!xlmBalance) {
```
aristidesstaffieri, line 1462: Should this condition trigger a user facing error so users know why a balance cannot be migrated?

extension/src/background/helpers/migration.ts
piyalbasu: Yup - good call. In the next PR I streamline this whole thing, I'll make this fix there

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: Yup, I'll work this into the UI

PR #1039, opened 2023-11-27
simplify trustline transfer and account for removing trustlines before merge

piyalbasu:


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1526,6 +1535,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
           destination: newKeyPair.publicKey,
           error: e,
         });
+        return { error: JSON.stringify(migrationErrors) };
```
piyalbasu, line 1538: catch any errors from the migrating the trustline and send them back to the UI

PR #1045, opened 2023-12-05
SignTransaction memo updated + added tests

quietbits:


extension/src/popup/helpers/parseTransaction.ts
```diff
@@ -5,18 +5,35 @@ import { MEMO_TYPES } from "popup/constants/memoTypes";
 
 import { ErrorMessage } from "@shared/api/types";
 
-export const decodeMemo = (memo: {}) => {
-  const memoType = get(memo, "_switch.name", "");
+type Memo = {
```
aristidesstaffieri, line 8: Should we use the type from stellar base here?
https://github.com/stellar/js-stellar-base/blob/cf370f4e627e2b0e296ceceaebd0164dede396bd/types/index.d.ts#L212

extension/src/popup/components/signTransaction/TransactionInfo/index.tsx
```diff
@@ -3,18 +3,26 @@ import { Icon, IconButton } from "@stellar/design-system";
...
+import { MEMO_TYPES } from "popup/constants/memoTypes";
 
 import "./styles.scss";
 
 const MemoDisplay = ({
   memo,
   isMemoRequired,
 }: {
-  memo: string;
+  memo: { value: string; type: MEMO_TYPES };
```
aristidesstaffieri, line 14: Should this type be a `Memo` from stellar-base?
https://github.com/stellar/js-stellar-base/blob/cf370f4e627e2b0e296ceceaebd0164dede396bd/types/index.d.ts#L212

extension/src/popup/constants/memoTypes.ts
```diff
@@ -2,4 +2,5 @@ export enum MEMO_TYPES {
   MEMO_ID = "memoId",
```
aristidesstaffieri, line 2: I think this type can be replaced with the stellar-base equivalent
https://github.com/stellar/js-stellar-base/blob/cf370f4e627e2b0e296ceceaebd0164dede396bd/types/index.d.ts#L204

extension/src/popup/helpers/__tests__/decodeHash.test.js
```diff
@@ -4,34 +4,38 @@ import { decodeMemo } from "../parseTransaction";
 describe("decodeMemo", () => {
   it("should return a memo id", () => {
     const MEMO = {
```
aristidesstaffieri, line 6: Should we use the actual Memo class here instead of stubbing it? I think typically when we see the properties that are prefixed with an underscore, like `_type` or `_value`, the convention is to use the class getter to access it as they intend the field to be private to the class.
https://github.com/stellar/js-stellar-base/blob/cf370f4e627e2b0e296ceceaebd0164dede396bd/src/memo.js#L64

extension/src/popup/components/signTransaction/TransactionInfo/index.tsx
piyalbasu: Good call - I think we should leverage types from the SDK wherever we can

piyalbasu (approved):

extension/src/popup/components/signTransaction/TransactionInfo/index.tsx
quietbits: Here we need to keep `memo: { value: string; type: MemoType };` because this is for decoded memo.

aristidesstaffieri (approved):

PR #1046, opened 2023-12-05
switch on Mainnet for migration and show a streamlined password prompt

piyalbasu:


extension/src/popup/components/accountMigration/ReviewMigration/index.tsx
```diff
@@ -83,7 +83,7 @@ const isReadyToMigrate = ({
         recommendedFee,
         hasTrustlineBalances: Boolean(trustlineBalancesLength),
         isMergeSelected,
-      }) < new BigNumber(xlmBalance),
+      }) < new BigNumber(xlmBalance).minus(minBalance),
```
piyalbasu, line 86: also fix a minor bug where the UI was allowing you to try a migration even if you don't have enough funds

quietbits (approved):

PR #1047, opened 2023-12-06
Swap payment tests

quietbits:


extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -332,15 +332,12 @@ export const getBestPath = createAsyncThunk<
   "getBestPath",
   async ({ amount, sourceAsset, destAsset, networkDetails }, thunkApi) => {
     try {
-      const server = new Horizon.Server(networkDetails.networkUrl);
-      const builder = server.strictSendPaths(
-        getAssetFromCanonical(sourceAsset) as Asset,
+      return await horizonGetBestPath({
```
quietbits, line 335: I had to extract this function to be able to mock it for tests.

extension/src/popup/views/__tests__/Swap.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+  it("renders swap view initial state", () => {
+    expect(screen.getByTestId("AppHeaderPageTitle")).toHaveTextContent(
+      "Swap XLM",
+    );
+    expect(screen.getByTestId("AppHeaderPageSubtitle")).toHaveTextContent(
+      "220.49999 XLM available",
+    );
+    expect(screen.getByTestId("send-amount-amount-input")).toHaveValue("0");
+
+    const assetSelects = screen.getAllByTestId("AssetSelect");
```
quietbits, line 156: I noticed we're using two conventions for test IDs: camelcase and kebab case. I used camelcase, but I can change it to kebab if we prefer that.

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -290,8 +292,13 @@ export const SendAmount = ({
   ]);
 
   // for swaps we're loading and choosing the default destinationAsset here
+  // also, need to check if both source and destination are native
```
aristidesstaffieri, line 295: This is kind of curious, why would we allow a swap from native to native?

aristidesstaffieri (approved):
lgtm! Left a question about the tweak to the check in swap but it's probably inconsequential.

extension/src/popup/components/sendPayment/SendAmount/index.tsx
quietbits: The tests would sometimes fail without this check because both the source and destination would be set to XML. I noticed it in the app as well when refreshing the page when on that view. I'm not sure if this is the best fix, but it seemed the least risky.

aristidesstaffieri: Hmm yeah fair enough, I think it's probably an existing bug if we can setup a swap where both sides are the same asset but the check doesn't make it worse so I'll go hit the approve button, thanks!

extension/src/popup/views/__tests__/Swap.test.tsx
piyalbasu: Oh yeah, this isn't consistent in the repo 😬 I think camel is fine!

piyalbasu (approved):

PR #1056, opened 2023-12-11
fix trustline error view and scroll long sc values

piyalbasu:


extension/src/popup/components/manageAssets/TrustlineError/index.tsx
```diff
@@ -143,15 +144,19 @@ export const TrustlineError = ({
...
-    <div className="TrustlineError">
-      <div className="TrustlineError__body">
-        <RenderedError
-          errorState={errorState}
-          assetBalance={assetBalance}
-          resultCodes={JSON.stringify(getResultCodes(error))}
-        />
-      </div>
-      <div className="TrustlineError__button">
+    <View>
```
piyalbasu, line 147: without this - the extension had no width and was "collapsing" into a long skinny screen

extension/src/popup/components/signTransaction/Operations/styles.scss
```diff
@@ -42,6 +42,7 @@
     font-size: 0.875rem;
     height: 7rem;
     margin: 0.5rem 0;
+    overflow: scroll;
```
piyalbasu, line 45: some sc values are really long, so allowing this div to scroll to allow users to see the value

extension/src/popup/components/manageAssets/TrustlineError/index.tsx
quietbits: Thanks for fixing this! I noticed it when writing tests for trustlines.

quietbits (approved):

PR #1058, opened 2023-12-12
Manage trustline tests

quietbits:


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -205,64 +198,10 @@ export const ManageAssetRows = ({
       dispatch(resetSubmission());
       navigateTo(ROUTES.account);
     }
-  }, [submitStatus, assetSubmitting, dispatch]);
```
quietbits, line 208: I removed `assetSubmitting` because it's not used. We can put it back if it needs to trigger something.

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -479,11 +485,22 @@ export const NewAssetWarning = ({
 
   const { isRevocable, isNewAsset, isInvalidDomain } = newAssetFlags;
 
+  useEffect(
+    () => () => {
```
aristidesstaffieri, line 489: Is this closure intentional?

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
aristidesstaffieri: oh weird, I'm surprised that this didn't trigger a warning before 🤔

aristidesstaffieri (approved):
Thanks for cleaning up these helpers, much better! 👏

extension/src/popup/components/WarningMessages/index.tsx
quietbits: Yes, `eslint` prefers it this way over
```typescript
useEffect(() => {
    return () => {
      setIsSubmitting(false);
    };
}, []);
```

piyalbasu (approved):

PR #1067, opened 2023-12-19
fix some broken tests and disable some flaky ones

piyalbasu:


extension/src/popup/views/__tests__/ManageAssets.test.tsx
```diff
@@ -174,16 +174,19 @@ jest.mock("popup/constants/history", () => ({
...
+        loadAccount() {
+          return {
+            sequenceNumber: () => 1,
+            accountId: () => publicKey,
+            incrementSequenceNumber: () => {},
+          };
+        }
+      },
     },
+    SorobanRpc: original.SorobanRpc,
```
piyalbasu, line 189: fixing the structure on these SDK methods.

Also, there's a weird issue where `...original` wasn't working because there's something wrong with the Getter for `ContractSpec` 🤔 I'll investigate further, but it's probably best to disable any SDK methods we're not using to prevent unexpected calls to Horizon (more on this later)

extension/src/popup/views/__tests__/SendTokenPayment.test.tsx
```diff
@@ -169,7 +177,7 @@ describe("SendTokenPayment", () => {
     </Wrapper>,
   );
 
-  it("can send a payment using a Soroban token", async () => {
+  it.skip("can send a payment using a Soroban token", async () => {
```
piyalbasu, line 180: looks like this test is dependent on a real testnet account and a real contract ID. We should probably mock these - otherwise we'll run into issues with this test every time there's a testnet reset, like today

extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -359,7 +363,7 @@ describe("SignTransactions", () => {
     );
   });
 
-  it("memo: render memo return", async () => {
+  xit("memo: render memo return", async () => {
```
piyalbasu, line 366: I'm still confused as to why these tests are failing - the test runner is saying it's because Horizon is returning a 405 (too many requests), but this view doesn't make any Horizon requests. I only get this error when I run all the tests together. Running these tests individually works as expected.

My theory is that some other tests are actually hitting Horizon a bunch, prompting the 405, and the error is surfacing here?

quietbits (approved):

PR #1068, opened 2023-12-19
Fix more flaky tests

quietbits:


extension/src/popup/views/__tests__/Swap.test.tsx
```diff
@@ -108,6 +108,28 @@ jest.mock("popup/helpers/horizonGetBestPath", () => ({
 
 jest.mock("lodash/debounce", () => jest.fn((fn) => fn));
 
+jest.mock("stellar-sdk", () => {
```
quietbits, line 111: This might have been the culprit for flaky tests. I missed that it was called somewhere down the line. :see_no_evil:

piyalbasu: 🎉 🎉 🎉

piyalbasu (approved):

PR #1070, opened 2023-12-22
add 24 word mnemonic phrase to recover-account

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/9b8208e6-e33c-49ad-92e1-ba908b1256e4


extension/src/popup/components/AutoSave/styles.scss
```diff
@@ -6,7 +6,7 @@
     position: fixed;
     top: 0;
     width: 100%;
-    transition: opacity var(--anim-transition-default);
+    transition: opacity var(--dropdown-animation);
```
piyalbasu, line 9: apparently this var doesn't exist in the repo anymore, so replacing it with this other animation timing var

config/jest/resolver.js
```diff
@@ -0,0 +1,18 @@
+module.exports = (path, options) =>
+  options.defaultResolver(path, {
+    ...options,
+    // Use packageFilter to process parsed `package.json` before the resolution (see https://www.npmjs.com/package/resolve#resolveid-opts-cb)
+    packageFilter: (pkg) => {
+      if (pkg.name === "@stellar/design-system") {
```
piyalbasu, line 6: interesting use-case here:

Webopack/Rollup default to using the file listed under `module` in package.json, but Jest uses `main`. I set this up because I'd like Jest to use the same file path that Webpack will be using

package.json
```diff
@@ -12,9 +12,6 @@
     "extension",
     "@stellar/freighter-api"
   ],
-  "resolutions": {
-    "**/terser-webpack-plugin": "^1.4.3"
```
piyalbasu, line 16: Webpack 5 comes with this included. This was causing an issue as this old version of Terser contained an outdated JS spec

extension/e2e-tests/onboarding.test.ts
```diff
@@ -89,6 +89,57 @@ test("Import wallet", async ({ page }) => {
...
+test("Import 24 word wallet", async ({ page }) => {
+  await expect(page).toHaveScreenshot("welcome-page.png");
+  await page.getByText("Import Wallet").click();
+  await expect(
+    page.getByText("Import wallet from recovery phrase"),
+  ).toBeVisible();
+  await page.locator(".RecoverAccount__phrase-toggle > label").click();
+
+  // time out to accomodate animation
+  await page.waitForTimeout(150);
```
aristidesstaffieri, line 101: this probably works well enough but wanted to point out that there is also `page.waitForSelector('yourselector')` and `page.waitForLoadState('domcontentloaded')` which could give you less of a racy condition to wait for in the case of waiting for css animations.

aristidesstaffieri (approved):

extension/e2e-tests/onboarding.test.ts
piyalbasu: Ah, nice! That's much better. I was not psyched about doing this, but couldn't get it to work reliably otherwise

PR #1090, opened 2024-01-25
only goToReview on button click

piyalbasu:


extension/src/popup/components/sendPayment/SendSettings/index.tsx
```diff
@@ -152,7 +152,6 @@ export const SendSettings = ({
         initialValues={{ memo }}
         onSubmit={(values) => {
```
piyalbasu, line 153: we call `onSubmit` when we click the buttons to nav to the change transaction fee and change slippage views. The app was navigating you to Review before it could nav to the fee/slippage update views, so users were not able to update either of these values

By moving `goToReview` to the button down below, we make sure we don't navigate to Review til we hit the Continue btn at the bottom

aristidesstaffieri (approved):

PR #1095, opened 2024-02-01
Constrain token payment

aristidesstaffieri:
This makes SendSettings work more like ConfirmSettings.
Moves the token-simulation action from the view to a new redux slice.
Uses new failure screen for any simulation rejections.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendSettings/SettingsFail/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendSettings/SettingsFail/index.tsx
@@ -1,0 +1,56 @@
+import React, { useEffect } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { Button } from "@stellar/design-system";
+import { useTranslation } from "react-i18next";
+
+import { navigateTo } from "popup/helpers/navigate";
+import { ROUTES } from "popup/constants/routes";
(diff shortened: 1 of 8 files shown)


extension/src/popup/components/sendPayment/SendSettings/SettingsFail/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useDispatch, useSelector } from "react-redux";
+import { Button } from "@stellar/design-system";
+import { useTranslation } from "react-i18next";
+
+import { navigateTo } from "popup/helpers/navigate";
+import { ROUTES } from "popup/constants/routes";
+import { View } from "popup/basics/layout/View";
+import IconFail from "popup/assets/icon-fail.svg";
+
+import "./styles.scss";
```
piyalbasu, line 11: this will probably trigger an eslint error - should be the last import

piyalbasu (approved):

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendSettings/SettingsFail/index.tsx
aristidesstaffieri: Hmm, didn't for me but I'll move it anyways.

PR #1098, opened 2024-02-02
updates changelog for 5.14.0

aristidesstaffieri:
Updates changelog for 5.14.0


piyalbasu (approved):
Thanks for adding! It’d be good to add a ticket to the release that describes that this is the release where we start using the Indexer

PR #1115, opened 2024-02-14
Feature/token allowlist

piyalbasu:
Screenshot 2024-02-14 at 2 14 01 PM


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -279,78 +314,122 @@ export const ManageAssetRows = ({
...
+                          handleRowClick(
+                            { code, issuer, image, domain },
+                            isTrustlineActive,
+                          );
+                        }
+                      }}
+                      type="button"
+                      data-testid="ManageAssetRowButton"
+                    >
+                      {isTrustlineActive ? (
```
aristidesstaffieri, line 380: Should this still be `isTrustlineActive || isContract`?

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -666,6 +670,138 @@ export const NewAssetWarning = ({
...
+            <div className="UnverifiedTokenWarning__icon">
+              <SorobanTokenIcon code={code} noMargin />
+            </div>
+            <div className="UnverifiedTokenWarning__code">{code}</div>
+            <div className="UnverifiedTokenWarning__domain">{domain}</div>
+            <div className="UnverifiedTokenWarning__description">
+              <div className="UnverifiedTokenWarning__description__icon">
+                <Icon.VerifiedUser />
+              </div>
+              <div className="UnverifiedTokenWarning__description__text">
```
aristidesstaffieri, line 743: Would we show this even when adding a Soroban token? My thought here is that they don't have trustlines so maybe this inly belongs on classic assets or SACs?

aristidesstaffieri (approved):

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
piyalbasu: Ah, good callout. I actually refactored `isTrustlineActive` up above to include all assets in `balances` (classic or token) and `tokensWithNoBalance`, so this should cover all asset/tokens a user has associated with their account. (at least in my testing it does 😬 )

extension/src/popup/components/WarningMessages/index.tsx
piyalbasu: Yup, we would show this if a user is adding a contract ID that doesn't appear on this list: 
https://api.stellar.expert/explorer/testnet/asset-list/top50

So we just flag that a user is adding a contract ID that has not been vetted

Lmk if that answers your question!

aristidesstaffieri: Sounds good. I was mainly curious if the badge could be confused with the concept of having a trustline to a classic asset. As in this case we're "trusting" the asset by means of it being on the list which is different than having a trust line on the classic asset side.

PR #1122, opened 2024-02-16
Better UX for contract IDs

aristidesstaffieri:
What
Upgrades `KeyIdenticon` to allow for icons on both sides and copy mode with an icon indicator.
Updates `truncateString` to use a parameter for the truncation amount, allowing for more flexibility to show more of the truncated string.
Adds a copy text and icon indicator to all contract IDs.

Why
To address feedback that we should have more interaction with contract IDs, and that we should show as much of them as possible.

Examples

Screenshot 2024-03-22 at 2 42 56 PM
Screenshot 2024-03-22 at 2 43 11 PM

Screenshot 2024-03-22 at 2 43 20 PM

Notes
There is one case where we can see a contract ID on the screen but it's hard for us to key off of in order to add copying. When signing a transaction that has a struct/vec as a param and that contains a contractId. Like in the case below -
Screenshot 2024-03-22 at 2 46 02 PM

I think this can be improved by rendering structs and arrays in a more structured way and doing some comparison on their members, right now `scValByType` returns them as a string-ified json object.


extension/src/popup/components/WarningMessages/styles.scss
```diff
@@ -385,6 +385,28 @@
       line-height: 1.375rem;
       margin: 0;
     }
+
+    .CopyText {
```
aristidesstaffieri, line 389: You'll notice we had to add very similar styles in several places. This is done to handle the slight differences in layout needed from the CopyText SDS component. This resulted in a change in CopyText to allow for a headless and/or block mode which will allow us to consolidate this a bit.

extension/src/popup/components/accountHistory/AssetNetworkInfo/index.tsx
```diff
@@ -60,15 +61,20 @@ export const AssetNetworkInfo = ({
     if (networkIconUrl || assetType === "native") {
       return <img src={networkIconUrl || StellarLogo} alt="Network icon" />;
     }
-    return <div className="AssetNetworkInfo__network__icon" />;
+    return null;
```
aristidesstaffieri, line 64: I can't remember the intention behind this but it seemed weird, especially in the case where you have no home domain it was more or less a floating circle with no context.

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -386,7 +391,16 @@ export const KeyValueInvokeHostFnArgs = ({ args }: { args: xdr.ScVal[] }) => (
     <div className="OperationParameters">
       {args.map((arg) => (
         <div className="Parameter" key={arg.toXDR().toString()}>
-          {scValByType(arg)}
+          {arg.switch() === xdr.ScValType.scvAddress() ? (
```
aristidesstaffieri, line 394: It seemed more appropriate to handle this here instead of in `scValByType` so we can keep that generic enough to work outside of components.

extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -218,12 +218,15 @@ describe("SignTransactions", () => {
 
     expect(
       opDetails.includes(
-        `Parameters${args?.from.toString()}${args?.to.toString()}${args?.amount.toString()}`,
```
aristidesstaffieri, line 221: These strings look like this now because of the tooltip added by the CopyContext component, it renders it inline with your element and show/hides.

extension/src/popup/components/accountHistory/AssetNetworkInfo/index.tsx
piyalbasu: So this is used when we can't find an icon for an asset. Maybe this should only return `null` if there is no home domain? 
Screenshot 2024-03-22 at 5 37 32 PM

extension/src/popup/components/identicons/KeyIdenticon/index.tsx
```diff
@@ -15,41 +15,76 @@ interface IdenticonWrapperElProps {
 }
 
 interface KeyIdenticonProps extends IdenticonWrapperElProps {
+  allowCopy?: boolean;
```
piyalbasu, line 18: NIT: we should follow boolean naming convention with `isCopyAllowed` (or something similar)

extension/src/popup/components/identicons/KeyIdenticon/index.tsx
```diff
@@ -15,41 +15,76 @@ interface IdenticonWrapperElProps {
...
-      <span {...props}>{shortPublicKey}</span>
+      {iconSide === "left" && (
+        <div className="KeyIdenticon--icon" style={customStyle}>
+          <IdenticonImg publicKey={publicKey} />
+        </div>
+      )}
+      {allowCopy ? (
+        <CopyText textToCopy={publicKey}>
+          <span {...props} className="KeyIdenticon--key">
+            <div className="CopyContractId">
```
piyalbasu, line 70: should this just be a custom component `<CopyContractId value= {truncateString(publicKey, keyTruncationAmount)} />` to avoid repeating the jsx/styles?

piyalbasu: Or if there are too many variants, maybe you could create a mixin or something to reuse the relevant styles. My concern is there's a lot of nested `.CopyContractId`'s which could lead to naming collisions and messiness in the future

extension/src/popup/components/accountHistory/AssetNetworkInfo/index.tsx
aristidesstaffieri: Makes sense, added that check in 4b135e5b899b32ec2040d54ce0438b559fe3b135

extension/src/popup/components/identicons/KeyIdenticon/index.tsx
aristidesstaffieri: I made it more generic since there are already a few places where we need it for other string types, so now we have `<CopyValue />`. There are still a few places where we need to tweak the styles for it inline for certain layouts but hopefully this cleans it up a bit. Done in 4b135e5b899b32ec2040d54ce0438b559fe3b135

extension/src/popup/components/identicons/KeyIdenticon/index.tsx
aristidesstaffieri: Done in 4b135e5b899b32ec2040d54ce0438b559fe3b135

piyalbasu (approved):

PR #1126, opened 2024-02-20
splits logic for auth entry invocations to account for contract creation

aristidesstaffieri:
Fixes auth entry invocation tree renderer to account for contract creation

### DIFF
--- a/extension/src/popup/helpers/soroban.ts
+++ b/extension/src/popup/helpers/soroban.ts
@@ -327,18 +327,90 @@ export function pickTransfers(invocationTree: InvocationTree) {
 export function getInvocationDetails(
   invocation: xdr.SorobanAuthorizedInvocation,
 ) {
-  return [
+  const invocations = [
     getInvocationArgs(invocation),
     ...invocation.subInvocations().map(getInvocationArgs),
   ];
+  return invocations.filter(isInvocationArg);
 }
 
-function getInvocationArgs(invocation: xdr.SorobanAuthorizedInvocation) {
-  const _invocation = invocation.function().contractFn();
-  const contractId = StrKey.encodeContract(
-    _invocation.contractAddress().contractId(),
-  );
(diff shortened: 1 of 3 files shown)


extension/src/popup/helpers/soroban.ts
```diff
@@ -327,18 +327,90 @@ export function pickTransfers(invocationTree: InvocationTree) {
...
+  asset: string;
+}
+
+type InvocationArgs = FnArgsInvoke | FnArgsCreateWasm | FnArgsCreateSac;
+
+const isInvocationArg = (
+  invocation: InvocationArgs | undefined,
+): invocation is InvocationArgs => !!invocation;
+
+function getInvocationArgs(
```
aristidesstaffieri, line 362: This fn has a good amount of overlap with 'buildInvocationTree', there should be the opportunity to combine these in the future but I didn't do it today in order to keep the change set as small as possible before the vote.

extension/src/popup/views/ReviewAuth/index.tsx
```diff
@@ -195,16 +200,25 @@ const AuthDetail = ({
...
   const { t } = useTranslation();
   const details = getInvocationDetails(authEntry.rootInvocation());
+  const invocations = details.filter(
+    (detail) => detail.type === "invoke",
+  ) as FnArgsInvoke[];
+  const createWasms = details.filter(
+    (detail) => detail.type === "wasm",
+  ) as FnArgsCreateWasm[];
+  const createSacs = details.filter(
+    (detail) => detail.type === "sac",
```
piyalbasu, line 210: future improvement: might be nice to store these arg types in an enum for easier reuse

aristidesstaffieri: makes sense, I think we can refactor this to possibly just drop all of this in favor of some tweaks to `buildInvocationTree`

PR #1137, opened 2024-02-22
Improves API error handling & states

aristidesstaffieri:
Adds a global error boundary component that wraps the app high up in the dom tree and catches any unhandled exceptions, falling back to a generic error message and action.
Captures all API related errors in Sentry.
Adds an error state for account balance when the API hard errors, no longer defaults to `!isFunded` base state and no longer incorrectly shows your account as not funded when the API is down.

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -19,6 +19,7 @@ import {
   browserLocalStorage,
 } from "background/helpers/dataStorage";
 import { INDEXER_URL } from "@shared/constants/mercury";
+import { captureException } from "@sentry/browser";
 
 const localStore = dataStorageAccess(browserLocalStorage);
 
@@ -155,6 +156,9 @@ export const subscribeAccount = async (publicKey: string) => {
     await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, subsByKeyId);
   } catch (e) {
     console.error(e);
+    captureException(
+      `Failed to subscribe account with Mercury - ${JSON.stringify(e)}`,
+    );
     throw new Error("Error subscribing account");
   }
 
@@ -181,6 +185,9 @@ export const subscribeTokenBalance = async (
     await fetch(`${INDEXER_URL}/subscription/token-balance`, options);
   } catch (e) {
     console.error(e);
+    captureException(
+      `Failed to subscribe token balance - ${JSON.stringify(e)}`,
+    );
     throw new Error(`Error subscribing to token: ${contractId}`);
   }
 };
@@ -200,6 +207,9 @@ export const subscribeTokenHistory = async (
     await fetch(`${INDEXER_URL}/subscription/token`, options);
   } catch (e) {
     console.error(e);
+    captureException(
+      `Failed to subscribe token history - ${JSON.stringify(e)}`,
+    );
     throw new Error(`Error subscribing to token: ${contractId}`);
   }
 };
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -126,11 +126,6 @@ export const getNetworksList = async () => {
   return networksList;
 };
 
-export const getIsSorobanSupported = async () => {
-  const networkDetails = await getNetworkDetails();
-  return !!networkDetails.sorobanRpcUrl;
-};
-
 export const subscribeAccount = async (publicKey: string) => {
   // if pub key already has a subscription setup, skip this
   const keyId = await localStore.getItem(KEY_ID);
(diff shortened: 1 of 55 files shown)


extension/src/background/helpers/account.ts
```diff
@@ -155,6 +156,9 @@ export const subscribeAccount = async (publicKey: string) => {
     await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, subsByKeyId);
   } catch (e) {
     console.error(e);
+    captureException(
```
aristidesstaffieri, line 159: I'm not actually super sure about what we should do if we fail to subscribe an account. When/if this happens we will end up showing an empty account when in reality we could just be failing to index it. It might be worth thinking about a way to toggle Mercury off when this happens so that we can investigate/fix the missed subscription?

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -242,48 +242,61 @@ export const HistoryItem = ({
...
 
-              const _token = {
-                contractId: attrs.contractId,
-                total: isRecieving ? attrs.amount : 0,
-                decimals: tokenDetails.decimals,
-                name: tokenDetails.name,
-                symbol: tokenDetails.symbol,
-              };
+                setRowText(operationString);
+                setTxDetails((_state) => ({
```
aristidesstaffieri, line 251: This should have been falling back to a generic tx row in this edge case failure.

extension/src/popup/views/Account/index.tsx
```diff
@@ -206,15 +213,26 @@ export const Account = () => {
                   </div>
                 </div>
               </div>
-              {isFunded ? (
+              {hasError && (
```
aristidesstaffieri, line 216: I split these out to avoid showing an NotFundedMessage when we fail to fetch account details.

extension/src/background/helpers/account.ts
piyalbasu: This is an interesting thought and related to something I've been thinking about, too. If Mercury/Freighter BE is down (or fails for some reason), should we keep some basic Horizon functionality to fallback to in this extension so users don't think somehow their funds have disappeared when really it was an API error?

extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -54,13 +54,11 @@ export const shouldAccountDoesntExistWarning = (
 
 export const AccountDoesntExistWarning = () => {
   const { t } = useTranslation();
+  const notificationTitle = t("The destination account doesn’t exist");
```
piyalbasu, line 57: ah think you just need to merge master. Think there's a small diff issue here

extension/src/popup/components/manageAssets/AddToken/index.tsx
```diff
@@ -145,17 +146,26 @@ export const AddToken = () => {
...
+                issuer: contractId,
+                domain: "",
+                name: resJson.name,
+              },
+            ]);
+          }
         } catch (e) {
           setAssetRows([]);
+          captureException(
+            `Failed to fetch token details - ${JSON.stringify(e)}`,
```
piyalbasu, line 167: Maybe we should surface this error to users - let them know their assets aren't gone, we just couldn't fetch them

piyalbasu: similar to what you're doing below with Account Balances

piyalbasu (approved):

extension/src/background/helpers/account.ts
aristidesstaffieri: Well so if Mercury fails in the lifecycle where we fetch their data like account details or history, then the backend falls back to using Horizon or rpc. If the API hard errors, then we show an error screen, but in the case that we fail to subscribe but the Mercury still responds to read queries then we don't have a good story since they would see the empty account in that case.
I think this edge case is complex though, we would likely need to take action when a sub fails and either use Horizon/rpc for the pub key that failed or use it for everything while we investigate the sub failure and restore it.

extension/src/popup/components/sendPayment/SendTo/index.tsx
aristidesstaffieri: Hmm I did merge to master 🤔 let me see why this diff still exists.

extension/src/popup/components/manageAssets/AddToken/index.tsx
aristidesstaffieri: hmm yeah so I think we could show something like "Token unknown" instead of the token details?

aristidesstaffieri: so thinking about this more, I think in the case that we fail to fetch token details we can do a couple of things -
We can surface the error and let them add it still, and label it "unknown" or something.
We can surface the error and not let them add the token.

We don't really have a pattern for inline errors so I'm not super sure what we would want here display wise. I would imagine an inline error and either one of the options above but let me know if you had something else in mind.

PR #1138, opened 2024-02-23
Bugfix/restore standalone balances

piyalbasu:


@shared/api/internal.ts
```diff
@@ -467,6 +481,156 @@ export const getAccountIndexerBalances = async (
...
+
+  return {
+    balance,
+    decimals,
+    name,
+    symbol,
+  };
+};
+
+export const getAccountBalancesStandalone = async ({
```
piyalbasu, line 517: creating a standalone fetch for balances that combines our old logic of fetching classic assets from a Horizon endpoint and token balances from a Soroban RPC endpoint

@shared/api/internal.ts
```diff
@@ -467,6 +481,156 @@ export const getAccountIndexerBalances = async (
...
+            decimals: await sorobanClientTxBuilder(),
+            symbol: await sorobanClientTxBuilder(),
+          },
+          params,
+        );
+        /* eslint-enable no-await-in-loop */
+
+        const total = new BigNumber(balance);
+
+        tokenBalances[`${symbol}:${tokenId}`] = {
```
piyalbasu, line 612: here I'm conforming the token balance to the shape the indexer provides so we don't run into type issues elsewhere

@shared/api/types.ts
```diff
@@ -177,6 +177,7 @@ export interface SorobanBalance {
   name: string;
   symbol: string;
   decimals: number;
+  token?: { code: string; issuer: { key: string } };
```
piyalbasu, line 180: this is added specifically to conform to the indexer's balance response

@shared/helpers/soroban/server.ts
```diff
@@ -0,0 +1,21 @@
+import {
```
piyalbasu, line 1: restoring this file

@shared/helpers/soroban/token.ts
```diff
@@ -0,0 +1,93 @@
+import {
```
piyalbasu, line 1: same. just restoring this file that existed before

extension/src/popup/SorobanContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const SorobanProvider = ({
+  children,
+  pubKey,
+}: {
+  children: React.ReactNode;
+  pubKey: string;
+}) => {
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const { sorobanRpcUrl = "" } = networkDetails;
```
piyalbasu, line 29: I simplified the logic here as before we were dealing with edge cases where network's didn't have an RPC. I think we can assume everyone will be including an RPC url going forward, but I'll test the fail case of this

extension/src/popup/components/manageAssets/AddToken/index.tsx
```diff
@@ -129,6 +135,26 @@ export const AddToken = () => {
             domain: record.domain,
           })),
         );
+      } else if (isCustomNetwork(networkDetails)) {
+        const name = await getName(
```
piyalbasu, line 139: hit the RPC to find some details about the token ID a user is trying to add

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -277,20 +279,36 @@ export const getAccountBalances = createAsyncThunk<
...
-    return thunkApi.rejectWithValue({ errorMessage: e });
-  }
-});
+>(
+  "getAccountBalances",
+  async ({ publicKey, networkDetails, sorobanClient }, thunkApi) => {
+    try {
+      let balances;
+
+      if (isCustomNetwork(networkDetails)) {
```
piyalbasu, line 291: splitting the logic between an indexer fetch and a local fetch

aristidesstaffieri (approved):

PR #1139, opened 2024-02-23
adds fallback to local mode in account history for custom networks

aristidesstaffieri:
Adds fallback to use local connections to Horizon/rpc for account history when using a custom network.


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -256,7 +288,7 @@ export const HistoryItem = ({
 
               const formattedTokenAmount = formatTokenAmount(
                 new BigNumber(attrs.amount),
-                _token.decimals,
+                _token.decimals!,
```
aristidesstaffieri, line 291: This should be ok because if any of the rpc calls to get token details fail, we will hit the catch block which renders a more generic history item row without token details.

piyalbasu (approved):

PR #1141, opened 2024-02-23
restoring sign/send for standalone networks

piyalbasu:


@shared/api/internal.ts
```diff
@@ -877,6 +879,80 @@ export const signFreighterSorobanTransaction = async ({
...
+    tx = TransactionBuilder.fromXDR(
+      signedXDR,
+      networkDetails.networkPassphrase,
+    );
+  } catch (e) {
+    console.error(e);
+  }
+
+  if (
+    !networkDetails.sorobanRpcUrl &&
```
aristidesstaffieri, line 916: should probably now just be `if (!networkDetails.sorobanRpcUrl) {..}`

PR #1143, opened 2024-02-25
Tests/add soroban token

piyalbasu:


extension/src/popup/components/manageAssets/AddToken/styles.scss
```diff
@@ -10,6 +10,7 @@
       margin-left: 0.15625rem;
 
       a {
+        display: inline;
```
piyalbasu, line 13: small css fix while I'm here

aristidesstaffieri (approved):

PR #1146, opened 2024-02-26
add lighter weight getpublickey with docs

piyalbasu:


@shared/api/external.ts
```diff
@@ -3,7 +3,7 @@ import { NetworkDetails } from "../constants/stellar";
 import { sendMessageToContentScript } from "./helpers/extensionMessaging";
 import { UserInfo } from "./types";
 
-export const requestPublicKey = async (): Promise<string> => {
+export const requestAccess = async (): Promise<string> => {
```
piyalbasu, line 6: requestAccess is now the method that will prompt the user to grant access

@shared/api/external.ts
```diff
@@ -21,6 +21,24 @@ export const requestPublicKey = async (): Promise<string> => {
   return publicKey;
 };
 
+export const requestPublicKey = async (): Promise<string> => {
```
piyalbasu, line 24: getPublicKey will return the public key or simply return an empty string now

docs/docs/guide/usingFreighterWebApp.md
```diff
@@ -1,13 +1,13 @@
 ---
-id: usingFreighterNode
-title: Using Freighter in node.js
```
piyalbasu, line 3: I don't know why I called this `node.js` before. This is pretty wrong 🤦

aristidesstaffieri (approved):

PR #1147, opened 2024-02-27
Feature/disable soroban based on health

piyalbasu:


extension/src/background/helpers/account.ts
```diff
@@ -127,6 +132,28 @@ export const getNetworksList = async () => {
   return networksList;
 };
 
+export const getIsRpcHealthy = async (networkDetails: NetworkDetails) => {
+  let rpcHealth = { status: "" };
+  if (isCustomNetwork(networkDetails)) {
+    // TODO: use server.getHealth method to get accurate result for standalone network
+    rpcHealth = { status: "healthy" };
```
piyalbasu, line 139: weirdly, I was having trouble using the sdk method for checking rpc health for my local Standalone. I'll take another look at this, but I think Mainnet/Testnet is the priority here

extension/src/popup/components/manageAssets/SelectAssetRows/index.tsx
```diff
@@ -45,6 +45,17 @@ export const SelectAssetRows = ({ assetRows }: SelectAssetRowsProps) => {
     return "";
   };
 
+  const getTokenBalanceFromCanonical = (canonical: string) => {
```
piyalbasu, line 48: small unrelated fix: when choosing an asset to Send Payment with, token balance wasn't being formatted with decimals correctly

aristidesstaffieri (approved):

PR #1152, opened 2024-02-27
use the network passphrase from the passed xdr to properly figure out…

piyalbasu:
use the network passphrase from the passed xdr to properly figure out what network the tx was built for

### DIFF
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -76,6 +76,7 @@ export const SignTransaction = () => {
 
   const {
     accountToSign: _accountToSign,
+    transaction: { _fee, _networkPassphrase },
     transactionXdr,
     domain,
     isDomainListedAllowed,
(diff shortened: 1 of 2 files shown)


extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -76,20 +76,20 @@ export const SignTransaction = () => {
...
   } = tx;
 
+  // rebuild transaction to get Transaction prototypes
   const transaction = TransactionBuilder.fromXDR(
     transactionXdr,
-    networkPassphrase,
+    _networkPassphrase,
   ) as Transaction | FeeBumpTransaction;
 
-  const { fee: _fee, networkPassphrase: _networkPassphrase } = transaction;
```
piyalbasu, line 91: before, we were building the xdr with the network that Freighter was configured to and then comparing the network from that Transaction with Freighter's network. It was always coming back as true, of course.

Now, we grab the network from the passed Transaction. Then we rebuild it the Transaction using the xdr and the passphrase from the old Transaction

aristidesstaffieri (approved):

PR #1159, opened 2024-03-01
make sure we always have sorobanRpcUrl and improve error messaging

piyalbasu:
Screenshot 2024-03-01 at 3 10 31 PM

### DIFF
--- a/extension/src/popup/SorobanContext.tsx
+++ b/extension/src/popup/SorobanContext.tsx
@@ -46,7 +46,7 @@ export const SorobanProvider = ({
         serverUrl = SOROBAN_RPC_URLS[NETWORKS.PUBLIC];
         break;
       default:
-        serverUrl = SOROBAN_RPC_URLS[NETWORKS.PUBLIC];
+        serverUrl = "";
     }
 
     server = new SorobanRpc.Server(serverUrl, {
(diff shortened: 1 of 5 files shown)


extension/src/popup/SorobanContext.tsx
```diff
@@ -26,19 +28,51 @@ export const SorobanProvider = ({
...
+        serverUrl = SOROBAN_RPC_URLS[NETWORKS.FUTURENET];
+        break;
+      case NETWORKS.TESTNET:
+        serverUrl = SOROBAN_RPC_URLS[NETWORKS.TESTNET];
+        break;
+      case NETWORKS.PUBLIC:
+        serverUrl = SOROBAN_RPC_URLS[NETWORKS.PUBLIC];
+        break;
+      default:
+        serverUrl = SOROBAN_RPC_URLS[NETWORKS.PUBLIC];
```
piyalbasu, line 49: this is a bit heavy handed, but we'll be refactoring this whole thing into a helper, so this is a band-aid to unblock users who get stuck

piyalbasu: using a switch because indexing SOROBAN_RPC_URLS with an enum was kind of messy

aristidesstaffieri: sounds good! Wondering if we should still throw in the default case since it will error downstream anyway?

aristidesstaffieri (approved):

extension/src/popup/SorobanContext.tsx
piyalbasu: yeah that's a good idea. Using the mainnet rpc with the wrong network will lead to weird user experience

mtwtfss:
@piyalbasu can you add some tests for this change?

aristidesstaffieri (approved):

PR #1165, opened 2024-03-04
Feautre/improve sentry tracking

piyalbasu:


aristidesstaffieri (approved):

mtwtfss:
is there anything else we can do to reduce the frequency of these errors?
do we know why exactly this is happening? is it 5XX errors from Amplitude? is it rate limiting?
could we consider increasing the timeout for these Amplitude requests? do you think that would help?

piyalbasu:
> is there anything else we can do to reduce the frequency of these errors? do we know why exactly this is happening? is it 5XX errors from Amplitude? is it rate limiting? could we consider increasing the timeout for these Amplitude requests? do you think that would help?

This is a good question. Sentry notoriously does not provide much information about `failed to fetch` errors: https://github.com/getsentry/sentry-javascript/issues/4146

My hunch is that these are a combination of rate-limiting and 5xx errors from Amplitude (https://www.docs.developers.amplitude.com/analytics/apis/http-v2-api/#429-response-toomanyrequestsfordeviceerror) as these are documented issues.

The payloads that we send to Amplitude are very small (usually just the name of the view that has rendered) so it's unlikely to be a payload size issue.

We do have a throttle of 500ms for requests to Amplitude. I could certainly try just increasing this timeout, but the downside is that we may lose some analytic events if a user is navigating through the app quickly

mtwtfss:
it would be great if we can get more concrete evidence of what is causing these issues with Amplitude before completely filtering all of them out

I'm worried that any issues in reporting analytic events might not be obvious just from looking at the dashboards if we filter these events out

piyalbasu:
Closing in favor of better Amplitude error tracking

PR #1166, opened 2024-03-04
Feature/improve router loading state

aristidesstaffieri:
What
Moves our BottomNav component to a central place instead of rendering it individual views.
Creates an "Outlet" component to house the BottomNav, global loader and fetchers, and Routes.

Why
Right now, when you load Freighter you see 2 different loading states, first the router loading with no nav and then the page loading with nav. This can be jarring because you expect 1 loading state as an end user. Pulling out the BottomNav to be rendered persistently at the router level lets our 2 loading states blend together better.

Notes
As I hinted above, this doesn't totally fix it because we do still have 2 loading screens where there should be 1(notice the flicker of the loader in the "after" recording). In order to fix that, we would need to move the calls and/or loading states for `loadSettings/loadAccount` into each view in order to only show 1 loading state from the combined loading states of the page data and settings/account data.

After -

https://github.com/stellar/freighter/assets/6886006/0b2867d1-2ffe-44bb-9e99-691d5c5966fa

### DIFF
--- a/extension/src/popup/Router.tsx
+++ b/extension/src/popup/Router.tsx
@@ -239,6 +239,27 @@ const RouteListener = () => {
   return null;
 };
 
+const SHOW_NAV_ROUTES = [
+  ROUTES.account,
+  ROUTES.accountHistory,
+  ROUTES.settings,
+  ROUTES.connectWallet,
+  ROUTES.connectWalletPlugin,
+  ROUTES.swapSettings,
+  ROUTES.sendPaymentAmount,
+];
+
+const NO_APP_LAYOUT_ROUTES = [
+  ROUTES.mnemonicPhrase,
+  ROUTES.mnemonicPhraseConfirmed,
+  ROUTES.accountCreator,
+  ROUTES.accountMigration,
+  ROUTES.recoverAccount,
+  ROUTES.recoverAccountSuccess,
+  ROUTES.pinExtension,
+  ROUTES.welcome,
+];
+
 const Outlet = () => {
   const dispatch = useDispatch();
   const location = useLocation();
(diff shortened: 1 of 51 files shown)


extension/src/popup/Router.tsx
```diff
@@ -248,137 +253,164 @@ export const Router = () => {
...
+    location.pathname &&
+    ((location.pathname === "/" &&
+      applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED) ||
+      location.pathname === ROUTES.account ||
+      location.pathname === ROUTES.accountHistory ||
+      location.pathname === ROUTES.settings ||
+      location.pathname === ROUTES.connectWallet ||
+      location.pathname === ROUTES.connectWalletPlugin ||
+      location.pathname === ROUTES.swapSettings ||
+      location.pathname === ROUTES.sendPaymentAmount ||
```
piyalbasu, line 266: think it'd be a little cleaner to create an array of the ROUTES that need to showNav and an array for ROUTES that need isAppLayout. it'll be a little easier to edit a list in array form

extension/src/popup/components/ErrorBoundary/index.tsx
```diff
@@ -54,7 +54,7 @@ export const UnhandledError = ({
   // expected to work outside of <Router />
   const isOnAccount = window.location.hash === "#/account";
   return (
-    <View>
+    <React.Fragment>
```
piyalbasu, line 57: does removing `View` here and in subsequent views still make full screen mode work?

aristidesstaffieri: yeah it does. The <View> is still above these components in the Outlet now, we have to remove these because having 2 Views nested created weird layout bugs but the rendered dom should be very similar in hierarchy.

piyalbasu:
> @piyalbasu wanted to get your opinion on the note in the PR note above -
> 
> ```
> As I hinted above, this doesn't totally fix it because we do still have 2 loading screens where there should be 1(notice the flicker of the loader in the "after" recording). In order to fix that, we would need to move the calls and/or loading states for loadSettings/loadAccount into each view in order to only show 1 loading state from the combined loading states of the page data and settings/account data.
> ```
> 
> I think in order to get the loading state correct here 100%, we need to go from 2 loading screens to 1. This is a much bigger refactor because it means every view needs to either account for the extra loading vars from `loadAccount/loadSettings` or it needs to roll those up into its data lifecycle. Likely, this is too much to include in this same PR but I think worth doing to get the UX correct here.

Agreed! I think it'd be great to get this UX right, but probably too big of a refactor for one PR. I think iterative improvement is a solid strategy here.  Do you want to create a follow up ticket to finish this work? We can release this improvement in the meantime

piyalbasu (approved):

PR #1171, opened 2024-03-06
Sentry not properly alerting on non-200 rpc health

piyalbasu:


aristidesstaffieri (approved):

extension/src/background/helpers/account.ts
```diff
@@ -143,6 +143,10 @@ export const getIsRpcHealthy = async (networkDetails: NetworkDetails) => {
       const res = await fetch(
         `${INDEXER_URL}/rpc-health?network=${networkDetails.network}`,
       );
+
+      if (!res.ok) {
+        captureException(`Failed to load rpc health for Soroban`);
+      }
```
piyalbasu, line 149: a non-200 doesn't actually throw an error, so it didn't reach the catch block to alert

PR #1176, opened 2024-03-12
[BUG] Handle not ok case in subscriptions

aristidesstaffieri:
What?
Handles the !res.ok() case when subscribing to an account and/or token details.
Resets the HAS_ACCOUNT_SUBSCRIPTION storage object.
Commented out the Sentry log for these failures as we expect this to fail when Mercury is off.

Why?
I was looking into our Sentry logs and realized that we should expect to see a ton of [these logs](https://github.com/stellar/freighter/blob/master/extension/src/background/helpers/account.ts#L207) because we try to subscribe to accounts/token details even when mercury is not enabled, but we saw none of those logs.
The reason is because the 400 thrown response in this case, when using fetch, does not throw but instead returns the field ok as false. Since we were not checking this, when any user hit this error we ignored it and then set their account to true in HAS_ACCOUNT_SUBSCRIPTION incorrectly.

Since Mercury is not "on" yet, this doesn't affect any users but needs to be resolved before turning it on. Resetting the accounts is necessary in this case since many accounts are likely marked as having subs when they do not.

Todo
Should we add the useMercury flag from the backend to the feature-flags route? Right now, I don't think we have a way for the UI to know when to try to subscribe or not, we could expose the flag from the backend in order to use that to skip these calls when Mercury is disabled.

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -196,18 +196,24 @@ export const subscribeAccount = async (publicKey: string) => {
         network: networkDetails.network,
       }),
     };
-    await fetch(`${INDEXER_URL}/subscription/account`, options);
+    const res = await fetch(`${INDEXER_URL}/subscription/account`, options);
     const subsByKeyId = {
       ...hasAccountSubByKeyId,
       [keyId]: true,
     };
-    await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, subsByKeyId);
+
+    if (res.ok) {
+      await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, subsByKeyId);
+    } else {
+      const resJson = await res.json();
+      throw new Error(resJson);
+    }
   } catch (e) {
     console.error(e);
-    captureException(
-      `Failed to subscribe account with Mercury - ${JSON.stringify(e)}`,
-    );
-    throw new Error("Error subscribing account");
+    // Turn on when Mercury is enabled
+    // captureException(
+    //   `Failed to subscribe account with Mercury - ${JSON.stringify(e)}`,
+    // );
   }
 
   return { publicKey };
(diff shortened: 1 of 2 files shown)


extension/src/background/helpers/account.ts
```diff
@@ -196,18 +196,24 @@ export const subscribeAccount = async (publicKey: string) => {
...
     };
-    await fetch(`${INDEXER_URL}/subscription/account`, options);
+    const res = await fetch(`${INDEXER_URL}/subscription/account`, options);
     const subsByKeyId = {
       ...hasAccountSubByKeyId,
       [keyId]: true,
     };
-    await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, subsByKeyId);
+
+    if (res.ok) {
```
aristidesstaffieri, line 205: we 400 when Mercury is turned off which does not go to the catch block and was getting missed.

extension/src/background/helpers/account.ts
```diff
@@ -196,18 +196,24 @@ export const subscribeAccount = async (publicKey: string) => {
...
+      const resJson = await res.json();
+      throw new Error(resJson);
+    }
   } catch (e) {
     console.error(e);
-    captureException(
-      `Failed to subscribe account with Mercury - ${JSON.stringify(e)}`,
-    );
-    throw new Error("Error subscribing account");
+    // Turn on when Mercury is enabled
```
aristidesstaffieri, line 213: Since we can't skip around this call yet, I turned this off but if we add the feature flag to avoid calling this when Mercury is off then we can turn it back on before Mercury is on.

extension/src/background/helpers/account.ts
```diff
@@ -230,13 +236,20 @@ export const subscribeTokenBalance = async (
...
+    if (!res.ok) {
+      const resJson = await res.json();
+      throw new Error(resJson);
+    }
   } catch (e) {
     console.error(e);
     captureException(
       `Failed to subscribe token balance - ${JSON.stringify(e)}`,
     );
-    throw new Error(`Error subscribing to token: ${contractId}`);
```
aristidesstaffieri, line 239: We don't need to throw here because we display the op as a generic case in the case of this failure.

extension/src/background/helpers/dataStorage.ts
```diff
@@ -246,6 +246,16 @@ const migrateSorobanRpcUrlNetwork = async () => {
   }
 };
 
+export const resetAccountSubscriptions = async () => {
```
aristidesstaffieri, line 249: We need to reset this object because of the error handling bug when subscribing to accounts. This has no effect on users since Mercury has not been turned on yet.

extension/src/background/helpers/account.ts
piyalbasu: This is a good call - we should add `useMercury` to the feature-flags endpoint

piyalbasu (approved):

extension/src/background/helpers/account.ts
aristidesstaffieri: cool I'll get that added to the feature-flags route.

PR #1179, opened 2024-03-12
Bugfix/allowlist native contract

piyalbasu:
Screenshot 2024-03-12 at 6 25 56 PM

Do not flag XLM as verified or unverified


extension/src/popup/components/manageAssets/SearchAsset/styles.scss
```diff
@@ -23,16 +23,12 @@
 
   &__loader {
     position: absolute;
-    top: calc(50% - var(--SearchAsset--loader--dimension));
```
piyalbasu, line 26: this fixes an issue across adding trustlines and soroban tokens where the spinning loader wasn't aligned properly

extension/src/popup/components/manageAssets/AddToken/index.tsx
```diff
@@ -106,6 +107,9 @@ export const AddToken = () => {
   const [isSearching, setIsSearching] = useState(false);
   const [hasNoResults, setHasNoResults] = useState(false);
   const [isVerifiedToken, setIsVerifiedToken] = useState(false);
+  const [isVerificationInfoShowing, setIsVerificationInfoShowing] = useState(
```
piyalbasu, line 110: e2e tests for this whole flow will be added in the manifest v3 testing issue: https://stellarorg.atlassian.net/browse/WAL-1375

extension/src/popup/helpers/searchAsset.ts
```diff
@@ -21,6 +21,32 @@ export const searchAsset = async ({
...
+      return {
+        ...NATIVE_CONTRACT_DEFAULTS,
+        contract: "CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA",
+        issuer: "GDMTVHLWJTHSUDMZVVMXXH6VJHA2ZV3HNG5LYNAZ6RTWB7GISM6PGTUV",
+      };
+    case NETWORKS.TESTNET:
+      return {
+        ...NATIVE_CONTRACT_DEFAULTS,
+        contract: "CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC",
+        issuer: "",
```
aristidesstaffieri, line 43: curious why there's an issuer on mainnet but not testnet?

aristidesstaffieri (approved):

extension/src/popup/helpers/searchAsset.ts
piyalbasu: When Testnet resets, my understanding is that this contract will need to be redeployed, at which point the issuer could be different. I believe this value will be out of date eventually.

We don't actually do anything with this value, so this shouldn't pose a problem. The `issuer` key is more useful for trustlines, I'm just following the common interface here

PR #1182, opened 2024-03-13
[FEATURE] Token transfer summary in auth entry review

aristidesstaffieri:
What
Adds a special case in auth entry review for token transfers, adds a transfer summary in these cases.

Why
This was included in the original design but was left out of the original implementation

Details
Mocks - https://www.figma.com/file/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?type=design&node-id=228-16058&mode=design

The mocks won't match exactly due to some changes in design and other lines of work, notable differences include:
We've added both from/to in all cases and renamed to sender/receiver
The contract ID gravatar is still on the left because there is another line of work that will include moving this and other usability improvements for IDs
Icons are different because they were designed on sds v3 but we are still on v2

Example:
Screenshot 2024-03-13 at 12 14 05 PM

### DIFF
--- a/extension/src/popup/views/ReviewAuth/index.tsx
+++ b/extension/src/popup/views/ReviewAuth/index.tsx
@@ -211,52 +211,44 @@ type TokenDetailMap = Record<string, { name: string; symbol: string }>;
 const TransferSummary = ({
   transfer,
   symbol,
-  publicKey,
 }: {
   transfer: {
     contractId: string;
     amount: string;
     to: string;
+    from: string;
   };
   symbol: string;
-  publicKey: string;
-}) => {
-  const isSending = transfer.to !== publicKey;
-
-  return (
-    <div className="AuthDetail__InfoBlock TransferSummary">
-      {isSending ? (
-        <div className="SummaryBlock">
-          <div className="SummaryBlock__Title">
-            <Icon.ArrowCircleRight />
-            <p>Sending to</p>
-          </div>
-          <KeyIdenticon publicKey={transfer.to} isSmall />
-        </div>
-      ) : (
-        <div className="SummaryBlock">
-          <div className="SummaryBlock__Title">
-            <Icon.ArrowCircleLeft />
-            <p>From</p>
-          </div>
-          <KeyIdenticon publicKey={transfer.to} isSmall />
-        </div>
-      )}
-      <div className="SummaryBlock">
-        <div className="SummaryBlock__Title">
-          <Icon.Toll />
-          <p>Amount</p>
-        </div>
-        <div className="SummaryBlock__Title">
-          <p>
-            {transfer.amount} {symbol}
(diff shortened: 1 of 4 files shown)


extension/src/popup/views/ReviewAuth/index.tsx
```diff
@@ -216,69 +273,127 @@ const AuthDetail = ({
...
+    [publicKey, networkDetails.network],
+  );
+
+  const transfersDepKey = JSON.stringify(transfers);
+  React.useEffect(() => {
+    async function getTokenDetails() {
+      setLoading(true);
+      const _tokenDetails = {} as TokenDetailMap;
+
+      // eslint-disable-next-line
```
aristidesstaffieri, line 294: @piyalbasu this is one the rules that I would vote that we drop. 
`For..of` is the least convoluted way to do async on a list, otherwise to avoid the loop we would need to do something like an async fn in a map or for each and then `Promise.all` on them which is easier to get wrong since you it's two separate steps. 

Let me know if you disagree though.

aristidesstaffieri:
@piyalbasu one question that I have after working on this -

What icon do we show for a SAC?
I don't know of a way to go from a SAC contract address to knowing which classic asset it is, if there's no way to do that reliably then I'm not sure that we can show the asset from the issuer's toml config? This could cause a disconnect between seeing a classic asset in your account view and seeing their custom icon, vs then seeing a SAC for that same asset but not seeing the same icon.

extension/src/popup/views/ReviewAuth/index.tsx
piyalbasu: Yeah, I agree - I don't love the `Promise.all` paradigm either. I think we have a good oppty now to update some of our linting rules. I'd think we can drop this rule as well as the similar `no await inside of a for loop`

piyalbasu (approved):

aristidesstaffieri:
> > @piyalbasu one question that I have after working on this -
> > What icon do we show for a SAC? I don't know of a way to go from a SAC contract address to knowing which classic asset it is, if there's no way to do that reliably then I'm not sure that we can show the asset from the issuer's toml config? This could cause a disconnect between seeing a classic asset in your account view and seeing their custom icon, vs then seeing a SAC for that same asset but not seeing the same icon.
> 
> This is a great question. Stellar.expert has found a way to do this (https://api.stellar.expert/explorer/public/asset-list/top50), so there must be a way to trace back to the original issuer to get their icon. We have a similar issue on the Manage Assets screen where we don't know what to show if a user adds a SAC by contract id. Let's take this on as a follow up task bc I agree, if it's a SAC, we should be able to show the same icon as the trustline

Make sense, I'll go dig up how OrbitLens tied them together in the list and then will create a task to review this everywhere that a SAC can appear in the UI.

PR #1184, opened 2024-03-14
clean up UI for showing "destination acct doesn't exist" warning

piyalbasu:
Fixing some UI jumpiness where the "destination acct doesn't exist" was showing for a split second while we were looking up the account to determine if it actually was funded or not
Screenshot 2024-03-14 at 12 09 49 PM
Screenshot 2024-03-14 at 12 09 37 PM

### DIFF
--- a/extension/src/popup/components/sendPayment/SendTo/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendTo/index.tsx
@@ -41,6 +41,7 @@ import {
 import { SorobanContext } from "popup/SorobanContext";
 
 import "../styles.scss";
+import { ActionStatus } from "@shared/api/types";
 
 const baseReserve = new BigNumber(1);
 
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/sendPayment/SendTo/index.tsx
```diff
@@ -287,24 +290,29 @@ export const SendTo = ({ previous }: { previous: ROUTES }) => {
                 <div>
                   {formik.isValid ? (
                     <>
-                      {!destinationBalances.isFunded && (
-                        <AccountDoesntExistWarning />
-                      )}
-                      {isFederationAddress(formik.values.destination) && (
+                      {destinationAccountBalanceStatus ===
+                      ActionStatus.SUCCESS ? (
```
piyalbasu, line 294: wait until the request has completed before we show anything

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -580,7 +582,7 @@ export const initialState: InitialState = {
   destinationBalances: {
     tokensWithNoBalance: [],
     balances: null,
-    isFunded: false,
+    isFunded: true,
```
piyalbasu, line 585: default to true because the negative case shows an error. Let's not show it until we're sure

aristidesstaffieri: I've been thinking that `isFunded` should actually be `boolean | null` because we've run into several unrelated bugs where the default boolean is incorrect due to an error in trying to populate it. Not recommending doing that now but wanted to surface for discussion.

PR #1187, opened 2024-03-14
[FIX] adds layout styles to form container to avoid button overflow in footer

aristidesstaffieri:
What
Adds some layout styles to the "manually add asset" form container

Why
To avoid the button in the footer overflowing

Before
Screenshot 2024-03-14 at 12 08 28 PM

After
Screenshot 2024-03-14 at 12 11 25 PM

### DIFF
--- a/extension/src/popup/components/manageAssets/AddAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/AddAsset/index.tsx
@@ -80,7 +80,7 @@ export const AddAsset = () => {
   return (
     <Formik initialValues={initialValues} onSubmit={handleSubmit}>
       {({ dirty, errors, isSubmitting, isValid, touched }) => (
-        <Form>
+        <Form className="FormContainer">
           <React.Fragment>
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -80,7 +80,7 @@ export const AddAsset = () => {
   return (
     <Formik initialValues={initialValues} onSubmit={handleSubmit}>
       {({ dirty, errors, isSubmitting, isValid, touched }) => (
-        <Form>
+        <Form className="FormContainer">
```
piyalbasu, line 83: can you namespace this is as `AddAsset__FormContainer` to prevent possible collision? I could see .FormContainer being reused

aristidesstaffieri: yup sure thing, done in e0b38d8e779eb7766a64537240d3247c2a53ae55

piyalbasu (approved):

PR #1190, opened 2024-03-15
[FIX] Account loading glitch

aristidesstaffieri:
What
Resets account balances any time the account balance status is reset
Moves isLoading to a useState var in Account
Adds a check for `balances` before showing not funded message

Why
When switching accounts, you would experience a flicker of the last state due to a mismatch in the rendering lifecycle and data fetching lifecycle.

### DIFF
--- a/extension/src/popup/components/identicons/AccountListIdenticon/index.tsx
+++ b/extension/src/popup/components/identicons/AccountListIdenticon/index.tsx
@@ -16,6 +16,7 @@ interface KeyIdenticonProps {
   publicKey: string;
   displayKey?: boolean;
   setIsDropdownOpen?: (IsDropdownOpen: boolean) => void;
+  setLoading?: (isLoading: boolean) => void;
 }
 
 export const AccountListIdenticon = ({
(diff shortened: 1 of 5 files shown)


extension/src/popup/views/Account/index.tsx
```diff
@@ -273,7 +284,7 @@ export const Account = () => {
                   />
                 </div>
               )}
-              {!isFunded && !hasError && (
+              {balances !== null && !isFunded && !hasError && (
```
aristidesstaffieri, line 287: @piyalbasu I'd like to chat through why this check is needed because I still think it should not be needed but the rendering steps still cause this to flash in between account changes without it, the state at that point doesn't make a ton of sense to me yet but this check avoids the flash because `balances` is only null before fetching balances or in the event of an error, which both are cases where this should not be shown.

extension/src/popup/components/identicons/AccountListIdenticon/index.tsx
```diff
@@ -29,12 +30,14 @@ export const AccountListIdenticon = ({
   const shortPublicKey = truncatedPublicKey(publicKey);
 
   const handleMakeAccountActive = () => {
-    if (!active) {
-      dispatch(makeAccountActive(publicKey));
-    }
     if (setIsDropdownOpen) {
       setIsDropdownOpen(false);
     }
+
```
aristidesstaffieri, line 36: The account page loads a bit smoother if we swap the order of these steps

piyalbasu (approved):

PR #1191, opened 2024-03-15
[FIX] HistoryItem/TransactionDetail load glitch

aristidesstaffieri:
What
Adds missing loading states in TransactionDetail & HistoryItem to avoid the layout shift.

Why
In both of these spots, we rendered null for a loading state which caused a layout shift as the nav moved up where the missing space is.

### DIFF
--- a/extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
+++ b/extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
@@ -6,6 +6,7 @@ import { Button } from "@stellar/design-system";
 import { KeyIdenticon } from "popup/components/identicons/KeyIdenticon";
 import { SubviewHeader } from "popup/components/SubviewHeader";
 import { AssetNetworkInfo } from "popup/components/accountHistory/AssetNetworkInfo";
+import { Loading } from "popup/components/Loading";
 import { View } from "popup/basics/layout/View";
 
 import { emitMetric } from "helpers/metrics";
@@ -72,12 +73,15 @@ export const TransactionDetail = ({
 
   const { t } = useTranslation();
 
-  const { assetDomain } = useAssetDomain({
+  const { assetDomain, error: assetError } = useAssetDomain({
     assetIssuer,
   });
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
+  const showContent = assetIssuer && !assetDomain && !assetError;
 
-  return assetIssuer && !assetDomain ? null : (
+  return showContent ? (
+    <Loading />
+  ) : (
     <React.Fragment>
       <SubviewHeader
         customBackAction={() => setIsDetailViewShowing(false)}
--- a/extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
@@ -420,7 +420,7 @@ export const SubmitFail = () => {
             <img src={IconFail} alt="Icon Fail" />
           </div>
           <div className="SubmitResult__error-code">
-            {errDetails.status ? `${errDetails.status}:` : ""}{" "}
+            {errDetails.status ? `Status ${errDetails.status}:` : ""}{" "}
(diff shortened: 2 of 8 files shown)


extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -72,12 +73,15 @@ export const TransactionDetail = ({
 
   const { t } = useTranslation();
 
-  const { assetDomain } = useAssetDomain({
+  const { assetDomain, error: assetError } = useAssetDomain({
```
aristidesstaffieri, line 76: @piyalbasu A good follow up here would be to handle the asset error more explicitly imo. If we fail to fetch an asset domain when we have an asset issuer, in `AssetNetworkInfo` we will incorrectly display these as `Stellar Lumens`.

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -336,12 +336,7 @@ export const SubmitFail = () => {
           "Destination account does not accept this asset",
         );
         errorDetails.errorBlock = (
-          <Notification
-            variant="error"
-            title={t(
-              "The destination account does not accept the asset you’re sending",
-            )}
-          >
+          <Notification variant="error" title="">
```
aristidesstaffieri, line 339: I happened to hit this while testing and notice that we display the same text as the title several times on the error page already. It seemed appropriate enough to leave the title off from the notification since the layout around it includes a title already.
Before:
Screenshot 2024-03-15 at 10 55 41 AM

After:
Screenshot 2024-03-15 at 10 57 50 AM

Let me know if you think this should be different though.

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -608,7 +608,6 @@ const transactionSubmissionSlice = createSlice({
     resetSubmission: () => initialState,
     resetAccountBalanceStatus: (state) => {
       state.accountBalanceStatus = initialState.accountBalanceStatus;
-      state.accountBalances = initialState.accountBalances;
```
aristidesstaffieri, line 611: This is a bug from the fix to the loading glitch in AccountList.
We can't actually reset balances here(and don't need to really) because not all other pages fetch the account balances before using them.

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
aristidesstaffieri: Sorry correction, it just shows an empty space using the empty string. Still looks a bit strange but maybe not as urgent as the mislabeling I thought would happen.

piyalbasu: Great - yeah, a slightly less optimal UI in an error case is preferable to incorrect info

piyalbasu: but we can follow up and fix this

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
piyalbasu: Oh yeah, we had a similar issue with another notification: https://github.com/stellar/freighter/blob/master/extension/src/popup/components/sendPayment/SendTo/index.tsx#L63

We actually handled by leaving the title and just omitting the repeated line in the body of the text

piyalbasu: Also, the `key ... returned an object...` error is a weird issue with `react-i18next` where if you try to translate a single line that exists verbatim in another body of text, it shows that error. So it's best in these cases to make sure the title text doesn't repeat in the text body

aristidesstaffieri: thanks, makes sense. Updated in f8c6d5b4a261db0e6b04dabd31da94e399928932

PR #1193, opened 2024-03-15
[FEATURE] send token from detail

aristidesstaffieri:
What
Removes restriction on send from asset detail for Soroban tokens.

Why
We support sending Soroban tokens as a feature and were missing the path to do it from asset detail.

### DIFF
--- a/extension/src/popup/components/account/AssetDetail/index.tsx
+++ b/extension/src/popup/components/account/AssetDetail/index.tsx
@@ -17,7 +17,7 @@ import {
 } from "popup/helpers/account";
 import { useAssetDomain } from "popup/helpers/useAssetDomain";
 import { navigateTo } from "popup/helpers/navigate";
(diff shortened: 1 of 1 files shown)


extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -183,17 +186,19 @@ export const AssetDetail = ({
...
-                    }}
-                  >
-                    {t("SEND")}
-                  </PillButton>
-                )}
+                <PillButton
+                  onClick={() => {
+                    dispatch(saveAsset(selectedAsset));
+                    if (isContract) {
+                      dispatch(saveIsToken(true));
```
aristidesstaffieri, line 193: @piyalbasu we really don't need to have `saveIsToken` since we can just check the issuer inline and avoid having to set this flag. I added a follow up task to refactor that away.

piyalbasu (approved):

PR #1195, opened 2024-03-15
[WIP] Feature/auto horizon error

piyalbasu:


extension/src/popup/locales/en/Please note the following the message.json
```diff
@@ -0,0 +1 @@
+{}
```
aristidesstaffieri, line 1: Translations?

extension/src/background/helpers/account.ts
```diff
@@ -164,6 +164,22 @@ export const getIsRpcHealthy = async (networkDetails: NetworkDetails) => {
...
 
+export const getUserNotification = async () => {
+  let response = { enabled: false, message: "" };
+
+  try {
+    const res = await fetch(`${INDEXER_URL}/user-notification`);
+    response = await res.json();
+  } catch (e) {
+    captureException(
+      `Failed to load feature flag for Soroban mainnet - ${JSON.stringify(e)}`,
```
aristidesstaffieri, line 175: Nit but feature flags are not per network right now

aristidesstaffieri (approved):
lgtm

extension/src/background/helpers/account.ts
piyalbasu: Oh good catch - bad copy/paste job

@shared/api/internal.ts
```diff
@@ -455,7 +455,6 @@ export const getAccountIndexerBalances = async (
   if ("error" in data && (data?.error?.horizon || data?.error?.soroban)) {
     const _err = JSON.stringify(data.error);
     captureException(`Failed to fetch account balances - ${_err}`);
-    throw new Error(_err);
```
piyalbasu: throw an error here leads to the redux thunk rejecting - which just leads to the generic `Failed to fetch account balances error`. If we don't throw, we can show the user a Horizon specific error

extension/src/popup/views/Account/index.tsx
```diff
@@ -309,7 +309,7 @@ export const Account = () => {
                   />
                 </div>
               )}
-              {!isFunded && !hasError && (
+              {!isFunded && !hasError && !error?.horizon && (
```
piyalbasu, line 312: don't show the unfunded message if we have a horizon message bc we're not sure if you're funded or not

PR #1199, opened 2024-03-18
fix auto-scrolling in add trustlines views

piyalbasu:


extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -81,7 +81,7 @@ export const AddAsset = () => {
     <Formik initialValues={initialValues} onSubmit={handleSubmit}>
       {({ dirty, errors, isSubmitting, isValid, touched }) => (
         <Form className="AddAsset__FormContainer">
-          <React.Fragment>
+          <View>
```
piyalbasu, line 84: because this is a UI bug, not a DOM issue, we'll need to build an e2e test for this view and then test this using Playwright. I'll be looking at this as part of our Manifest v3 testing

quietbits (approved):

PR #1209, opened 2024-03-22
before we close the sign transaction window, complete metrics upload

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/8b119640-3291-4cf0-8f7a-083a9f565290

A hypothesis for why we're seeing so many Amplitude errors in Sentry is that the popup is closing before the call can be completed. We seem to be getting a good deal more errors for SignTransaction. This may be because we were emitting a metric, signing/transmitting the tx, and then closing the window. The latter 2 steps happen very quickly. Maybe too quickly for the metric emit to complete.

I was able to simulate this by throttling my network to a slow speed and observing that at slow network speeds, I was pretty reliably generating a Sentry error because the Amplitude request was always still in process when the window closed. 

The change above shows what happens at a slow network speed now. The window waits to complete any metric emission before closing. This is at a throttled speed, so the delay is pretty pronounced. On a normal speed, it's almost completely unnoticeable.

Let's see if this reduces the amount of Sentry errors we get. We will still have the problem of Amplitude requests being cut off if someone quickly closes the popup. But this will hopefully a) reduce some noise and b) lend more credence to our hypothesis

### DIFF
--- a/extension/src/popup/helpers/useSetupSigningFlow.ts
+++ b/extension/src/popup/helpers/useSetupSigningFlow.ts
@@ -55,7 +55,6 @@ export function useSetupSigningFlow(
   };
 
   const signAndClose = async () => {
-    emitMetric(METRIC_NAMES.approveSign);
     if (isHardwareWallet) {
       dispatch(
         startHwSign({ transactionXDR: transactionXdr, shouldSubmit: false }),
@@ -63,6 +62,7 @@ export function useSetupSigningFlow(
       setStartedHwSign(true);
     } else {
       await dispatch(signFn());
+      await emitMetric(METRIC_NAMES.approveSign);
       window.close();
     }
   };
--- a/extension/src/helpers/metrics.ts
+++ b/extension/src/helpers/metrics.ts
@@ -147,7 +147,7 @@ const getUserId = () => {
  * @param {object?} body An optional object containing event metadata
  * @returns {void}
  */
-export const emitMetric = (name: string, body?: any) => {
+export const emitMetric = async (name: string, body?: any) => {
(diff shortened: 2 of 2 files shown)


extension/src/popup/helpers/useSetupSigningFlow.ts
```diff
@@ -55,14 +55,14 @@ export function useSetupSigningFlow(
...
   const signAndClose = async () => {
-    emitMetric(METRIC_NAMES.approveSign);
     if (isHardwareWallet) {
       dispatch(
         startHwSign({ transactionXDR: transactionXdr, shouldSubmit: false }),
       );
       setStartedHwSign(true);
     } else {
       await dispatch(signFn());
+      await emitMetric(METRIC_NAMES.approveSign);
```
aristidesstaffieri, line 65: Nice, I think this makes sense. I wonder what the UX is like if `emitMetric` takes a long time though. I don't think it will typically but in that case the user might experience a hung popup feeling because there isn't a loading state associated with `emitMetric`. 
Would this edge case be handled by emitting the metric from the action itself so that it can piggy back on that loading/success/error lifecycle?

aristidesstaffieri (approved):

extension/src/popup/helpers/useSetupSigningFlow.ts
piyalbasu: Good thought. If you check the video, the loading spinner does continue to spin on the `Sign` button while metric is being emitted to give the visual cue that something is still happening.

The signed xdr is transmitted immediately, so if for some odd reason the emit metric takes a realllllly long time, the dapp already has the xdr and the user can continue whatever they're doing. The popup will just continue to operate and close whenever it finishes

PR #1210, opened 2024-03-25
Feature/multiple allow list

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/2c6a91fe-e90c-4721-b325-257c72b46b7d


.eslintrc.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
     "react/jsx-filename-extension": ["error", { extensions: [".tsx", ".jsx"] }],
     "jsdoc/newline-after-description": "off",
     "max-len": "off",
+    "no-await-in-loop": "off",
```
piyalbasu, line 34: we can get rid of this, right? it'll be one last lint rule to ignore in the code lol

@shared/api/internal.ts
```diff
@@ -1162,7 +1163,9 @@ export const editCustomNetwork = async ({
   return response;
 };
 
-export const loadSettings = (): Promise<Settings & IndexerSettings> =>
+export const loadSettings = (): Promise<
+  Settings & IndexerSettings & { assetsLists: AssetsLists }
```
piyalbasu, line 1167: assets lists don't quite fall into the paradigm of general app settings or settings from the indexer. Maybe if this "3rd thing" category grows, we can add a separate service to deal with it

@shared/api/types.ts
```diff
@@ -66,7 +68,7 @@ export interface Response {
   network: string;
   networkIndex: number;
   networkName: string;
-  recentAddresses: Array<string>;
+  recentAddresses: string[];
```
piyalbasu, line 71: fixing a couple linting warnings while I'm in here

extension/src/background/helpers/dataStorage.ts
```diff
@@ -254,6 +256,18 @@ export const resetAccountSubscriptions = async () => {
     // once account is unlocked, setup Mercury account subscription if !HAS_ACCOUNT_SUBSCRIPTION
     await localStore.setItem(HAS_ACCOUNT_SUBSCRIPTION, {});
   }
+  await migrateDataStorageVersion("4.0.2");
```
piyalbasu, line 259: the `resetAccountSubscription` migration forgot to migrate the version, so adding this in here now

.eslintrc.js
aristidesstaffieri: 😄 yes please

@shared/constants/soroban/token.ts
```diff
@@ -26,3 +28,35 @@ export interface SorobanToken {
...
+};
+
+export interface AssetsListItem {
+  url: string;
+  enabled: boolean;
+}
+
+export const DEFAULT_ASSETS_LISTS: AssetsLists = {
+  [NETWORKS.PUBLIC]: [
+    {
```
aristidesstaffieri, line 45: In the mocks, it looks like we will allow deleting a list but always want to retain one list as the "default" or one list that the user cannot delete. Should we add a field to the stellar expert list in order to key that one as the default one?

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: omg, thank you

extension/src/popup/views/ManageAssetsLists/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+            </ListNavLinkWrapper>
+          </div>
+        )}
+      </View.Content>
+      <View.Footer>
+        <Button
+          size="md"
+          isFullWidth
+          variant="tertiary"
+          onClick={() => navigateTo(ROUTES.account)}
```
aristidesstaffieri, line 113: assuming the add list is "TODO"

aristidesstaffieri (approved):

extension/src/popup/views/ManageAssetsLists/index.tsx
piyalbasu: Yup, just placeholder until the next view is created

@shared/constants/soroban/token.ts
piyalbasu: Yeah, good question. I need to follow up with product on this. Not sure which of these should be "undeletable"

PR #1211, opened 2024-03-26
[CHORE] Refactor standalone workflows

aristidesstaffieri:
What
Adds getTokenDetails, and replaces all inline calls to get token details
Makes getAccountHistory, to abstract over standalone mode vs indexer history
Moves all data related uses of `isCustomNetwork` into the data layer
Moves uses of Soroban server/tx builder into standalone workflows
Removes SorobanContext, replaces uses with helpers in standalone workflows

Why
Cleans up logic after the move to client side standalone mode. Soroban server and tx builder is only needed in standalone mode now, but we still used old logic and constructs with the assumptions that they were needed in all cases.


@shared/api/tsconfig.json
```diff
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "baseUrl": "."
   },
-  "include": ["."]
+  "include": [".", "../helpers/stellar.ts"]
```
aristidesstaffieri, line 6: Looks like this was auto-added because I include this helper in the api layer usage now.

extension/src/popup/components/account/AccountAssets/styles.scss
```diff
@@ -44,6 +44,8 @@ $loader-light-color: #444961;
     }
 
     &--error {
+      width: 32px;
```
aristidesstaffieri, line 47: Unrelated but I snuck these in here since these 10liners fix the loading state for assets on account details.

@shared/api/internal.ts
```diff
@@ -687,6 +695,79 @@ export const getIndexerAccountHistory = async ({
...
+      networkDetails,
+    });
+  }
+  return await getIndexerAccountHistory({
+    publicKey,
+    networkDetails,
+  });
+};
+
+export const getIndexerTokenDetails = async ({
```
piyalbasu, line 714: should this be called `getTokenDetails1 because it handles both the standalone/custom case and the indexer case?

piyalbasu (approved):

@shared/api/internal.ts
aristidesstaffieri: yeah for sure it should 😄 , done in 797a2c5d99761ba9bb8948c7f11e8de8fd769c57

PR #1221, opened 2024-03-29
Feature/check multiple tokenlist

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/af708e8c-6a95-47e9-8798-0dfb3805abad

This covers 3 scenarios:
1) Verified asset: shows what lists the token appears on in the modal
2) Unverified asset: shows unverified warning
3) native asset: shows no notification at all


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -746,70 +749,78 @@ export const UnverifiedTokenWarning = ({
...
-                  <br />
-                  <Link
-                    isUnderline
-                    variant="secondary"
-                    href="https://www.freighter.app/faq"
-                  >
-                    {t("Learn more")}
-                  </Link>
+                <div className="TokenWarning__flag__content">
+                  {t("This is a relatively new asset")}
```
piyalbasu, line 817: This "relatively new asset" is hardcoded for now (as all the assets are "relatively new" 😄 ). I'm looking into how we can actually figure out if an asset is now

extension/src/popup/helpers/searchAsset.ts
```diff
@@ -86,39 +58,77 @@ export interface TokenRecord {
...
+  let verifiedToken = {} as TokenRecord;
+  const verifiedLists: string[] = [];
 
-    verifiedTokens = verifiedTokenRes.filter((record: TokenRecord) => {
-      const regex = new RegExp(contractId, "i");
-      if (record.contract.match(regex)) {
-        return true;
+  promiseRes.forEach((r) => {
+    if (r.status === "fulfilled") {
+      const list = r.value?.tokens ? r.value?.tokens : r.value?.assets;
```
piyalbasu, line 109: in the future, these should all adhere to the SEP0042 spec where the key will be called `tokens`. Adding this check for now while I wait for Soroswap to fix this on their end

piyalbasu:
Also, something interesting to note is that when adding a SAC, the indexer is coming back with 2 separate balances: 1 for the classic Horizon balance and 1 for the contract balance. I've created a ticket here: https://stellarorg.atlassian.net/browse/WAL-1418

extension/src/popup/components/WarningMessages/index.tsx
aristidesstaffieri: We could index for new token deployments in the future but yeah not sure what else we could do right now.

aristidesstaffieri:
> Also, something interesting to note is that when adding a SAC, the indexer is coming back with 2 separate balances: 1 for the classic Horizon balance and 1 for the contract balance. I've created a ticket here: https://stellarorg.atlassian.net/browse/WAL-1418

Hmm yeah I guess really we shouldn't let SACs be added through the Soroban token flows since they are already represented in the wallet through the classic side. The SAC stores the `assetId:issuer` as the `name` of then token so I think we could look that up to know when a user is trying to add a SAC instance.

aristidesstaffieri (approved):

piyalbasu:
> > Also, something interesting to note is that when adding a SAC, the indexer is coming back with 2 separate balances: 1 for the classic Horizon balance and 1 for the contract balance. I've created a ticket here: https://stellarorg.atlassian.net/browse/WAL-1418
> 
> Hmm yeah I guess really we shouldn't let SACs be added through the Soroban token flows since they are already represented in the wallet through the classic side. The SAC stores the `assetId:issuer` as the `name` of then token so I think we could look that up to know when a user is trying to add a SAC instance.

Bruno's working on a spec for the SAC experience. The idea is basically if you add a SAC contract ID, it'll just add the trustline for it and we'll just use Horizon for balance & payment. If you already have the trustline, it likely just won't do anything

PR #1223, opened 2024-04-01
[FEATURE] Onboarding keyboard interaction

aristidesstaffieri:
What
Fixes/restores keyboard navigation/interaction in both onboarding flows.
Adds outline indicators to focused checkbox labels to match native browser behavior on our hidden input checkboxes
Adds enter key handling for missing places in onboarding: tos checking, mnemonic phrase word selection

Why
Users without a mouse currently will not be able to complete onboarding.

### DIFF
--- a/extension/src/popup/components/Onboarding/index.tsx
+++ b/extension/src/popup/components/Onboarding/index.tsx
@@ -97,7 +97,7 @@ export const OnboardingButtons = ({
           {showBackButton ? (
             <BackButton
               customButtonComponent={
-                <Button variant="secondary" size="md">
+                <Button variant="secondary" size="md" type="button">
                   {t("Back")}
                 </Button>
               }
--- a/extension/src/popup/components/mnemonicPhrase/CheckButton/index.tsx
+++ b/extension/src/popup/components/mnemonicPhrase/CheckButton/index.tsx
@@ -5,12 +5,18 @@ import "./styles.scss";
 
 type CheckButtonProps = {
   onChange: (e: any) => void;
+  onKeyDown?: (e: any) => void;
   wordKey: string;
   word: string;
 };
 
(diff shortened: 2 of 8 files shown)


extension/src/popup/components/mnemonicPhrase/CheckButton/index.tsx
```diff
@@ -5,12 +5,18 @@ import "./styles.scss";
...
 
-export const CheckButton = ({ onChange, wordKey, word }: CheckButtonProps) => (
-  <>
+export const CheckButton = ({
+  onChange,
+  onKeyDown,
+  wordKey,
+  word,
+}: CheckButtonProps) => (
+  <span>
```
aristidesstaffieri, line 19: We need to use a wrapping element here in order to use the sibling css selector to make the label outlined when the hidden input is focused. Using a span lets them still render inline and retains the current layout.

extension/src/popup/components/mnemonicPhrase/CheckButton/styles.scss
```diff
@@ -12,7 +12,14 @@
 }
 
 .CheckButton {
-  display: none;
+  opacity: 0;
+
+  &:focus + label {
```
aristidesstaffieri, line 17: matching the default element focus styles from the browser

extension/src/popup/components/Onboarding/index.tsx
```diff
@@ -97,7 +97,7 @@ export const OnboardingButtons = ({
           {showBackButton ? (
             <BackButton
               customButtonComponent={
-                <Button variant="secondary" size="md">
+                <Button variant="secondary" size="md" type="button">
```
aristidesstaffieri, line 100: in some states, this would submit the form since the default type is "submit"

piyalbasu (approved):

PR #1224, opened 2024-04-04
Feature/modify asset list

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/85c109e5-0c2e-44cc-a731-777115f50af7


@shared/constants/soroban/token.ts
```diff
@@ -37,26 +37,29 @@ export type AssetsLists = {
 
 export interface AssetsListItem {
   url: string;
-  enabled: boolean;
+  isEnabled: boolean;
```
piyalbasu, line 40: fixing boolean naming convention

extension/src/background/helpers/dataStorage.ts
```diff
@@ -151,8 +151,8 @@ const migrateTokenIdList = async () => {
       [NETWORKS.FUTURENET]: tokenIdsByKey,
     };
     await localStore.setItem(TOKEN_ID_LIST, newTokenList);
+    await migrateDataStorageVersion("1.0.0");
```
piyalbasu, line 154: oof - this was kind of a gotcha. Because this wasn't inside the conditional, on every update, the version was being set to `1.0.0`, which meant is going down to the whole list of migrations. Could this be why we've had a couple reports of data changing?

piyalbasu: I believe we should only be upping the migration version if a migration has been applied, but check my logic on this!

extension/src/popup/components/manageAssetsLists/AssetLists/index.tsx
```diff
@@ -0,0 +1,96 @@
+import React from "react";
```
piyalbasu, line 1: this is just moving all the code from views/ManageAssetsLists to a component

extension/src/popup/components/manageAssetsLists/ModifyAssetList/index.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const [submitErrorString, setSubmitErrorString] = useState("");
+  const [assetListInfo, setAssetListInfo] = useState({} as AssetsListsData);
+  const [isFetchingAssetList, setIsFetchingAssetList] = useState(false);
+  const [isEditing, setIsEditing] = useState(false);
+  const [isDefaultAssetList, setIsDefaultAssetList] = useState(false);
+
+  const defaultAssetsListIndex = DEFAULT_ASSETS_LISTS[selectedNetwork].length;
+
+  useEffect(() => {
+    if (assetListIndex) {
```
piyalbasu, line 54: we are in "Edit" mode in this block, so prepopulating some info

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: ah man yeah that makes sense, looks like the last few were always incrementing it 😞 
I still don't see how this could cause the network switch. I see that `migrateSorobanRpcUrlNetwork` is the last network related migration to run but it only updates the selected network if it is futurenet so doesn't totally explain the reported behavior. This def should be only incrementing inside the version check though.

extension/src/popup/views/ManageAssetsLists/index.tsx
```diff
@@ -1,62 +1,63 @@
 import React, { useEffect, useState } from "react";
```
piyalbasu, line 1: migrating most of this view to components

aristidesstaffieri:
It looks like we will still only use the stellar-expert list to search for verified tokens, just want to confirm that it is intended to not use the newly added lists yet? @piyalbasu

aristidesstaffieri (approved):

aristidesstaffieri:
> > It looks like we will still only use the stellar-expert list to search for verified tokens, just want to confirm that it is intended to not use the newly added lists yet? @piyalbasu
> 
> This is actually already working. I added this in a previous PR: https://github.com/stellar/freighter/blob/release/5.18.0/extension/src/popup/helpers/searchAsset.ts#L83
> 
> Screenshot 2024-04-04 at 3 29 00 PM

oh yup, I hadn't pulled that down yet. Thanks!

PR #1229, opened 2024-04-09
default to selected network if network is verifiable

piyalbasu:
* When the user is on a verifiable network, default the asset list dropdown to that network. If not, default to Mainnet
* If the asset list field is empty, make sure to show error string

### DIFF
--- a/extension/src/popup/views/ManageAssetsLists/index.tsx
+++ b/extension/src/popup/views/ManageAssetsLists/index.tsx
@@ -24,17 +24,18 @@ export interface AssetsListsData {
 }
 
 export const ManageAssetsLists = () => {
-  const [selectedNetwork, setSelectedNetwork] = useState(
-    NETWORKS.PUBLIC as AssetsListKey,
-  );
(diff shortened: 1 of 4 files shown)


extension/src/popup/views/ManageAssetsLists/index.tsx
```diff
@@ -63,6 +64,18 @@ export const ManageAssetsLists = () => {
     fetchLists();
   }, [selectedNetwork, assetsLists]);
 
+  useEffect(() => {
+    if (networkDetails.network === NETWORKS.TESTNET) {
```
aristidesstaffieri, line 68: Is this check already taken care of by the next `setSelectedNetwork`?

aristidesstaffieri (approved):

extension/src/popup/views/ManageAssetsLists/index.tsx
piyalbasu: Oh whoops - my bad, meant to remove this!

PR #1232, opened 2024-04-11
Feature/redesign settings

piyalbasu:
Screenshot 2024-04-10 at 1 29 35 PM
Screenshot 2024-04-10 at 1 30 00 PM
Screenshot 2024-04-10 at 1 29 44 PM


piyalbasu:
I went down a little bit of a rabbit hole on this PR. We generally load svg's using an img tag as opposed to loading the svg directly in the tsx, which works fine. But when you do this, you get a little flash of no content while the img tag loads. Svg's written directly into the markup don't do this; they just load with the html. Hard to notice the img tag flash in other parts of the app because we're generally waiting for data to load, which gives the img tag a second to catch up.

I noticed it more in the Settings screens as these are static pages. It was even _more_ noticeable because the Icons from SDS are loaded directly in the tsx as React components, so they were loading first while the img tags had to catch up.

I'm using svgr to load some svg's directly into the tsx as react components. I set this up so if you add `?react` to the end of an svg import, you can load it in as a React component, which just writes the svg directly into the html.

I left in the option to load an svg as an img because we sometimes need to dynamically load an image where we don't know if it'll be an svg or png or what (like for asset icons). In this case, we'll want to continue using img tag's for this

extension/src/popup/views/ReviewAuth/index.tsx
```diff
@@ -18,6 +18,8 @@ import { getTokenDetails } from "@shared/api/internal";
 import { PunycodedDomain } from "popup/components/PunycodedDomain";
```
aristidesstaffieri, line 18: This file is the only non settings related change so just confirming that this is intended for this PR? Lgtm if it is.

aristidesstaffieri (approved):
👏

extension/src/popup/views/ReviewAuth/index.tsx
piyalbasu: ooof - my bad. thanks for catching this! I was testing something with Ledger and this ended up on the branch 🤦 Removing this

PR #1234, opened 2024-04-12
Validate asset list on every fetch

piyalbasu:


extension/src/popup/components/manageAssetsLists/ModifyAssetList/index.tsx
```diff
@@ -90,21 +90,6 @@ export const ModifyAssetList = ({
...
-      schemaRes = await fetch(
-        "https://raw.githubusercontent.com/orbitlens/stellar-protocol/sep-0042-token-lists/contents/sep-0042/assetlist.schema.json",
-      );
-    } catch (err) {
-      captureException("Unable to fetch SEP-0042 JSON schema");
-      setFetchErrorString("Unable to validate asset asset list");
-      return;
-    }
-
-    const schemaResJson = await schemaRes?.json();
```
piyalbasu, line 106: moving this check to a helper that runs every time we fetch a list. This is to guard against instances where a user added a list that was valid, and then at some point, the list became invalid. In that case, we should fail gracefully

aristidesstaffieri (approved):

PR #1235, opened 2024-04-15
asset list UI cleanup

piyalbasu:


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -482,6 +482,9 @@ export const ManageAssetRow = ({
   const { blockedDomains } = useSelector(transactionSubmissionSelector);
   const canonicalAsset = getCanonicalFromAsset(code, issuer);
   const isScamAsset = !!blockedDomains.domains[domain];
+  const assetCode = name || code;
+  const truncatedAssetCode =
+    assetCode.length > 20 ? truncateString(assetCode) : assetCode;
```
piyalbasu, line 487: for SAC's, if they're not an asset list, they come back with {asset_code:issuer_key}, which gets pretty long and breaks the UI

aristidesstaffieri (approved):

PR #1237, opened 2024-04-15
break once we found the contract in asset list

piyalbasu:


extension/src/popup/helpers/searchAsset.ts
```diff
@@ -151,6 +151,7 @@ export const getVerifiedTokens = async ({
           if (record.contract && record.contract.match(regex)) {
             verifiedToken = record;
             verifiedLists.push(r.value.name as string);
+            break;
```
piyalbasu, line 154: found this edge case when testing against stellar.expert's testnet list. If a contract ID shows up multiple times on an asset list, it can cause some confusing results in the UI. Let's just stop check an asset list if we find one occurrence of an asset on it. 

The asset list spec shouldn't allow multiple entries for the same contract ID. This guards against a mistake in the asset list schema. Also, it provides a small perf boost

aristidesstaffieri (approved):

PR #1250, opened 2024-04-23
[BUG] fixes fall through in mnemonic phrase component

aristidesstaffieri:
What
Adds a redirect to account creation in the fall through case in the MnemonicPhrase component

Why
The previous logic rendered a null when falling through, this happened whenever the MnemonicPhrase was rendered outside of AccountCreator without passing in a mnemonic.
If you closed onboarding at the "confirm mnemonic" step, then you would hit this fall through and render a dead screen.
Now we redirect to account creator in this case so that the user can re-enter their password and get back to their mnemonic when this happens.


aristidesstaffieri:
I pointed this at 5.18 to make the diff easy but I can rebase this onto the next release since 5.18 is on the finish line @piyalbasu

extension/src/popup/views/MnemonicPhrase.tsx
```diff
@@ -55,5 +55,11 @@ export const MnemonicPhrase = ({
     );
   }
 
-  return null;
+  return (
```
piyalbasu, line 58: Just want to confirm - what happens if you create an account, close the window at this stage, come back (are redirected to ` ROUTES.accountCreator` as stated here) and then create a new account instead of importing your old phrase?

Kind of an edge case, but just want to make sure that if a user does something weird, it doesn't put their account data into a weird state. I _think_ it should work fine and ignore the old keystore, but not 100% sure

piyalbasu (approved):

extension/src/popup/views/MnemonicPhrase.tsx
aristidesstaffieri: yeah so I think you're right that it does take you back to the create a password step if this happens. I think maybe we could also handle this by doing this instead - 
instead of routing back to account creator, we could add a new branch to that component that shows a password field and does a `SHOW_BACKUP_PHRASE` after collecting it.

My original thoughts were that at this stage if they abandon and come back, they has no way of key being used yet so easy enough to add this fallback but I could see how it would be better to do what I listed above. It is an edge case but maybe worth switching to getting the existing phrase?

piyalbasu: Yeah, I think maybe the most complete solution here is to have the user enter password and get back to `SHOW_BACKUP_PHRASE` 👍

aristidesstaffieri: @piyalbasu I added this in [7c276d2](https://github.com/stellar/freighter/pull/1250/commits/7c276d281ddc8d5bab71d4873cd8528f8bf5eb61)

PR #1262, opened 2024-04-30
check for existence of localStore before doing migration

piyalbasu:
https://github.com/stellar/freighter/assets/6789586/808f21a9-76d6-408d-8c26-d104fb141a11

Fixes an issue where previous local storage data was continually overwriting current settings via migration

### DIFF
--- a/extension/src/background/helpers/dataStorage.ts
+++ b/extension/src/background/helpers/dataStorage.ts
@@ -2,6 +2,7 @@ import browser from "webextension-polyfill";
 import semver from "semver";
 
 import {
+  APPLICATION_ID,
   HAS_ACCOUNT_SUBSCRIPTION,
   NETWORK_ID,
   NETWORKS_LIST_ID,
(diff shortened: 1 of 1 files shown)


extension/src/background/helpers/dataStorage.ts
```diff
@@ -80,6 +81,13 @@ export const normalizeMigratedData = async () => {
   const localStore = dataStorageAccess(browserLocalStorage);
   const localStorageEntries = Object.entries(localStorage);
 
+  const applicationState = await localStore.getItem(APPLICATION_ID);
+  const isLocalStoreSetup = !!applicationState?.length;
```
piyalbasu, line 85: checking the existence of the `applicationState` to determine if this migration has been run. `applicationState` is a base setting that is set right away in user flow, so we should be able to rely on the existence of this to determine that the user has interacted with Freighter before

piyalbasu: The interface we set up for chrome.storage.local only allows getting by string key, it doesn't allow you to get the whole object by passing `null`. Because of that, I decided to check by key, which is a more specific check anyway and should be more deterministic

aristidesstaffieri (approved):

PR #1267, opened 2024-05-01
[WIP] Chore/ts wallet sdk data provider

aristidesstaffieri:
What
Replaces all remaining instances of wallet-sdk

Why
wallet-sdk is deprecated

### DIFF
--- a/@shared/api/types.ts
+++ b/@shared/api/types.ts
@@ -1,7 +1,5 @@
 import BigNumber from "bignumber.js";
-import { Horizon } from "stellar-sdk";
-import { Types } from "@stellar/wallet-sdk";
-import { AssetBalance, NativeBalance } from "@stellar/wallet-sdk/dist/types";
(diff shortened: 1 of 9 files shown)


@shared/api/types.ts
```diff
@@ -181,6 +179,53 @@ export interface AssetDomains {
   [code: string]: string;
 }
 
+export interface NativeToken {
```
aristidesstaffieri, line 182: These never moved to ts-wallet-sdk, will they?

@shared/helpers/stellar.ts
```diff
@@ -7,3 +15,112 @@ export const isCustomNetwork = (networkDetails: NetworkDetails) => {
 
   return network === CUSTOM_NETWORK;
 };
+
+export function getBalanceIdentifier(
```
aristidesstaffieri, line 19: more helpers that never made it to ts-wallet-sdk, will they be ported?

PR #1275, opened 2024-05-06
[FEATURE] Standalone contract spec

aristidesstaffieri:
What
Adds getContractSpec for standalone custom networks

Why
getContractSpec is missing for custom networks


@shared/api/helpers/soroban.ts
```diff
@@ -25,3 +33,108 @@ export const decodeU32 = (b64: string) =>
...
+  let requestBody = {
+    jsonrpc: "2.0",
+    id: id,
+    method: "getLedgerEntries",
+    params: {
+      keys: [entryKey],
+    },
+  };
+
+  let res = await fetch(rpcUrl, {
```
piyalbasu, line 65: just want to check what happens if `rpcUrl` is undefined. WIll this throw an error? May be best to wrap this in a try/catch

piyalbasu (approved):

@shared/api/helpers/soroban.ts
aristidesstaffieri: yeah it will throw an error. Any error thrown from the `getContractSpec` right now is caught by the caller since this isn't part of a store and holds no state to put an error for the views to use. So if this errors, the views that calls it falls back to not showing the param names like before.

I think the alternative if we want to move the error handling to the internal of this would be to make a store and actions and then store/use that error in the view instead.

Issue #1259, opened 2024-04-26
Make transaction timeout configurable


tomerweller:
Timeout is currently set pretty low, which makes sense for hot wallet operations. More sophisticated users might take the raw xdr to be signed elsewhere and that might take some time. 

This can be a global setting.

aristidesstaffieri:
This was published in version 5.18.4.

PR #1276, opened 2024-05-07
[FEATURE] Configure transaction timeout

aristidesstaffieri:
What
Adds transaction timeout configuration to tx settings for payments and swaps.

Why
To allow users to configure different time bounds for transactions built by Freighter
https://github.com/stellar/freighter/issues/1259


extension/src/popup/components/sendPayment/SendSettings/TxTimeout/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+          ),
+        })}
+      >
+        {({ values, isValid, errors }) => (
+          <Form className="View__contentAndFooterWrapper">
+            <View.Content>
+              <FormRows>
+                <Field name="transactionTimeout">
+                  {({ field }: FieldProps) => (
+                    <Input
```
piyalbasu, line 68: I think we should offer a default value here. If I didn't know much about timeouts, I wouldn't know what to put here

piyalbasu (approved):

extension/src/popup/components/sendPayment/SendSettings/TxTimeout/index.tsx
aristidesstaffieri: We do get the default from the tx store here so it defaults to the previous hardcoded value of 180

PR #1280, opened 2024-05-09
fix issue with metric middleware not running on page load

piyalbasu:
This fixes an issue where the loading of the Sign Transaction view was not being logged correctly in Amplitude. We were getting events for `loaded screen: sign transaction`, but they were all missing the `domain` property, which we use to see what dapps people are signing tx's on.

Context: https://stellarfoundation.slack.com/archives/C016CPSLVLG/p1715202813917719?thread_ts=1715200459.695319&cid=C016CPSLVLG

There were 2 bugs here:

1) the /review-auth route was incorrectly emitting the `loaded screen: sign transaction` metric. This route isn't configured to report `domain`, so it was sending that event but without the custom property.

2) /sign-transaction, the view that's supposed to be emitting that metric, wasn't reporting at all. That's because it was reporting _before_ the user loaded their settings. Before a user loads their settings, `isDataSharingAllowed` defaults to `false`. So, it would get ready to emit the correct event (with the `domain` custom property), but it would abort when it hit the `isDataSharingAllowed` flag.

### DIFF
--- a/extension/src/popup/Router.tsx
+++ b/extension/src/popup/Router.tsx
@@ -227,10 +227,13 @@ const HomeRoute = () => {
 const RouteListener = () => {
   const dispatch = useDispatch();
   const location = useLocation();
+  const settingsState = useSelector(settingsStateSelector);
 
   useEffect(() => {
-    dispatch(navigate(location));
-  }, [dispatch, location]);
+    if (settingsState === SettingsState.SUCCESS) {
+      dispatch(navigate(location));
+    }
+  }, [dispatch, location, settingsState]);
 
   return null;
 };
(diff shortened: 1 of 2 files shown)


extension/src/popup/Router.tsx
```diff
@@ -227,10 +227,13 @@ const HomeRoute = () => {
 const RouteListener = () => {
   const dispatch = useDispatch();
   const location = useLocation();
+  const settingsState = useSelector(settingsStateSelector);
 
   useEffect(() => {
-    dispatch(navigate(location));
-  }, [dispatch, location]);
+    if (settingsState === SettingsState.SUCCESS) {
+      dispatch(navigate(location));
```
piyalbasu, line 234: This action is used to emit `loaded screen:` events. None of them were firing on initial page load because the settings weren't loaded yet. When a user would nav to a different view, this would fire correctly bc the settings would have loaded by then. But you would lose the initial page load, which is important for views like `/sign-transaction` that pop up from freighter-api

piyalbasu:
Before:
Screenshot 2024-05-08 at 9 27 53 PM

After:
Screenshot 2024-05-08 at 9 57 07 PM

extension/src/popup/Router.tsx
piyalbasu: This `RouteListener` used to instantiate after all account and settings were loaded

aristidesstaffieri (approved):

PR #1297, opened 2024-05-16
don't babel-polyfill contentScript

piyalbasu:
Firefox has a strict Content Security Policy where they don't allow `eval()` in an extension content script. `eval` is added as a byproduct of Webpack compiling our code with `babel-polyfill`. I simply removed the `babel-polyfill` compilation for `contentScript`.

I don't think we'll lose much here, as we're not supporting old versions of (for ex) IE. Also, FF and Chrome do autoupdate. We'll need to be mindful of using only ES approved features. ES2023 has been supported by Chrome and FF since 2022

### DIFF
--- a/extension/public/static/manifest/v3.json
+++ b/extension/public/static/manifest/v3.json
@@ -3,6 +3,12 @@
   "version": "5.18.5",
   "version_name": "5.18.5",
   "description": "Freighter is a non-custodial wallet extension that enables you to sign Stellar transactions via your browser.",
+  "browser_specific_settings": {
+    "gecko": {
+      "id": "{3ee0dd4e-8c64-4b92-b539-25718a10f62f}",
+      "strict_min_version": "48.0"
+    }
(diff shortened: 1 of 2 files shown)


extension/public/static/manifest/v3.json
```diff
@@ -3,6 +3,12 @@
   "version": "5.18.5",
   "version_name": "5.18.5",
   "description": "Freighter is a non-custodial wallet extension that enables you to sign Stellar transactions via your browser.",
+  "browser_specific_settings": {
+    "gecko": {
+      "id": "{3ee0dd4e-8c64-4b92-b539-25718a10f62f}",
+      "strict_min_version": "48.0"
```
piyalbasu, line 9: this is already present in the Manifest v2 json. We need to have this in v3 as well, specifically for Firefox

aristidesstaffieri (approved):

PR #1298, opened 2024-05-17
use yarn to build docs

piyalbasu:
The yarn-specific aliases I created for stellar-sdk and stellar-sdk-next weren't playing nice with the docs build because they use `npm` instead of yarn. 

I had been using `npm` to make sure the docs grabbed freighter-api from the registry rather than trying to symlink to the workspace in this repo. This PR finds a different workaround for this

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "@docusaurus/core": "2.4.1",
     "@docusaurus/preset-classic": "2.4.1",
-    "@stellar/freighter-api": "^2.0.0",
+    "@stellar/freighter-api": "latest",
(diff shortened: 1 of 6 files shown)


docs/package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "@docusaurus/core": "2.4.1",
     "@docusaurus/preset-classic": "2.4.1",
-    "@stellar/freighter-api": "^2.0.0",
+    "@stellar/freighter-api": "latest",
```
piyalbasu, line 28: apparently if you use `latest`, yarn will ignore the workspace and just grab from the registry :tada

aristidesstaffieri (approved):

PR #1303, opened 2024-05-28
Feature/hash signing 2

piyalbasu:
Enables hash signing in Ledgers, which allows users to sign next-gen XDR's by simply signing the hash rather than decoding the XDR and reviewing individual operations

Screenshot 2024-05-28 at 1 35 52 PM
Screenshot 2024-05-28 at 1 36 05 PM
Screenshot 2024-05-28 at 1 36 20 PM


extension/public/static/manifest/v3.json
```diff
@@ -3,17 +3,20 @@
   "version": "5.18.5",
   "version_name": "5.18.5",
   "description": "Freighter is a non-custodial wallet extension that enables you to sign Stellar transactions via your browser.",
+  "browser_specific_settings": {
```
aristidesstaffieri, line 6: this is probably fine to leave in but looks like it's from a different line of work.

extension/src/background/helpers/dataStorage.ts
```diff
@@ -39,7 +40,7 @@ export const browserLocalStorage = storage?.local;
 export const browserSessionStorage = storage?.session;
 
 // Session Storage Feature Flag - turn on when storage.session is supported
-export const SESSION_STORAGE_ENABLED = false;
+export const SESSION_STORAGE_ENABLED = true;
```
aristidesstaffieri, line 43: Should this flag get flipped in this PR?

extension/src/popup/components/AutoSave/index.tsx
```diff
@@ -31,7 +33,13 @@ export const AutoSaveFields = ({ debounceMs = 500 }: AutoSaveFieldsProps) => {
 
   useEffect(() => {
     if (formik.isValid && formik.dirty && !formik.isSubmitting) {
-      debouncedSubmit(formik);
+      console.log(values);
```
aristidesstaffieri, line 36: forgotten debug logs?

extension/src/popup/views/ExperimentalFeatures/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    isExperimentalModeEnabledValue: boolean;
+    isHashSigningEnabledValue: boolean;
+  }
+
+  const initialValues: SettingValues = {
+    isExperimentalModeEnabledValue: isExperimentalModeEnabled,
+    isHashSigningEnabledValue: isHashSigningEnabled,
+  };
+
+  const handleSubmit = async (formValue: SettingValues) => {
```
aristidesstaffieri, line 37: Does this action need a loading state?

aristidesstaffieri:
Seems like we have a bunch of formatting changes, am wondering if you're local prettier version is out of date?

extension/src/popup/views/ExperimentalFeatures/index.tsx
piyalbasu: I added this, though it just shows for a split second so may be hard to see

Screenshot 2024-05-28 at 4 22 13 PM

aristidesstaffieri: Thank yeah I did honestly mean that as a question cause if it doesn't make a network trip, even though it's an action, sometimes it can be janky with the very fast loading state but I don't think we have a real pattern for when it is or isn't appropriate I think.

aristidesstaffieri (approved):

PR #1304, opened 2024-05-28
Test/data storage

aristidesstaffieri:
What
Adds tests for the data storage wrappers and migrations

Why
We've hit several storage/migration bugs in the past and had no test coverage for this area.


extension/src/background/helpers/dataStorage.ts
```diff
@@ -105,7 +105,7 @@ export const normalizeMigratedData = async () => {
 
 // This migration adds a friendbotUrl to testnet and futurenet network details
 export const migrateFriendBotUrlNetworkDetails = async () => {
-  const localStore = dataStorageAccess(browserLocalStorage);
+  const localStore = exports.dataStorageAccess(browserLocalStorage);
```
aristidesstaffieri, line 108: we have to assign these to "exports" in order to be able to properly mock them with jest. This should be equivalent to local reference.

piyalbasu: Interesting - I've never had to do this. Can you point me to some documentation on this?

piyalbasu (approved):

PR #1317, opened 2024-06-04
Fix manage assets tests

piyalbasu:


extension/src/popup/views/__tests__/ManageAssets.test.tsx
```diff
@@ -5,6 +5,7 @@ import {
   fireEvent,
   screen,
   within,
+  getByTestId,
```
aristidesstaffieri, line 8: looks like this is failing because this import is unused.

PR #1319, opened 2024-06-06
add better error messaging for liabilities when removing an asset

piyalbasu:
Screenshot 2024-06-06 at 6 22 21 PM


extension/src/popup/components/manageAssets/TrustlineError/index.tsx
```diff
@@ -78,13 +91,32 @@ const RenderedError = ({
...
+      return (
+        <>
+          <div className="TrustlineError__title">
+            {t("This asset has buying liabilities")}
+          </div>
+          <div
+            className="TrustlineError__body"
+            data-testid="TrustlineError__body"
+          >
+            {t("You still have a buying liability of")} {buyingLiabilities}.{" "}
```
aristidesstaffieri, line 114: Should the buying liability message also include the liability currency, like instead of `You still have a buying liability of 1`, should it say `You still have a buying liability of 1 BLND`? 

I'm not sure if that's obvious from the context of that page already.

aristidesstaffieri (approved):

extension/src/popup/components/manageAssets/TrustlineError/index.tsx
piyalbasu: It _should_ be obvious, I think. You have to specifically click `remove asset` next to the asset to get to this screen. 

I'm mostly jsut matching the pattern set by the `this asset has a balance` error above

PR #1320, opened 2024-06-07
verify params before fetching issuer info

piyalbasu:
This reduces the number of failing calls we get from tokens on the Account screen. With classic assets, we look up the issuer to get the home domain, and then load the toml to get the icon. Tokens don't follow this structure, so users were getting a lot of failed HTTP calls. 

The errors were handled gracefully, so it didn't disrupt UX. This PR will give a performance boost as we can make sure to not make HTTP calls that we know will fail


aristidesstaffieri (approved):
ah nice thank you!!

PR #1324, opened 2024-06-13
fix add asset padding; fix ledger trustline error; fix token-spec check

piyalbasu:


@shared/api/internal.ts
```diff
@@ -745,12 +745,12 @@ export const getIsTokenSpec = async ({
     `${INDEXER_URL}/token-spec/${contractId}?network=${networkDetails.network}`,
   );
   const response = await fetch(url.href);
-  const { data, error } = await response.json();
+  const { isSep41Compliant, error } = await response.json();
```
aristidesstaffieri, line 748: ah weird not sure how that got through, thanks!

aristidesstaffieri (approved):

@shared/api/internal.ts
aristidesstaffieri: actually since this hasn't been released yet, should we change the APi response to match the other instead with the schema `{data, error}`?

piyalbasu: Yup, good call. I'll revert this change

aristidesstaffieri: I made the schema change here https://github.com/stellar/freighter-backend/pull/83/commits/be87a3cf0952de9da56ec42157267a7475ea9a20

PR #1333, opened 2024-06-18
upgrade redux toolkit

piyalbasu:


extension/src/helpers/metrics.ts
```diff
@@ -30,17 +29,6 @@ export function metricsMiddleware<State>(): Middleware<object, State> {
...
-// without creating an intermediate value
-// eslint-disable-next-line
-const dummyThunk = createAsyncThunk<any, any>("dummy", () => {});
-const dummyAction = createAction<any>("also dummy");
-type ActionType =
-  | typeof dummyThunk.fulfilled
-  | typeof dummyThunk.rejected
-  | typeof dummyThunk.pending
-  | typeof dummyAction;
-
```
piyalbasu, line 43: this doesn't play well with the newer version of redux toolkit. This was pretty hacky anyway, I don't think we need to work this hard to deal with redux toolkit type weirdness

aristidesstaffieri (approved):

PR #1356, opened 2024-07-02
Feature/insufficient fee warning

aristidesstaffieri:
What
Adds a check and warning associated with a user trying to sign a transaction with an account that does not have enough xlm to cover the associated tx fee.

Why
This keeps users from going all the way through the flow just to get rejected on submission.

Screenshot 2024-07-02 at 8 52 25 AM

### DIFF
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -1,8 +1,9 @@
 import React, { useCallback, useEffect, useState } from "react";
 import { useLocation } from "react-router-dom";
-import { useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import { useTranslation, Trans } from "react-i18next";
 import { Button, Icon, Notification } from "@stellar/design-system";
+import BigNumber from "bignumber.js";
 import {
   MuxedAccount,
   Transaction,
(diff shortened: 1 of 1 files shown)


extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -212,6 +237,37 @@ export const SignTransaction = () => {
...
+    return <Loading />;
+  }
+
+  const hasBalance =
+    hasLoadedBalances && accountBalanceStatus !== ActionStatus.ERROR;
+  const hasEnoughXlm = accountBalances.balances?.native.available.gt(
+    stroopToXlm(_fee as string),
+  );
+  if (hasBalance && currentAccount.publicKey && !hasEnoughXlm) {
+    return (
```
piyalbasu, line 254: hm, i don't think we should block signing. I think maybe we should just warn, and if the user still wants to sign, they can. 

Thinking of a scenario where a different account will be submitting and therefore will be responsible for the fee

aristidesstaffieri: hmm yeah good call, let me see if I can make this non blocking.

aristidesstaffieri: @piyalbasu what do you think of this? Made the warning dismissable in c1cd38e8ad7a97d0a4faa37a654804d1d1cba4b5

piyalbasu (approved):

PR #1357, opened 2024-07-03
Feature/standard wallet interface

piyalbasu:
For ref: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0043.md


@shared/api/external.ts
```diff
@@ -31,25 +37,23 @@ export const requestPublicKey = async (): Promise<string> => {
     console.error(e);
   }
 
-  const { publicKey, error } = response;
-
-  if (error) {
-    throw error;
-  }
-  return publicKey;
+  return { publicKey: response?.publicKey || "", error: response?.apiError };
```
piyalbasu, line 40: A similar change happens in each of the functions below:

instead of `throw`ing on error, we're returning the error as a possible key on the payloaf

@shared/api/external.ts
```diff
@@ -105,59 +112,79 @@ export const submitBlob = async (
...
+export const requestNetwork = async (): Promise<{
+  network: string;
+  networkPassphrase: string;
+  error?: FreighterApiError;
+}> => {
+  let response;
   try {
     response = await sendMessageToContentScript({
-      type: EXTERNAL_SERVICE_TYPES.REQUEST_NETWORK,
+      type: EXTERNAL_SERVICE_TYPES.REQUEST_NETWORK_DETAILS,
```
piyalbasu, line 172: this method now return the network and the network passphrase. I'm opting to use the `REQUEST_NETWORK_DETAILS` service so i can grab both those values at once.

`REQUEST_NETWORK`, though not used, will need to stay in the codebase for older versions of freighter-api that still make use of it

@stellar/freighter-api/src/index.ts
```diff
@@ -1,6 +1,7 @@
+import { getAddress } from "./getAddress";
 import { getPublicKey } from "./getPublicKey";
 import { signTransaction } from "./signTransaction";
-import { signBlob } from "./signBlob";
+import { signMessage } from "./signMessage";
```
piyalbasu, line 4: another significant change here is that I'm repurposing `signBlob` to become `signMessage`. We've talked about removing `signBlob` for a bit, and since `signMessage` is a very similar API, I'm making that change now

@shared/api/external.ts
```diff
@@ -13,16 +19,16 @@ export const requestAccess = async (): Promise<string> => {
...
   }
 
-  const { publicKey, error } = response;
+  const { publicKey } = response || { publicKey: "" };
 
-  if (error) {
-    throw error;
-  }
-  return publicKey;
+  return { publicKey, error: response?.apiError };
```
piyalbasu, line 24: A similar change happens in each of the functions below:

instead of throwing on error, we're returning the error as a possible key on the payload

@shared/api/external.ts
```diff
@@ -244,28 +277,9 @@ export const setAllowedStatus = async (): Promise<boolean> => {
...
 
-  const { isAllowed, error } = response;
-
-  if (error) {
-    throw error;
-  }
-  return isAllowed;
-};
-
-export const requestUserInfo = async (): Promise<UserInfo> => {
```
piyalbasu, line 255: I'm removing this method as it's duplicative of the `getAddress` method. I'm leaving the underlying service to maintain backward compatibility

@stellar/freighter-api/src/getAddress.ts
```diff
@@ -0,0 +1,20 @@
...
+    const req = await requestPublicKey();
+
+    if (req.error) {
+      return { error: req.error };
+    }
+
+    return { address: req.publicKey };
+  }
+
+  return { error: FreighterApiNodeError };
```
piyalbasu, line 19: similar changes are happening to each of the client facing api methods:

If there's an error, return it.

If not, return the requested data.

Also, in the case of a node env, return a "node is not supported" error

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -95,282 +99,347 @@ export const freighterApiMessageListener = (
           resolve({ publicKey: publicKeySelector(sessionStore.getState()) });
         }
 
-        resolve({ error: "User declined access" });
+        resolve({
```
piyalbasu, line 102: This file has a huge diff but the main changes here are:

1. Returning 2 kinds of errors to maintain backwards compat
2. Wrapping all these methods in try...catch so on any type of runtime error, we can surface an error

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1068,7 +1070,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
       const transactionResponse = responseQueue.pop();
 
       if (typeof transactionResponse === "function") {
-        transactionResponse(response);
+        transactionResponse(response, sourceKeys.publicKey());
```
piyalbasu, line 1073: now the signing methods also return the address that signed them - so adding this as a second param

extension/src/helpers/urls.ts
```diff
@@ -2,13 +2,14 @@ import punycode from "punycode";
 import browser from "webextension-polyfill";
 import { TransactionInfo } from "../types/transactions";
 
-export interface BlobToSign {
+export interface MessageToSign {
```
piyalbasu, line 5: below is a lot of renaming from `signBlob` to `signMessage`

extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -77,6 +83,23 @@ export const SignAuthEntry = () => {
...
+      >
+        <p>
+          {t("The requester expects you to sign this auth entry on")}{" "}
+          {entryNetworkPassphrase}.
+        </p>
+        <p>{t("Signing this transaction is not possible at the moment.")}</p>
+      </WarningMessage>
+    );
+  }
+
```
piyalbasu, line 102: now that signAuthEntry and signMessage accept a desired network, they should also show a warning when the user is on the wrong network. Lifted from the SignTransaction UI

@stellar/freighter-api/src/signMessage.ts
```diff
@@ -0,0 +1,29 @@
...
+  blob: string,
+  opts?: {
+    accountToSign?: string;
+  }
+): Promise<
+  | { signedMessage: string; signerAddress: string }
+  | { error: FreighterApiError }
+> => {
+  if (isBrowser) {
+    const req = await submitBlob(blob, opts);
```
aristidesstaffieri, line 16: should this be renamed to `submitMessage` to match the other reference changes?

extension/src/helpers/urls.ts
```diff
@@ -2,13 +2,14 @@ import punycode from "punycode";
...
   isDomainListedAllowed: boolean;
   domain: string;
   tab?: browser.Tabs.Tab;
   blob: string;
```
aristidesstaffieri, line 9: should this field be named `message` now to match the update references?

aristidesstaffieri (approved):

PR #1358, opened 2024-07-08
[FEATURE] Connect malicious dapp warning

aristidesstaffieri:
What
Adds a variant to the GrantAccess modal, `malicious`. This prop will toggle the modal to show the "malicious dapp" warning and is triggered from a blockaid query to scan a site.

Why
Initial blockaid integration to warn about scam/malicious sites.

Depends on: https://github.com/stellar/freighter-backend/pull/99

Design: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1617-5469&t=eB9KEEBCcXKr7c02-0

Screenshot 2024-07-08 at 11 55 31 AM


extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -85,6 +85,7 @@ export const freighterApiMessageListener = (
     browser.windows.create({
       url: chrome.runtime.getURL(`/index.html#/grant-access?${encodeOrigin}`),
       ...WINDOW_SETTINGS,
+      width: 400,
```
aristidesstaffieri, line 88: needed to make the window a little wider to accomodate design changes to the buttons.

extension/src/popup/components/ModalInfo/styles.scss
```diff
@@ -24,10 +34,10 @@
 
   &--subject {
     font-size: 0.75rem;
-    color: var(--color-gray-70);
-    background-color: var(--color-gray-30);
+    color: #a0a0a0;
```
aristidesstaffieri, line 37: Should we add these colors to the SDS? These are the ones used in the mocks.

extension/src/popup/components/PunycodedDomain/index.tsx
```diff
@@ -37,7 +35,6 @@ export const PunycodedDomain = ({
           {isDomainValid ? punycodedDomain : `xn-${punycodedDomain}`}
         </strong>
       </div>
-      <div className="PunycodedDomain__title">{domainTitle}</div>
```
aristidesstaffieri, line 40: Removed full URL per design recommendation.

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1053,3 +1053,12 @@ const WarningMessageTokenDetails = ({
     </div>
   );
 };
+
+export const MaliciousDomainWarning = ({ message }: { message: string }) => (
+  <div className="MaliciousDomainWarning">
+    <div className="Icon">
+      <Icon.Warning className="WarningMessage__icon" />
```
aristidesstaffieri, line 1060: This icon is different from the one used in the mocks, I think this is an SDS v1 vs v2 difference.

extension/src/popup/components/ModalInfo/styles.scss
piyalbasu: Yeah, I think so - but it's probably okay to leave as-is for now. We need to update quite a few SDS components/colors. I'm working with Joseph to try to schedule this for Q3

extension/src/popup/ducks/blockaid.ts
```diff
@@ -0,0 +1,92 @@
+import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
+
+import { ActionStatus, ErrorMessage } from "@shared/api/types";
+import { INDEXER_URL } from "@shared/constants/mercury";
+import { NetworkDetails } from "@shared/constants/stellar";
+import { isCustomNetwork } from "@shared/helpers/stellar";
+
+export const scanSite = createAsyncThunk<
```
piyalbasu, line 8: just curious - what do you see as the value of saving the blockaid result in redux? it seems like we only need this result in the one view

aristidesstaffieri: yeah actually I wondered the same thing and I think this is true for the balance/history ducks as well. The reason I did it here is because it's the most common pattern in this app for fetching data and using a data fetching lifecycle(ActionStatus). I think the alternative would be to either remove the `scanSiteResults` key and assign it locally in the component or maybe do away with the duck altogether and use a helper + local loading states. 

The downside of the helper is that then later if we do need this call in another place, we're duplicating the loading logic but that seems fine to me.

piyalbasu: Yeah, I def see your point about the ActionStatus pattern. Maybe we should just use the helper/local loading state pattern for now and optimize to use redux if we find out later on we need this data in a few different places?

I don't think your approach is wrong or anything, I think redux just causes us to write a lot of lines code with all the slice/reducers stuff. Might be nicer to keep it a little simpler if there isn't a big advantage to using redux

aristidesstaffieri: 👍 refactored to a helper hook in b8f20f03c413cfb70e6c40bce1500ca860ac9f05

piyalbasu (approved):

PR #1370, opened 2024-07-12
Feature/standard wallet interface docs

piyalbasu:


@stellar/freighter-api/src/getAddress.ts
```diff
@@ -4,17 +4,19 @@ import { FreighterApiNodeError } from "@shared/api/helpers/extensionMessaging";
 import { isBrowser } from ".";
 
 export const getAddress = async (): Promise<
-  { address: string } | { error: FreighterApiError }
+  { address: string } & { error?: FreighterApiError }
```
piyalbasu, line 7: I found it a little more ergonomic to change all the payloads to the `Payload&{error?}` instead. Otherwise, in TS, you needed to check what keys were going to be in the payload by doing `if ("error" in payload)` or `if ("address" in payload)` which I always find a little ugly.

Now you can just do `if (payload.error)` to check an error. And since the payload is always required, `payload.address` will always have a value

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -238,7 +239,7 @@ export const freighterApiMessageListener = (
         isDomainListedAllowed,
         url: tabUrl,
         flaggedKeys,
-        accountToSign,
+        accountToSign: accountToSign || addressToSign,
```
piyalbasu, line 242: this will maintain backwards compatibility with older versions of freighter-api

aristidesstaffieri (approved):

PR #1374, opened 2024-07-17
add isNonSSELEnabled flag and use that to block non-https tx's

piyalbasu:
Screenshot 2024-07-17 at 3 38 24 PM
Screenshot 2024-07-17 at 3 09 08 PM

### DIFF
--- a/extension/src/background/helpers/dataStorage.ts
+++ b/extension/src/background/helpers/dataStorage.ts
@@ -9,6 +9,7 @@ import {
   TOKEN_ID_LIST,
   ASSETS_LISTS_ID,
   IS_HASH_SIGNING_ENABLED_ID,
+  IS_NON_SSL_ENABLED_ID,
 } from "constants/localStorageTypes";
 import {
   DEFAULT_NETWORKS,
@@ -238,6 +239,17 @@ export const addIsHashSigningEnabled = async () => {
   }
 };
 
+export const addIsNonSSLEnabled = async () => {
+  const localStore = dataStorageAccess(browserLocalStorage);
+  const storageVersion = (await localStore.getItem(STORAGE_VERSION)) as string;
+
+  if (!storageVersion || semver.lt(storageVersion, "4.2.0")) {
+    // add the base asset lists
+    await localStore.setItem(IS_NON_SSL_ENABLED_ID, false);
+    await migrateDataStorageVersion("4.2.0)");
+  }
+};
+
 export const versionedMigration = async () => {
   // sequentially call migrations in order to enforce smooth schema upgrades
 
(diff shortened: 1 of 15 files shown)


extension/src/helpers/stellar.ts
```diff
@@ -128,11 +128,9 @@ export const isMuxedAccount = (publicKey: string) => publicKey.startsWith("M");
 export const isFederationAddress = (address: string) => address.includes("*");
 
 export const isMainnet = (networkDetails: NetworkDetails) => {
-  const { networkPassphrase, networkUrl } = networkDetails;
+  const { networkPassphrase } = networkDetails;
 
-  return (
-    networkPassphrase === Networks.PUBLIC && networkUrl === NETWORK_URLS.PUBLIC
-  );
+  return networkPassphrase === Networks.PUBLIC;
```
piyalbasu, line 133: Making this change because otherwise, if a user adds their own custom Mainnet connection, it won't be considered `isMainnet` because the networkUrl could be different even though their still connected to Mainnet

extension/src/popup/ducks/settings.ts
```diff
@@ -599,3 +608,8 @@ export const settingsStateSelector = createSelector(
   settingsSelector,
   (settings) => settings.settingsState,
 );
+
+export const isNonSSLEnabledSelector = createSelector(
+  settingsSelector,
+  (settings) => !isMainnet(settings.networkDetails) || settings.isNonSSLEnabled,
```
piyalbasu, line 614: talking with the devx folks and some users on discord, it seems like the SSL check is annoying for a lot of devs testing local dapps on Testnet. So, we're only going to check the setting on non-Mainnet networks

extension/src/background/helpers/dataStorage.ts
```diff
@@ -238,6 +239,17 @@ export const addIsHashSigningEnabled = async () => {
   }
 };
 
+export const addIsNonSSLEnabled = async () => {
+  const localStore = dataStorageAccess(browserLocalStorage);
+  const storageVersion = (await localStore.getItem(STORAGE_VERSION)) as string;
+
+  if (!storageVersion || semver.lt(storageVersion, "4.2.0")) {
+    // add the base asset lists
```
aristidesstaffieri, line 247: comment copy pasta?

extension/src/helpers/stellar.ts
aristidesstaffieri: ah interesting yeah I've wondered about this case before, checking the passphrase seems makes sense to me as long there isn't some way to create another network option and re-use the passphrase which I think is not possible without changing core so 👍

aristidesstaffieri (approved):

PR #1377, opened 2024-07-22
Run snapshot tests in CI

aristidesstaffieri:
What
Moves end to end tests to their own action, running in a different container in order to support snapshot tests.
Snapshot tests are triggered when test suite runs or manually
Manual option to update snapshots in run

Why
Snapshot tests were failing in Ubuntu due to OS differences in chromium.


aristidesstaffieri:
Passing snapshot https://github.com/stellar/freighter/actions/runs/10046508092

piyalbasu (approved):

PR #1392, opened 2024-07-30
Bugfix/sign msg auth entry buffer

piyalbasu:
After working with the new freighter-api for a bit, I'm starting to think it's actually best to just return the signed Buffer when you sign an auth entry or a message. It matches the previous api, so it will help with backwards compatibility. And just returning the Buffer allows users to decide whether they want to work with the Buffer or if they want to convert it to a string, using the encoding of their choice

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -1069,7 +1069,8 @@ export const SSLWarningMessage = ({ url }: { url: string }) => {
           The website <strong>{url}</strong> does not use an SSL certificate.
           For additional safety Freighter only works with websites that provide
           an SSL certificate by default. You may enable connection to domains
-          that do not use an SSL certificate in Preferences.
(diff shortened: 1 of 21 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1069,7 +1069,8 @@ export const SSLWarningMessage = ({ url }: { url: string }) => {
           The website <strong>{url}</strong> does not use an SSL certificate.
           For additional safety Freighter only works with websites that provide
           an SSL certificate by default. You may enable connection to domains
-          that do not use an SSL certificate in Preferences.
+          that do not use an SSL certificate in Settings &gt; Security &gt;
+          Advanced Settings.
```
piyalbasu, line 1073: this copy change is unrelated, but I noticed this fell of somewhere during the course of adding to this release. Adding it back now

aristidesstaffieri (approved):

PR #1393, opened 2024-07-30
[FEATURE] Tx scanning

aristidesstaffieri:
What
Adds calls to the tx-scan endpoint in order to start feeding blockaid tx data before adding UI elements triggered from results.

### DIFF
--- a/extension/src/popup/helpers/blockaid.ts
+++ b/extension/src/popup/helpers/blockaid.ts
@@ -17,6 +17,11 @@ interface BlockAidScanSiteResult {
   // ...
 }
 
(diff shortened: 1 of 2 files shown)


extension/src/popup/helpers/blockaid.ts
```diff
@@ -17,6 +17,11 @@ interface BlockAidScanSiteResult {
   // ...
 }
 
+interface BlockAidScanTxResult {
```
aristidesstaffieri, line 20: Not using the result yet, but when we add the UI pieces of this we can expand this type.

piyalbasu (approved):

PR #1405, opened 2024-08-08
scan asset during Add Asset flow and Sign classic tx flow

piyalbasu:
_What_

* Adding scan-asset check for adding assets and signing classic tx's
* Emitting an Amplitude metric every time we get a response from Blockaid

### DIFF
--- a/extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
+++ b/extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
@@ -9,10 +9,10 @@ import { AppDispatch } from "popup/App";
 import { navigateTo } from "popup/helpers/navigate";
 import { stellarSdkServer } from "@shared/api/helpers/stellarSdkServer";
 import { emitMetric } from "helpers/metrics";
-import { useNetworkFees } from "popup/helpers/useNetworkFees";
 import { getCanonicalFromAsset } from "helpers/stellar";
 import { getManageAssetXDR } from "popup/helpers/getManageAssetXDR";
 import { checkForSuspiciousAsset } from "popup/helpers/checkForSuspiciousAsset";
+import { scanAsset } from "popup/helpers/blockaid";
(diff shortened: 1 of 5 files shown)


extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
```diff
@@ -95,7 +97,6 @@ export const ManageAssetRowButton = ({
   const walletType = useSelector(hardwareWalletTypeSelector);
   const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const publicKey = useSelector(publicKeySelector);
-  const { recommendedFee } = useNetworkFees();
```
piyalbasu, line 98: small optimization here - because we render one of these components for every asset we show, we were hitting the fees endpoint many times in a short amount of time.

Changed to call the fee endpoint once in the top level and just pass it to all the child rows

extension/src/popup/helpers/blockaid.ts
```diff
@@ -109,3 +114,31 @@ export const useScanTx = () => {
     scanTx,
   };
 };
+
+export const scanAsset = async (
```
piyalbasu, line 118: I moved away from using the hook pattern you were using because I was getting a lot of infinite re-renders when using that pattern with scanAsset. Not sure if you ran into this, too? Looks like you had to disable the exhaustive-deps check on your implementations

Just exporting the `scanAsset` method allowed me to fit it into the existing flows a little more cleanly. These flows already handled loading states, so there wasn't much to gain from using the hook

aristidesstaffieri: Disabling `exhaustive-deps` is only used to have the effect trigger once on render, so in my cases the deps for the effect triggered re-renders but I only want to run the effect "on mount".

I think this does more or less the same though.

aristidesstaffieri (approved):

PR #1407, opened 2024-08-09
upgrading ledger SDK

piyalbasu:
Upgrading Ledger's deps to the latest. This will allow signing Soroban XDR's without hash signing

### DIFF
--- a/extension/src/popup/helpers/hardwareConnect.ts
+++ b/extension/src/popup/helpers/hardwareConnect.ts
@@ -4,6 +4,7 @@ import {
   MemoType,
   Operation,
   Transaction,
+  StrKey,
 } from "stellar-sdk";
 import {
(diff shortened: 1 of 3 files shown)


extension/src/popup/helpers/hardwareConnect.ts
```diff
@@ -36,7 +37,7 @@ export const createWalletConnection: CreateWalletConnection = {
     const ledgerApi = new LedgerApi(transport);
     const response = await ledgerApi.getPublicKey(bipPath);
 
-    return response.publicKey;
+    return StrKey.encodeEd25519PublicKey(response.rawPublicKey);
```
piyalbasu, line 40: the API now just returns the raw public key, so we need to encode it on our own

PR #1413, opened 2024-08-12
Bugfix/show soroswap sim rejection error

piyalbasu:
We weren't properly showing the simulation rejection error during the Soroswap flow

Screenshot 2024-08-12 at 6 08 06 PM


extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -129,29 +127,25 @@ export const Settings = ({
     }
 
     if (isSoroswap) {
-      let simulatedTx;
-      try {
-        simulatedTx = await buildAndSimulateSoroswapTx({
+      const simulatedTx = await dispatch(
```
piyalbasu, line 130: following the pattern we set with simulating soroban token transfers, which is already set up to show sim tx errors

extension/src/popup/views/Preferences/styles.scss
```diff
@@ -5,12 +5,12 @@
     display: flex;
     flex-direction: column;
     gap: 1rem;
+    margin-bottom: 2rem;
```
piyalbasu, line 8: this was throwing a sass-loader error after the dep update

aristidesstaffieri (approved):
lgtm. Imo we should later consider parsing a shorter, friendlier message to use in the error display and maybe allow copying the larger error response. I think the big blob from rpc could be hard for users to parse out.

piyalbasu:
> lgtm. Imo we should later consider parsing a shorter, friendlier message to use in the error display and maybe allow copying the larger error response. I think the big blob from rpc could be hard for users to parse out.

Yeah, totally. This is a tough error to parse, and even working through it with the Palta Labs folks, we couldn't really figure out what was actually wrong. We'll need to do some thinking about how this can actually be useful to users

PR #1415, opened 2024-08-13
fix bug with soroswapping back to xlm

piyalbasu:
Fixes a bug where soroswapping to XLM was incorrectly going down the Send Token code path

### DIFF
--- a/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
@@ -297,7 +297,7 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
             sourceAsset: sourceAsset.code,
           });
 
-          if (isSoroswap) {
+          if (isSoroswap && destAsset.issuer) {
             await dispatch(
               addTokenId({
                 publicKey,
--- a/extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
@@ -346,6 +346,7 @@ export const Settings = ({
             </View.Content>
             <View.Footer>
               <Button
+                disabled={!transactionFee}
                 size="md"
                 isFullWidth
                 onClick={goToReview}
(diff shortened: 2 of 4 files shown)


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -297,7 +297,7 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
             sourceAsset: sourceAsset.code,
           });
 
-          if (isSoroswap) {
+          if (isSoroswap && destAsset.issuer) {
```
piyalbasu, line 300: only add the token ID if there is a token ID found (in the case of native, this will be skipped)

extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -87,6 +87,28 @@ export const Settings = ({
   const showSlippage = (isPathPayment || isSwap) && !isSoroswap;
 
   async function goToReview() {
+    if (isSoroswap) {
```
piyalbasu, line 90: just moving this up to make sure we go down this path first and return if it's a soroswap

extension/src/popup/helpers/useNetworkFees.ts
```diff
@@ -15,7 +15,7 @@ export const useNetworkFees = () => {
   const { networkUrl, networkPassphrase } = useSelector(
     settingsNetworkDetailsSelector,
   );
-  const [recommendedFee, setRecommendedFee] = useState("100");
+  const [recommendedFee, setRecommendedFee] = useState("");
```
piyalbasu, line 18: @aristidesstaffieri I noticed this defaulted to "100" from your PR here: https://github.com/stellar/freighter/pull/1376/files#diff-48e5eb06e07934afbdf9c6be1ad5d87bb1658ee0c72641914788d5655df34d2dR18

Was this for testing or was this supposed to fix a bug? I noticed that I was always getting my fee set to 100 XLM even though the fee stats from Horizon were coming back lower

aristidesstaffieri: Yeah this was meant to fix a bug, I believe it was in the event that we couldn't get a fee then we would have no default but this value should be updated when the fees are fetched. 

After looking at this hook again, I think it's still flawed because there is no loading state so probably the initial value gets set and fee stats call never updates it due to the potential race condition if you confirm the screen before the fee resolves from the rpc.

piyalbasu: Oh gotcha. Cool - I'll add a fix in the next commit

aristidesstaffieri (approved):

PR #1427, opened 2024-08-19
[FEATURE] Scrub logs

aristidesstaffieri:
What
Tweaks calls to `captureException` to only use allowed identifiers
Adds Sentry hook to scrub route paths with G addresses

Why
Remove G addresses from Sentry logs

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -256,10 +256,6 @@ export const subscribeAccount = async (publicKey: string) => {
     }
   } catch (e) {
     console.error(e);
-    // Turn on when Mercury is enabled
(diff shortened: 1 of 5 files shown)


extension/src/background/helpers/account.ts
```diff
@@ -256,10 +256,6 @@ export const subscribeAccount = async (publicKey: string) => {
     }
   } catch (e) {
     console.error(e);
-    // Turn on when Mercury is enabled
```
aristidesstaffieri, line 259: We shouldn't need these, we can get these logs from our server logs and probably don't want this triggering when Mercury is turned off.

piyalbasu (approved):

PR #1434, opened 2024-08-21
[FIX] Tweaks listeners for new chrome types

aristidesstaffieri:
What
Tweaks listeners to align with new types from chrome

### DIFF
--- a/extension/src/background/index.ts
+++ b/extension/src/background/index.ts
@@ -16,6 +16,8 @@ import {
   migrateSorobanRpcUrlNetworkDetails,
   versionedMigration,
 } from "./helpers/dataStorage";
+import { ExternalRequest, Response } from "@shared/api/types";
+import { isExternalMessage, isResponse } from "./helpers/message";
 
 export const initContentScriptMessageListener = () => {
   browser?.runtime?.onMessage?.addListener((message) => {
@@ -24,6 +26,7 @@ export const initContentScriptMessageListener = () => {
         file: "contentScript.min.js",
       });
     }
+    return undefined;
   });
 };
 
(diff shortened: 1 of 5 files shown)


extension/src/background/index.ts
```diff
@@ -24,25 +26,30 @@ export const initContentScriptMessageListener = () => {
         file: "contentScript.min.js",
       });
     }
+    return undefined;
```
aristidesstaffieri, line 29: looks like we should be returning `true | undefined` here, returning true means that there is an async response expected so I believe in our case `undefined` is the correct return value.

extension/src/background/index.ts
```diff
@@ -24,25 +26,30 @@ export const initContentScriptMessageListener = () => {
...
   });
 };
 
 export const initExtensionMessageListener = () => {
   browser?.runtime?.onMessage?.addListener(async (request, sender) => {
     const sessionStore = await buildStore();
     // todo this is kinda ugly
+    const req = request as ExternalRequest | Response;
```
aristidesstaffieri, line 37: Since they now call `request` unknown we have to cast the expected value.

extension/src/background/index.ts
piyalbasu: We do `await` a response from the content script, so I think maybe `true` would be correct here?

https://github.com/stellar/freighter/blob/master/%40shared/api/external.ts#L15

aristidesstaffieri: I _think_ they mean that in the case where you got an event in your listener, and that listener's callback plans to `sendResponse` async then you should return `true` to signal that.
https://stackoverflow.com/questions/70353944/chrome-runtime-onmessage-returns-undefined-even-when-value-is-known-for-asynch

PR #1435, opened 2024-08-21
fix types

piyalbasu:


extension/src/background/helpers/dataStorage.ts
```diff
@@ -28,7 +28,9 @@ export const SESSION_STORAGE_ENABLED = true;
 
 export const normalizeMigratedData = async () => {
```
piyalbasu, line 29: this is a helper that really shouldn't be running anymore. This was used to migrate data from localStorage to chrome.storage. Apparently there was an error in the code here when we switched over to v3, but we never noticed because if a user is on v3, this function returns early on ln 39

aristidesstaffieri (approved):

PR #1436, opened 2024-08-21
fix check for message listeners

piyalbasu:


extension/src/background/helpers/dataStorage.ts
```diff
@@ -247,7 +247,7 @@ export const addIsNonSSLEnabled = async () => {
 
   if (!storageVersion || semver.lt(storageVersion, "4.2.0")) {
     await localStore.setItem(IS_NON_SSL_ENABLED_ID, false);
-    await migrateDataStorageVersion("4.2.0)");
+    await migrateDataStorageVersion("4.2.0");
```
piyalbasu, line 250: noticed this issue with migrations, so adding the fix while i'm here

extension/src/background/index.ts
```diff
@@ -36,16 +35,22 @@ export const initExtensionMessageListener = () => {
     // todo this is kinda ugly
     const req = request as ExternalRequest | Response;
     let res;
-    if (isResponse(req) && Object.values(SERVICE_TYPES).includes(req.type)) {
```
piyalbasu, line 39: the isResponse vs isExternalMessage checks didn't quite work because they were keying off of `messageId`, which actually is only an optional field on the type `Response`. We only include it on requests that come through freighter-api and requests that come through localhost when you're in fullscreen mode in dev.

Unfortunately this meant requests that were coming from the extension itself weren't making it to the appropriate handler.

The type casting is a little ugly, but accomplishes routing the requests to the appropriate handler. The comment `// todo this is kinda ugly` definitely still stands. I'll take another shot at cleaning this up in the next release

aristidesstaffieri (approved):

PR #1446, opened 2024-08-27
[CHORE] adds fetchJson to handle non json fetch responses

aristidesstaffieri:
What
Adds `fetchJson` to handle non json response types and general error handling in fetch calls

Why
We have an error that is hard to trace due to the text response handling

### DIFF
--- a/extension/src/popup/helpers/fetch.ts
+++ b/extension/src/popup/helpers/fetch.ts
@@ -1,0 +1,14 @@
+export const fetchJson = async <T>(url: string, options?: RequestInit) => {
+  const res = await fetch(url, options);
+  if (!res.ok) {
+    throw new Error(res.statusText);
+  }
+
+  if (res.headers.get("content-type") !== "application/json") {
+    const content = await res.text();
+    throw new Error(content);
+  }
+
+  const data = res.json() as T;
+  return data;
+};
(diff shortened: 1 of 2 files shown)


extension/src/popup/helpers/fetch.ts
```diff
@@ -0,0 +1,14 @@
+export const fetchJson = async <T>(url: string, options?: RequestInit) => {
```
aristidesstaffieri, line 1: We can probably use this everywhere that we use `fetch` now but I wanted to get this out first to make sure it makes to the upcoming patch.

extension/src/popup/helpers/fetch.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (!res.ok) {
+    throw new Error(res.statusText);
+  }
+
+  if (res.headers.get("content-type") !== "application/json") {
+    const content = await res.text();
+    throw new Error(content);
+  }
+
+  const data = res.json() as T;
```
piyalbasu, line 12: should this be `await res.json()`?

piyalbasu (approved):

extension/src/popup/helpers/fetch.ts
aristidesstaffieri: hey thanks! fixed in 5be529182068ae3c9a37ae0616fb7e5a16e4998b

PR #1449, opened 2024-08-27
[FEATURE] malicious tx warning, sign-tx

aristidesstaffieri:
What
Adds UI warning when blockaid transaction scan results in malicious tag

Screenshot 2024-08-27 at 1 20 03 PM


aristidesstaffieri:
The UI is still the old warning style but will be in the new style and match the figma after the SDS work lands.

piyalbasu (approved):

PR #1451, opened 2024-08-28
[BUG] SAC payment displayed as generic host fn

aristidesstaffieri:
What
Adds a tokenKey check using the derived SAC address when processing transfers in account history
Updates history item detail in transfer cases to use same display as classic side payments

Why
SAC payments are shown as `InvokeHostFunction` because we fail to find a key for the balance in the SAC case due to the balances being stored under keys that are `{assetCode}:{issuer}` where the issuer is a G address for classic assets.

Also, SEP41/SAC transfers can be shown as payments to avoid confusion.

Result: (Top 2 are classic payments, next two are a SAC transfer and SEP41 transfer)
Screenshot 2024-08-28 at 3 24 15 PM

### DIFF
--- a/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
+++ b/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
@@ -7,6 +7,7 @@ import camelCase from "lodash/camelCase";
 import { Icon, Loader } from "@stellar/design-system";
 import { BigNumber } from "bignumber.js";
 import { useTranslation } from "react-i18next";
+import { Asset } from "stellar-sdk";
 
 import { OPERATION_TYPES } from "constants/transaction";
 import { SorobanTokenInterface } from "@shared/constants/soroban/token";
@@ -16,6 +17,7 @@ import { emitMetric } from "helpers/metrics";
 import {
   formatTokenAmount,
   getAttrsFromSorobanHorizonOp,
+  isContractId,
 } from "popup/helpers/soroban";
 import { formatAmount } from "popup/helpers/formatters";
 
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -166,7 +168,7 @@ export const HistoryItem = ({
         setBodyComponent(
           <>
             {paymentDifference}
-            {formatAmount(new BigNumber(amount).toFixed(2, 1))} {destAssetCode}
```
aristidesstaffieri, line 169: So right now, if you send "1 XLM" using classic payments in Freighter, that item in history will show up as `Sent 1.00 XLM`.
This would differ to the formatting that we're doing in token cases, where the SAC payment would show up as `Sent 1 XLM`.
I'm curious why we round to the 2 place decimal in the classic payment case?
I would think that we would want to show the decimals and maybe remove trailing zeros for readability(what we do for tokens now).
@piyalbasu

piyalbasu: I think this was an oversight as we added `.toFixed` to cut strings like `1.91111` to `1.91`. But we should remove trailing zeroes if applicable here

piyalbasu (approved):

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -138,6 +138,8 @@ export const HistoryItem = ({
 
   const renderBodyComponent = () => BodyComponent;
   const renderIcon = () => IconComponent;
+  /* eslint-disable react-hooks/exhaustive-deps */
```
aristidesstaffieri, line 141: @piyalbasu maybe I'm missing something but this rule doesn't seem right. It wants to include `t` as a dep of the `useCallback` hook here but really we want an empty dep array. 
We want a memoized version of `t` for the whole lifecycle of the component since it never changes and we need it as a dep of useEffect, but adding `t` as a dep of the callback  or effect causes an infinite loop in the tests.
https://react.dev/reference/react/useCallback#parameters

Not sure why we don't see an infinite loop when actually running the component and only in Jest though 🤔

PR #1452, opened 2024-08-28
Feature/blockaid asset UI warning

piyalbasu:
Sorry for the lengthy PR, but as you can see from the screenshots below, the asset warnings touch a LOT of different views. It was a bit like whack-a-mole trying to catch all the different spots. I'm working on a PR to add tests for all these scenarios because there is a lot of them

Screenshot 2024-08-28 at 6 11 44 PM
Screenshot 2024-08-28 at 7 39 41 PM
Screenshot 2024-08-28 at 7 27 17 PM
Screenshot 2024-08-28 at 7 26 56 PM
Screenshot 2024-08-28 at 6 53 02 PM
Screenshot 2024-08-28 at 6 52 43 PM
Screenshot 2024-08-28 at 6 12 40 PM
Screenshot 2024-08-28 at 6 12 23 PM
Screenshot 2024-08-28 at 6 12 14 PM
Screenshot 2024-08-28 at 6 12 05 PM


extension/src/popup/components/ModalInfo/index.tsx
```diff
@@ -1,42 +1,77 @@
...
+        <div>{t("Add Asset trustlinet")}</div>
+      </>
+    );
+  }
+
+  return (
+    <>
+      <Icon.Link />
+      <div>{t("Connection Request")}</div>
+    </>
```
piyalbasu, line 42: I think this will end up colliding a bit with the domain check in this screen: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=2051-33657&node-type=FRAME&m=dev

But we can iron that out

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -70,18 +73,40 @@ export const AssetDetail = ({
...
   const isSorobanAsset = canonical.issuer && isSorobanIssuer(canonical.issuer);
-  const isOwnedScamAsset = useIsOwnedScamAsset(
-    canonical.code,
-    canonical.issuer,
-  );
 
   const { accountBalances: balances } = useSelector(
     transactionSubmissionSelector,
   );
+  const { scannedAsset } = useScanAsset(
```
aristidesstaffieri, line 85: Does this need a loading state?
It seems like since the hook does an async thing to set `scannedAsset`, without the loading state won't `scannedAsset` only ever get assigned to the initial value in the hook?

aristidesstaffieri: reviewed hook rules and see now that this value will get updated and cause a re-render, so never mind I think probably it does make sense to not represent the loading state.

aristidesstaffieri (approved):

extension/src/popup/components/account/AssetDetail/index.tsx
piyalbasu: Yup, it should be triggering a re-render on value update

PR #1453, opened 2024-08-30
[BUG]Allow fee config on token payment

aristidesstaffieri:
What
Removes constrain for token transfers on fee configuration in send settings

Why
Users should be able to configure fees when sending tokens as well as classic payments


extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -190,52 +190,50 @@ export const Settings = ({
           <Form className="View__contentAndFooterWrapper">
             <View.Content>
               <FormRows>
-                {!isToken ? (
```
aristidesstaffieri, line 193: I haven't been able to track down the original reason for doing this but this seems to work as is when removing this constraint.

piyalbasu (approved):

PR #1454, opened 2024-08-30
add Blockaid tests for account, manage assets, send and swap flows

piyalbasu:


extension/src/popup/views/__tests__/Account.test.tsx
```diff
@@ -136,21 +144,28 @@ describe("Account view", () => {
...
+        screen.getByTestId("AccountAssets__asset--loading-XLM"),
+      ).not.toContainElement(screen.getByTestId("ScamAssetIcon"));
+      expect(
+        screen.getByTestId("AccountAssets__asset--loading-USDC"),
+      ).toContainElement(screen.getByTestId("ScamAssetIcon"));
       expect(screen.getAllByText("USDC")).toBeDefined();
     });
   });
-  it.skip("goes to account details", async () => {
+  it("goes to account details", async () => {
```
piyalbasu, line 168: i was able to fix the issue with this test so I'm re-enabling 🎉

extension/src/popup/views/__tests__/SendPayment.test.tsx
```diff
@@ -68,7 +89,7 @@ jest.mock("popup/constants/history", () => ({
 
 const publicKey = "GA4UFF2WJM7KHHG4R5D5D2MZQ6FWMDOSVITVF7C5OLD5NFP6RBBW2FGV";
 
-describe.skip("SendPayment", () => {
+describe("SendPayment", () => {
```
piyalbasu, line 92: I was also able to fix and re-enable this suite of tests

extension/src/popup/views/__tests__/Swap.test.tsx
```diff
@@ -51,7 +49,40 @@ export const swapMockBalances = {
       total: new BigNumber("333"),
       available: new BigNumber("333"),
     },
-  } as any as Balances,
```
piyalbasu, line 54: I removed the typecasting to simplify as it doesn't provide a ton of value in a test imo

extension/src/popup/views/__tests__/Swap.test.tsx
```diff
@@ -138,44 +173,96 @@ describe("Swap", () => {
       } as any),
     );
   });
-  beforeEach(async () => {
-    const history = createMemoryHistory();
-    history.push(ROUTES.swap);
-    mockHistoryGetter.mockReturnValue(history);
-
-    act(() => {
```
piyalbasu, line 146: unfortunately doing this render in the `beforeEach` made it impossible to overwrite `jest
      .spyOn(ApiInternal, "getAccountIndexerBalances")` with different mock balances in individual tests. `render` in each test is a little more verbose, but it's easier to change data for each test

aristidesstaffieri (approved):

PR #1460, opened 2024-09-03
qa fixes for blockaid asset check

piyalbasu:


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -244,8 +244,11 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
       (tag) => tag === TRANSACTION_WARNING.memoRequired && !memo,
     );
 
-  const isAssetMalicious =
-    accountBalances.balances?.[asset].isMalicious || false;
+  let isAssetMalicious = accountBalances.balances?.[asset].isMalicious || false;
+  if (!isAssetMalicious && destinationAsset) {
+    isAssetMalicious =
+      accountBalances.balances?.[destinationAsset].isMalicious || false;
+  }
```
piyalbasu, line 251: This makes sure to check the destination asset (in a swap) to see if the user is swapping to a malicious asset

extension/src/popup/views/Preferences/index.tsx
```diff
@@ -74,7 +74,7 @@ export const Preferences = () => {
               <AutoSaveFields />
               <div className="Preferences--section">
                 <div className="Preferences--section--title">
-                  {t("Verification with")} stellar.expert
+                  {t("Verification with")} stellar.expert & blockaid.io
```
piyalbasu, line 77: Since we're using both stellar.expert and blockaid, I think it's appropriate to add this to the settings field. I'll have product weigh in

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -144,6 +144,12 @@ export const SignTransaction = () => {
   const isMemoRequired = flaggedKeyValues.some(
     ({ tags }) => tags.includes(TRANSACTION_WARNING.memoRequired) && !memo,
   );
+  const isUnsafe = flaggedKeyValues.some(({ tags }) =>
+    tags.includes(TRANSACTION_WARNING.unsafe),
+  );
+  const isMaliciousAddress = flaggedKeyValues.some(({ tags }) =>
+    tags.includes(TRANSACTION_WARNING.malicious),
+  );
```
piyalbasu, line 152: I mistakenly removed this in an earlier PR thinking we'd use Blockaid for this use case. These flags are actually if the address if malicious (not the asset). Since Blockaid doesn't offer address scanning for Stellar yet, we should continue using stellar.expert to scan malicious addresses

PR #1463, opened 2024-09-04
[FEATURE] adds tx scan during send payment flow tx details

aristidesstaffieri:
What?
Adds a `scanTx` to the payment flow

Why?
Scanning payment transactions could flag tokens or accounts involved in the payment tx.

### DIFF
--- a/extension/src/popup/components/ModalInfo/index.tsx
+++ b/extension/src/popup/components/ModalInfo/index.tsx
@@ -31,7 +31,7 @@ const PillCopy = ({ pillType }: PillyCopyProps) => {
     return (
       <>
         <img src={IconShieldPlus} alt="Add trustline icon" />
-        <div>{t("Add Asset trustlinet")}</div>
+        <div>{t("Add Asset trustline")}</div>
       </>
     );
   }
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -1147,7 +1147,7 @@ export const BlockAidSiteScanLabel = ({
   return <BlockAidBenignLabel />;
 };
 
-export const BlockaidMaliciousTxWarning = ({
+export const BlockaidMaliciousTxSignWarning = ({
   type,
 }: {
   type: BlockAidScanTxResult["validation"]["result_type"];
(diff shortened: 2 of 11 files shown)


extension/src/popup/components/ModalInfo/index.tsx
```diff
@@ -31,7 +31,7 @@ const PillCopy = ({ pillType }: PillyCopyProps) => {
     return (
       <>
         <img src={IconShieldPlus} alt="Add trustline icon" />
-        <div>{t("Add Asset trustlinet")}</div>
+        <div>{t("Add Asset trustline")}</div>
```
aristidesstaffieri, line 34: random typo I found 🤷‍♂️

extension/src/popup/helpers/fetch.ts
```diff
@@ -4,7 +4,7 @@ export const fetchJson = async <T>(url: string, options?: RequestInit) => {
     throw new Error(res.statusText);
   }
 
-  if (res.headers.get("content-type") !== "application/json") {
+  if (!res.headers.get("content-type")?.includes("application/json")) {
```
aristidesstaffieri, line 7: Seems like a fastify version tweaked this header slightly(added the charset) so maybe it's better to try to be looser with the content type here?

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1175,3 +1175,38 @@ export const BlockaidMaliciousTxWarning = ({
...
+          src={IconWarningBlockaid}
+          alt="icon warning blockaid"
+        />
+      </div>
+      <div>
+        <div className="ScamAssetWarning__description">
+          {t(
+            "This transaction was flagged by Blockaid for the following reasons:",
+          )}
+          <div>{description}</div>
```
aristidesstaffieri, line 1200: will take the description straight from Blockaid, so it can be something like
`Gaining account GA223OFHVKVAH2NBXP4AURJRVJTSOVHGBMKJNL6GRJWNN4SARVGSITYG is classified as malicious_link`

extension/src/popup/helpers/fetch.ts
piyalbasu: Oh interesting - looks like we only use this in `useScanTx` (so far), so should be safe to change this

piyalbasu (approved):

PR #1469, opened 2024-09-05
Bugfix/show non malicious blockaid info

piyalbasu:
Showing more info from Blockaid for all non-benign assets
Screenshot 2024-09-05 at 6 01 43 PM

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -300,9 +300,7 @@ export const BlockaidAssetWarning = ({
       <div>
         <div className="ScamAssetWarning__description">
           {t(
-            `This token was flagged as ${
-              scannedAssetData.result_type || "malicious"
-            } by Blockaid. Interacting with this token may result in loss of funds and is not recommended for the following reasons`,
+            `This token was flagged as ${scannedAssetData.result_type} by Blockaid. Interacting with this token may result in loss of funds and is not recommended for the following reasons`,
           )}
           :
           <ul className="ScamAssetWarning__list">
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -283,12 +300,14 @@ export const BlockaidAssetWarning = ({
...
+            `This token was flagged as ${
+              scannedAssetData.result_type || "malicious"
+            } by Blockaid. Interacting with this token may result in loss of funds and is not recommended for the following reasons`,
           )}
           :
           <ul className="ScamAssetWarning__list">
-            <li>{t("Identified as a scam")}</li>
-
+            {scannedAssetData.features &&
+              scannedAssetData.features.map((f) => <li>{f.description}</li>)}
```
piyalbasu, line 310: grab the `features` key from the Blockaid response and show the information here

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -283,12 +300,14 @@ export const BlockaidAssetWarning = ({
       <div>
         <div className="ScamAssetWarning__description">
           {t(
-            "This token was flagged as malicious by Blockaid. Interacting with this token may result in loss of funds and is not recommended for the following reasons",
+            `This token was flagged as ${
```
aristidesstaffieri, line 303: Does this text show up for Benign scans as well?
Also I don't think `result_type` is optional so not sure if we need the fallback case to `malicious` here.

piyalbasu: This does not show up for Benign because we check for non-benign before we instantiate this component

piyalbasu: Yeah, good point about `result_type` not being optional. I'm probably being overly cautious here

aristidesstaffieri (approved):

PR #1475, opened 2024-09-10
add better typing for standalone balances

piyalbasu:
Motivation:

I added `blockaidData` to the BE's balances which caused the balances we get from the BE to be out of sync with the balances we get using a custom network. This should have raised type errors throughout the app, but it did not because we used a lot of casting as `any` with balances to work our way out of some tricky type issues.

I'm removing the use of `any` when dealing with balances so we have a strongly typed (and more predictable) format for balances


@shared/api/internal.ts
```diff
@@ -556,20 +556,20 @@ export const getAccountBalancesStandalone = async ({
     // eslint-disable-next-line no-plusplus
     for (let i = 0; i < Object.keys(resp.balances).length; i++) {
       const k = Object.keys(resp.balances)[i];
-      const v: any = resp.balances[k];
-      if (v.liquidity_pool_id) {
+      const v = resp.balances[k];
+      if (v.liquidityPoolId) {
```
piyalbasu, line 560: fixed this key to be camelcase rather than snake

@shared/helpers/stellar.ts
```diff
@@ -42,93 +46,99 @@ export function getBalanceIdentifier(
...
+};
+
 export function makeDisplayableBalances(
   accountDetails: StellarSdk.Horizon.ServerApi.AccountRecord,
-): BalanceMap {
+) {
   const { balances, subentry_count, num_sponsored, num_sponsoring } =
     accountDetails;
 
-  const displayableBalances = Object.values(balances).reduce(
```
piyalbasu, line 51: using `reduce` made it hard for TS to typecheck the different balance types. Switching to a for loop for more clarity.

Also, at some point `balances` must have changed from an object to an array. No need to do this `Object.values` conversion

@shared/api/internal.ts
```diff
@@ -556,20 +556,19 @@ export const getAccountBalancesStandalone = async ({
...
-          .liquidityPoolId(v.liquidity_pool_id)
+          .liquidityPoolId(v.liquidityPoolId)
           .call();
         balances[k] = {
           ...balances[k],
-          liquidityPoolId: v.liquidity_pool_id,
+          liquidityPoolId: v.liquidityPoolId,
           reserves: lp.reserves,
         };
-        delete balances[k].liquidity_pool_id;
```
piyalbasu, line 572: initially, we assigned `liquidity_pool_id`'s value to `liquidityPoolId` and then deleted the old `liquidity_pool_id`. I'm just simplifying the logic in `makeDisplayableBalances` so we don't need to do this anymore

aristidesstaffieri (approved):

PR #1476, opened 2024-09-12
add blockaid scan for custom mainnet and use blockaid lib for types

piyalbasu:
We should make sure that users who are using Mainnet with a custom network (a common occurrence with people wanting to use their rpc/horizon instances) still get the benefit of Blockaid asset scanning

### DIFF
--- a/@shared/api/package.json
+++ b/@shared/api/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "@blockaid/client": "^0.25.0",
     "@stellar/js-xdr": "^3.1.1",
     "bignumber.js": "^9.1.1",
     "prettier": "^2.0.5",
--- a/@shared/api/types.ts
+++ b/@shared/api/types.ts
@@ -1,5 +1,6 @@
 import BigNumber from "bignumber.js";
 import { AssetType as SdkAssetType, Horizon } from "stellar-sdk";
+import Blockaid from "@blockaid/client";
 
 import { SERVICE_TYPES, EXTERNAL_SERVICE_TYPES } from "../constants/services";
 import { APPLICATION_STATE } from "../constants/applicationState";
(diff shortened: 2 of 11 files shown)


@shared/api/package.json
```diff
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "@blockaid/client": "^0.25.0",
```
piyalbasu, line 7: I'm adding the lib as a dep strictly for the types. I thought it'd be nice to not have to worry about updating types with subsequent releases. The cost here is the added bundle size, but imo not a major concern with an extension

@shared/api/types.ts
```diff
@@ -245,10 +246,9 @@ export interface Balance {
   blockaidData: BlockAidScanAssetResult;
 }
 
-export interface BlockAidScanAssetResult {
-  result_type: "Benign" | "Warning" | "Malicious" | "Spam";
-  features: { description: string }[];
-}
+export type BlockAidScanAssetResult = Blockaid.TokenScanResponse;
+export type BlockAidScanSiteResult = Blockaid.SiteScanResponse;
+export type BlockAidScanTxResult = Blockaid.StellarTransactionScanResponse;
```
piyalbasu, line 251: I'm exporting all the types from this file so we only have to worry about keeping "@blockaid/client" updated in one subrepo

extension/src/popup/views/GrantAccess/index.tsx
```diff
@@ -48,6 +48,9 @@ export const GrantAccess = () => {
     window.close();
   };
 
+  const isMalicious =
+    data && "is_malicious" in data ? data.is_malicious : false;
```
piyalbasu, line 52: was getting a type error with the new type as "is_malicious" doesn't exist when Blockaid returns a miss

extension/src/popup/views/__tests__/Account.test.tsx
```diff
@@ -168,6 +256,85 @@ describe("Account view", () => {
       expect(screen.getAllByText("USDC")).toBeDefined();
     });
   });
+  it("displays balances and scam notifications on custom Mainnet network", async () => {
```
piyalbasu, line 259: these tests also give us a little cover on the custom network experience, which we sorely need

aristidesstaffieri: nice, thanks

PR #1479, opened 2024-09-16
[BUG] send dest suspicious check

aristidesstaffieri:
What
Makes `isAssetSuspicous` stricter

Why
In some "send" scenarios, this check would return false positive because of the dest asset response


piyalbasu (approved):
Nice! Yeah, I think returning `true` on `{}` was the real issue here

PR #1483, opened 2024-09-17
Feature/rm stellar.expert safety checks

piyalbasu:
Removing stellar.expert safety checks because Blockaid uses stellar.expert's data to inform their own warnings


extension/src/background/constants/apiUrls.ts
```diff
@@ -1,5 +1,2 @@
 export const STELLAR_EXPERT_BLOCKED_ACCOUNTS_URL =
```
piyalbasu, line 1: we're still going to use stellar.expert to check for `memo required` accounts to warn users if an xdr will require a memo.

This var is a misnomer now - I'm going to rename it to `STELLAR_EXPERT_MEMO_REQUIRED_ACCOUNTS_URL` in the next PR. Holding off because this PR is already pretty large

@shared/api/internal.ts
```diff
@@ -1356,13 +1348,6 @@ export const loadSettings = (): Promise<
     type: SERVICE_TYPES.LOAD_SETTINGS,
   });
 
-export const getBlockedDomains = async () => {
-  const resp = await sendMessageToBackground({
```
piyalbasu, line 1360: blockedDomains were used to check the domain of an asset against a list of malicious domains. Blockaid will handle this check now

@shared/api/internal.ts
```diff
@@ -1356,13 +1348,6 @@ export const loadSettings = (): Promise<
...
-    type: SERVICE_TYPES.GET_BLOCKED_DOMAINS,
-  });
-  return resp;
-};
-
 export const getBlockedAccounts = async () => {
```
piyalbasu, line 1351: We'll continue using `blockedAccounts`, but this will be renamed. More info below

extension/src/background/constants/apiUrls.ts
piyalbasu, line 1: we're still going to use stellar.expert to check for memo required accounts to warn users if an xdr will require a memo.

This var is a misnomer now - I'm going to rename it to STELLAR_EXPERT_MEMO_REQUIRED_ACCOUNTS_URL in the next PR. Holding off because this PR is already pretty large

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -244,46 +231,13 @@ export const BackupPhraseWarningMessage = () => {
...
-      );
-      const resp = await checkForSuspiciousAsset({
-        code: assetCode,
-        issuer: assetIssuer,
-        domain: "",
-        server,
-        networkDetails,
-      });
-
-      setIsNewAssetState(resp.isNewAsset);
```
piyalbasu, line 280: we don't need to manually check for new assets with stellar.expert - blockaid handles this now

extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
```diff
@@ -217,11 +217,10 @@ export const ManageAssetRowButton = ({
       setSuspiciousAssetData({
         ...assetRowData,
         blockaidData: scannedAsset,
-        isNewAsset: resp.isNewAsset,
       });
     } else if (
       !isTrustlineActive &&
-      (resp.isInvalidDomain || resp.isRevocable || resp.isNewAsset)
+      (resp.isInvalidDomain || resp.isRevocable)
```
piyalbasu, line 223: we'll still be checking for invalid domains and revocable assets using Horizon

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -183,33 +182,23 @@ export const freighterApiMessageListener = (
...
+      if (isValidatingMemo) {
         _operations.forEach((operation: StellarSdk.Operation) => {
           accountData.forEach(
             ({ address, tags }: { address: string; tags: string[] }) => {
               if (
                 "destination" in operation &&
                 address === operation.destination
               ) {
-                let collectedTags = [...tags];
+                const collectedTags = [...tags];
```
aristidesstaffieri, line 194: If this no longer gets mutated, then is it just `tags` now?

piyalbasu: It can still be mutated because we're filtering if `!isValidatingMemo` on ln 197. It's good practice to not mutate a passed param

aristidesstaffieri: hmm that seems like a bug on 197 then, filter does not mutate the array you are filtering, it makes a shallow copy so on 197 we're not assigning the result of that filter to anything and `collectedTags` is not mutated there.

piyalbasu: Oh man - you're right. Let me take a closer look 👀

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -318,12 +316,7 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
     accountBalances.balances?.[asset].blockaidData,
   );
 
-  const isTxMalicious =
-    (isValidatingSafety && isSourceAssetSuspicious) || false;
-  const isUnsafe =
-    isValidatingSafety &&
-    matchingBlockedTags.some((tag) => tag === TRANSACTION_WARNING.unsafe);
-  const isSubmitDisabled = isMemoRequired || isTxMalicious || isUnsafe;
+  const isSubmitDisabled = isMemoRequired || isSourceAssetSuspicious;
```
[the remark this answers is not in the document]
piyalbasu: Yup - good catch

extension/src/background/messageListener/freighterApiMessageListener.ts
piyalbasu: Addressed in the latest commit. Looks like we had been reassigning the var correctly for the other safety checks, but not for memo required 🤦

aristidesstaffieri: Thanks, lgtm!

PR #1485, opened 2024-09-18
Rename memo required

piyalbasu:
* Renames "blocked accounts" to be "memo required accounts" to make the code more clear
* Adds a migration to remove now unneeded cached stellar.expert payloads
* Removes the Blockaid welcome modal


extension/src/background/helpers/dataStorage.ts
```diff
@@ -252,12 +252,17 @@ export const addIsNonSSLEnabled = async () => {
   }
 };
 
-export const addIsBlockaidAnnounced = async () => {
+export const removeStellarExpertData = async () => {
```
piyalbasu, line 255: since we're going to launch Blockaid silently - I'm removing the migration needed to show the modal. I'm lumping it into this PR because I wanted to add a cleanup migration and I wanted to use the migration version number I had previously added for the Blockaid announcement

extension/src/popup/views/Account/index.tsx
```diff
@@ -53,7 +53,7 @@ import { AccountHeader } from "popup/components/account/AccountHeader";
 import { AssetDetail } from "popup/components/account/AssetDetail";
 import { Loading } from "popup/components/Loading";
 import { NotFundedMessage } from "popup/components/account/NotFundedMessage";
-import { BlockaidAnnouncement } from "popup/components/account/BlockaidAnnouncement";
+// import { BlockaidAnnouncement } from "popup/components/account/BlockaidAnnouncement";
```
piyalbasu, line 56: commenting out so we can quickly turn this back on when we're ready

aristidesstaffieri (approved):

PR #1486, opened 2024-09-19
better loading states while loading blockaid data

piyalbasu:


extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
```diff
@@ -129,9 +129,6 @@ export const ManageAssetRowButton = ({
         }),
```
piyalbasu, line 129: these changes do a better job of showing a loading state on the asset row's button while Blockaid is scanning the asset

aristidesstaffieri (approved):

PR #1498, opened 2024-09-26
Protocol 22 upgrade

aristidesstaffieri:
Depends on https://github.com/stellar/js-stellar-base/pull/777

What
Adds handling and display components for create contract v2 constructor arguments.
Upgrades stellar-sdk to p22 compatible RC

Why
Protocol 22 will be released soon and will add this update to contract creation.

Screenshot 2024-10-11 at 1 46 45 PM


extension/src/popup/helpers/soroban.ts
```diff
@@ -508,3 +508,21 @@ function getInvocationArgs(
     }
   }
 }
+
+export const getCreateContractArgs = (hostFn: xdr.HostFunction) => {
+  if (
+    hostFn.switch() !== xdr.HostFunctionType.hostFunctionTypeCreateContractV2()
+  ) {
```
Shaptic, line 515: I would recommend just using an empty constructor argument list in this case if possible, ie don’t allow creation w/o ctor path.

aristidesstaffieri: well, so this helper mainly helps with the type assertion. Otherwise I would have to do [2 inline assertions in the function that renders the args.](https://github.com/stellar/freighter/blob/ddccf105290f10408a59d0408fb3019ca69cb055/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx#L477).

It seems cleaner to me to have one assertion helper than to have 2 inline assertions in order to use am empty array for the v1 case. What do you think?

aristidesstaffieri: oh wait I see, you're saying return an empty array in my helper. Sure yeah I think that makes enough sense, will add that.

aristidesstaffieri: actually after looking at it again, I do think it makes more sense to leave as is because our `KeyValueInvokeHostFnArgs` component that is used to render keyvals is not meant to be called with an empty `args` array. It uses react internal hooks, so to step around it we would have to run the hooks first and then render an empty params list(hooks in same order rule).

There's no reason to render this component in v1 cases since it would just show a title that says "Parameters" with no params in every case, and as it is, it would also fetch the contract spec in these cases even though we don't need it.

It seems cleaner to me to stick closer to the xdr and not render the params component in v1 cases, what do you think @Shaptic ?

Shaptic: Ah yeah I think I misunderstood - this is for rendering the invocation to the user, right? So you don't really have control over the invocation itself. You aren't _building_ said invocation, so my comment doesn't make any sense. Carry on :facepalm:

aristidesstaffieri: that's exactly it, this is a display rendering flow but I do appreciate the review!

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -471,16 +472,19 @@ export const KeyValueInvokeHostFn = ({
...
   function renderDetails() {
     switch (hostfn.switch()) {
+      case xdr.HostFunctionType.hostFunctionTypeCreateContractV2():
       case xdr.HostFunctionType.hostFunctionTypeCreateContract(): {
-        const createContractArgs = hostfn.createContract();
-        const preimage = createContractArgs.contractIdPreimage();
-        const executable = createContractArgs.executable();
+        const createContractArgs = getCreateContractArgs(hostfn);
+        const preimage = createContractArgs.contractIdPreimage;
+        const executable = createContractArgs.executable;
```
piyalbasu, line 479: noticing that we're not calling contractIdPreimage() and createContractArgs.executable() anymore. Is this an API change from the SDK? Wondering if this is a code smell that users who don't upgrade Freighter immediately when P22 hits might have issues

piyalbasu:
Could we please add a test for this change? Thanks!

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
aristidesstaffieri: no it's not an API change, I just added `getCreateContractArgs` helper that does the unpacking of all of those and the constructor args so they're just not called there anymore.

aristidesstaffieri:
> Could we please add a test for this change? Thanks!

added in db801d8ac40f5d60e16e040c2c7e939a6900059b

piyalbasu (approved):

PR #1503, opened 2024-10-01
redesign popup blockaid warnings

piyalbasu:
Screenshot 2024-10-01 at 3 57 40 PM
Screenshot 2024-10-01 at 3 57 55 PM
Screenshot 2024-10-01 at 3 57 47 PM


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1146,6 +1185,142 @@ export const BlockaidTxScanLabel = ({
...
+        <Icon.ChevronRight className="WarningMessage__link-icon" />
+      </div>
+    </WarningMessageHeader>
+  );
+
+  const truncatedDescription = description.map((d) => {
+    const arr = d.split(" ");
+
+    const truncatedArr = arr.map((word) => {
+      if (word.length > 30) {
```
piyalbasu, line 1249: this is meant to truncate any public keys/contract id's as they come in at full length from Blockaid, but look pretty ugly inside such a small modal. A long string either breaks outside of the modal or would need to break on to 2 lines

aristidesstaffieri: Is the user still able to get the full pub key or contract ID when this truncation happens?

Also, I think this is very unlikely(seems like there's only [a few real English words](https://en.wikipedia.org/wiki/Longest_word_in_English) that are even that long) but should we explicitly check for a word being a contract ID or public key in order to avoid accidentally truncating other strings?

piyalbasu: This is a good point about getting the full pub key/contract id. Let me thinking about how to handle that 🤔 Maybe I can add a float over tooltip that shows the whole string

I started to go down the route of checking for contract ID and public key, but was a little concerned my logic might be too strict. Since this description can be any string, I thought it might be safer to be a little looser with what I'm checking for since really any long word will break the UI

aristidesstaffieri: sounds good, yeah I think that's likely fine. I don't expect that we would actually hit this but wanted to note since it would be weird if it did happen.

aristidesstaffieri: 👏 nice, thanks!

aristidesstaffieri (approved):

PR #1505, opened 2024-10-03
[FIX] Update E2E test token and related tests

aristidesstaffieri:
What
Updates the E2E test token address for tests that use it.
Adds fetching of token details in history rows that are transfers in all cases.
Makes `getArgsForTokenInvocation` work with transfers to C addresses.

Why
The address was stale after the testnet reset.
Transfers to C addresses uncovered a few edge cases in history.


extension/e2e-tests/helpers/login.ts
```diff
@@ -22,7 +22,7 @@ export const loginAndFund = async ({ page, extensionId }) => {
   });
 
   await page.goto(`chrome-extension://${extensionId}/index.html#/account`);
-  await page.getByTestId("BlockaidAnnouncement__accept").click();
+  // await page.getByTestId("BlockaidAnnouncement__accept").click();
```
aristidesstaffieri, line 25: This isn't shown right now and also commented out so I assume we will need this again soon.

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -397,17 +397,26 @@ export const HistoryItem = ({
...
-            // TODO: attempt to fetch token details, not stored
-            setRowText(operationString);
-            setTxDetails((_state) => ({
-              ..._state,
-              headerTitle: translations("Transaction"),
-              operationText: operationString,
-            }));
-          } else {
-            const { token, decimals } = balances[tokenKey] as TokenBalance;
+          try {
```
aristidesstaffieri, line 402: Now that sending transfers from payments is better supported, it became more obvious that we can't rely on client side cached token details for this path because we more often don't have them(user has a classic balance in Freighter but send to C address through SAC transfer for example). This update just fetches token details for every row that is a transfer inline to avoid the inconsistencies.

extension/src/popup/helpers/soroban.ts
```diff
@@ -121,18 +128,12 @@ export const getArgsForTokenInvocation = (
 
   switch (fnName) {
     case SorobanTokenInterface.transfer:
-      from = StrKey.encodeEd25519PublicKey(
-        args[0].address().accountId().ed25519(),
-      );
-      to = StrKey.encodeEd25519PublicKey(
-        args[1].address().accountId().ed25519(),
-      );
+      from = addressToString(args[0].address());
```
aristidesstaffieri, line 131: Now that we support transfers to C addresses, I ran into this where from and to are actually not always `accountId` and can also be `contractId`.

extension/e2e-tests/sendPayment.test.ts
```diff
@@ -1,8 +1,8 @@
 import { test, expect, expectPageToHaveScreenshot } from "./test-fixtures";
 import { loginAndFund, loginToTestAccount, PASSWORD } from "./helpers/login";
+import { TEST_TOKEN_ADDRESS } from "./helpers/test-token";
 
 test("Send XLM payment to G address", async ({ page, extensionId }) => {
-  test.slow();
```
piyalbasu, line 5: is there a specific reason to drop this? if I recall, it just extends the wait time of a test. I believe I had added this in the event of slow responses from Horizon

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -357,7 +357,9 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
           url,
           networkDetails,
         );
+        return;
```
piyalbasu, line 360: does this do something the `return` on ln 388 does not?

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -357,7 +357,9 @@ export const TransactionDetails = ({ goBack }: { goBack: () => void }) => {
...
       }
+      setLoading(false);
```
piyalbasu, line 362: do we need to worry about this loading state for classic tx's, as well?

piyalbasu (approved):

extension/e2e-tests/sendPayment.test.ts
aristidesstaffieri: I just didn't observe that it added any longer of a timeout but no qualms with restoring, will add back.

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: well I think we could safely drop it but in the case that we `scanTx` we don't need to manually `setLoading(false)` because it happens in the hook call. So the bug was that we relied on `isLoading` but didn't always set it correctly in the case here.
Dropping it should only result in calling that state setter once unnecessarily which is probably fine.

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: no we don't because we seem to be calling that in every case here. I guess maybe I should about the intention of the condition to not `scanSorobanTx` if `submission.submitStatus === ActionStatus.IDLE && transactionSimulation.preparedTransaction`? I think that's actually correct since we don't really get anything from re-scanning it in tx details after the tx has been done. Maybe the miss here is that we should also be doing this in `scanClassicTx`?

extension/e2e-tests/sendPayment.test.ts
piyalbasu: Cool - once we stub out external calls, we shouldn't need this anymore

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: oh wait I wrote that! lol Yeah curious if you think that's right? I think we probably also want to skip the scan for classic in the same condition.

piyalbasu: Okay took me a second to understand what was happening, but I get it now:

This conditional is just to make sure we've finished simming the tx, which is only necessary for Soroban tx's. So it makes sense that we only have the conditional on scanSorobanTx

aristidesstaffieri: I changed this in fd2e8dcc617d3c27b72ff2d0ca2de6084d75044f
I think it makes sense to not scan either tx type in tx details if we are rendering it after the tx has been sent.

aristidesstaffieri: ah maybe I misunderstand this conditional. I think really what we want here is to never scan the tx is we are rendering the details at confirmation so I will just add a prop that we can pass in explicitly at confirmation to make this less brittle.

piyalbasu: But if we go back to tx details from the Tx Success screen, we'll re-scan a classic tx, but we won't re-scan a soroban tx. And that's kind of a weird discrepancy. We probably shouldn't re-scan in either case

aristidesstaffieri: ok cool I think we're aligned here, 62cd1a6cd66f2a512d4b2850cf08932d01e6eaee is the version that I think handles this best.

piyalbasu: Nice - this lgtm!

PR #1506, opened 2024-10-04
make sure to show sim errors in send flow; add tests

piyalbasu:
Screenshot 2024-10-04 at 3 45 31 PM

Seems as though Blockaid's tx scanning isn't responding properly, so I stubbed some data to show what this would look like if the endpoint came back with a sim error

This PR fixes an issue with showing Benign warnings and not showing tx sim errors

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -1326,7 +1326,7 @@ export const BlockaidWarningModal = ({
               :
               <ul className="ScamAssetWarning__list">
                 {description.map((d) => (
-                  <li key={d}>{truncatedDescription(d)}</li>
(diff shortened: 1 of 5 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1126,18 +1126,27 @@ export const BlockAidSiteScanLabel = ({
...
 }: {
   scanResult: BlockAidScanTxResult;
+  isPopup?: boolean;
 }) => {
   const { t } = useTranslation();
   const { simulation, validation } = scanResult;
 
   if (simulation && "error" in simulation) {
+    const header = t("This transaction is expected to fail");
+    if (isPopup) {
```
piyalbasu, line 1139: the main change here is to re-use this logic for showing sim and validation errors. In the send/swap flow, we just show a different a component

aristidesstaffieri (approved):

PR #1509, opened 2024-10-08
Fix flakey tests

piyalbasu:


extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -102,6 +102,9 @@ export const Settings = ({
 
   useEffect(() => {
     async function simulateTx() {
+      if (!recommendedFee) {
```
piyalbasu, line 105: it appears that part of the issue was the request to Horizon to get the recommended fee can sometimes take a long time to resolve, which will cause this hook to run but not be able to pass the conditional at ln 184 because the `recommendedFee` hasn't been set yet

aristidesstaffieri (approved):

PR #1514, opened 2024-10-09
upgrade soroswap and fix bug

piyalbasu:


extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -213,7 +213,7 @@ export const SendAmount = ({
     if (values.destinationAsset) {
       dispatch(saveDestinationAsset(values.destinationAsset));
       isDestAssetScam = isAssetSuspicious(
-        accountBalances.balances?.[destinationAsset].blockaidData,
+        accountBalances.balances?.[destinationAsset]?.blockaidData,
```
piyalbasu, line 216: this was throwing an error for soroswap assets because the user doesn't actually have these assets in their balances yet

extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -133,11 +133,13 @@ export const Settings = ({
             ),
             CLASSIC_ASSET_DECIMALS,
           );
-          dispatch(
-            saveTransactionFee(
-              baseFee.plus(new BigNumber(minResourceFee)).toString(),
-            ),
-          );
+          if (!transactionFee) {
```
piyalbasu, line 136: we do a similar check in the other pathways - this keeps the effect from running indefinitely

extension/src/popup/helpers/sorobanSwap.ts
```diff
@@ -107,12 +107,27 @@ export const soroswapGetBestPath = async ({
 
           const data = await res.json();
 
+          return data;
+        },
+      },
+      {
+        protocol: Protocol.PHOENIX,
```
piyalbasu, line 114: apparently we needed to add another liquidity provider to ensure we could get a usable pathway for swap

aristidesstaffieri (approved):

PR #1520, opened 2024-10-10
[BUG] Sign transaction should allow unfunded signer

aristidesstaffieri:
What
This improves the check for `hasEnoughXlm` during transaction signing in order to allow for an unfunded account to be the signer.

Why
In multi signature scenarios, this should be allowed given that the the signer may not be the tx source account.

Fixes: https://github.com/stellar/freighter/issues/1490


config/jest/setupTests.tsx
```diff
@@ -35,13 +35,12 @@ jest.mock("react-i18next", () => ({
       changeLanguage: () => new Promise(() => {}),
     },
   }),
-  Trans: ({ children }: { children: React.ReactElement[] }) =>
-    children.map((child, i) => {
-      if (typeof child === "string") {
-        return <span key={i}>{child}</span>;
-      }
-      return "";
-    }),
```
piyalbasu, line 44: just curious - why was this necessary?

piyalbasu (approved):

config/jest/setupTests.tsx
aristidesstaffieri: So it looks to me like this was incorrect but no test had hit this path to exercise it yet. In every case that we use Trans, we use a string as its child so expecting an array broke the test I was writing as it tried to render that warning.

PR #1522, opened 2024-10-10
Add define plugin for freighter api

JakeUrban:
After running `yarn start`, and chrome opening the docs, selecting the "Playground" dropdown menu will trigger the following error:

Screenshot 2024-10-10 at 10 51 15 AM

This is a result of `DEV_SERVER` and `DEV_EXTENSION` not being defined in the `@stellar/freighter-api` package in the same way its defined in `extension` -- using `webpack.DefinePlugin`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,5 +128,5 @@
       "webpack>watchpack>watchpack-chokidar2>chokidar>fsevents": true
     }
   },
-  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -127,5 +127,6 @@
       "webpack-dev-server>chokidar>fsevents": true,
       "webpack>watchpack>watchpack-chokidar2>chokidar>fsevents": true
     }
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
```
piyalbasu, line 131: looks like the the latest is actually 1.22.22 so let's point to the sha of that

piyalbasu (approved):

PR #1528, opened 2024-10-14
Release/5.23.4

aristidesstaffieri:


aristidesstaffieri:
Looks like we now see a warning from dart sass
```
Module Warning (from ../node_modules/sass-loader/dist/cjs.js):
[start:unpacked     ] Deprecation The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
```

I'm still trying to figure out if we can tweak the build to move off the legacy API but we can also silence it in the meantime.

PR #1529, opened 2024-10-15
add configuration to hide dust payments

piyalbasu:
Screenshot 2024-10-15 at 11 57 53 AM
Screenshot 2024-10-15 at 11 58 03 AM
Screenshot 2024-10-15 at 11 58 16 AM


@shared/api/internal.ts
```diff
@@ -1176,11 +1176,11 @@ export const saveAllowList = async ({
 export const saveSettings = async ({
   isDataSharingAllowed,
   isMemoValidationEnabled,
-  isNonSSLEnabled,
+  isHideDustEnabled,
```
piyalbasu, line 1179: there was a bug where, i think through a bad merge, the `isNonSSLEnabled` configuration was in both the Preferences and in the Advanced Settings. It used to be in Preferences but was later moved to Advanced Settings.

Taking this opportunity to replace errant `isNonSSLEnabled` config with the new `isHideDustEnabled`

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -99,13 +104,25 @@ export const AccountHistory = () => {
           operation.type ===
             Horizon.HorizonApi.OperationResponseType.createAccount &&
           operation.account !== publicKey;
+        const isDustPayment =
+          isPayment &&
+          "asset_type" in operation &&
+          operation.asset_type === "native" &&
+          "to" in operation &&
+          operation.to === publicKey &&
```
piyalbasu, line 112: checking for a native payment to the user's account

aristidesstaffieri: should this be flipped? `operation.asset_type === "native"`

I think dust payments would typically not be xlm right, as the spam is to try to get bunk tokens to show up in the interface by sending dust amounts of them.

extension/src/popup/views/Preferences/index.tsx
```diff
@@ -15,34 +15,35 @@ import "./styles.scss";
...
   };
 
   const handleSubmit = async (formValue: SettingValues) => {
     const {
       isValidatingMemoValue,
       isDataSharingAllowedValue,
-      isNonSSLEnabledValue,
+      isHideDustEnabledValue,
     } = formValue;
+    console.log(formValue);
```
aristidesstaffieri, line 39: stray log

extension/src/popup/views/__tests__/AccountHistory.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+        }}
+      >
+        <AccountHistory />
+      </Wrapper>,
+    );
+
+    await waitFor(() => screen.getByTestId("AccountHistory"));
+    expect(screen.getByTestId("AccountHistory")).toBeDefined();
+    const historyNodes = screen.getAllByTestId("history-item");
+    expect(historyNodes.length).toEqual(2);
```
aristidesstaffieri, line 72: along with the inferred length of the history item without the dust item, should we explicitly check that the correct history item is filtered out?

piyalbasu: Good call! I'll add

extension/src/popup/views/AccountHistory/index.tsx
piyalbasu: Actually, according to this [doc](https://docs.google.com/document/d/1tpGsVx3ZlRZbdhedkOfj_cUE2Ixdzct1v9xZP9zj3FE/edit), we're specifically on the lookout for small XLM payments

aristidesstaffieri: oh cause of the memo, yup got it!

aristidesstaffieri (approved):

Issue #1558, opened 2024-10-21
Add Options dropdown to Account View


piyalbasu:
Add dropdown that includes:
* Account details
* Connected apps
* Manage Assets
* Expand view

Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5036&node-type=frame&t=Zs0wZH6xg9GQJwnv-0

Issue #1548, opened 2024-10-18
Implement fullscreen UI button


piyalbasu:
Designs here: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5036&t=9rCjbUdcZRzRcu5C-1

piyalbasu:
Blocked by #1558

PR #1580, opened 2024-10-24
Feature/fullscreen btn

piyalbasu:
This PR adds a button to open up the extension in the user's browser. Visually, most of the views look okay in fullscreen mode, but there are some instances where a user might run into issues with stale data. For example, if a user opens Freighter in a new tab, and then leaves and submits some other tx's in a different tab, when they come back to the Freighter tab, their balances could be out of date. In a follow up PR, I'm going to explore solutions for solving this (maybe polling? maybe throwing up a warning saying "please refresh the app, you've been idle for too long")

We've always counted on user session's being pretty short since the app closes when a user clicks out of the extension. This may not be the case with this feature


piyalbasu:
Screenshot 2024-10-24 at 12 56 09 PM
Screenshot 2024-10-24 at 12 56 18 PM

This PR incorporates #1558 and #1548

extension/e2e-tests/addAsset.test.ts
```diff
@@ -6,14 +6,15 @@ test("Adding unverified Soroban token", async ({ page, extensionId }) => {
   test.slow();
   await loginToTestAccount({ page, extensionId });
 
+  await page.getByTestId("account-options-dropdown").click();
   await page.getByText("Manage Assets").click({ force: true });
   await page.getByPlaceholder("Enter password").fill(PASSWORD);
   await page.getByText("Log In").click({ force: true });
+  await expect(page.getByText("Your assets")).toBeVisible();
```
piyalbasu, line 13: This fixes a bug with snapshot: https://github.com/stellar/freighter/pull/1580/files#diff-93b044b86208b123f27589535905ad725da70f6b467ee0559c988a9f866e71e1

It was mistakenly snapshotting the password confirm screen instead of the manage assets screen because it wasn't waiting for the view to load

extension/src/popup/components/account/AccountOptionsDropdown/styles.scss
```diff
@@ -0,0 +1,37 @@
+@use "../../../styles/utils.scss" as *;
+
+.AccountOptionsDropdown {
+  &__modal {
+    background: var(--sds-clr-gray-01);
+    border-radius: #{pxToRem(6px)};
```
piyalbasu, line 6: I'm trying to make a conscious effort to start using this pxToRem helper more. It will probably make our css a little more foolproof if we let the helper do the math for us

aristidesstaffieri:
> This PR adds a button to open up the extension in the user's browser. Visually, most of the views look okay in fullscreen mode, but there are some instances where a user might run into issues with stale data. For example, if a user opens Freighter in a new tab, and then leaves and submits some other tx's in a different tab, when they come back to the Freighter tab, their balances could be out of date. In a follow up PR, I'm going to explore solutions for solving this (maybe polling? maybe throwing up a warning saying "please refresh the app, you've been idle for too long")
> 
> We've always counted on user session's being pretty short since the app closes when a user clicks out of the extension. This may not be the case with this feature

Yeah I'm not sure if users expect the UI to be a live feed or not. I think this problem exists even without full screen mode though because you can open your Freighter extension and then through some other means change the state of that account and the extension will also not poll or do anything to refresh balance without a user navigation. 

I guess maybe though we expect tabs could hang around longer than an open extension popup.

aristidesstaffieri (approved):

PR #1581, opened 2024-10-25
Fullscreen ux improvements

piyalbasu:
For users in full screen mode, they run the risk of leaving Freighter open in a tab for long periods of time (previously impossible). This can lead to a couple of issues:

1. They are at risk of having stale data as we only fetch account balances when the app starts
2. They could potentially leave the window open and forget about it, allowing a bad actor to get access to their account(s)

To help mitigate issue 1, I am showing a notification to the user that they might have stale data after a period of time and will need to reload. The _real_ solution is to implement streaming or polling to keep the app up to date, but that is a bigger ticket and I think should come in an iterative follow up.

For 2, when the user is in full screen mode, we should log out accounts that are inactive, similar to any banking app. This inactivity check is only enabled when the user is in fullscreen mode. The inactivity timer resets when a user navigates to a different page. In the demo video below, I set the timer to be 10 seconds to illustrate the UX. Once the user stops navigating, the app waits 10 seconds and then logs them out. In real life, they would be something like 20 minutes.

### DIFF
--- a/extension/src/popup/Router.tsx
+++ b/extension/src/popup/Router.tsx
@@ -12,7 +12,7 @@ import { useSelector, useDispatch } from "react-redux";
 import { APPLICATION_STATE } from "@shared/constants/applicationState";
 import { POPUP_WIDTH } from "constants/dimensions";
 import { newTabHref } from "helpers/urls";
-import { navigateTo, openTab } from "popup/helpers/navigate";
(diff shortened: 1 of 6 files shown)


piyalbasu:
https://github.com/user-attachments/assets/354c85ba-7dd1-47ed-9720-c04b42cfef91

Issue #1431, opened 2024-08-20
Prevent having to login and authenticate after generating new account


piyalbasu:
Steps to reproduce:

1. Start with a fresh install
2. At Welcome, either create a new account or import an existing account by mnemoic
3. Fund account and click "Manage Assets"
4. Observe that it asks for your password

AC:
User should not be prompted for password within 24 hours of creating account

PR #1590, opened 2024-10-29
start session properly after create/recover account

piyalbasu:
#1431 

Makes sure to start a 24h session after create/recover account


extension/e2e-tests/addAsset.test.ts
```diff
@@ -8,8 +8,6 @@ test("Adding unverified Soroban token", async ({ page, extensionId }) => {
 
   await page.getByTestId("account-options-dropdown").click();
   await page.getByText("Manage Assets").click({ force: true });
-  await page.getByPlaceholder("Enter password").fill(PASSWORD);
-  await page.getByText("Log In").click({ force: true });
```
piyalbasu, line 12: one byproduct of this ticket is that we are no longer testing the experience when a user's session has timed out. We'll have to think about how we can actually test this in the future 🤔 

Maybe we need to pass a configuration to the build step to make the session timeout immediately instead of wating the normal 24h

aristidesstaffieri: maybe one way would be to have `sessionTimer.startSession` take the session length as an arg so that in the test we could just pass in a `0`.

piyalbasu: That's a good call. I think the question is how we want to pass this configuration. Maybe the easiest path would be just appending a query param to `chrome-extension://${extensionId}/index.html` and using that to pass an arg to the background script.

We're going to have a ticket coming up for users to be able to configure how long they want their session to be. This can be added as part of that ticket

aristidesstaffieri (approved):

PR #1599, opened 2024-10-30
fix firefox pipeline

piyalbasu:


.github/workflows/submitProduction.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           channel: listed
           apiKey: ${{ secrets.AMO_SIGN_KEY }}
           apiSecret: ${{ secrets.AMO_SIGN_SECRET }}
-          timeout: 900000
+          timeout: 0 # this timeout waits for the .xpi to be approved/downloaded, but since we're submitting a zip, this will always fail
```
piyalbasu, line 94: So, this pipeline supports submitting a zip file (which we do) and an xpi file (a compiled firefox addon's file extension, which we don't use).

Their approval timeout logic appears to assume that all users of this pipeline always do the latter. They are expecting that we compile our code into an .xpi using their `build` cmd and then forwarding that on to Mozilla.As we can see in [previous failures](https://github.com/stellar/freighter/actions/runs/11488953879/job/31976780909), the pipeline succeeded in uploading a zip file to Mozilla, but it failed in finding the xpi file because it did not exist.

Setting this timeout to 0 should just upload the zip file to Mozilla and not wait for a signed xpi file to be spit out afterwards https://github.com/kewisch/action-web-ext/blob/main/node_modules/web-ext/lib/util/submit-addon.js#L194

aristidesstaffieri: interesting, thanks!

aristidesstaffieri (approved):

Issue #829, opened 2023-05-21
docs: unclear which freighter api version to use


tomerweller:
In the "Using Freighter in the browser" resource the quoted `<script>` tag leaves finding which version to use as an exercise to the dev. It would be great to have the link contain the latest version or at least tell the dev where to go to find this number.

piyalbasu:
This is live on the docs now: https://docs.freighter.app/docs/guide/usingFreighterBrowser

PR #1629, opened 2024-11-08
[Docs] General updates

CassioMG:
This PR suggests some doc updates and fixes some typos.

The main updates would be:
- [Update node version](https://github.com/stellar/freighter/pull/1629/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R18)
- Link to [extension's README urls](https://github.com/stellar/freighter/pull/1629/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R74-R77) which would help installing and running it locally

### DIFF
--- a/docs/docs/guide/basics/VersionedUrl.tsx
+++ b/docs/docs/guide/basics/VersionedUrl.tsx
@@ -1,8 +1,7 @@
 import React from "react";
-import packageJson from "../../../../@stellar/freighter-api/package.json";
 
 export const VersionedUrl = () => (
   <p>
-    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/${packageJson.version}/index.min.js' /></head>`}</code>
+    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/{version}/index.min.js' /></head>`}</code>
   </p>
 );
--- a/docs/docs/guide/basics/VersionedUrl.tsx
+++ b/docs/docs/guide/basics/VersionedUrl.tsx
@@ -1,7 +1,8 @@
 import React from "react";
+import packageJson from "../../../../@stellar/freighter-api/package.json";
 
 export const VersionedUrl = () => (
   <p>
-    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/{version}/index.min.js' /></head>`}</code>
+    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/${packageJson.version}/index.min.js' /></head>`}</code>
   </p>
 );
--- a/extension/src/popup/locales/en/translation.json
+++ b/extension/src/popup/locales/en/translation.json
@@ -14,6 +14,7 @@
   "Add": "Add",
   "Add address": "Add address",
   "Add an asset": "Add an asset",
+  "Add anyway": "Add anyway",
   "Add asset": "Add asset",
   "Add Asset trustline": "Add Asset trustline",
   "Add Asset Trustline": "Add Asset Trustline",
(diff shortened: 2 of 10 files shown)


docs/docs/guide/basics/VersionedUrl.tsx
```diff
@@ -1,8 +1,7 @@
 import React from "react";
-import packageJson from "../../../../@stellar/freighter-api/package.json";
 
 export const VersionedUrl = () => (
   <p>
-    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/${packageJson.version}/index.min.js' /></head>`}</code>
+    <code>{`<head><script src='https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/{version}/index.min.js' /></head>`}</code>
```
aristidesstaffieri, line 5: What's the difference between these two? I couldn't find much on the global `version`.

piyalbasu: For context, I added the original logic around grabbing the package json file because cdnjs creates a url based on the latest version. So, for ex, to get v3.0.0 of freighter-api, you would need a url of: https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/3.0.0/index.min.js

How does the `version` var accomplish that?

CassioMG: Ah yeah, so there isn't a global `version`. I changed it to be consistent with the text instructions [here](https://docs.freighter.app/docs/guide/usingFreighterBrowser) which mentions `{version}` . This is the only place importing this `VersionedUrl` component.

CassioMG: Humm maybe we should just put a plain `html` code like it's being done [here](https://github.com/stellar/freighter/blob/master/docs/docs/guide/gettingStarted.md) instead of importing this `VersionedUrl` component. Wdyt?

Something like:

```html
<head><script src="https://cdnjs.cloudflare.com/ajax/libs/stellar-freighter-api/{version}/index.min.js"></script></head>
```

piyalbasu: Oh, just to give further context: I did originally have it as plain html with the `{version}` var in it, but I got complaints from people in the org that the link should be dynamic and include the latest version so users can just copy and paste without having to insert the version number themself

piyalbasu: This is the original ticket that caused me to add this: https://github.com/stellar/freighter/issues/829

CassioMG: Ah, good to know! Ok, I'll revert my change and suggest another change on the body text

CassioMG: Tweaked the text a little bit here: https://github.com/stellar/freighter/pull/1629/commits/7375c2182a0bc5a897097f56186b84b815bc943f

CassioMG: Also reverted the `VersionedUrl` change here: https://github.com/stellar/freighter/pull/1629/commits/217eda9119a391419cc064295b68d60aa0747cfb

piyalbasu (approved):

extension/src/popup/locales/en/translation.json
```diff
@@ -200,6 +201,10 @@
   "Got it": "Got it",
   "Hash": "Hash",
   "Help": "Help",
+  "Hide payments smaller than 0": {
+    "1 XLM": "Hide payments smaller than 0.1 XLM"
+  },
+  "Hide small payments": "Hide small payments",
```
piyalbasu, line 207: Not 100% sure why this translation strings weren't already committed, but no problem adding them now

Issue #1495, opened 2024-09-24
Unlock all accounts when entering password


piyalbasu:
Description

When entering user password, we currently only unlock the active keypair. So, when a user switches the account, they must re-enter their password to get the new keypair.

Instead of saving just the active account’s privateKey as string in redux, let’s save an object of type {[keyId]: privateKey }

that contains all available private keys. When we confirmPassword, we can iterate over all the key stores, unlocking each one and saving them into this one object.

Acceptance Criteria:
* A user should only have to enter their password once and be able to freely switch between accounts without having to authenticate for 24h

* after 24h, all private keys are removed from redux

* when signing tx’s, we grab the private key associated with active keyId from the redux object

PR #1630, opened 2024-11-08
[DO NOT MERGE] check for password in redux when switching account

piyalbasu:
#1495 

Let's hold this for the next release as we've already QA'ed the 5.25.0 release

This will prevent the user from having to re-auth when they switch between accounts. I actually didn't really need to make any fundamental changes since we already hold the password in the memory. I'm just using the stored password to unlock the new account's keystore when we change accounts.

Also, making sure to clear out the stored password when the session time outs


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -572,7 +573,19 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     const { publicKey } = request;
     await _activatePublicKey({ publicKey });
 
-    sessionStore.dispatch(timeoutAccountAccess());
+    const password = passwordSelector(sessionStore.getState()) || "";
```
piyalbasu, line 576: this is the biggest change here: storing the new account's private key when we change accounts. Just in case something goes wrong in the `try`, we'll still be able to change the account but the user will need to manually re-auth (which is the current UX)

piyalbasu:
#1495

Issue #1579, opened 2024-10-24
Move e2e tests from on push to run in CI


piyalbasu:
Get e2e tests running in CI and block PR's if the tests fail

* Are the tests resilient enough to stand up to being run many times by many users?
* Should we add a toggle to run between a "stateless" vs "full integration" mode

aristidesstaffieri:
This should result in sub issues and/or a spike to figure out how we want these tests to work in the future.

PR #1631, opened 2024-11-08
[DON'T MERGE] add e2e tests to ci pipeline

piyalbasu:
#1579 

* Moves from the e2e tests from a pre-push git hook to running in CI on every commit
* Because we use the same Testnet account for a few different tests, we've had some issues where a failing test would leave the test account in a weird state. [more details](https://github.com/stellar/freighter/issues/1653). To combat that, I created some "cleanup jobs" that run only if one of these tests fails. They basically just revert the account back to it's base state


extension/e2e-tests/addAsset.test.ts
```diff
@@ -81,3 +81,22 @@ test("Adding Soroban verified token", async ({ page, extensionId }) => {
     timeout: 30000,
   });
 });
+test.afterAll(async ({ page, extensionId }) => {
+  if (
+    test.info().status !== test.info().expectedStatus &&
+    test.info().title === "Adding Soroban verified token"
```
piyalbasu, line 87: `Adding Soroban verified token` adds a trustline to the account, so if it fails, we might end up in a state where the account still has the trustline. This `afterAll` removes it

aristidesstaffieri (approved):

PR #1642, opened 2024-11-12
Bugfix/recommended fee

piyalbasu:
`recommendedFee` wasn't being set from Horizon's fee_stats. We've had this issue before, and fixed it. Looks like this bug made it's way back due to a bad merge because we jumped ahead of this release with the Blockaid release.

I've added a test to make sure we can catch this automatically in the future


extension/src/popup/helpers/useNetworkFees.ts
```diff
@@ -15,7 +15,7 @@ export const useNetworkFees = () => {
   const { networkUrl, networkPassphrase } = useSelector(
     settingsNetworkDetailsSelector,
   );
-  const [recommendedFee, setRecommendedFee] = useState("100");
+  const [recommendedFee, setRecommendedFee] = useState("");
```
piyalbasu, line 18: the bad merge that introduced this: https://github.com/stellar/freighter/pull/1398/files#diff-48e5eb06e07934afbdf9c6be1ad5d87bb1658ee0c72641914788d5655df34d2dR18

aristidesstaffieri (approved):

PR #1661, opened 2024-11-13
[BUG] fixes an incorrect reference to the old create contract class for v2

aristidesstaffieri:
Reroll of #1655

### DIFF
--- a/extension/src/popup/helpers/soroban.ts
+++ b/extension/src/popup/helpers/soroban.ts
@@ -444,7 +444,7 @@ const isInvocationArg = (
   invocation: InvocationArgs | undefined,
 ): invocation is InvocationArgs => !!invocation;
 
-function getInvocationArgs(
+export function getInvocationArgs(
   invocation: xdr.SorobanAuthorizedInvocation,
 ): InvocationArgs | undefined {
   const fn = invocation.function();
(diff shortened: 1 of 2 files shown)


piyalbasu (approved):

extension/src/popup/helpers/soroban.ts
```diff
@@ -496,11 +508,20 @@ export function getInvocationArgs(
...
-        case 1:
-          return {
+        case 1: {
+          const sacDetails = {
             type: "sac",
             asset: Asset.fromOperation(preimage.fromAsset()).toString(),
-          };
+          } as FnArgsCreateSac;
+
+          if (fn.switch().value === 2) {
```
aristidesstaffieri, line 517: It wasn't super clear to me through the xdr if this was possible for the SAC case but I think there's no reason to not include this line. If SACs will keep using v1 then this conditional won't ever be hit.

extension/src/popup/views/ReviewAuth/index.tsx
```diff
@@ -351,10 +351,12 @@ const AuthDetail = ({
     async function getIsToken() {
       try {
         const transfers = [];
-        const isToken = await getIsTokenSpec({
-          contractId: rootJson.args.source,
-          networkDetails,
-        });
+        const isToken = !rootJson.args.source
```
aristidesstaffieri, line 354: for both the invocation and sub invocation this should account for not having a source in the root args.

Issue #923, opened 2023-08-02
The words on password popup are confusing 


GHesericsu:
### What version are you using?

5.2.5

### What did you do?

I tried to submit a transaction and freighter asked me for password

### What did you expect to see?
A popup that asked me for password with clear instruction is asking for password

### What did you see instead?

Do you want my password or want an address? :) 
![Screen Shot 2023-08-01 at 6 20 04 PM](https://github.com/stellar/freighter/assets/60202455/b6d1cf75-5bc5-452b-bdb6-ff7d495ee1df)

CassioMG:
This seems like a UX issue. Whenever a user tries to create a new stellar address this screen on the right is displayed:
![Image](https://github.com/user-attachments/assets/f8d7c3dc-8259-4480-a740-b19aa6d284a9)
It could sound a bit confusing as users _could_ interpret the "Add" as a command to input an existing stellar address to the input field which actually has "Enter Password" as the placeholder.

I'd like to propose we change the copy a little bit on this screen so it's consistent with the button on the left which has "Create a new Stellar address" as the title, instead of using the "Add" verb.
So the second screen would look like this:
- Title: "Enter password to create a new Stellar address"
- Placeholder: "Enter Password"
- CTA: "Create New Address"
![Image](https://github.com/user-attachments/assets/d268a5a6-a3e0-47df-a01d-b78e450eb713)

@GHesericsu @sdfcharles @piyalbasu @aristidesstaffieri @JakeUrban curious about your thoughts here

JakeUrban:
Is there a reason why we require the user's password to create an address?

Regardless, I would probably vote for a less verbose prompt. So instead of "Create a new Stellar address" --> "Enter Password to create a new Stellar address" --> "Create New Address", I would vote for "Create a new Stellar address" --> "Enter Password" --> "Create New Address".

aristidesstaffieri:
> This seems like a UX issue. Whenever a user tries to create a new stellar address this screen on the right is displayed: ![Image](https://github.com/user-attachments/assets/f8d7c3dc-8259-4480-a740-b19aa6d284a9) It could sound a bit confusing as users _could_ interpret the "Add" as a command to input an existing stellar address to the input field which actually has "Enter Password" as the placeholder.
> 
> I'd like to propose we change the copy a little bit on this screen so it's consistent with the button on the left which has "Create a new Stellar address" as the title, instead of using the "Add" verb. So the second screen would look like this:
> 
[...]

PR #1682, opened 2024-11-18
Skip password when creating a new address

CassioMG:
Closes https://github.com/stellar/freighter/issues/923

This PR implements the following changes as [discussed on the issue](https://github.com/stellar/freighter/issues/923#issuecomment-2471429766):
- In case the `password is STILL saved` in current session store users won't need to input their password when creating a new Stellar address. Instead, the application will automatically [read it from the session state](ADD LINK).
- In case the `password is NOT saved` (e.g. the session timed out) then it'll ask for the password.
   - ^ this PR has also changed a `couple copies` on this "Enter Password" screen:
       - Screen title update: **"Add a new Stellar address" => "Enter Password"**
       - Button title update: **"Add New Address" => "Create New Address"** 

_*Note: this PR is based on [`feature/maintain-password-2`](https://github.com/stellar/freighter/tree/feature/maintain-password-2) **branch** since it relies on changes from [this other PR](https://github.com/stellar/freighter/pull/1677)._

### Creating a new address when password is saved
https://github.com/user-attachments/assets/c7d6f33a-5899-4c3e-bc9e-535d5d5f4115

### Creating a new address when password is NOT saved
https://github.com/user-attachments/assets/83436042-c4a3-4127-8a98-f056d1fd5ca7

### Auth error while creating a new address
https://github.com/user-attachments/assets/3dc6caa4-fcc3-440b-96e9-0b28b7f60c9f


extension/src/popup/views/AddAccount/AddAccount/index.tsx
```diff
@@ -35,30 +36,60 @@ export const AddAccount = () => {
...
     [dispatch],
   );
 
+  // No need to ask for password if it's already stored, so let's just briefly
+  // display a loading screen while user is being redirected to next screen
+  if (hasPrivateKey) {
+    return (
+      <React.Fragment>
+        <SubviewHeader title="" />
+        {!authError && <Loading />}
```
aristidesstaffieri, line 82: I'm not sure if this is worth a change or not but just for thought -

We have other places in Freighter that use a similar pattern but also produce the "loading stutter" effect we get when creating an account here. This happens because there are 2 loaders for 1 transition in state. 

Is there an easy way to make that feel more like 1 loading state?

CassioMG:
After the automatic base branch switch it appears this PR is showing more changes then it should have.

So closing this in favor of https://github.com/stellar/freighter/pull/1685 which contains only the changes related to this task.

extension/src/popup/views/AddAccount/AddAccount/index.tsx
CassioMG: @aristidesstaffieri Hummm I think I know what you mean. I _believe_ there is probably a way of making this transition look nicer but I'm _not sure_ it'll be a straightforward fix.

I'm still getting used to the codebase and to the React Web syntax (which is different from React Native syntax) so I think it'd take some time for me to figure this out. In case you think it's not a blocker, I'd like to move ahead "as is" and create a ticket so we could tackle it later. What do you think?

Btw, I've moved this PR [to here](https://github.com/stellar/freighter/pull/1685) since this one got messed up after the automatic rebase.

Issue #1653, opened 2024-11-13
[SPIKE] Investigate splitting e2e tests into stubbed vs full e2e mode


piyalbasu:
Currently in our e2e tests, we use a Testnet account to run through a few scenarios like sending a payment, adding a trustline, etc. These scenarios generate tx's and actually submit to Testnet.

The benefits of re-using the same account and not stubbing out these calls are:
1. We get signal that all our integrated API's are continuing to work as we expect. This gives us a more complete picture of if Freighter is actually working for the end user and we're not missing anything with some stubbing that is overly optimistic 
2. Tests run a bit faster because we don't need to be constantly creating and funding new accounts to test with
3. For Soroban token testing, we don't need to worry about minting to this newly generated account
4. Having to stub out calls makes creating tests slower and more complex. It's very fast to create an e2e test without stubbing. This is important as we really need a lot more coverage of our code.

The cons are:
1. The tests are non-deterministic. If a test fails in an unexpected way, the test account might end up in a weird state that causes subsequent test failures. This could be come even more of a problem with multiple devs running tests at the same time
2. Issues with any 3rd party API, even if transient, can cause a test failure. For example, in times of high congestion on Testnet, a test could fail just for taking too long

Possible solution:
We default to stubbing out these calls when the tests run in CI and, maybe before we cut a beta/prod release, we flip a toggle that allows these tests to run against actual API's. These full API test runs would happen infrequently and by only one dev at a time.

Open Questions:
1. How easy/hard is it to create this kind of toggle in Playwright? 
2. Can we run it from a Github Action?
3. How much work/effort will it be to stub out all the existing calls in our current e2e tests?

PR #1684, opened 2024-11-18
enable e2e tests in CI and add cleanup tasks

piyalbasu:
re-roll of https://github.com/stellar/freighter/pull/1631

https://github.com/stellar/freighter/issues/1579

Moves from the e2e tests from a pre-push git hook to running in CI on every commit
Because we use the same Testnet account for a few different tests, we've had some issues where a failing test would leave the test account in a weird state. https://github.com/stellar/freighter/issues/1653. To combat that, I created some "cleanup jobs" that run only if one of these tests fails. They basically just revert the account back to it's base state

### DIFF
--- a/.github/workflows/runTests.yml
+++ b/.github/workflows/runTests.yml
@@ -6,16 +6,16 @@ jobs:
   test-ci:
     name: test
     timeout-minutes: 20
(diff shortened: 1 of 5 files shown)


JakeUrban:
resolves #1579

PR #1685, opened 2024-11-18
Skip password when creating a new address

CassioMG:
Closes https://github.com/stellar/freighter/issues/923

This PR implements the following changes as [discussed on the issue](https://github.com/stellar/freighter/issues/923#issuecomment-2471429766):
- In case the `password is STILL saved` in current session store users won't need to input their password when creating a new Stellar address. Instead, the application will automatically [read it from the session state](https://github.com/stellar/freighter/pull/1682/files#diff-c40b026a1f6fd2502a2c1e271bfeb9ab9ed49061314f479451369d8a69b54ea8R463).
- In case the `password is NOT saved` (e.g. the session timed out) then it'll ask for the password.
   - ^ this PR has also changed a `couple copies` on this "Enter Password" screen:
       - Screen title update: **"Add a new Stellar address" => "Enter Password"**
       - Button title update: **"Add New Address" => "Create New Address"** 

### Creating a new address when password is saved
https://github.com/user-attachments/assets/c7d6f33a-5899-4c3e-bc9e-535d5d5f4115

### Creating a new address when password is NOT saved
https://github.com/user-attachments/assets/83436042-c4a3-4127-8a98-f056d1fd5ca7

### Auth error while creating a new address
https://github.com/user-attachments/assets/3dc6caa4-fcc3-440b-96e9-0b28b7f60c9f


extension/src/popup/views/AddAccount/AddAccount/index.tsx
```diff
@@ -35,30 +36,60 @@ export const AddAccount = () => {
...
   );
 
+  // No need to ask for password if it's already stored, so let's just briefly
+  // display a loading screen while user is being redirected to next screen
+  if (hasPrivateKey) {
+    return (
+      <React.Fragment>
+        <SubviewHeader title="" />
+        {!authError && <Loading />}
+        {authError && <AppError>{authError}</AppError>}
```
piyalbasu, line 83: What are the scenarios where this might happen? I think generally it'd be great to give the user a way to fix whatever went wrong. For ex, if we think that this was due to us somehow having an incorrect password in the `background`, maybe we should show them the `Enter Password` field so they can put the correct password in

extension/src/popup/views/AddAccount/AddAccount/index.tsx
```diff
@@ -35,30 +36,60 @@ export const AddAccount = () => {
...
     [dispatch],
   );
 
+  // No need to ask for password if it's already stored, so let's just briefly
+  // display a loading screen while user is being redirected to next screen
+  if (hasPrivateKey) {
+    return (
+      <React.Fragment>
+        <SubviewHeader title="" />
+        {!authError && <Loading />}
```
aristidesstaffieri, line 82: I'm not sure if this is worth a change or not but just for thought -

We have other places in Freighter that use a similar pattern but also produce the "loading stutter" effect we get when creating an account here. This happens because there are 2 loaders for 1 transition in state.

Is there an easy way to make that feel more like 1 loading state?

CassioMG: @aristidesstaffieri Hummm I think I know what you mean. I _believe_ there is probably a way of making this transition look nicer but I'm _not sure_ it'll be a straightforward fix.

I'm still getting used to the codebase and to the React Web syntax (which is different from React Native syntax) so I think it'd take some time for me to figure this out. In case you think it's not a blocker, I'd like to move ahead "as is" and create a ticket so we could tackle it later. What do you think?

extension/src/popup/views/AddAccount/AddAccount/index.tsx
CassioMG: @piyalbasu Tbh I'm not sure yet about all the scenarios which could result in an auth error here. I was basically trying to guard against this error since [we were already doing it](https://github.com/stellar/freighter/blob/master/extension/src/popup/views/AddAccount/AddAccount/index.tsx#L37). But I see what you mean, in the old UX the user would always have the chance to input the password again. 🤔 

I think I'll make a change so that if there is an auth error the user will be redirected to the "Enter Password" screen with the error message right below the input inbox so we keep the same old UX in this case allowing users to input the password again. What do you think?

extension/src/popup/views/AddAccount/AddAccount/index.tsx
CassioMG: Thanks for the review btw!

aristidesstaffieri: sure thing and yeah I think it's not a blocker, I do wonder if it's better to leave this loading state off for now. It feels a bit less choppy to me and I don't see that we had one in this flow before.

extension/src/popup/views/AddAccount/AddAccount/index.tsx
CassioMG: @piyalbasu on this commit I've changed the error UX to be like this:

extension/src/popup/views/AddAccount/AddAccount/index.tsx
CassioMG: @aristidesstaffieri  on [this commit](https://github.com/stellar/freighter/pull/1685/commits/2df35014371a5ef29b43aecc2d4d7041fb05e4df) I've removed the duplicate spinner so the UX is looking like the below. I think it's looking less chopy now :)

https://github.com/user-attachments/assets/f1ba36ce-aa01-432d-a499-8256a4887265

aristidesstaffieri: nice, thanks for entertaining that. We can think about how to improve the lag between the transition in a follow up.

aristidesstaffieri (approved):

piyalbasu (approved):

PR #1696, opened 2024-11-22
Improve error boundary

### LINKED ISSUE #1595: Improve error boundary
Ou error boundary has issues in some of it's states. It's not always a uniform size and sometimes the action to exit does not work.

aristidesstaffieri:
Closes #1595 

What
Makes the error boundary close in every scenario
Updates error boundary styles to not overflow header and better display error messages

Why
The navigation never works here because the error boundary sits at a higher level than the router. 
Navigation patterns using the global navigation context don't work very well either because of the hash routing model so we would have to reload the window anyways.

### DIFF
--- a/extension/src/popup/components/ErrorBoundary/index.tsx
+++ b/extension/src/popup/components/ErrorBoundary/index.tsx
@@ -5,8 +5,6 @@ import { useTranslation } from "react-i18next";
 
 import IconFail from "popup/assets/icon-fail.svg";
 import { Button } from "@stellar/design-system";
-import { navigateTo } from "popup/helpers/navigate";
-import { ROUTES } from "popup/constants/routes";
 
 import "./styles.scss";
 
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/ErrorBoundary/index.tsx
```diff
@@ -67,28 +63,19 @@ export const UnhandledError = ({
...
-            {t("Got it")}
-          </Button>
-        )}
+        <Button
+          isFullWidth
+          variant="tertiary"
+          size="md"
+          onClick={() => {
+            // https://stackoverflow.com/questions/57854/how-can-i-close-a-browser-window-without-receiving-the-do-you-want-to-close-thi
+            window.open("", "_self", "");
```
piyalbasu, line 72: oh interesting!

piyalbasu (approved):

CassioMG (approved):

PR #1698, opened 2024-11-25
[REFACTOR] History data fetching

### LINKED ISSUE #1672: Fix loading state for account history
**Problem:**
The account history page suffers from a jittery loading state. The reason this happens is because we model the loading state with a boolean, and this value is initialized as `false` then set to `true` while history data is being fetched, then set to `false` after the response is handled.

**Potential Solutions:**
We can "fix" this with a one line change, initializing [the loading flag](https://github.com/stellar/freighter/blob/master/extension/src/popup/views/AccountHistory/index.tsx#L79) to `true`.

This improves the UX but makes the code brittle and less understandable. We can improve this by modeling our data fetching state differently, for example - 

We could model our loading state more accurately with something like
```
export enum RequestState {
  IDLE = 'IDLE',
  LOADING = 'LOADING',
  SUCCESS = 'SUCCESS',
  ERROR = 'ERROR'
}
```
and we could write a hook that is reusable and encapsulates the loading/error/data logic, a simplified version could be -
```
export function useGetHistory() {
  const [state, setState] = React.useState(RequestState.IDLE)
  const [data, setData] = React.useState([] as HistoryItem[])
  const [error, setError] = React.useState(null as null | string)

  async function getHistory(
    ...args
  ) {
    setState(RequestState.LOADING)
    try {

      ...fetch data...

      setState(RequestState.SUCCESS)
      setData(data)
    } catch (error) {
      setState(RequestState.ERROR)
      setError(error)
      setData([])
    }
  }

  return {
    data,
    error,
    state,
    getHistory
  }
}
```

Then the loading state can be applied in the view more accurately and better communicate what it should be like - 
```
{ hookLoadingState === RequestState.IDLE || RequestState.LOADING ? <Loader /> : <..view stuff... /> }
```

aristidesstaffieri:
Closes #1672 

What
Adds `use-get-history` hook to fetch history items.
Uses it in AccountHistory to replace in view data fetching.

Why
We've talked about moving to a hook based data fetching pattern for views that need fresh data on every render. Also, account history suffers from a glitchy loading state due to it's current data fetching/loading pattern.

Todo
We should follow up with design tasks for error and empty states for this screen.

https://github.com/user-attachments/assets/d5b69eef-b9bc-474f-892c-c0d69602cb09


extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -76,14 +76,16 @@ export const AccountHistory = () => {
...
-  const isAccountHistoryLoading =
-    historySegments === null ||
-    accountBalanceStatus === ActionStatus.IDLE ||
-    accountBalanceStatus === ActionStatus.PENDING;
+  useEffect(() => {
+    const getData = async () => {
+      await fetchData();
+    };
+    getData();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aristidesstaffieri, line 87: If we use `fetchData` as a dep of the `useEffect` it triggers another load of the data. React officially[ recommends using an empty array ](https://react.dev/reference/react/useEffect#passing-an-empty-dependency-array)to run effects only on mount.

I think we should remove this lint rule but if we really don't want to we can also use the loading state to step around the extra call, but we won't be able to avoid the extra render.

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -141,86 +143,72 @@ export const AccountHistory = () => {
...
+    getHistoryState.data,
+    publicKey,
+    networkDetails,
+    isHideDustEnabled,
+  ]);
+
+  const showLoader =
+    getHistoryState.state === RequestState.IDLE ||
+    getHistoryState.state === RequestState.LOADING;
+  const hasEmptyHistory = !showLoader || !getHistoryState.data;
```
CassioMG, line 160: should we be checking for `data.length` here?
do we really need to check the `showLoader` value here?

e.g.:
`const hasEmptyHistory = !getHistoryState?.data?.length;`

JakeUrban:
> We should follow up with design tasks for error and empty states for this screen.

What do these cases look like now?

extension/src/popup/views/AccountHistory/index.tsx
CassioMG: I'm fine using the empty array as userEffect dependency, I think it makes sense for cases where we need to run the code only once on mount 👍 

I'm on the fence about removing the lint rule, it could be useful in other cases where it's important to add more deps

aristidesstaffieri:
> > We should follow up with design tasks for error and empty states for this screen.
> 
> What do these cases look like now?

The empty case just prints "No transactions to show" and an error to fetch history results in an empty history pane. I don't think we've ever had UI state for errors on this screen.

extension/src/popup/views/AccountHistory/index.tsx
aristidesstaffieri: we can't rely on the history length alone because then we would show this message during loading or idle states. I believe the correct logic is "show this when data is not loading and there is no history".

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -141,86 +143,72 @@ export const AccountHistory = () => {
...
+
+  if (isDetailViewShowing) {
+    return <TransactionDetail {...detailViewProps} />;
+  }
+
+  return showLoader ? (
```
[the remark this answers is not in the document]
aristidesstaffieri: oh yeah good call, I pulled out `isDetailViewShowing` but not sure why I didn't pull out the loading state. I'll make that change. 👍

extension/src/helpers/hooks/use-get-history.tsx
```diff
@@ -0,0 +1,85 @@
...
+    case "FETCH_DATA_SUCCESS":
+      return { error: null, state: RequestState.SUCCESS, data: action.payload };
+    case "FETCH_DATA_ERROR":
+      return { data: null, state: RequestState.ERROR, error: action.payload };
+    default:
+      return state;
+  }
+};
+
+function useGetHistory(publicKey: string, networkDetails: NetworkDetails) {
```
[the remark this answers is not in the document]
aristidesstaffieri: It's a little different than the example we discussed in the issue but this seemed easy to work with, open to any feedback on this pattern!

extension/src/popup/views/AccountHistory/index.tsx
CassioMG: got it, so maybe something like this? 
`const hasEmptyHistory = !showLoader && !getHistoryState?.data?.length;`

extension/src/popup/views/AccountHistory/index.tsx
CassioMG: +1

extension/src/popup/views/AccountHistory/index.tsx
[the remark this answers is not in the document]
aristidesstaffieri: updated in 8812240f903a3f43507b9630b95c3cbdd1cd0826

extension/src/popup/views/AccountHistory/index.tsx
aristidesstaffieri: actually since you suggested we pull out the loading render branch, it did make it easier to use your original suggestions. Updated in c1612f50a5b634df9cc37b607ff19754169872cd

CassioMG (approved):

extension/src/helpers/hooks/use-get-history.tsx
```diff
@@ -0,0 +1,85 @@
+import { useReducer } from "react";
```
piyalbasu, line 1: NIT: can you call this file useGetHistory.tsx to conform to the naming convention in the rest of the repo?

extension/src/popup/views/AccountHistory/index.tsx
piyalbasu: Agreed - let's keep the lint rule and just disable when we explicitly just want to run on mount

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -141,86 +143,75 @@ export const AccountHistory = () => {
...
+    }
+  }, [
+    getHistoryState.state,
+    getHistoryState.data,
+    publicKey,
+    networkDetails,
+    isHideDustEnabled,
+  ]);
+
+  const showLoader =
```
piyalbasu, line 157: NIT: let's call this `isLoaderShowing` to follow code convention

aristidesstaffieri: yup, renamed in 68003d30e1193bec7f21757fb3cc7afaee619967

piyalbasu (approved):

PR #1699, opened 2024-11-25
redesign the Create New Wallet screens and update tests

### LINKED ISSUE #1401: Redesign onboarding and recovery flows
This includes allowing skipping the recovery , using an easier confirmation, allowing pasting, using the same layout for backup and recovery

piyalbasu:
Closes #1401 

Apologies for the lengthy PR, but there are very few functional changes; the majority of the work here is simply reskinning the onboarding flow to match the new designs.

One functional change here is that we now allow users to skip through the "confirm mnemonic phrase" step by confirming the mnemonic for them.

The next PR will reskin the Import Wallet flow

https://github.com/user-attachments/assets/c3ed9b44-f482-4acb-b4f2-6105f7baec4b

https://github.com/user-attachments/assets/6fba1860-0370-4f4b-884d-2993d5d7147a


extension/src/popup/Router.tsx
```diff
@@ -363,9 +361,6 @@ const Outlet = () => {
           <PublicKeyRoute path={ROUTES.mnemonicPhraseConfirmed}>
             <FullscreenSuccessMessage />
           </PublicKeyRoute>
-          <PublicKeyRoute path={ROUTES.pinExtension}>
-            <PinExtension />
-          </PublicKeyRoute>
```
piyalbasu, line 368: we used to have some logic to redirect users to a specific "You're all set" page on Chrome and a different one on Firefox. We now just have one unified view for both, so this `PinExtension` is now no longer needed

extension/src/popup/components/mnemonicPhrase/ConfirmMnemonicPhrase/index.tsx
```diff
@@ -84,79 +84,111 @@ export const ConfirmMnemonicPhrase = ({
...
+                        wordKey={wordKey}
+                        word={convertToWord(wordKey)}
+                        wordNumber={() => {
+                          const wordIndex = selectedWords.findIndex(
+                            (selectedWord) => selectedWord === wordKey,
+                          );
+
+                          if (wordIndex > -1) {
+                            return String(wordIndex + 1);
+                          }
```
piyalbasu, line 141: logic to add the number of selection next to the word when a user clicks a word

extension/src/popup/views/FullscreenSuccessMessage/index.tsx
```diff
@@ -1,165 +1,91 @@
...
-import { Onboarding, OnboardingHeader } from "popup/components/Onboarding";
+import { Onboarding, OnboardingModal } from "popup/components/Onboarding";
 import { View } from "popup/basics/layout/View";
-import ExtensionIllo from "popup/assets/illo-extension.png";
 
-import "./styles.scss";
-
-// userAgent sniffing is not foolproof so shouldn't expect this method
-// to be 100% accurate.
-const isChrome = () =>
```
piyalbasu, line 20: removing all the logic related to splitting between Chrome and Firefox 🎉

extension/src/popup/views/MnemonicPhrase/index.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useState } from "react";
```
piyalbasu, line 1: this file is just migrating the code for MnemonicPhrase.tsx to index.tsx

extension/src/popup/views/MnemonicPhrase/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    dispatch(confirmMnemonicPhrase(mnemonicPhrase));
+  };
+
+  if (applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED) {
+    return <Redirect to={ROUTES.mnemonicPhraseConfirmed} />;
+  }
+
+  if (mnemonicPhrase) {
+    if (isModalShowing) {
```
piyalbasu, line 45: compare this file to the deleted `MnemonicPhrase.tsx` file above. It's essentially the same while adding everything in this `isModalShowing` conditional

extension/src/popup/views/Preferences/index.tsx
```diff
@@ -36,7 +36,6 @@ export const Preferences = () => {
       isDataSharingAllowedValue,
       isHideDustEnabledValue,
     } = formValue;
-    console.log(formValue);
```
piyalbasu, line 39: unrelated - I just found this errant console.log in the codebase

extension/src/popup/views/AccountCreator/index.tsx
```diff
@@ -67,106 +72,127 @@ export const AccountCreator = () => {
...
+                  </Heading>
+                  <div className="AccountCreator__subheading">
+                    {t("This will be used to unlock your wallet")}
+                  </div>
+                  <Form>
+                    <OnboardingOneCol>
+                      <div className="AccountCreator__form-rows">
+                        <Field name="password">
+                          {({ field }: FieldProps) => (
+                            <Input
```
aristidesstaffieri, line 100: I'm noticing a style inconsistency in the active state for the field. It seems like if you tab from the second password field to the checkbox, the checkbox does not show the "active" border even though it is the currently active element. You can see in the video that once I tab back, it seems to trigger that style on subsequent tab navs to the field. It seems like maybe this style does not get set properly when the form is not in a dirty state?

https://github.com/user-attachments/assets/9dfbf2ca-1e24-45e2-9880-50605541720f

extension/src/popup/components/mnemonicPhrase/DisplayMnemonicPhrase/index.tsx
```diff
@@ -15,67 +11,38 @@ import "./styles.scss";
...
-          variant="secondary"
-          onClick={() => {
-            setIsConfirmed(true);
-          }}
-        >
-          {t("Next")}
-        </Button>
-      </OnboardingOneCol>
-    </OnboardingTwoCol>
+    <OnboardingModal
```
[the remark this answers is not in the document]
piyalbasu: Oh yeah, I should have called that out - we removed the `copy to clipboard` in the last version due to a security vuln. I need to chat with the Security team to think if this is worth bringing back

extension/src/popup/views/AccountCreator/index.tsx
piyalbasu: Good callout! I'll take a look at this

piyalbasu: Hm, I do seem to see an active state when I tab to the checkbox. It's pretty light, but there is a difference. We could explore making this difference more apparent in SDS

Inactive:
Screenshot 2024-11-26 at 3 22 03 PM


Active:
Screenshot 2024-11-26 at 3 21 54 PM

extension/src/popup/components/mnemonicPhrase/DisplayMnemonicPhrase/index.tsx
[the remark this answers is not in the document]
piyalbasu: Here's the thread in the Security channel: https://stellarfoundation.slack.com/archives/C01Q6L6NR36/p1732650914431899

It's looking like we aren't going to be adding this feature

extension/src/popup/views/AccountCreator/index.tsx
aristidesstaffieri: Hmm yeah I see the shadow now.  If you tab back after the first time you get the red active state border on the checkbox but for some reason not on the first time. I don't see any css for this though so it's an artifact of the sds styles for the checkbox.

piyalbasu: Oh the red outline is to signify that it's a required field and you haven't filled it out yet. That styles comes from Formik. So in this instance, that interaction makes sense as you would only want to alert the user that they missed a required field after the first time they tabbed away from the field

PR #1704, opened 2024-12-02
Feature/redesign onboarding recover wallet

piyalbasu:
Relates to #1401 

This redesigns the `Import Wallet` flow during onboarding. This reuses components from the `Create wallet` flow, namely the password entry form and the success screen. I also tweaked some CSS in the entire Onboarding flow to make sure the screens stay vertically centered. That resulted in a couple of updated snapshots in the `Create Wallet` flow

One difference from the designs: I am keeping the `Show` button from the last iteration. It's more secure to use password input fields (rather than plain text inputs as the designs call for) so I'm maintaining the current UX of defaulting the fields to password inputs and allowing a user to opt into using a text field if they choose

https://github.com/user-attachments/assets/a9da8094-ad97-435c-b5f1-024772c7828c

https://github.com/user-attachments/assets/22e31a6e-1f8c-4790-8717-abe7fd8e9ca9


extension/src/popup/components/Onboarding/index.tsx
```diff
@@ -36,16 +36,6 @@ export const Onboarding = ({
   );
 };
 
-interface OnboardingHeaderProps {
```
piyalbasu, line 39: removing some now unused shared components

extension/src/popup/basics/Forms/index.tsx
```diff
@@ -25,10 +25,3 @@ export const SubmitButtonWrapper = ({
     {children}
   </div>
 );
-
-interface ErrorMessageProps {
```
piyalbasu, line 29: this is no longer used

extension/src/popup/components/accountCreator/PasswordForm/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  handleSubmit?: (values: FormValues) => void;
+}
+
+export const initialValues: FormValues = {
+  password: "",
+  confirmPassword: "",
+  termsOfUse: false,
+};
+
+export const PasswordForm = ({
```
piyalbasu, line 46: this is just breaking out some code I wrote for the Create Wallet form into a reusable component

extension/src/popup/views/RecoverAccount/index.tsx
```diff
@@ -162,161 +156,117 @@ export const RecoverAccount = () => {
...
+                                  index={i}
+                                />
+                              ))}
+                            </div>
+                          </div>
+                        </Card>
+                        <div className="RecoverAccount__mnemonic-footer">
+                          <div className="RecoverAccount__phrase-toggle">
+                            <div>{SHORT_PHRASE} word</div>
+                            <Toggle
```
aristidesstaffieri, line 211: It looks like this element is not navigable without a mouse, tabbing through the page goes from the last word input to the show/hide button. We should make this a selection target as well.

piyalbasu: Ah good catch! I'll take a look

aristidesstaffieri (approved):

PR #1706, opened 2024-12-04
Redesign Login and Re-Auth screens

### LINKED ISSUE #1545: Redesign login and re-auth screens
* Redesign for login screen
* Redesign re-authenticate screen (used in send flow when we don't have user's S key)

Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5003&node-type=frame&m=dev

CassioMG:
Closes #1545 

Included in this PR:
- Created reusable `EnterPassword` component to be used on login and re-auth screens
- Redesign of `Login` screen
  - Had to create this new `lastUsedAccount` local storage prop to be able to render the `Identicon` component on login screen (i.e. for logged out users)
- Redesign of `Re-Auth` screens used on `Send`, `Swap` and `Create new address` flows
- Redesign of `View.Header` component to display `new freighter logo` centered on nav bar
  - This `new freighter logo` is now also being used on the `About` screen
- Tweak `View.Content` to have a `horizontal padding of 1.5rem` to match Figma designs
- Wrap `NotFundedMessage` component with `View.Footer` for consistency instead of having to manually set its absolute positioning

### Login screen
Screenshot 2024-12-05 at 15 21 37

### Send and Swap Re-Auth screens
Screenshot 2024-12-05 at 15 32 01

### Create new address Re-Auth screen
Screenshot 2024-12-05 at 15 31 35

### Not funded message on footer component (no apparent change)
Screenshot 2024-12-04 at 20 45 56

### About screen with new logo
Screenshot 2024-12-04 at 20 37 47


extension/src/popup/basics/layout/View/styles.scss
```diff
@@ -134,8 +126,8 @@
     margin: 0 auto;
     // 848px = 796 + 16 * 2 (padding) + 20 (scrollbar)
     max-width: pxToRem(848px);
-    padding-left: 1rem;
-    padding-right: 1rem;
+    padding-left: 1.5rem;
+    padding-right: 1.5rem;
```
piyalbasu, line 130: very nit: would you mind using pxToRem here so we can continue replacing all rem instances with pxToRem?

extension/src/popup/views/VerifyAccount/index.tsx
```diff
@@ -32,70 +28,49 @@ export const VerifyAccount = ({
...
-      </Formik>
+      <div className="VerifyAccount">
+        <EnterPassword
+          accountAddress={publicKey}
+          description={
+            isApproval
+              ? undefined
+              : t(
+                  "Enter your account password to authorize this transaction. You won’t be asked to do this for the next 24 hours.",
+                )
```
piyalbasu, line 67: NIT: I think `You won’t be asked to do this for the next 24 hours.` should be part of the `EnterPassword` component since it's used by both the default `description` and this custom one.

The only reason I even bring it up is because we're hopefully going to make the amount of time configurable soon so it'd be nice to only have to deal with this copy in one place

extension/src/popup/views/Settings/index.tsx
```diff
@@ -25,10 +28,14 @@ export const Settings = () => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
 
+  const publicKey = useSelector(publicKeySelector);
+
   const signOutAndClose = async (e: React.FormEvent) => {
     e.preventDefault();
-    // eslint-disable-next-line
+    /* eslint-disable @typescript-eslint/await-thenable */
+    await dispatch(setLastUsedAccount({ publicKey }));
```
piyalbasu, line 36: A user isn't guaranteed to sign out. Their session may just timeout due to the 24h timer. 

I think maybe we should call `setLastUsedAccount` whenever a user switches account to make sure we always capture the last used account

piyalbasu:
One issue we're going to run into is what happens when a user opens the extension but `setLastUsedAccount` hadn't been called in the past. For ex, I just pulled down this branch, built the code and loaded the extension with this logic for the first. I got this screen:

Screenshot 2024-12-04 at 7 35 07 PM

Maybe we only want to show the identicon circle if we actually already have the last used account

extension/src/popup/basics/layout/View/styles.scss
CassioMG: Sure thing, will do. Is the idea to use pxToRem in 100% of times? Or just in a few specific cases?

extension/src/popup/views/VerifyAccount/index.tsx
CassioMG: Good point. I was in the fence about this too. Do you think we should also display this 24h warning on the login screen below?

Screenshot 2024-12-05 at 09 58 25

extension/src/popup/views/Settings/index.tsx
CassioMG: Nice catch, I'll make this improvement

CassioMG:
> One issue we're going to run into is what happens when a user opens the extension but `setLastUsedAccount` hadn't been called in the past. For ex, I just pulled down this branch, built the code and loaded the extension with this logic for the first. I got this screen:
> 

Oh no, I tested this case by deleting the extension and installing it again. But when you delete it it asks you to import or create a new account which skips this screen so I didn't catch this bug. But I forgot the case of existing users with a timed out session that just updated the app 🤦‍♂️  I'll fix it - thanks for the help testing!

extension/src/popup/basics/layout/View/styles.scss
CassioMG: https://github.com/stellar/freighter/pull/1706/commits/8932350118c187035d2b103d9c278a65da42ae00

CassioMG:
> Maybe we only want to show the identicon circle if we actually already have the last used account

Fixed here https://github.com/stellar/freighter/pull/1706/commits/2f133c100f02c4e78c1b6b0fe428aa410eed2ded
Screenshot 2024-12-05 at 10 14 09

extension/src/popup/views/Settings/index.tsx
CassioMG: How about this https://github.com/stellar/freighter/pull/1706/commits/20c6118673ec72535fdf359a06fbc5ceae582b24?
It's updating the `lastUsedAccount` every time users visit the main Account screen (which also happens every time users change account).

extension/src/popup/basics/layout/View/styles.scss
piyalbasu: Thanks! Yeah, I think it'd be nice to use pxToRem in all cases and hopefully over time, we'll replace all instances of rem with pxToRem for consistency

extension/src/popup/views/VerifyAccount/index.tsx
piyalbasu: Oh yeah, that's a good idea!

extension/src/popup/views/Settings/index.tsx
piyalbasu: Nice, this looks good! You could also add a conditional to only update `lastUsedAccount` if `lastUsedAccount !== publicKey` so you're not dispatching this action if the account hasn't changed

CassioMG: Humm yeah I thought about it too, but then we'd have to dispatch `loadLastUsedAccount()` every time too to fetch `lastUsedAccount` so it just seemed simpler/cleaner to overwrite it every time. Wdyt?

extension/src/popup/views/VerifyAccount/index.tsx
CassioMG: Here we go: https://github.com/stellar/freighter/pull/1706/commits/1ceebe6af730ea04a5952b473835382131939091
Screenshot 2024-12-05 at 15 36 14

extension/src/popup/views/Settings/index.tsx
aristidesstaffieri: I wonder if it would be better to dispatch `setLastUsedAccount` in the action to switch account instead of in this view layer? I realize that right now, every time a user switches accounts that this view re-renders but I think that coupling is loose and this dependency could be easy to miss in a refactor. Thoughts?

CassioMG: Which component could we consider as freighter's "Home" screen? My line of thought was to have the code below on a main component which we know would be rendered right after login and also after users switch account, so that the `publicKeySelector` would react to that and make sure to keep the `lastUsedAccount` value updated.
```

  useEffect(() => {
    if (publicKey) {
      dispatch(setLastUsedAccount({ publicKey }));
    }
  }, [publicKey, dispatch]);
```

If we add this code only to the switch account action I'm afraid users that never switch account wouldn't have the `lastUsedAccount` property set or updated, then in this case we'd need to `setLastUsedAccount` in 2 different places: the switch account action + login action. Which wouldn't be that bad at all, but would be 1 more place to maintain.

Wdyt?

aristidesstaffieri: yeah I think what you're saying makes sense, maybe there's a different way to keep it in 1 place but also in the action layer? My thought was that it would be good to have `lastUsedAccount` be set in 1 place that guarantees it is updated whenever the active account is set to a key. My only concern with doing this in the view is that is not really tied to the action, so right now this works but if we ever add a situation where we switch accounts and don't land back on this view after, we will have to implement this in those views as well where as if we had this in the action layer, it would not matter the view that gets rendered after switching accounts.

aristidesstaffieri: does `setActivePublicKey` get called in every case that we set a public key, maybe if that dispatched an action to set `lastUsedAccount` then we would not have to track that in a view?

CassioMG: I like this idea, I'll take a look

CassioMG: @aristidesstaffieri how about this https://github.com/stellar/freighter/pull/1706/commits/26f11ae0388f159b787c834b4e6040c622abca0e?
It's updating the `lastUsedAccount` every time user subscribes to an account, which happens exactly during login and also when user switches the active account. I've deleted the extension and tested in all scenarios, seems pretty solid!

aristidesstaffieri: nice 👏 yeah that lgtm!

aristidesstaffieri (approved):

Issue #1700, opened 2024-11-26
Add border to QR code


piyalbasu:
Add CSS border to QR code to enhance visibility

https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5872&t=7J6InwSrWH3wCDLf-1

Issue #1543, opened 2024-10-18
Redesign animations


piyalbasu:
Make animations for modals, etc. a bit quicker to allow users to move through the app faster. More details needed from Charles on this

PR #1707, opened 2024-12-05
redesign account view + account details QR code

### LINKED ISSUE #1539: Redesign Account view
Summary:
* bottom icons with title
* fade in account picker instead of sliding down; full width
* animations based on Rainbow
* Dropdown menu with manage assets, fullscreen mode, etc

Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-4983&node-type=frame&m=dev

piyalbasu:
This PR closes #1539 and #1700

It implements the new designs for Account view, without implementing the USD equivalents/gains as we don't have a service for this yet. That means I'm just implementing:
* An updated bottom nav
* An updated account list dropdown
* An updated network selector dropdown

Account view:
Screenshot 2024-12-05 at 3 38 55 PM

Account list dropdown:
Screenshot 2024-12-05 at 3 23 10 PM

Network selector dropdown:
Screenshot 2024-12-05 at 3 38 12 PM

Added white border to QR code (#1700):
Screenshot 2024-12-05 at 3 23 21 PM


sdfcharles:
@piyalbasu looks great!

CleanShot 2024-12-05 at 15 55 33@2x
can 


CleanShot 2024-12-05 at 15 55 28@2x

aristidesstaffieri:
I don't think this is a regression but since we're re-doing the dropdowns and trying to get closer to the UX of rainbow for those, maybe we should consider using a purpose built [drawer component like vaul](https://vaul.emilkowal.ski/) in the future? Our drawers transitions are a bit choppier that you would expect imo.

Maybe this will become more important when Freighter is actually used on mobile devices but libs like that do a really good job of getting the transition right and cover a [bunch of UX edge cases](https://emilkowal.ski/ui/building-a-drawer-component) across platforms.

aristidesstaffieri (approved):

piyalbasu:
> I don't think this is a regression but since we're re-doing the dropdowns and trying to get closer to the UX of rainbow for those, maybe we should consider using a purpose built [drawer component like vaul](https://vaul.emilkowal.ski/) in the future? Our drawers transitions are a bit choppier that you would expect imo.
> 
> Maybe this will become more important when Freighter is actually used on mobile devices but libs like that do a really good job of getting the transition right and cover a [bunch of UX edge cases](https://emilkowal.ski/ui/building-a-drawer-component) across platforms.

This is a great suggestion. We actually have a separate [ticket](#1543) for updating animations because we have a few other places where we do a slide up drawer (like when adding an asset, signing with hardware wallet, etc) that will need to be updated as well. I'm going to add your comment to that ticket. I can hopefully get to it before this sprint is over

PR #1723, opened 2024-12-09
allow signing auth entries using Ledger

### LINKED ISSUE #1352: Unable to hash sign Auth Entry with hardware wallet
See this line in code: https://github.com/stellar/freighter/blob/master/extension/src/popup/views/SignAuthEntry/index.tsx#L63

Should this be able to be signed using hash signing?

piyalbasu:
Closes #1352 

Adding the ability to sign auth entries with Ledger.

This previously wasn't possible due to Ledger not supporting Soroban authorizations. To keep our code pathways a little tidy, I tried to reuse as much of the existing `Sign with Hardware Wallet` code as possible

https://github.com/user-attachments/assets/ec3c8c65-6346-4373-9e10-4c5ac92ac09a


extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -244,9 +245,27 @@ export const signWithHardwareWallet = createAsyncThunk<
...
+      try {
+        const auth = Buffer.from(transactionXDR, "base64");
+
+        const signature = await hardwareSignAuth[walletType]({
+          bipPath,
+          auth,
+          isHashSigningEnabled,
+        });
+
+        return signature;
```
piyalbasu, line 262: returning a signed auth (buffer) is different than returning a signed xdr (string), but combining these allows us to reuse the UX in the view component that handles Ledger error messages/failures

@shared/api/internal.ts
```diff
@@ -942,15 +942,15 @@ export const grantAccess = async (url: string): Promise<void> => {
   }
 };
 
-export const handleSignedHwTransaction = async ({
-  signedTransaction,
+export const handleSignedHwPayload = async ({
```
piyalbasu, line 945: since this can now be a transaction OR a signed auth entry, renaming the function to be more indicative of what it actually does

aristidesstaffieri (approved):
nice 👏

PR #1725, opened 2024-12-10
Feature/dependabot bump all

aristidesstaffieri:
What
Makes dependabot keep package.json up to date with lock changes.
Unrelated translations.

Why
lock files were getting out of sync with package.json files and causing diffs for local installs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
+    versioning-strategy: "bump"
     directories:
       - "/"
       - "/@shared/api"
--- a/extension/src/popup/locales/en/translation.json
+++ b/extension/src/popup/locales/en/translation.json
@@ -6,6 +6,7 @@
   "A Stellar wallet for every website": "A Stellar wallet for every website",
   "About": "About",
   "Account": "Account",
+  "Account details": "Account details",
   "Account ID": "Account ID",
   "Account Migration": "Account Migration",
   "Account minimum balance is too low": "Account minimum balance is too low",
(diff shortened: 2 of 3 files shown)


.github/dependabot.yml
```diff
@@ -1,6 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
+    versioning-strategy: "bump"
```
piyalbasu, line 4: Hm, I actually didn't see a strategy called `bump` in the docs. Maybe I missed something? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

aristidesstaffieri: omg thanks I mis-read this. 
Ok [looking at this again](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) I think it should be `increase`. Does that sound right to you?

extension/src/popup/locales/en/translation.json
```diff
@@ -6,6 +6,7 @@
   "A Stellar wallet for every website": "A Stellar wallet for every website",
   "About": "About",
   "Account": "Account",
+  "Account details": "Account details",
```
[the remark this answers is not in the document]
aristidesstaffieri: I can back it out. This happens in the pre-push hook which I didn't expect to have any diff since I didn't touch any dom nodes so unclear why this happened but I can disable the hook and undo this.

.github/dependabot.yml
piyalbasu: np! yeah, I think `increase` is what we want here

aristidesstaffieri: done in 1b72dd4f9386d9d1e4a947ffc28f2199de29702e

extension/src/popup/locales/en/translation.json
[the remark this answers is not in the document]
piyalbasu: Oh weird - yeah not sure why this would trigger new translations if your branch was based off `master` 🤔

piyalbasu (approved):

PR #1737, opened 2024-12-12
Redesign Account History

### LINKED ISSUE #1542: Redesign Account History
The toggle to show "Pending transactions" will be added in a separate ticket

Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5226&node-type=frame&m=dev

CassioMG:
Closes #1542 

This PR redesigns the Account History screen on both the main History tab and the individual asset History.

Screenshot 2024-12-12 at 17 04 37
Screenshot 2024-12-12 at 17 05 19
Screenshot 2024-12-12 at 17 05 34
Screenshot 2024-12-12 at 17 09 07
Screenshot 2024-12-12 at 17 08 55
Screenshot 2024-12-12 at 17 08 36

### Navigating on main History tab
https://github.com/user-attachments/assets/44e1cf78-9c40-4718-aaa5-6c0fe7e1fcd3

### Navigating on each asset History
https://github.com/user-attachments/assets/6a0f42f1-d950-4429-91ff-278aac41efa1


JakeUrban:
Overall this looks great. Just two questions / suggestions:

- In the balances view, do we need the ticker after the amount, since the ticker is on the left as well?
- On the history view (both the global view and per-token), does it make sense to have the "N/A" icon, as opposed to just leaving that area blank?

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -506,33 +727,47 @@ export const HistoryItem = ({
   return (
     <div
       data-testid="history-item"
-      className="HistoryItem"
+      className="HistoryItem HistoryItem--gray12"
```
piyalbasu, line 730: I think it'd be nice to call this classname something more descriptive than `HistoryItem--gray12` as we may change the color to something other than `--gray12` at some point. For ex, something like `HistoryItem--highlight`

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -506,33 +727,47 @@ export const HistoryItem = ({
...
+        )}
+        {!isLoading && (
+          <div className="HistoryItem__row HistoryItem--space-between">
+            <div className="HistoryItem__row">
+              <div className="HistoryItem__icon">{renderIconComponent()}</div>
+              <Text
+                as="div"
+                size="md"
+                weight="regular"
+                addlClassName="HistoryItem__description HistoryItem--gray12"
```
piyalbasu, line 751: similar comment as above regarding color names in class names

extension/src/popup/views/AccountHistory/index.tsx
```diff
@@ -88,75 +67,79 @@ export const AccountHistory = () => {
...
+          const month = date.getMonth();
+          const year = date.getFullYear();
+          const monthYear = `${month}:${year}`;
 
-        segments[SELECTOR_OPTIONS.ALL].push(historyOperation);
-      });
+          // pop() is a very performant method to get the last element of an array
+          // but it actually removes the last element, so we need to make sure
+          // to put it back later.
+          const lastSection = sections.pop();
```
piyalbasu, line 111: maybe a safer way to do this is to do `lastSection = sections[sections.length -1]` to get the last item without transforming the array;

piyalbasu:
Here are some screenshots of what Soronban invocations look like with the new design. I think we may need to tweak some logic here. For ex, a sent Soroban transfer should use the same airplane icon. And I got different renders for 2 mint transactions when I'd expect them to be the same. (A `mint` is when the owner of a token creates new tokens and then sends them to a user) 

This is for account GBTYAFHGNZSTE4VBWZYAGB3SRGJEPTI5I4Y22KZ4JTVAN56LESB6JZOF on Testnet

Screenshot 2024-12-12 at 5 42 55 PM
Screenshot 2024-12-12 at 5 42 41 PM
Screenshot 2024-12-12 at 5 42 12 PM

CassioMG:
> * In the balances view, do we need the ticker after the amount, since the ticker is on the left as well?

@JakeUrban do you mean removing the token code on the right here? That's actually an interesting idea so we'd have more room for the amount number
Screenshot 2024-12-13 at 10 14 42

> * On the history view (both the global view and per-token), does it make sense to have the "N/A" icon, as opposed to just leaving that area blank?

@JakeUrban I'm not against it. But curious to hear @sdfcharles thoughts on it.

Thanks for reviewing!

CassioMG:
> Here are some screenshots of what Soronban invocations look like with the new design. I think we may need to tweak some logic here. For ex, a sent Soroban transfer should use the same airplane icon. And I got different renders for 2 mint transactions when I'd expect them to be the same. (A `mint` is when the owner of a token creates new tokens and then sends them to a user)
> 
> This is for account GBTYAFHGNZSTE4VBWZYAGB3SRGJEPTI5I4Y22KZ4JTVAN56LESB6JZOF on Testnet

@piyalbasu thanks a lot for the help testing those. I haven't changed our existing Soroban logic on this PR since it looks a bit complex so I basically added the `Icon.User01` component behind the existing arrow icons to keep the pattern. I'll try to repro those same Soroban transactions you submitted with your test user and tweak the logic a bit to accommodate it.

extension/src/popup/views/AccountHistory/index.tsx
CassioMG: https://github.com/stellar/freighter/pull/1737/commits/7bfb35047e5f463cae2155b76350c3f87990822f

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -424,18 +636,26 @@ export const HistoryItem = ({
...
+                {formattedAmount}
+              </Badge>,
             );
             setIconComponent(
-              _isRecipient ? (
-                <Icon.ArrowDown className="HistoryItem__icon--received" />
-              ) : (
-                <Icon.ArrowUp className="HistoryItem__icon--sent" />
-              ),
+              <div className="HistoryItem__icon__bordered HistoryItem--gray09">
```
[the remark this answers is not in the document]
CassioMG: https://github.com/stellar/freighter/pull/1737/commits/893112df39eb381fe715c2e68c821bdb6b58ce21

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
CassioMG: https://github.com/stellar/freighter/pull/1737/commits/893112df39eb381fe715c2e68c821bdb6b58ce21

aristidesstaffieri:
Screenshot 2024-12-13 at 8 45 42 AM

I think these rows have a potential to overflow their columns. Since we intend to show token transfers as sends, I think we should include some overflow/truncation of the columns. A custom token can define their `decimals` to be higher than what we see here which would cause the balances string to be longer.

sdfcharles:
> > * In the balances view, do we need the ticker after the amount, since the ticker is on the left as well?
> 
> @JakeUrban do you mean removing the token code on the right here? That's actually an interesting idea so we'd have more room for the amount number Screenshot 2024-12-13 at 10 14 42
> 
> > * On the history view (both the global view and per-token), does it make sense to have the "N/A" icon, as opposed to just leaving that area blank?
> 
> @JakeUrban I'm not against it. But curious to hear @sdfcharles thoughts on it.
> 
> Thanks for reviewing!

yeah i'm down to remove the ticker here @CassioMG

sdfcharles:
@CassioMG everything lgtm, only one change:

- update these buttons to tertiary stylling (dark bg)

CleanShot 2024-12-13 at 13 05 55

CassioMG:
> Screenshot 2024-12-13 at 8 45 42 AM
> 
> I think these rows have a potential to overflow their columns. Since we intend to show token transfers as sends, I think we should include some overflow/truncation of the columns. A custom token can define their `decimals` to be higher than what we see here which would cause the balances string to be longer.

@aristidesstaffieri that's a good point, thanks for raising it.
So below (image on the left) is how the SDS `Badge` component behaves for long texts. The `Badge` component does not seem very editable (e.g. it does not take `addlClassName` prop as other SDS components do =/) so that apparently we can't configure it to truncate with ellipsis. With that in mind I'm wondering if a a quick fix like reducing the Badge `max width` a little bit could be enough for this issue (image on the right) so that it doesn't get too close to the other label. Wdyt?
Screenshot 2024-12-13 at 18 31 16

aristidesstaffieri:
> > Screenshot 2024-12-13 at 8 45 42 AM
> > I think these rows have a potential to overflow their columns. Since we intend to show token transfers as sends, I think we should include some overflow/truncation of the columns. A custom token can define their `decimals` to be higher than what we see here which would cause the balances string to be longer.
> 
> @aristidesstaffieri that's a good point, thanks for raising it. So below (image on the left) is how the SDS `Badge` component behaves for long texts. The `Badge` component does not seem very editable (e.g. it does not take `addlClassName` prop as other SDS components do =/) so that apparently we can't configure it to truncate with ellipsis. With that in mind I'm wondering if a a quick fix like reducing the Badge `max width` a little bit could be enough for this issue (image on the right) so that it doesn't get too close to the other label. Wdyt? Screenshot 2024-12-13 at 18 31 16

Oh nice I thought the pill would overflow as well so that's not as bad. Yeah it's too bad we can't tweak the classname in order to make the pill truncate the text. I think this works, this should be a less common edge case. Thanks!

CassioMG:
Closing this in favor of https://github.com/stellar/freighter/pull/1785 since this one got too out of sync from master and rebasing this existing branch was too messy. I'll address the remainder PR comments on the new PR.

PR #1739, opened 2024-12-13
Bugfix/5.26.0 qa

### LINKED ISSUE #1736: QA Freighter v5.26.0-beta.0
Release + Context: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1734023353188429

piyalbasu:
Closes #1736 

Addresses comments from Charles.

Many of these fixes are simple CSS fixes that are holdovers from upgrading SDS v2 to v3


extension/src/popup/Router.tsx
```diff
@@ -277,7 +277,7 @@ const Outlet = () => {
     ((location.pathname === ROUTES.welcome &&
       applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED) ||
       SHOW_NAV_ROUTES.some((route) => location.pathname === route) ||
-      isSwap);
+      (isSwap && location.pathname !== ROUTES.unlockAccount));
```
piyalbasu, line 280: ugh - sorry for adding even more twisty logic to this conditional: but we're basically saying that if we show the password auth screen during the Swap flow (bc the user's session happened to time out), don't show the bottom nav

extension/src/popup/components/EnterPassword/index.tsx
```diff
@@ -39,7 +39,7 @@ export const EnterPassword = ({
   const titleLabel = title || t("Enter your password");
   const descriptionLabel = `${
     description || t("Enter your account password to verify your account.")
-  } ${t("You won’t be asked to do this again for the next 24 hours.")}`;
+  }`;
```
piyalbasu, line 42: @CassioMG Heads up that after some more discussion, we decided to remove this line during QA

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -151,8 +151,8 @@ export const AddAsset = () => {
           setVerifiedLists(verifiedTokens[0].verifiedLists);
           setAssetRows(
             verifiedTokens.map((record: VerifiedTokenRecord) => ({
-              code: record.code,
-              issuer: record.issuer,
+              code: record.code || record.contract,
+              issuer: record.issuer || record.contract,
```
piyalbasu, line 155: there is an edge case where a user's asset list can be missing fields but still pass the schema validation Orbit Lens created. In that case, we're adding some fallback values here

piyalbasu: I'll take to OL about tightening the schema rules

extension/src/popup/styles/global.scss
```diff
@@ -113,8 +113,3 @@ a {
 .sds-theme-dark ol {
   color: inherit !important;
 }
-
-// SDS overrides
-.Heading--lg {
-  line-height: normal;
```
piyalbasu, line 119: I double checked that this override was only being used in one place so removing this global override should not have any unexpected results

extension/src/popup/Router.tsx
aristidesstaffieri: makes sense!

aristidesstaffieri (approved):

extension/src/popup/components/EnterPassword/index.tsx
CassioMG: fine by me!

PR #1749, opened 2024-12-17
add API stubbing for addAsset and test gh action

### LINKED ISSUE #1653: [SPIKE] Investigate splitting e2e tests into stubbed vs full e2e mode

piyalbasu:
Closes #1653 

This is a strategy for being able to run our e2e tests in 2 different modes:

1) Run our tests making calls to services like Horizon, stellar.expert, etc. This is useful for making sure our integrations are actually working as expected and our tests are truly mimicking a real user's experience

2) Run our tests stubbing out the calls to external services. This is useful for making sure that we don't get false positives from those services having issues, experiencing network latency, etc. When we run our tests over and over again, we can sometimes hit these type of issues which slow down our development time as we need to investigate whether there's a bug or if it's just the test being flakey. Also, because we're making changes to a real account on Testnet, we can end up in an unexpected state if a test fails or if multiple people are running the same test at once.

This approach will run our tests in Mode 2 (stubbed mode) when we're working on PR's to a `release` branch. We anticipate these tests being run frequently as people open PR's/add commits, so we want these tests to be as stable as possible.

For PR's to `master` (.i.e., a change that we're planning on sending to production soon), we will run these tests in Mode 1, where we can get an actual sense of how the app is working using real services. We anticipate that this version of the tests will run more infrequently - only when a PR has been fully reviewed and approved, and pushed to the release branch.

Let me know your thoughts on this approach and if I'm missing anything here!


extension/e2e-tests/addAsset.test.ts
```diff
@@ -34,13 +64,12 @@ test.skip("Adding unverified Soroban token", async ({ page, extensionId }) => {
...
 });
 test("Adding Soroban verified token", async ({ page, extensionId }) => {
-  const assetsList = await fetch(
-    "https://api.stellar.expert/explorer/testnet/asset-list/top50",
-  );
-  const assetsListData = await assetsList.json();
-  const verifiedToken =
-    assetsListData?.assets[0]?.contract ||
-    "CBIELTK6YBZJU5UP2WWQEUCYKLPU6AUNZ2BQ4WWFEIE3USCIHMXQDAMA";
+  if (!process.env.IS_INTEGRATION_MODE) {
```
piyalbasu, line 67: we key off this param to decide whether to stub or not

.github/workflows/runTests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - run: yarn build:freighter-api
       - run: yarn build:extension
       - run: yarn test:ci
-      - run: yarn test:e2e
+      - run: ${{ github.base_ref == 'master' && 'IS_INTEGRATION_MODE=true' || '' }} yarn test:e2e
```
piyalbasu, line 23: for PR's based on `master`, we'll run the test in integration mode

extension/e2e-tests/addAsset.test.ts
```diff
@@ -34,13 +64,12 @@ test.skip("Adding unverified Soroban token", async ({ page, extensionId }) => {
   await expect(page.getByTestId("account-view")).toContainText("E2E");
 });
 test("Adding Soroban verified token", async ({ page, extensionId }) => {
-  const assetsList = await fetch(
-    "https://api.stellar.expert/explorer/testnet/asset-list/top50",
-  );
-  const assetsListData = await assetsList.json();
-  const verifiedToken =
-    assetsListData?.assets[0]?.contract ||
-    "CBIELTK6YBZJU5UP2WWQEUCYKLPU6AUNZ2BQ4WWFEIE3USCIHMXQDAMA";
```
piyalbasu, line 43: I'm removing some unnecessary data fetching here - it's more reliable to just hardcode the USDC token address

aristidesstaffieri:
I think generally this lgtm. One thought I have about this is that it could be nice to try to get all of the the things that rely on `IS_INTEGRATION_MODE` to toggle modes in one place.
We could import all of those deps into a common test file that uses the toggle and then import the toggled deps into the tests in order to keep from littering this logic all through the tests.

piyalbasu:
> I think generally this lgtm. One thought I have about this is that it could be nice to try to get all of the the things that rely on `IS_INTEGRATION_MODE` to toggle modes in one place. We could import all of those deps into a common test file that uses the toggle and then import the toggled deps into the tests in order to keep from littering this logic all through the tests.

Oh this is an interesting thought - I added a fixture in this commit to centralize all of our stubbing logic. There will probably still be instances where we need to override these base stubs to simulate certain interactions, but I think this might work for simple on/off situations

https://github.com/stellar/freighter/pull/1749/commits/dd34a60f068f0dd6bf233ddb6f41c251341ad9fc

aristidesstaffieri:
> > I think generally this lgtm. One thought I have about this is that it could be nice to try to get all of the the things that rely on `IS_INTEGRATION_MODE` to toggle modes in one place. We could import all of those deps into a common test file that uses the toggle and then import the toggled deps into the tests in order to keep from littering this logic all through the tests.
> 
> Oh this is an interesting thought - I added a fixture in this commit to centralize all of our stubbing logic. There will probably still be instances where we need to override these base stubs to simulate certain interactions, but I think this might work for simple on/off situations
> 
> [dd34a60](https://github.com/stellar/freighter/commit/dd34a60f068f0dd6bf233ddb6f41c251341ad9fc)

nice thanks yeah I recognize that it'll be hard to make this a 100% rule but it should help the readability a bit imo.

aristidesstaffieri (approved):

PR #1752, opened 2024-12-17
[FEATURE] add empty asset list message on manage asset page

### LINKED ISSUE #1619: Empty state for manage assets screen
When opening this screen without any asset, create a state that let's the user know they can start by adding an asset

relates to: https://github.com/orgs/stellar/projects/58/views/6?layout=board&pane=issue&itemId=81405984

aristidesstaffieri:
Closes #1619 

What
Adds empty asset list message on asset management screen.

Why
Missing empty state

Design: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=2932-3383&t=XdeMvXWoCGmqoOrr-0

Screen:
Screenshot 2024-12-17 at 9 59 23 AM

### DIFF
--- a/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
@@ -172,18 +172,27 @@ export const ChooseAsset = ({ balances }: ChooseAssetProps) => {
           </div>
         ) : (
           <div className="ChooseAsset__wrapper">
-            <div
-              className={`ChooseAsset__assets${
-                isManagingAssets && isSorobanSuported ? "--short" : ""
-              }`}
-              ref={ManageAssetRowsWrapperRef}
-            >
-              {isManagingAssets ? (
-                <ManageAssetRows assetRows={assetRows} />
-              ) : (
-                <SelectAssetRows assetRows={assetRows} />
-              )}
-            </div>
+            {!assetRows.length ? (
+              <div className="ChooseAsset__empty">
+                <p>
+                  You have no assets added. Get started by adding an asset
+                  below.
+                </p>
+              </div>
+            ) : (
+              <div
+                className={`ChooseAsset__assets${
+                  isManagingAssets && isSorobanSuported ? "--short" : ""
+                }`}
+                ref={ManageAssetRowsWrapperRef}
+              >
+                {isManagingAssets ? (
+                  <ManageAssetRows assetRows={assetRows} />
+                ) : (
+                  <SelectAssetRows assetRows={assetRows} />
+                )}
+              </div>
+            )}
           </div>
         )}
       </View.Content>
--- a/extension/src/popup/components/manageAssets/ChooseAsset/styles.scss
+++ b/extension/src/popup/components/manageAssets/ChooseAsset/styles.scss
@@ -12,6 +12,19 @@
     left: 0;
   }
 
+  &__empty {
+    display: flex;
+    justify-content: center;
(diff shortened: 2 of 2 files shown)


extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
```diff
@@ -172,18 +172,27 @@ export const ChooseAsset = ({ balances }: ChooseAssetProps) => {
...
+                }`}
+                ref={ManageAssetRowsWrapperRef}
+              >
+                {isManagingAssets ? (
+                  <ManageAssetRows assetRows={assetRows} />
+                ) : (
+                  <SelectAssetRows assetRows={assetRows} />
+                )}
+              </div>
+            )}
```
CassioMG, line 195: It's your call, just a formatting suggestion! :)

```suggestion
            {!assetRows.length && (
              <div className="ChooseAsset__empty">
                <p>
                  {t("You have no assets added. Get started by adding an asset below.")}
                </p>
              </div>
            )}
            {assetRows.length && (
              <div
                className={`ChooseAsset__assets${
                  isManagingAssets && isSorobanSuported ? "--short" : ""
                }`}
                ref={ManageAssetRowsWrapperRef}
              >
                {isManagingAssets && <ManageAssetRows assetRows={assetRows} />}
                {!isManagingAssets && <SelectAssetRows assetRows={assetRows} />}
              </div>
            )}
```

CassioMG (approved):

piyalbasu (approved):

extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
aristidesstaffieri: using `&&` with a length actually [has a few pitfalls](https://menaiala.medium.com/no-one-told-me-that-logical-operator-is-bad-in-react-1723caa46660). Scroll to the title "React meets && operator" for more info.

I tend to use the ternary because of the subtle behavior difference. Wdyt?

CassioMG: that's true, good point. In this case I'd create a const like `const hasAssets = assetRows.length > 0;` and use that instead. But I'll leave it to you!

CassioMG: when using the `&&` operator we need to make sure we are always using a boolean value, that should prevent any unexpected behavior

aristidesstaffieri: that does work but I think having the use of `&&` in jsx is maybe an anti pattern since it makes it easier for someone to change the var in the future without knowing the subtle difference of the expression vs logical operator.

I think if it's all the same to you, the ternary avoids that pitfall altogether. I wanted to propose we add a lint rule against this but that's a different conversation 😄

piyalbasu: My 2 cents is that I like the ternary because it draws a clearer association between the 2 pieces of tsx: in this case, the ternary says to the dev "if you have assetRows, show this, but if not, show that". It makes the code a little more understandable, IMO

aristidesstaffieri: cool thanks for the feedback everyone. It seems like there are a few benefits to keeping the ternary even though it doesn't read as nicely so will keep that for now.

PR #1755, opened 2024-12-19
[BUG] Make signMessage API SEP-43 compliant

### LINKED ISSUE #1634: signMessage API returns wrong type according to SEP43
SEP43 defines the response for signMessage to be 

```
Promise<{
      signedMessage: string;
      signerAddress: string;
    } & error?: Error>,
```
but our response today is 
```
Promise<{
      signedMessage: Buffer like object;
      signerAddress: string;
    } & error?: Error>,
```

https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0043.md

This was originally reported by Enrique here https://github.com/stellar/freighter/issues/1626

aristidesstaffieri:
Closes #1634 

What
`signMessage` returns a Buffer like object for the `signedMessage` key. [The SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0043.md) defines this to return a string instead.

Why
Implementation not aligned with standard.

This should be a major version bump due to the API change but was implemented to be backwards compatible for earlier versions. Versions < 4.0 will still return the old response.

### DIFF
--- a/@stellar/freighter-api/src/signMessage.ts
+++ b/@stellar/freighter-api/src/signMessage.ts
@@ -1,5 +1,4 @@
 import packageJson from "../package.json";
-import semver from "semver";
 import { Buffer } from "buffer";
 
 import { submitMessage } from "@shared/api/external";
@@ -29,18 +28,10 @@ export const signMessage = async (
   }
 ): Promise<SignMessageV3Response | SignMessageV4Response> => {
   if (isBrowser) {
-    const req = await submitMessage(message, opts);
+    const req = await submitMessage(message, packageJson.version, opts);
 
     if (req.error) {
       return { signedMessage: null, signerAddress: "", error: req.error };
-    }
-
-    if (semver.gte(packageJson.version, "4.0.0") && req.signedMessage) {
-      const signedMessage = Buffer.from(req.signedMessage).toString("base64");
-      return {
-        signedMessage,
-        signerAddress: req.signerAddress,
-      };
     }
 
     return {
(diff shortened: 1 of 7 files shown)


@stellar/freighter-api/src/signMessage.ts
```diff
@@ -1,26 +1,48 @@
...
-> => {
+): Promise<SignMessageV3Response | SignMessageV4Response> => {
   if (isBrowser) {
     const req = await submitMessage(message, opts);
 
     if (req.error) {
       return { signedMessage: null, signerAddress: "", error: req.error };
     }
 
+    if (semver.gte(packageJson.version, "4.0.0") && req.signedMessage) {
```
JakeUrban, line 38: Why not hold off on merging this PR until we're ready to release 4.0, rather than have code thats version-aware?

aristidesstaffieri: actually yeah on second thought, the code that should be version aware should be in the background script's message handler. We'll have 3.x and 4.x versions of the API installed on machines out there for some time so we do need to be backwards compatible here but I did it in the wrong place.

aristidesstaffieri: ok I fixed this in fe25f4c68bc0194700837264bb97b629f0c9f554

@stellar/freighter-api/src/signMessage.ts
```diff
@@ -1,21 +1,34 @@
...
   }
-): Promise<
-  { signedMessage: Buffer | null; signerAddress: string } & {
-    error?: FreighterApiError;
-  }
-> => {
+): Promise<SignMessageV3Response | SignMessageV4Response> => {
   if (isBrowser) {
-    const req = await submitMessage(message, opts);
+    const req = await submitMessage(message, packageJson.version, opts);
```
piyalbasu, line 31: nice - good workaround!

piyalbasu (approved):

PR #1758, opened 2025-01-02
Feature/show xdr during send swap

### LINKED ISSUE #1715: Show full xdr and signatures during send/swap flow
Design ticket: https://github.com/stellar/freighter/issues/1570

Designs: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=2859-10870&t=RtzfYbDtnpaOhi6K-1

piyalbasu:
closes #1715 

Shows the user the XDR that's about to be submitted to the network in the send/swap flow

Screenshot: https://github.com/stellar/freighter/blob/e68a8378451afe36a608c40063862c9a0e8011e2/extension/e2e-tests/sendPayment.test.ts-snapshots/send-payment-confirm-chromium-darwin.png


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -218,6 +218,7 @@ const getBuiltTx = async (
   fee: string,
   transactionTimeout: number,
   networkDetails: NetworkDetails,
+  memo: string,
```
aristidesstaffieri, line 221: should this be optional?

piyalbasu: Oh this is a good catch. I checked why this wasn't flagged by TS and it's because we actually have a deeper issue [here](https://github.com/stellar/freighter/blob/master/extension/src/popup/ducks/transactionSubmission.ts#L567) where in redux we don't define `memo` as optional.

Looks like we handle memo being optional properly in the code, we just mislabeled the type. Fixing that in this commit: https://github.com/stellar/freighter/pull/1758/commits/7558875de0fa0351d1eb386a2941545708cc1b98

aristidesstaffieri: ah nice, thanks for digging into that! lgtm

aristidesstaffieri (approved):

PR #1773, opened 2025-01-10
wrap sac match logic in try catch to gracefully fall through

piyalbasu:
User reported issue of SAC transfer to C address throwing an uncaught error when creating Asset instance. `try...catch` will catch the error and allow for gracefully falling through to next if statement

### DIFF
--- a/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
+++ b/extension/src/popup/components/accountHistory/HistoryItem/index.tsx
@@ -7,8 +7,6 @@ import camelCase from "lodash/camelCase";
 import { Icon, Loader } from "@stellar/design-system";
 import { BigNumber } from "bignumber.js";
 import { useTranslation } from "react-i18next";
-import { Asset } from "stellar-sdk";
-import * as Sentry from "@sentry/browser";
 
 import { OPERATION_TYPES } from "constants/transaction";
 import { SorobanTokenInterface } from "@shared/constants/soroban/token";
@@ -18,9 +16,9 @@ import { emitMetric } from "helpers/metrics";
 import {
   formatTokenAmount,
   getAttrsFromSorobanHorizonOp,
-  isContractId,
 } from "popup/helpers/soroban";
 import { formatAmount } from "popup/helpers/formatters";
+import { getBalanceByKey } from "popup/helpers/balance";
 
 import {
   AccountBalancesInterface,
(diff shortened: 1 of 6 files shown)


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -229,13 +230,20 @@ export const HistoryItem = ({
...
+            try {
+              const sacAddress = new Asset(code, issuer).contractId(
+                networkDetails.networkPassphrase,
+              );
+              const matchesSac = attrs?.contractId === sacAddress;
+              return matchesSac;
+            } catch (e) {
+              console.error(e);
+              Sentry.captureException(
+                `Error checking for SAC match with code ${code} and issuer ${issuer}. Error: ${e}`,
```
piyalbasu, line 244: Sentry did not automatically catch this error previously, so forcing Sentry to report here with details of the params that are causing the issue

aristidesstaffieri: nice thanks, I do think this logic is wrong too. 
`((issuer && !isContractId(issuer)) || code === "XLM")`
I think it should be 
`(!isContractId(issuer) || code === "XLM")`

Seems like right now we hit this if statement if we have no issuer where I believe we expect to skip this branch if we have no issuer. Wdyt?

Maybe better to keep this separate but wanted to note it.

extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -229,13 +230,20 @@ export const HistoryItem = ({
...
-          // if issuer if a G address or xlm, check for a SAC match
+          // if issuer is a G address or xlm, check for a SAC match
           if ((issuer && !isContractId(issuer)) || code === "XLM") {
-            const sacAddress = new Asset(code, issuer).contractId(
-              networkDetails.networkPassphrase,
-            );
-            const matchesSac = attrs?.contractId === sacAddress;
-            return matchesSac;
+            try {
+              const sacAddress = new Asset(code, issuer).contractId(
```
piyalbasu, line 236: The error user received: `Asset code is invalid: Maximum alphanumeric, 12 characters at max`.

Somehow a long string is being passed as `code` here

aristidesstaffieri: it looks like the regex used here will also fail an empty string
https://github.com/stellar/js-stellar-base/blob/1036eabf1f20a1154297c419fc2c663040338b22/src/asset.js#L18

I think it's possible that somehow an asset in the list is `:C...`.

aristidesstaffieri: This pattern doesn't really guard against malformed strings for `code:issuer`, should it?
https://github.com/stellar/freighter/blob/master/extension/src/popup/components/accountHistory/HistoryItem/index.tsx#L228

aristidesstaffieri:
Screenshot 2025-01-10 at 11 38 35 AM

PR #1797, opened 2025-01-16
[DONT MERGE] Hide assets from views

### LINKED ISSUE #1620: UX for hide tokens flow

aristidesstaffieri:
Closes #1620 
Design: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=2932-3383&t=KhdFFpbyyxRgbwuo-1

What
Adds a "Toggle Assets" page, where users can select assets to hide from their views.

Why
Users may no longer want to see assets where they have "dust" and no longer care or for whatever other reason they may want to remove the asset from view while still keeping it in their assets list.

Notable Changes:
Adds new internal actions `GET_HIDDEN_ASSETS` & `CHANGE_ASSET_VISIBILITY` which operate on a new storage key `HIDDEN_ASSETS`. `HIDDEN_ASSETS` is a map of `<issuer key, "hidden" | "visible">`.
Adds new "toggle assets" page and navigation action from the "Manage Assets" page.
Changes `getAccountBalances` to accept a new arg, `showHidden`. This can be used to filter out assets according to their state in `HIDDEN_ASSETS`.
Refactors `ChooseAsset` to fetch it's own balances, it now needs to account for hidden assets. 
Extracts a new hook, `useFetchDomains` to be used in both `ChooseAsset` and `AssetVisibility`.
Adds new test for toggling an asset's visibility.

Screenshot 2025-01-16 at 12 15 38 PM

Screenshot 2025-01-16 at 12 15 57 PM


extension/src/popup/components/manageAssets/ChooseAsset/index.tsx
```diff
@@ -1,178 +1,86 @@
...
-    balances,
-    isManagingAssets,
-    isSorobanSuported,
-    isSwap,
-    isSoroswapEnabled,
-    assetSelect.isSource,
-    soroswapTokens,
-    networkDetails,
-  ]);
+  const goBack = () => {
```
aristidesstaffieri, line 47: Now that we need to fetch balances here, this pattern suffers more from the stale store state so this resets the store in order for the state to be correct on the Account page once navigation lands there. I noted a possible refactor to help this in the new hook `fetchDomains`.

extension/src/popup/helpers/useFetchDomains.ts
```diff
@@ -0,0 +1,163 @@
...
+    useSelector(transactionSubmissionSelector);
+  const isSorobanSuported = useSelector(settingsSorobanSupportedSelector);
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
+
+  const isManagingAssets = assetSelect.type === AssetSelectType.MANAGE;
+  const { balances } = accountBalances;
+
+  const [assets, setAssets] = useState({
+    assetRows: [] as ManageAssetCurrency[],
+    // TODO: REFACTOR getAccountBalances to a hook
```
aristidesstaffieri, line 33: I think it would be good to line up `getAccountBalances` to move to a hook similar to `useGetHistory`. This shared state made it tricky to ensure the correct loading/initial state across several views sharing it. Each view relies on unrelated views to "clean up" the state in order for the pages to work correctly.

extension/src/popup/views/Account/index.tsx
```diff
@@ -53,14 +49,8 @@ import { Loading } from "popup/components/Loading";
 import { NotFundedMessage } from "popup/components/account/NotFundedMessage";
 
 import "popup/metrics/authServices";
-
 import "./styles.scss";
 
-export const defaultAccountBalances = {
```
aristidesstaffieri, line 59: Just getting rid of a little cruft here.

extension/src/popup/views/ManageAssets/index.tsx
```diff
@@ -1,56 +1,28 @@
...
-  AssetSelectType,
-} from "popup/ducks/transactionSubmission";
 import { ChooseAsset } from "popup/components/manageAssets/ChooseAsset";
 import { SearchAsset } from "popup/components/manageAssets/SearchAsset";
 import { AddAsset } from "popup/components/manageAssets/AddAsset";
+import { AssetVisibility } from "popup/components/manageAssets/AssetVisibility";
 import { PrivateKeyRoute } from "popup/Router";
 import { ROUTES } from "popup/constants/routes";
 
-export const ManageAssets = () => {
```
aristidesstaffieri, line 15: Both views that need balances here now need to fetch them on render, this is because `AssetVisibility` needs to not filter hidden assets, and `ChooseAsset` needs to take the hidden assets into account.

@shared/api/internal.ts
```diff
@@ -1529,3 +1532,39 @@ export const simulateTransaction = async (args: {
...
+}) => {
+  let response = {
+    error: "",
+    hiddenAssets: {} as Record<IssuerKey, AssetVisibility>,
+  };
+
+  response = await sendMessageToBackground({
+    type: SERVICE_TYPES.CHANGE_ASSET_VISIBILITY,
+    assetVisibility: {
+      issuer: assetIssuer,
```
piyalbasu, line 1564: We'll run into an issue here when a G address is the issuer for multiple assets.

Here's an example of an issuer on Mainnet who issues multiple assets under one address: GBX23RCWKV7DA23J2MA2OFMTRV3XZHHKOTHBQF6AKOEP7AGQUIZZXLMG

I think you might just want to use `{assetCode}:{assetIssuer}` as your key here

extension/src/popup/components/manageAssets/ToggleAssetRows/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    code={code}
+                    issuer={issuer}
+                    image={image}
+                    domain={domain}
+                    name={name}
+                    isSuspicious={isSuspicious}
+                  />
+                  <Toggle
+                    checked={isAssetVisible(hiddenAssets, issuer)}
+                    id="isVisible"
```
piyalbasu, line 70: NIT: you'll potentially end up with multiple elements with the same `id`. Not sure how much that actually matters here - but to be syntactically correct, you might wanna use something like `isVisible-${code}:${issuer}`

extension/src/popup/components/manageAssets/ToggleAssetRows/styles.scss
```diff
@@ -0,0 +1,67 @@
...
+    align-items: center;
+    height: 100%;
+
+    .EmptyMessage {
+      display: flex;
+      align-items: center;
+      color: var(--sds-clr-gray-09);
+
+      svg {
+        margin-right: 0.25rem;
```
piyalbasu, line 14: NIT: would you mind using pxToRem? It'd be nice to start removing `rem` altogether from the codebase over time

extension/src/popup/helpers/useFetchDomains.ts
piyalbasu: Yeah, good call - that'd be a much nicer improvement than having to manually rehydrate

extension/src/popup/views/Account/index.tsx
```diff
@@ -175,7 +165,6 @@ export const Account = () => {
   return (
     <>
       <AccountHeader
-        // accountDropDownRef={accountDropDownRef}
```
piyalbasu, line 178: 🎉

extension/src/popup/views/__tests__/ManageAssets.test.tsx
```diff
@@ -864,4 +854,28 @@ describe("Manage assets", () => {
...
+    expect(childInput).toBeEnabled();
+
+    if (!childInput) {
+      throw new Error("toggle not found");
+    }
+
+    expect(childInput).toBeChecked();
+    fireEvent.change(childInput, { target: { checked: false } });
+    expect(childInput).not.toBeChecked();
+  });
```
piyalbasu, line 880: It might be nice to have an e2e test for this. It would cover the UI stuff plus make sure the background is properly saving hidden assets

@shared/api/internal.ts
Mateusz9611: @

aristidesstaffieri: @piyalbasu very good point thank you, will update to use the composite key. 
Can an issuer have two assets that use the same ticker?

extension/src/popup/components/manageAssets/ToggleAssetRows/index.tsx
aristidesstaffieri: ah yeah I didn't use the ID in this case but will update for completeness

extension/src/popup/views/__tests__/ManageAssets.test.tsx
aristidesstaffieri: yeah good call it would make sense to navigate to the account view and test for the hidden asset being missing from the list also, let me see what I can add here.

@shared/api/internal.ts
aristidesstaffieri: done in 0512f6cfa6098b46575df70a61a7f180d26d8564

extension/src/popup/components/manageAssets/ToggleAssetRows/index.tsx
aristidesstaffieri: done in 0512f6cfa6098b46575df70a61a7f180d26d8564

extension/src/popup/components/manageAssets/ToggleAssetRows/styles.scss
aristidesstaffieri: done in 0512f6cfa6098b46575df70a61a7f180d26d8564

extension/src/constants/localStorageTypes.ts
```diff
@@ -23,3 +23,4 @@ export const IS_HASH_SIGNING_ENABLED_ID = "isHashSigningEnabled";
 export const IS_NON_SSL_ENABLED_ID = "isNonSSLEnabled";
 export const IS_BLOCKAID_ANNOUNCED_ID = "isBlockaidAnnounced";
 export const IS_HIDE_DUST_ENABLED_ID = "isHideDustEnabled";
+export const HIDDEN_ASSETS = "hiddenAsset";
```
CassioMG, line 26:
```suggestion
export const HIDDEN_ASSETS = "hiddenAssets";
```

extension/src/popup/components/manageAssets/AssetVisibility/styles.scss
```diff
@@ -0,0 +1,54 @@
...
+    flex-direction: column;
+    height: 100%;
+  }
+
+  &__assets {
+    flex-grow: 1;
+
+    &--short {
+      flex-grow: 1;
+    }
```
CassioMG, line 42: does it make sense to have this one since it's not adding any actual change?

extension/src/popup/components/manageAssets/ToggleAssetRows/styles.scss
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+  }
+  &__scrollbar {
+    height: 100%;
+  }
+  &__content {
+    display: flex;
+    flex-direction: column;
+    gap: 1.5rem;
```
CassioMG, line 27: should we use pxToRem() in all values from this file?

extension/src/constants/localStorageTypes.ts
aristidesstaffieri: 😄 thanks

extension/src/popup/components/manageAssets/AssetVisibility/styles.scss
aristidesstaffieri: oh good call this came from a similar view but its not needed here. Removing.

extension/src/popup/components/manageAssets/ToggleAssetRows/styles.scss
aristidesstaffieri: Yup! Adding...

@shared/api/internal.ts
piyalbasu: Oh, that's a good question about multiple assets using the same ticket. Off the top of my head, I'm not 100% sure. We may need to test it out and see if the network lets us do it

extension/src/constants/localStorageTypes.ts
aristidesstaffieri: done in 365d79bd01eaab8997857dee77402b4cda033a03

extension/src/popup/components/manageAssets/AssetVisibility/styles.scss
aristidesstaffieri: done in 365d79bd01eaab8997857dee77402b4cda033a03

extension/src/popup/components/manageAssets/ToggleAssetRows/styles.scss
aristidesstaffieri: done in 365d79bd01eaab8997857dee77402b4cda033a03

PR #1799, opened 2025-01-17
checks for the existence of a version before doing semver check

aristidesstaffieri:
What
Version 4.0 of the API introduced a breaking change in order to align with the SEP for `signMessage`.
The intention was for this change to be backwards compatible but it fails to work with older versions of the extension because of a faulty semver check.

### DIFF
--- a/docs/docs/playground/signMessage.mdx
+++ b/docs/docs/playground/signMessage.mdx
@@ -3,7 +3,7 @@ id: signMessage
 title: signMessage
 ---
 
-#### `signMessage(message: string, opts: { address: string }) -> <Promise<{ signedMessage: Buffer | null; signerAddress: string; } & { error?: string; }>>`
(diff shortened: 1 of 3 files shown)


docs/docs/playground/signMessage.mdx
```diff
@@ -3,7 +3,7 @@ id: signMessage
 title: signMessage
 ---
 
-#### `signMessage(message: string, opts: { address: string }) -> <Promise<{ signedMessage: Buffer | null; signerAddress: string; } & { error?: string; }>>`
+#### `signMessage(message: string, opts: { address: string }) -> <Promise<{ signedMessage: string | null; signerAddress: string; } & { error?: string; }>>`
```
piyalbasu, line 6: Can you plz add an explainer on how to convert the b64 string back to a Buffer if needed?

aristidesstaffieri: Added in 10b33fa89ef4703f4ad145bd7e8ae175753fd6a0
Screenshot 2025-01-17 at 10 02 32 AM

piyalbasu (approved):

Issue #1646, opened 2024-11-12
“Add Token” via freighter-api


sdfcharles:
Details: https://docs.google.com/document/d/1d4qZ2nCSauPbStRgfQ73JT5SqvT_-JQ972WWeicG0W4/edit?usp=sharing

This ticket encompasses 3 parts:
- Adding the API method
- Adding the UI that Charles has designed
- Updating the docs so dev know how to implement this API method themselves

sdfcharles:
Designs: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=2932-3383&t=35EKICTeCYU1MrVS-1

JakeUrban:
@sdfcharles The 'Transaction Info' section is a bit confusing to me, since there is no transaction being signed to add the asset to the user's balances within Freighter, right? Are there any other fields other than amount that can be shown within that section? If not, I would consider changing it to 'Balance Info' to show the current balance credited to the user's address.

piyalbasu:
> [@piyalbasu](https://github.com/piyalbasu) I'm trying to get a sense for what is in-scope for this task. I assume this task includes the work to add the `addToken` method to the freighter API dapps use, as well as the updates needed to the API documentation?
> 
> What about adding the `addToken` method to the SEP-43 spec? It may be wise to decouple this work from this issue since it may take some time to get input from other wallet builders.

This ticket encompasses 3 parts:

1) Adding the API method 
2) Adding the UI that Charles has designed
3) Updating the docs so dev know how to implement this API method themselves

I think the SEP work should be handled separately as we’ll need to present the issue to the ecosystem, collect feedback, and then possibly make changes.

JakeUrban:
Lets create a task to kick off the SEP discussion if we don't have that already

CassioMG:
@sdfcharles could we have a "Token not found" screen for this flow to handle cases where it's not found? E.g. if an user passes a random C address which does not correspond to a Token - Thanks

PR #1815, opened 2025-01-25
“Add Token” via freighter-api

CassioMG:
Related issue: https://github.com/stellar/freighter/issues/1646

This PR adds the `addToken` functionality via freighter-api, it uses the existing `signTransaction` and `signMessage` APIs as inspiration.

The `addToken` API takes both a `contractId` and a `networkPassphrase` as input which Freighter will use to load the token details (e.g. `symbol, name, decimals and balance`) for the user.

Freighter will then show these token details along with any applicable warnings. The user will need to approve it in order for the token to be added.

This PR extracts the `token lookup` and `change trustline` functionalities into the [`useTokenLookup`](https://github.com/stellar/freighter/pull/1815/files#diff-91491052e1db4956dc122844de0bb5947c3aa5822cd871b588328dad8cf6a67fR23) and [`useChangeTrustline`](https://github.com/stellar/freighter/pull/1815/files#diff-f604925e030dfaef86db8ce4442da3cbd2dd71656abe6251fb55a0dc12d6ae87R26) hooks respectively so we can reuse the "add token" logic in both our API and manual approaches.

### Adding a custom token with the `addToken` API

https://github.com/user-attachments/assets/581fbf13-5c9d-48eb-8276-ce521ce19125

### Adding a SAC token with the `addToken` API

https://github.com/user-attachments/assets/4e48baef-c5f1-4440-ae3a-31e4fb198ac1

### Asset lists and Blockaid warning messages

Screenshot 2025-02-03 at 16 34 53

Screenshot 2025-02-03 at 16 21 31

Screenshot 2025-02-03 at 16 18 50

Screenshot 2025-02-03 at 16 22 57

Screenshot 2025-02-03 at 16 23 24

### Documentation update

Screenshot 2025-02-03 at 17 50 10
Screenshot 2025-02-03 at 17 50 28


@stellar/freighter-api/src/addToken.ts
```diff
@@ -0,0 +1,23 @@
...
+  error?: FreighterApiError;
+}> => {
+  if (isBrowser) {
+    const req = await submitToken(args);
+
+    if (req.error) {
+      return { error: req.error };
+    }
+
+    return {};
```
aristidesstaffieri, line 19: should this return a response that indicates a success? I think since we don't implement something like a response code it might be nice for clients to get back something other than `{}`.

extension/src/background/helpers/account.ts
```diff
@@ -238,11 +239,9 @@ export const subscribeAccount = async (publicKey: string) => {
     const options = {
       method: "POST",
       headers: {
-        // eslint-disable-next-line
```
aristidesstaffieri, line 241: I think you'll need these due to the lint rule for object keys

extension/src/background/helpers/account.ts
```diff
@@ -301,20 +302,28 @@ export const subscribeTokenBalance = async (
...
+}: {
+  publicKey: string;
+  contractId: string;
+  network: string;
+}) => {
   try {
     const options = {
       method: "POST",
       headers: {
-        // eslint-disable-next-line
```
aristidesstaffieri, line 312: same as the other eslint comment

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -210,7 +294,7 @@ export const freighterApiMessageListener = (
         });
       }
 
-      const server = stellarSdkServer(networkUrl, networkPassphrase);
+      const server = stellarSdkServer(networkUrl, networkPassphrase || "");
```
aristidesstaffieri, line 297: is this just need to apease the type checker? I think we should always have a`networkPassphrase` here.

extension/src/popup/helpers/useSetupAddTokenFlow.ts
```diff
@@ -0,0 +1,97 @@
...
+} => {
+  const [isConfirming, setIsConfirming] = useState(false);
+  const [isPasswordRequired, setIsPasswordRequired] = useState(false);
+
+  const dispatch: AppDispatch = useDispatch();
+  const hasPrivateKey = useSelector(hasPrivateKeySelector);
+
+  const { changeTrustline } = useChangeTrustline({ assetCode, assetIssuer });
+
+  const rejectAndClose = () => {
```
aristidesstaffieri, line 43: is this the same as the prop by the same name?

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1071,6 +1065,38 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
+      const tokenInfo = tokenQueue.pop();
+
+      if (!tokenInfo?.contractId) {
+        throw Error("Missing contract id");
+      }
+
+      const response = await addTokenWithContractId({
+        contractId: tokenInfo.contractId,
+        network: networkDetails.network,
+        publicKey: sourceKeys.publicKey(),
```
piyalbasu, line 1085: an easier way to get the public key is to just use `publicKeySelector`. That way, you don't need to bother with the private key at all. In fact, it's preferable because I'm about to open a PR that removes `privateKeySelector` 😃

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -62,120 +46,18 @@ export const AddAsset = () => {
...
-    setIsSearching(false);
-    setIsVerificationInfoShowing(isAllowListVerificationEnabled);
-  };
+  const { handleTokenLookup } = useTokenLookup({
+    setAssetRows,
+    setIsSearching,
+    setIsVerifiedToken,
+    setIsVerificationInfoShowing,
+    setVerifiedLists,
+  });
```
piyalbasu, line 60: 🎉 🎉 🎉

@stellar/freighter-api/src/addToken.ts
```diff
@@ -0,0 +1,23 @@
+import { submitToken } from "@shared/api/external";
+import { FreighterApiError } from "@shared/api/types";
+import { FreighterApiNodeError } from "@shared/api/helpers/extensionMessaging";
+import { isBrowser } from ".";
+
+export const addToken = async (args: {
+  contractId: string;
+  networkPassphrase?: string;
```
piyalbasu, line 8: if you want to keep this param optional (which makes sense; in our API methods, we tend to keep network passphrase optional), I think you'll need some logic [here](https://github.com/stellar/freighter/pull/1815/files#diff-96f7a3637d5d67d1c18eb2b9f7c8312edd3eaa1706e0c0467f922626010c7271R145) to handle what to do if the network passphrase isn't included. In `submitTx`, for ex, we just default it to Mainnet

extension/src/background/helpers/account.ts
CassioMG: I've disabled the naming convention globally on this file [here](https://github.com/stellar/freighter/pull/1815/files#diff-337c0386fa91dd3917d65696501c2ab7b726af5b72dfda3615d07a8a49293e65R1) for simplicity given we were using many "next-line" disables. Wdyt?

extension/src/popup/helpers/useSetupAddTokenFlow.ts
CassioMG: by prop do you mean this [returned prop](https://github.com/stellar/freighter/pull/1815/files#diff-e5c5a0337e655c2badcce2f7f200f2ae8e016bb30629ccf76b82132a9c0bbcb7R33)?

extension/src/background/messageListener/popupMessageListener.ts
CassioMG: https://github.com/stellar/freighter/pull/1815/commits/2188240c6033a6ba5d3409c14dc3f86899345f30

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -130,6 +140,80 @@ export const freighterApiMessageListener = (
...
+          `/index.html#/add-token?${encodedTokenInfo}`,
+        ),
+        ...WINDOW_SETTINGS,
+      });
+
+      return new Promise((resolve) => {
+        if (!popup) {
+          resolve({
+            // return 2 error formats: one for clients running older versions of freighter-api, and one to adhere to the standard wallet interface
+            apiError: FreighterApiInternalError,
```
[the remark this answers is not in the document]
CassioMG: https://github.com/stellar/freighter/pull/1815/commits/04c21cd950f76a1997462ed8438a620a1073b62d

@stellar/freighter-api/src/addToken.ts
CassioMG: Nice catch! How about returning the contractId? https://github.com/stellar/freighter/pull/1815/commits/1addc55863ef331a62604e882e87db14eef2fe94

@stellar/freighter-api/src/addToken.ts
CassioMG: Nice catch, how does this look? https://github.com/stellar/freighter/pull/1815/commits/8e0e204937d76bf13ce0a073113a43bf644c06fe

@stellar/freighter-api/src/addToken.ts
aristidesstaffieri: NIce thanks, that response lgtm

extension/src/background/helpers/account.ts
aristidesstaffieri: well curious to hear from Piyal too but I think we cant to keep the rule, with the exception being objects that we need to send to the backend which require a snake case key. Wdyt?

extension/src/popup/helpers/useSetupAddTokenFlow.ts
aristidesstaffieri: yeah so it looks like this takes a prop `rejectAndClose` and then we also define a local function by the same name, and return that local function at the end. So it seems we shadow the prop by name but don't use it.

extension/src/background/helpers/account.ts
CassioMG: I think there is typo here which is making the sentence a little confusing, just want to make sure I get the correct idea: "I think we cant to keep the rule"

extension/src/popup/helpers/useSetupAddTokenFlow.ts
CassioMG: It does not take a rejectAndClose prop, that's the function return type. I'll rewrite it to make it more readable!

CassioMG: https://github.com/stellar/freighter/pull/1815/commits/3cc5b015c76a42f04715960a75597c188c452725

extension/src/background/helpers/account.ts
CassioMG: Got it. I'm fine disabling only the necessary lines as well. Let my try making it in blocks to see how it looks

CassioMG: https://github.com/stellar/freighter/pull/1815/commits/49aaacc4322e6f37b1e484436a35842c176f68aa

extension/src/popup/helpers/useSetupAddTokenFlow.ts
aristidesstaffieri: Ah sorry you're right I was confused by the diff. Apologies!

extension/src/background/messageListener/freighterApiMessageListener.ts
aristidesstaffieri: yeah I think works as well but was mainly curious why we could not just do
`const server = stellarSdkServer(networkUrl, networkPassphrase);`

Seems like `networkPassphrase` is a non optional string in the function signature.

CassioMG: Got it.

So this PR [changes the ExternalRequestBase interface a little bit ](https://github.com/stellar/freighter/pull/1815/files#diff-4df78ad811c49ead77d71379dcbeab4deb38073ad58a12b18bd88d3706ba5c99R105-R110) to consider `networkPassphrase` as optional since on the [external API calls](https://github.com/stellar/freighter/pull/1815/files#diff-a07d7b44912cd9de72ec5cbce0973a3ffa42d8f1d9d80096eb4de1efcd044300R69-L56) they are also optional and the code was setting them as empty string `""` in case it was not passed. I thought about adjusting it to increase consistency so that the optional params are actually treated as optional so devs don't need to worry if an empty string is being forwarded as a "valid" value. How does that sound?

I've just pushed [this commit](https://github.com/stellar/freighter/pull/1815/commits/38d05a142d370a458d65413422af145255ceefeb) to remove the empty strings since the param interfaces are already considering them as optional. :)

aristidesstaffieri: Thanks! Yeah I agree I tent to not like the `value || ""` pattern, it just seemed wrong to pass en empty string to `stellarSdkServer` as the passphrase but this makes sense.

@shared/api/internal.ts
```diff
@@ -780,12 +780,34 @@ export const getTokenDetails = async ({
...
+  fetchBalance?: boolean;
+}): Promise<{
+  name: string;
+  decimals: number;
+  symbol: string;
+  balance?: number;
+} | null> => {
   try {
+    let balance;
+    if (fetchBalance && networkDetails.sorobanRpcUrl) {
```
aristidesstaffieri, line 797: is there any reason to not use the balance from the backend in this case? We only want to use the rpc locally in the case of a custom network in general.

extension/src/popup/views/AddToken/index.tsx
```diff
@@ -335,31 +354,32 @@ export const AddToken = () => {
...
+            {hasBalance && (
+              <div className="AddToken__wrapper__info">
                 <Text
                   as="div"
                   size="xs"
-                  addlClassName="AddToken__wrapper__info--amount AddToken__wrapper__info__row__right_label"
+                  addlClassName="AddToken__wrapper__info--title"
                 >
-                  +1000.00 GO
+                  {t("Simulated Balance Changes")}
```
aristidesstaffieri, line 384: Curious why this says "Simulated Balance Changes"?

Isn't adding a token only bringing into your Freighter balances? I don't think this changes a balance or simulates a balance change.

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -130,6 +140,78 @@ export const freighterApiMessageListener = (
     }
   };
 
+  const submitToken = async () => {
+    try {
+      const { contractId, networkPassphrase: reqNetworkPassphrase } =
+        request as ExternalRequestToken;
+
+      const networkPassphrase =
```
aristidesstaffieri, line 148: Seems like defaulting to mainnet when there is no `networkPassphrase` is an assumption likely to fail. What about failing early if we have no network passphrase here? I think even though the type for `ExternalRequestToken` is a bit loose, in practice we should always have a network passphrase right?

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -210,7 +292,10 @@ export const freighterApiMessageListener = (
         });
       }
 
-      const server = stellarSdkServer(networkUrl, networkPassphrase);
+      const server = stellarSdkServer(
```
aristidesstaffieri, line 295: If we align on my previous suggestion regarding the network passphrase then we should be able to not need a fallback for one here.

extension/src/background/messageListener/freighterApiMessageListener.ts
piyalbasu: This comment definitely makes sense @aristidesstaffieri , but we do actually tend to default to MAINNET in the case of `no network passphrase` in our other API methods (like signTransaction). This is actually a model that a lot of wallets in the ecosystem follow, as well. So, probably makes sense to keep this

piyalbasu:
I'm still taking a look at this (sorry for the delay!). But could you please add documentation for this new method in `docs/docs/guide/usingFreighterWebApp.md`? It'd be great to have this outlined next to all the other methods.

@shared/api/internal.ts
piyalbasu: Second this. We should use the backend to fetch balances and token details (unless we're on a custom network).

 Also, using `buildSorobanServer` with `networkDetails.sorobanRpcUrl` will not work on mainnet because our Soroban RPC url for Mainnet is on a private kube cluster that only our backend can reach. You'll see that if you hit this method on Mainnet that the call will fail

CassioMG: Thanks for raising this!

So I ended up adding this fetch here because the backend request was not returning the balance. It's currently returning an object like this:
```
{
  name: string;
  decimals: number;
  symbol: string;
}
```
I'll take a look at the backend code to see what we could do there.

extension/src/popup/views/AddToken/index.tsx
CassioMG: Good point, I was simply following the [Designs here](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=3359-32729&m=dev). Looks like @sdfcharles tried using the [same wording from other screens](https://github.com/stellar/freighter/issues/1646#issuecomment-2593530587), but thinking about this now I think something like "Balance Info" would probably make more sense here.

What do you think?

CassioMG:
> usingFreighterWebApp

great catch, I'll add it

CassioMG:
@piyalbasu I've added the missing documentation for `addToken` and I noticed it was also missing for the `getNetworkDetails` api so I went ahead and added both ([here](https://github.com/stellar/freighter/pull/1815/commits/5578efb4e4790e6600353cecabdc4c2e9888279c) and [here](https://github.com/stellar/freighter/pull/1815/commits/4fcde95ac86d0f7d722c3efa3a2611472d637839)). Let me know what you think:

Screenshot 2025-02-06 at 15 25 36

Screenshot 2025-02-06 at 15 25 09

extension/src/popup/views/AddToken/index.tsx
CassioMG: Screenshot 2025-02-06 at 15 31 05

@shared/api/internal.ts
CassioMG: PR opened here: https://github.com/stellar/freighter-backend/pull/204

CassioMG: FYI, I've merged the backend PR and tested it on this frontend branch using the indexer url to access the new `should_fetch_balance` param and it worked great!
Frontend changes are here: https://github.com/stellar/freighter/pull/1815/commits/41f614da51f91044ba13bbedf0ff1b4872eda00b

piyalbasu (approved):

PR #1820, opened 2025-01-27
Feature/memory security improvement 3

piyalbasu:
*What*
* Uses built-in crypto library to hash the user's password before storing in Redux
* Uses built-in crypto lib to then use that hashed password to encrypt sensitive data in a temporary storage object in browser.storage
* Any place where we used `privateKeySelector` or `mnemonicPhraseSelector` in `popupMessageListener` is now replaced with a call to `getEncryptedTemporaryData` to get the equivalent data

*Caveat*
* As discussed, I had to remove the "no password" flow for creating a new Stellar address :(


extension/e2e-tests/sendPayment.test.ts
```diff
@@ -39,71 +39,38 @@ test("Send doesn't throw error when account is unfunded", async ({
   );
 });
 
-test("Send XLM payment to G address", async ({ page, extensionId }) => {
+test("Send XLM payments from multiple accounts to G Address", async ({
```
piyalbasu, line 42: this test now creates 2 accounts and sends payments with both to ensure that when we switch accounts, we properly retrieve the correct private key

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+
+interface HashString {
+  str: string;
+  keyObject: { iv: ArrayBuffer; key: CryptoKey };
+}
+
+const TEMPORARY_STORE_ENCRYPTION_NAME = "AES-CBC";
+const HASH_KEY_ENCRYPTION_PARAMS = { name: "PBKDF2", hash: "SHA-256" };
+
+export const encryptHashString = ({ str, keyObject }: HashString) => {
```
piyalbasu, line 40: this is the method used to encrypt sensitive data using the hashed key

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+    keyObject.key,
+    hash,
+  );
+
+  const textDecoder = new TextDecoder();
+
+  return textDecoder.decode(decrypted);
+};
+
+export const deriveKeyFromString = async (str: string) => {
```
piyalbasu, line 77: this method derives the hash key from the user's password

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+
+interface StoreActiveHashKey {
+  sessionStore: Store;
+  hashKey: {
+    key: CryptoKey;
+    iv: ArrayBuffer;
+  };
+}
+
+export const storeActiveHashKey = async ({
```
piyalbasu, line 132: it's a bit of a pain to store a type CryptoKey in browser.storage as it tries to serialize all its inputs, so this method encodes the CryptoKey into something we can reliably retrieve later

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+      },
+    }),
+  );
+};
+
+interface GetActiveHashKey {
+  sessionStore: Store;
+}
+
+export const getActiveHashKeyCryptoKey = async ({
```
piyalbasu, line 157: this does the reverse of above - get the encode version of the CryptoKey from redux and then decode it into something we can use for encryption/decryption

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+  localStore: DataStorageAccess;
+  keyName: string;
+  temporaryData: string;
+  hashKey: {
+    key: CryptoKey;
+    iv: ArrayBuffer;
+  };
+}
+
+export const storeEncryptedTemporaryData = async ({
```
piyalbasu, line 198: utility method for store in browser.storage. It takes in a string, does all the encryption for you using the hashkey, and then stores it

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+  });
+};
+
+interface GetEncryptedTemporaryData {
+  sessionStore: Store;
+  localStore: DataStorageAccess;
+  keyName: string;
+}
+
+export const getEncryptedTemporaryData = async ({
```
piyalbasu, line 224: And this is the opposite: takes a keyName and then will decrypt the data for you using the current hashKey (if one is saved).

extension/src/background/helpers/session.ts
```diff
@@ -17,3 +28,249 @@ export class SessionTimer {
...
+  localStore: DataStorageAccess;
+  sessionStore: Store;
+}
+
+export const clearSession = async ({
+  localStore,
+  sessionStore,
+}: ClearSession) => {
+  sessionStore.dispatch(timeoutAccountAccess());
+  await localStore.remove(TEMPORARY_STORE_ID);
```
piyalbasu, line 275: when we end sessions, we now also clear `TEMPORARY_STORE_ID` in addition to dispatching `timeoutAccountAccess`

extension/package.json
```diff
@@ -41,6 +41,7 @@
     "@types/testing-library__jest-dom": "^5.14.5",
     "@types/webextension-polyfill": "^0.12.0",
     "@types/yup": "^0.32.0",
+    "base64-arraybuffer-es6": "3.1.0",
```
aristidesstaffieri, line 44: Is this used? I don't see an import but also would expect that we could use the native ArrayBuffer.

extension/src/background/helpers/session.ts
aristidesstaffieri: does it make sense for this to be stores in session storage instead of local storage?
It seems like this is used only during the session and is temporary.

extension/package.json
piyalbasu: Yup, it's here: https://github.com/stellar/freighter/pull/1820/files#diff-a54d899cec48734cfbe57e1190c3e9d5480cc97bd23b0a7d55d711bc05158212R2

It's used to encode an ArrayBuffer into a serializable format for redux/browser.storage

extension/src/background/helpers/session.ts
piyalbasu: Oh that's a good idea. Let me think about if there would be any unintended consequences of that 🤔

extension/package.json
aristidesstaffieri: oh thanks missed that, but is it different than https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -292,13 +290,37 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
     let keyStore = { id: "" };
+    let activeHashKey = await getActiveHashKeyCryptoKey({ sessionStore });
+    if (activeHashKey === null) {
+      activeHashKey = await deriveKeyFromString(password);
+    }
 
+    // store encrypted extra data
     try {
       keyStore = await keyManager.storeKey(keyMetadata);
+      await storeEncryptedTemporaryData({
```
aristidesstaffieri, line 301: This can throw by way of `crypto.subtle.encrypt` throwing an exception. I don't think we have a UI state for this exception though. Does it make sense to handle this more explicitly than logging the error to the console?

extension/package.json
piyalbasu: Yeah it is - TextEncoder/TextDecoder didn't quite do what I needed to do here. I kept getting slightly different results when encoding/decoding

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -292,13 +290,37 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     };
 
     let keyStore = { id: "" };
+    let activeHashKey = await getActiveHashKeyCryptoKey({ sessionStore });
+    if (activeHashKey === null) {
+      activeHashKey = await deriveKeyFromString(password);
```
aristidesstaffieri, line 295: this seems unlikely but its possible that both `getActiveHashKeyCryptoKey` and `deriveKeyFromString` can throw an exception. Do we need to handle this case?

extension/src/background/helpers/session.ts
aristidesstaffieri: this has the potential to throw an exception(doesn't seem likely but several inner functions can throw), should we explicitly handle these exceptions in the places where we call this?

Right now, it's not clear to me what happens in the various workflows if this does throw but likely it's unexpected behavior in the message listener.

extension/package.json
piyalbasu: This has been updated here: https://github.com/stellar/freighter/pull/1820/commits/d568671635b0183019e0769c04d622c17038c988

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: I added better error handling (and sentry capturing) to try to log errors if any of these throw errors: https://github.com/stellar/freighter/pull/1820/commits/033c7b84815001916e4a7f3cc740a2b78a4a882d

piyalbasu:
Reopened in #1827 to remove release/5.28.0 features

PR #1830, opened 2025-02-03
Core dependency upgrades

### LINKED ISSUE #1824: Update packages in `Security > Dependabot`

aristidesstaffieri:
Closes #1824 

Major dependency upgrade for - 
react
react-router
typescript
styled-components
clsx
@testing-library/*
@reduxjs-toolkit

Notable changes - 
Improved store types
new routing API from react-router v7, significant changes to the routing paradigm and architecture, mostly found in `Router.tsx` but also impacts all nested routes.

Regressions - 
Send/Swap tests are not easily compatible with these changes due to the changes to nested routes. I've commented and detailed the situation and potential ways to revive them in the change set diffs for the tests.


extension/src/popup/components/BottomNav/index.tsx
```diff
@@ -38,7 +39,7 @@ export const BottomNav = () => {
           <Icon.ClockRewind />
           {t("History")}
         </BottomNavLink>
-        <BottomNavLink to={ROUTES.swap}>
+        <BottomNavLink to={ROUTES.swapAmount}>
```
aristidesstaffieri, line 42: The nested routing logic impacts the way we used to treat `/swap` as the index route while in practice it rendered the swap amount route as a redirect. This causes us to need to change the index route for swaps to `swap/amount`

extension/src/popup/components/Onboarding/index.tsx
```diff
@@ -58,10 +57,9 @@ export const OnboardingButtons = ({
   customBackAction,
   children,
 }: OnboardingButtonsProps) => {
-  const history = useHistory();
   const { t } = useTranslation();
 
-  const isNewTabSession = history.length === 1;
+  const isNewTabSession = window.history.length === 1;
```
aristidesstaffieri, line 62: The new router patterns prevents us from injecting our own history, so we need to rely on the window's history for this check.

extension/src/popup/helpers/hardwareConnect.ts
```diff
@@ -87,13 +87,10 @@ export const hardwareSign: HardwareSign = {
   }: HardwareSignParams) => {
     const transport = await TransportWebUSB.create();
     const ledgerApi = new LedgerApi(transport);
-    let result = { signature: Buffer.from([]) };
+    const result = isHashSigningEnabled
```
aristidesstaffieri, line 90: This change allows the Buffer type to align to the expected type from the ledger API.

extension/src/popup/helpers/navigate.ts
```diff
@@ -1,11 +1,16 @@
...
 
-export const navigateTo = (path: ROUTES, queryParams?: string) => {
+export const navigateTo = (
+  path: ROUTES,
+  navigate: NavigateFunction,
+  queryParams?: string,
+) => {
   const pathname = queryParams ? `${path}${queryParams}` : path;
-  history.push({ pathname });
+  navigate(pathname);
```
aristidesstaffieri, line 13: We can no longer rely on the history we inject into the router, so now we take a navigate function in order to replace this.

extension/src/popup/views/AccountMigration/index.tsx
```diff
@@ -13,43 +13,73 @@ import { MigrationComplete } from "popup/components/accountMigration/MigrationCo
 
 import "./styles.scss";
 
-export const AccountMigration = () => (
-  <>
+export const AccountMigration = () => {
```
aristidesstaffieri, line 16: Because of the changes to nested routes in v7, we need to only use the path after the base path from the nested structure. In order to keep the existing mappings in `routes.ts`, we can split away the base path and pass the page specific path to each nested route. 

This pattern is used for all nested routes but only commented about here.

extension/src/popup/views/__tests__/SendPayment.test.tsx
```diff
@@ -99,16 +98,10 @@ jest.mock("react-router-dom", () => {
...
-jest.mock("popup/constants/history", () => ({
-  get history() {
-    return mockHistoryGetter();
-  },
-}));
 
 const publicKey = "GA4UFF2WJM7KHHG4R5D5D2MZQ6FWMDOSVITVF7C5OLD5NFP6RBBW2FGV";
 
-describe("SendPayment", () => {
+describe.skip("SendPayment", () => {
```
aristidesstaffieri, line 104: I haven't found a way to maintain this test as written. Because of the nested router changes and the way the send/swap component relies on location to render the appropriate copy/components, the test fails to render the correct screen. The top level router hasn't been part of the tests and the changes to the global history make maintaining this pattern difficult. 

I think we can rewrite the send/swap components to rely on an injected location from the `useNavigate` hook while also refactoring the tests in order to render use a test router to mimic the nested routing structure in the full application.

extension/src/popup/views/__tests__/Swap.test.tsx
```diff
@@ -175,11 +167,7 @@ jest.mock("stellar-sdk", () => {
 
 const publicKey = "GCXRLIZUQNZ3YYJDGX6Z445P7FG5WXT7UILBO5CFIYYM7Z7YTIOELC6O";
 
-const history = createMemoryHistory();
-history.push(ROUTES.swap);
-mockHistoryGetter.mockReturnValue(history);
-
-describe("Swap", () => {
+describe.skip("Swap", () => {
```
aristidesstaffieri, line 170: This is subject to the same details as the SendPayment tests and can be revived using the same solutions.

extension/src/popup/views/__tests__/SwapUnfunded.test.tsx
```diff
@@ -91,7 +83,7 @@ jest.mock("stellar-sdk", () => {
   };
 });
 
-describe("Swap unfunded account", () => {
+describe.skip("Swap unfunded account", () => {
```
aristidesstaffieri, line 86: Same comments as other skipped tests

extension/webpack.common.js
```diff
@@ -128,10 +128,10 @@ const commonConfig = (
     ],
   },
   plugins: [
-    new ESLintPlugin({
-      extensions: [".ts", ".tsx"],
-      failOnWarning: true,
-    }),
+    // new ESLintPlugin({
```
aristidesstaffieri, line 131: Our current linting pipeline relies on @stellar/config which is not compatible with the upgraded version of typescript/react. We have to upgrade eslint & friends in that repo before we can use that config in this repo.

extension/src/popup/components/BottomNav/index.tsx
aristidesstaffieri: I found a way to resolve this and restore the base route in 03bcd79f93c948748b06509328f8d260eaf9175c

extension/src/popup/views/__tests__/SendPayment.test.tsx
aristidesstaffieri: I haven't done this because of the massive change set already, it seemed like this could be broken off into another PR but I am open to suggestions.

extension/src/popup/Router.tsx
```diff
@@ -104,19 +105,17 @@ export const PublicKeyRoute = (props: RouteProps) => {
...
-        }}
+      <Navigate
+        to={`${ROUTES.unlockAccount}${location.search}`}
+        state={{ from: location }}
+        replace
       />
     );
   }
-  return <Route {...props} />;
+  return children;
```
piyalbasu, line 115: just calling some things out as I see them - I'll pull this branch and try to see if this actually has an effect:

I notice we're not passing the props here. Wondering if this will break some views

extension/src/popup/helpers/navigate.ts
```diff
@@ -1,11 +1,16 @@
 import browser from "webextension-polyfill";
 
 import { ROUTES } from "popup/constants/routes";
-import { history } from "popup/constants/history";
+// import { history } from "popup/constants/history";
```
piyalbasu, line 4: let's rm this comment

extension/src/popup/views/AccountMigration/index.tsx
piyalbasu: Can we create a helper for this where we pass in the nested path and the base path?

`getNestedRoute(ROUTES.accountMigratonReviewMigration, ROUTES.accountMigration)`

extension/webpack.common.js
piyalbasu: Do we know what the lift is on upgrading @stellar/config? I think we should do that first if it's reasonable. If we put it off, we could introduce a bunch of linting errors we'll need to address down the line

aristidesstaffieri: I think it requires a significant rewrite of the config but not sure on the t-shirt size of it. Open to suggestions here but it seemed like doing this could introduce quite a few more changes to this already massive PR due to the likelihood of rules changing across several major versions of the tooling.

extension/src/popup/views/AccountMigration/index.tsx
aristidesstaffieri: What do you have in mind?
It's not always the same prefix or the same number of prefix paths so I think the helper would essentially just be a wrapper around String.split.

.github/workflows/runTests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 21
+          node-version: 22
```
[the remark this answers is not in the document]
aristidesstaffieri: Not that I've found, but yeah maybe the build ones would too I can upgrade those to 22.

extension/src/popup/Router.tsx
aristidesstaffieri: I don't believe it does because the re-organization and use of the the outlet makes those props no longer needed but would love another glance at that as it departed significantly from the previous patterns.

extension/src/popup/views/AccountMigration/index.tsx
piyalbasu: Oh gotcha - it looked like the nested route was always `${baseRoute}/{subRoute}`, where you're always just extracting `subRoute`. The nested routes are _generally_ set up that way, but I see we actually have another level of nesting on some of them which makes this non-standard.

Hm, this does feel a little messy having these `.split`'s at the top. Maybe the real fix here is just moving away from this nested router structure 🤔 But that's maybe a new ticket

For now, to avoid duplicating the same string a bunch of times, can we at least do:

```
const baseRoute = `${ROUTES.accountMigration}/`;
const review Slug = ROUTES.accountMigrationReviewMigration.split(
    baseRoute,
  )[1];
```

aristidesstaffieri: hmm not sure I see the example, aren't they all different sub paths already?

Yeah there is def a discussion about the routing because the sub routes don't work well for "full screen mode" but yeah I feel that's outside of this scope and would be a significant refactor.

aristidesstaffieri: Will pull out the base path into a var here, incoming patch.

aristidesstaffieri: this is done in 96ef7f52036f8615d60ad60470a9215049324305

.github/workflows/runTests.yml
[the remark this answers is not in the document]
aristidesstaffieri: this is done in add3f0cf8f726d80ffc1fe5849e9cf1e74d87c1c

extension/src/popup/helpers/navigate.ts
aristidesstaffieri: done in d88f696ecc0d555945db92aac982083c51dc7c2b

extension/webpack.common.js
aristidesstaffieri: this is a pretty big departure since eslint 9 is wildly different but I think this gets us as close as is reasonable given the changes.

done in commit 9134812a80b45aadef1d87648e9bc6480d194a44

piyalbasu (approved):

extension/e2e-tests/sendPayment.test.ts
```diff
@@ -282,7 +282,7 @@ test("Send token payment to C address", async ({ page, extensionId }) => {
   await page.getByText("Details").click({ force: true });
 
   await expect(page.getByText("Sent E2E")).toBeVisible();
-  await expect(page.getByTestId("asset-amount")).toContainText(".001 E2E");
+  await expect(page.getByTestId("asset-amount")).toContainText("0.001");
```
CassioMG, line 285: nice catch on those "asset-amount" string tweaks

extension/src/popup/views/__tests__/SendPayment.test.tsx
CassioMG: Agree on breaking this off into another PR, sounds like the best approach for simplicity

CassioMG (approved):
Lgtm, thanks for working on this!

PR #1832, opened 2025-02-03
Fix/import acct when timed out

piyalbasu:
Similar to `add account`, if our session has timed out, make sure to log back in before adding a new account to the mix

### DIFF
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -1074,7 +1074,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
       );
     }
 
-    // clear the temporary store so we can replace it with the new encrypted data
+    // clear the temporary store (if it exists) so we can replace it with the new encrypted data
     await localStore.remove(TEMPORARY_STORE_ID);
 
     try {
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -248,6 +248,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     );
   };
 
+  /* Append an additional account to user's account list */
   const _storeAccount = async ({
     mnemonicPhrase,
     password,
@@ -264,6 +265,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     const allAccounts = allAccountsSelector(sessionStore.getState());
     const accountName = `Account ${allAccounts.length + 1}`;
 
+    // set the active public key
     await sessionStore.dispatch(
       logIn({
         publicKey,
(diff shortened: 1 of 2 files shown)


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -466,6 +469,8 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
 
     const currentState = sessionStore.getState();
 
+    sessionTimer.startSession();
```
piyalbasu, line 472: Small oversight - we were not starting the session timer on account creation

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -491,9 +496,6 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
           keyName: TEMPORARY_STORE_EXTRA_ID,
         });
       } catch (e) {
-        captureException(
-          `Could not login when adding account: ${JSON.stringify(e)}`,
-        );
```
piyalbasu, line 496: removing this Sentry alert because this will be triggered every time a user enters an incorrect password. This will pollute our Sentry dashboard with non-errors

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -502,9 +504,10 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
       ? await _getNonHwKeyID()
       : (await localStore.getItem(KEY_ID)) || "";
 
+    // if the session is active, confirm that the password is correct and the hashkey properly unlocks
     let activePrivateKey = "";
     try {
-      // await _unlockKeystore({ keyID, password });
+      await _unlockKeystore({ keyID, password });
```
piyalbasu, line 510: mistakenly left this out - make sure to check the user's password is correct

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -555,12 +560,39 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
   const importAccount = async () => {
     const { password, privateKey } = request;
     let sourceKeys;
+
+    let mnemonicPhrase = await getEncryptedTemporaryData({
```
piyalbasu, line 564: these changes are similar to the changes I added here for `add account`: https://github.com/stellar/freighter/pull/1827/commits/cb9d65c3308f6bc4968ef89c85d9a8764fb12f75

We're making sure that if our session has timed out, we log back in with the user's password before continuing

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1121,9 +1155,6 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     try {
       await loginToAllAccounts(password);
     } catch (e) {
-      captureException(
-        `Could not login when confirming password: ${JSON.stringify(e)}`,
-      );
```
piyalbasu, line 1126: same change as above - don't report to Sentry every time a user enter's an incorrect password

aristidesstaffieri (approved):

PR #1834, opened 2025-02-04
login before showing mnemonic phrase

piyalbasu:
Similar to the last bugfix, making sure we log in before trying to show the user their backup phrase if their session has timed out

### DIFF
--- a/extension/src/background/ducks/session.ts
+++ b/extension/src/background/ducks/session.ts
@@ -78,15 +78,6 @@ export const sessionSlice = createSlice({
   reducers: {
     reset: () => initialState,
     logOut: () => initialState,
-    setActivePrivateKey: (state, action: { payload: AppData }) => {
-      const { privateKey = "", password = "" } = action.payload;
-
-      return {
-        ...state,
-        privateKey,
-        password,
-      };
-    },
     setActiveHashKey: (state, action: { payload: AppData }) => {
       const {
         hashKey = {
@@ -166,7 +157,6 @@ export const {
   actions: {
     reset,
     logOut,
-    setActivePrivateKey,
     setActiveHashKey,
     timeoutAccountAccess,
     updateAllAccountsAccountName,
--- a/extension/src/background/messageListener/popupMessageListener.ts
+++ b/extension/src/background/messageListener/popupMessageListener.ts
@@ -254,16 +254,28 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     password,
     keyPair,
     imported = false,
+    isSettingHashKey = false,
   }: {
     mnemonicPhrase: string;
     password: string;
     keyPair: KeyPair;
     imported?: boolean;
+    isSettingHashKey?: boolean;
   }) => {
     const { publicKey, privateKey } = keyPair;
 
     const allAccounts = allAccountsSelector(sessionStore.getState());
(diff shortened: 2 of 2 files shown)


extension/src/background/ducks/session.ts
```diff
@@ -78,15 +78,6 @@ export const sessionSlice = createSlice({
   reducers: {
     reset: () => initialState,
     logOut: () => initialState,
-    setActivePrivateKey: (state, action: { payload: AppData }) => {
```
piyalbasu, line 81: this is no longer used with the new flow, so removing

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -254,17 +254,29 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
     imported?: boolean;
+    isSettingHashKey?: boolean;
   }) => {
     const { publicKey, privateKey } = keyPair;
 
     const allAccounts = allAccountsSelector(sessionStore.getState());
     const accountName = `Account ${allAccounts.length + 1}`;
 
+    let activeHashKey = await getActiveHashKeyCryptoKey({ sessionStore });
+    if (activeHashKey === null && isSettingHashKey) {
```
piyalbasu, line 271: I'm adding an explicit flag here just to avoid any weird edge cases where a user gets to this block of code and unexpectedly doesn't have their hash key already. Deriving a new hash key would cause a user to lose access to anything already stored in `temporaryDataStorage`.

This is an edge case that I don't think was really possible, but I'd rather be explicit about it just in case I'm missing some scenario in which this might happen

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -254,17 +254,29 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
+      // this should only happen on account creation & account recovery
+      activeHashKey = await deriveKeyFromString(password);
+    }
+
+    if (activeHashKey === null) {
+      throw new Error("Error deriving hash key");
```
piyalbasu, line 277: Short circuit if we unexpectedly don't have the hash key. The error will be picked up and reported by the caller

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -976,12 +985,27 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
       return { error: "Incorrect Password" };
     }
 
+    let mnemonicPhrase = await getEncryptedTemporaryData({
```
piyalbasu, line 988: same change as the last bugfix: if we don't have the mnemonic, let's log in and get it

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1428,6 +1452,7 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
 
   const signOut = async () => {
     sessionStore.dispatch(logOut());
+    await localStore.remove(TEMPORARY_STORE_ID);
```
piyalbasu, line 1455: `temporaryStore` should be overwritten properly when a user logs back in, but no harm in explicitly removing this to prevent any edge cases

aristidesstaffieri (approved):

Issue #1791, opened 2025-01-14
Token list refresh


aristidesstaffieri:
We currently use Stellar Experts top 50 token list as a way to warn users when they interact with tokens that are not trusted by this list.

We want to move to a model that is not derived from those metrics to allow inclusion of tokens like Paxos' USDL before seeing much use in the network.

One option is to adopt https://github.com/soroswap/token-list as a new default and collaborate to add new tokens
Another option could be to maintain our own list.

In a related effort, we need to find a way to support icons for custom tokens, can this list serve as a source for icons for the tokens on the list?

PR #1839, opened 2025-02-07
[FEATURE] Split assets by verification status

### LINKED ISSUE #1553: Sort assets by on/not on your asset lists
Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-6039&node-type=frame&m=dev

aristidesstaffieri:
Closes #1553 

What
Splits assets by verification status on the add asset workflows

Notes:
`getAssetListForAsset` was introduced separately to `splitVerifiedAssetCurrency` because we will need to get asset lists for an asset separately from processing an asset list for [the next task to refactor our asset workflows](https://github.com/stellar/freighter/issues/1791). This sets us up to be able to use icons from the asset lists in that task.

Open questions:
What should happen if we fail to fetch one or all of the asset lists? Right now we silently fail but this will in effect label all assets as "Not on your list". Should we fallback to the unlabeled list in these cases?

Screenshot 2025-02-07 at 3 06 08 PM

Screenshot 2025-02-07 at 3 07 39 PM

### DIFF
--- a/extension/src/popup/components/manageAssets/AddAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/AddAsset/index.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-argument */
 import React, { useEffect, useCallback, useRef, useState } from "react";
 import { useSelector } from "react-redux";
 import { Networks, StellarToml, StrKey } from "stellar-sdk";
@@ -10,11 +9,14 @@ import { stellarSdkServer } from "@shared/api/helpers/stellarSdkServer";
 import { isSacContractExecutable } from "@shared/helpers/soroban/token";
 
 import { FormRows } from "popup/basics/Forms";
-import { settingsNetworkDetailsSelector, settingsSelector } from "popup/ducks/settings";
+import {
+  settingsNetworkDetailsSelector,
+  settingsSelector,
+} from "popup/ducks/settings";
 import { isMainnet, isTestnet } from "helpers/stellar";
 import { getNativeContractDetails } from "popup/helpers/searchAsset";
 import {
-  getAssetListForAsset,
+  getAssetListsForAsset,
   splitVerifiedAssetCurrency,
 } from "popup/helpers/assetList";
 import { isContractId } from "popup/helpers/soroban";
@@ -151,7 +153,7 @@ export const AddAsset = () => {
         setVerifiedAssetRows(verifiedAssets);
         setUnverifiedAssetRows(unverifiedAssets);
 
-        const assetListsForToken = await getAssetListForAsset({
+        const assetListsForToken = await getAssetListsForAsset({
           asset: token,
           assetsListsDetails: assetsLists,
           networkDetails,
@@ -216,7 +218,7 @@ export const AddAsset = () => {
               isSuspicious: isAssetSuspicious(
                 scannedAssets.results[`${record.code}-${record.issuer}`],
               ),
-            } as ManageAssetCurrency),
+            }) as ManageAssetCurrency,
         );
 
         const { verifiedAssets, unverifiedAssets } =
--- a/extension/src/popup/components/manageAssets/AddAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/AddAsset/index.tsx
@@ -18,8 +18,12 @@ import {
   splitVerifiedAssetCurrency,
 } from "popup/helpers/searchAsset";
 import { isContractId } from "popup/helpers/soroban";
-import { isAssetSuspicious, scanAsset, scanAssetBulk } from "popup/helpers/blockaid";
-import { AssetNotifcation } from "popup/components/AssetNotification";
+import {
+  isAssetSuspicious,
+  scanAsset,
+  scanAssetBulk,
+} from "popup/helpers/blockaid";
+
 import { SubviewHeader } from "popup/components/SubviewHeader";
 import { View } from "popup/basics/layout/View";
 import { publicKeySelector } from "popup/ducks/accountServices";
@@ -257,7 +261,7 @@ export const AddAsset = () => {
     setIsVerificationInfoShowing(isAllowListVerificationEnabled);
   }, [isAllowListVerificationEnabled]);
 
-  const hasAssets = verifiedAssetRows.length && unverifiedAssetRows.length;
+  const hasAssets = verifiedAssetRows.length || unverifiedAssetRows.length;
 
   return (
     <Formik initialValues={initialValues} onSubmit={() => {}}>
@@ -295,10 +299,6 @@ export const AddAsset = () => {
                   isSearching={isSearching}
                   resultsRef={ResultsRef}
                 >
-                  {hasAssets && isVerificationInfoShowing ? (
-                    <AssetNotifcation isVerified={isVerifiedToken} />
-                  ) : null}
-
                   {hasAssets ? (
                     <ManageAssetRows
                       header={null}
--- a/extension/src/popup/components/manageAssets/AddAsset/index.tsx
+++ b/extension/src/popup/components/manageAssets/AddAsset/index.tsx
@@ -13,6 +13,7 @@ import { FormRows } from "popup/basics/Forms";
 import { settingsNetworkDetailsSelector, settingsSelector } from "popup/ducks/settings";
 import { isMainnet, isTestnet } from "helpers/stellar";
 import {
+  getAssetListForAsset,
   getNativeContractDetails,
   splitVerifiedAssetCurrency,
 } from "popup/helpers/searchAsset";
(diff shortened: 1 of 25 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -246,13 +250,202 @@ export const ManageAssetRows = ({
...
+  }) => React.ReactNode;
+}) => {
+  if (shouldSplitAssetsByVerificationStatus) {
+    return (
+      <>
+        {verifiedAssetRows.length > 0 && (
+          <InfoTooltip
+            infoText={
+              <span>
+                Freighter uses asset lists to verify assets before interactions.
```
aristidesstaffieri, line 300: @sdfcharles any qualms with this copy tweak? Just trying not to end a sentence with "with" 😄

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -246,23 +252,235 @@ export const ManageAssetRows = ({
...
   domain,
   name,
   isSuspicious = false,
+  contractId,
 }: AssetRowData) => {
+  const networkDetails = useSelector(settingsNetworkDetailsSelector);
   const canonicalAsset = getCanonicalFromAsset(code, issuer);
-  const assetCode = name || code;
+  // use the name unless the name is SAC, format "code:issuer"
+  const assetCode =
```
aristidesstaffieri, line 474: SAC contracts store their name as `code:issuer` so we need to account for this by detecting if tokens are a SAC and then choosing to only use the code in those cases.

CassioMG:
> What should happen if we fail to fetch one or all of the asset lists? Right now we silently fail but this will in effect label all assets as "Not on your list". Should we fallback to the unlabeled list in these cases?

I think falling back to the unlabeled list in these cases makes sense since it's a more neutral approach, otherwise Freighter could unintentionally scare users for some trustful assets. Maybe we could even display some warning letting users know Freighter is having trouble fetching the lists but I'm also not sure if it'd would be worth adding more complexity to it since it should almost never happen and this warning could _still_ scare users.

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -115,63 +119,55 @@ export const AddAsset = () => {
...
-            contract: contractId,
-            issuer,
-            domain: "",
-            name: tokenDetailsResponse.name,
-            isSuspicious: isAssetSuspicious(scannedAsset),
-          },
-        ]);
+        setVerifiedAssetRows([]);
+        setUnverifiedAssetRows([]);
+        return null;
```
CassioMG, line 124: +1 for using multiple early returns instead of `if else`

extension/src/popup/helpers/assetList.ts
```diff
@@ -0,0 +1,105 @@
...
+  // eslint-disable-next-line no-restricted-syntax
+  const validatedAssets = [] as AssetListReponseItem[];
+  // eslint-disable-next-line no-restricted-syntax
+  for (const responses of settledResponses) {
+    if (responses.status === "fulfilled") {
+      // confirm that this list still adheres to the agreed upon schema
+      const validatedList = await schemaValidatedAssetList(responses.value);
+      validatedAssets.push(...validatedList.assets);
+    }
+  }
```
CassioMG, line 33:
```suggestion
  for (const response of settledResponses) {
    if (response.status === "fulfilled") {
      // confirm that this list still adheres to the agreed upon schema
      const validatedList = await schemaValidatedAssetList(response.value);
      validatedAssets.push(...validatedList.assets);
    }
  }
```

extension/src/popup/helpers/assetList.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+    if (!verifiedIds.has(validAsset.issuer)) {
+      verifiedIds.add(validAsset.issuer);
+    }
+  }
+
+  const [verifiedAssets, unverifiedAssets] = assets.reduce<
+    [typeof assets, typeof assets]
+  >(
+    ([inC, notInC], item) => {
```
CassioMG, line 49: Could we rename `[inC, notInC]` to something a little more explicit? Maybe `[inVerifiedsList, notInVerifiedsList]`. I'm not sure what the `C` means here but it gave me the impression that it could be handling only the `C`ontract ids which was not the case

extension/src/popup/helpers/assetList.ts
```diff
@@ -0,0 +1,105 @@
...
+  // eslint-disable-next-line no-restricted-syntax
+  const validatedAssets = {} as Record<string, AssetListReponseItem[]>;
+  // eslint-disable-next-line no-restricted-syntax
+  for (const responses of settledResponses) {
+    if (responses.status === "fulfilled") {
+      // confirm that this list still adheres to the agreed upon schema
+      const validatedList = await schemaValidatedAssetList(responses.value);
+      validatedAssets[responses.value.name] = validatedList.assets;
+    }
+  }
```
CassioMG, line 95:
```suggestion
  const validatedAssets = {} as Record<string, AssetListReponseItem[]>;
  // eslint-disable-next-line no-restricted-syntax
  for (const response of settledResponses) {
    if (response.status === "fulfilled") {
      // confirm that this list still adheres to the agreed upon schema
      const validatedList = await schemaValidatedAssetList(response.value);
      validatedAssets[response.value.name] = validatedList.assets;
    }
  }
```

extension/src/popup/helpers/assetList.ts
aristidesstaffieri: ah yeah good call those var names are terrible 😅 I'll update that.

extension/src/popup/helpers/assetList.ts
```diff
@@ -0,0 +1,105 @@
...
+    [[], []],
+  );
+
+  return {
+    verifiedAssets,
+    unverifiedAssets,
+  };
+};
+
+export const getAssetListForAsset = async ({
```
CassioMG, line 72:
```suggestion
export const getAssetListsForAsset = async ({
```

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -246,23 +252,235 @@ export const ManageAssetRows = ({
...
+  }) => React.ReactNode;
+}) => {
+  if (shouldSplitAssetsByVerificationStatus) {
+    return (
+      <>
+        {verifiedAssetRows.length > 0 && (
+          <InfoTooltip
+            infoText={
+              <span>
+                Freighter uses asset lists to verify assets before interactions.
```
CassioMG, line 303: should we wrap all copies with `t()` in this component?

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -121,40 +132,58 @@ export const SearchAsset = () => {
...
+                record.domain && /\S/.test(record.domain),
+            )
+            .map(
+              (record: {
+                asset: string;
+                tomlInfo: { image: string };
+                domain: string;
+              }) => {
+                const [code, issuer] = record.asset.split("-");
+                const assetId = `${code}-${issuer}`;
```
CassioMG, line 150:
```suggestion
                const assetId = record.asset;
                const [code, issuer] = assetId.split("-");
```

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
aristidesstaffieri: ya know it's unclear to me if we're still going to keep putting effort into supporting the translation layer. I think @JakeUrban mentioned we're moving in the direction to ditch this. @JakeUrban can you confirm or deny?
Happy to wrap it otherwise.

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
aristidesstaffieri: hmm is this right? I think the assetId should be format `${code}-${issuer}`

CassioMG: yeah so looking solely at those 2 lines of code, it appears you are splitting a string and then concatenating it to the exact same thing afterwards. My suggestion was intended to avoid doing that concatenation which apparently seemed unnecessary.

There is some edge cases though where my suggestion and the original code would diverge in the result.
E.g. when `record.asset` does not contain a dash (`-`).

Does that make sense?

I'm not strong about my suggestion though!

CassioMG (approved):
Lgtm - left some comments/suggestions

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
aristidesstaffieri: ah yeah good call, I think actually I can just avoid the `assetId` var altogether if that works for you, I just didn't see it when refactoring this but it's only used as a key right below it so I don't see the need to assign the extra var. Wdyt?

CassioMG: sounds good to me as well!

aristidesstaffieri: so there was a subtle reason to do it this way that I didn't realize before. The format from items in the search are actually not `{code}:{issuer}`, they are `{code}:{issuer}{type 1|2}` so that's why the split was done this way in this past. 

Search example - https://api.stellar.expert/explorer/public/asset?search=usdc

PR #1847, opened 2025-02-12
Fix/replace alert

### LINKED ISSUE #1846: Blockaid Form QA - replace Alert notification
The SDS currently has a different `<Alert />` component than what is in Figma. While we wait for the Figma design to be added to the SDS codebase, let's do some custom styling to ensure the designs match

piyalbasu:
Closes #1846 

Until the new `<Alert />` component gets moved to SDS, we'll need to do some custom styling to match the existing design

Screenshot 2025-02-11 at 7 09 39 PM

### DIFF
--- a/extension/src/popup/components/WarningMessages/index.tsx
+++ b/extension/src/popup/components/WarningMessages/index.tsx
@@ -2,7 +2,6 @@ import React, { useState, useRef, useEffect } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { createPortal } from "react-dom";
 import {
-  Alert,
   Button,
   Card,
   Icon,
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1405,7 +1404,7 @@ export const BlockaidAssetScanLabel = ({
 
   return (
     <BlockaidWarningModal
-      header={`This asset was flagged as ${blockaidData.result_type}`}
+      header={`This asset was flagged as ${blockaidData.result_type.toLowerCase()}`}
```
piyalbasu, line 1407: slightly unrelated - but I just noticed that occasionally we use uppercase and lower case when denoting something as malicious, so making sure it's uniform throughout

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -1437,18 +1436,31 @@ export const BlockaidWarningModal = ({
...
+      data-testid="BlockaidWarningModal__alert"
     >
-      <BlockaidByLine
-        hasArrow={hasArrow}
-        handleClick={() => setIsModalActive(true)}
-        requestId={requestId}
-        address={address}
-      />
-    </Alert>
+      <div className="ScamAssetWarning__box__content">
```
piyalbasu, line 1445: Since this is a temporary fix (we should be able to use the <Alert> component in the future), I'm re-using the styles we use on the other Blockaid warnings

JakeUrban:
I don't want my question / comment to block this work, but it seems redundant to show two warnings for a transaction (one for the transaction, another for the asset). Is there a way to determine if a transaction is flagged as malicious _due to_ a malicious asset being transacted? Or is it blockaid's policy that all transactions that use malicious assets are malicious?

If either are true, it feels like we should be showing one warning, rather than two, that communicates the asset being transacted is malicious. Maliciousness of the transaction is a transitive property in this case.

What do you think @sdfcharles ?

sdfcharles:
hmm good point, agreed would be nice if it was only one error so wondering about the above questions as well.

cc @piyalbasu 

> I don't want my question / comment to block this work, but it seems redundant to show two warnings for a transaction (one for the transaction, another for the asset). Is there a way to determine if a transaction is flagged as malicious _due to_ a malicious asset being transacted? Or is it blockaid's policy that all transactions that use malicious assets are malicious?
> 
> If either are true, it feels like we should be showing one warning, rather than two, that communicates the asset being transacted is malicious. Maliciousness of the transaction is a transitive property in this case.
> 
> What do you think @sdfcharles ?

aristidesstaffieri:
> > I don't want my question / comment to block this work, but it seems redundant to show two warnings for a transaction (one for the transaction, another for the asset). Is there a way to determine if a transaction is flagged as malicious _due to_ a malicious asset being transacted? Or is it blockaid's policy that all transactions that use malicious assets are malicious?
> > If either are true, it feels like we should be showing one warning, rather than two, that communicates the asset being transacted is malicious. Maliciousness of the transaction is a transitive property in this case.
> > What do you think @sdfcharles ?
> 
> Let's chat with Blockaid about this. They mentioned today that they'll be starting to provide UX best practices for wallets, which I think this fits under. Until we have a clearer answer from them, I'd be hesitant to try to be too smart with hiding any warning.
> 
> Fwiw, the tx warning just tells you that the token is malicious, but the asset warning tells you why the token is malicious

I agree that this shouldn't be a part of this change set but I do wonder if we can aggregate blockaid response types by type sop we at most show 1 "malicious" warning for example, where in the details we expose both the asset and tx details from each response item.

aristidesstaffieri (approved):

PR #1848, opened 2025-02-12
Feature/account mismatch 2

### LINKED ISSUE #1821: Better handling of multiple active accounts
When a user is in fullscreen mode, they can conceivably still have stale data present in the browser if they go to to the `popup` and switch accounts or import a new account

`Acceptance Criteria`:

1. When a user switches accounts, if they have an open tab with the old account activated, don't let that account make changes.
2. Inform the user that their sessions has gone stale

piyalbasu:
Closes #1821 

This is a 2 part solution to users being able to have multiple fullscreen views open at once that have different accounts active. When this happens, the background (source of truth) has one specific account active, and the popup may have a different account active:

1) If the account active in the popup is different than the one set in the background, block the user from making any requests to the background

2) In the event of the popup having a different account active, show the user a message letting them know they need to refresh the page

https://github.com/user-attachments/assets/0d6bf3ef-d90d-4114-ac2c-77c8743b6350


@shared/api/helpers/extensionMessaging.ts
```diff
@@ -67,11 +68,21 @@ export const sendMessageToContentScript = (msg: Msg): Promise<Response> => {
 
 export const sendMessageToBackground = async (msg: Msg): Promise<Response> => {
   let res;
+
+  // grab the active public key from the popup and pass it along with every message.
+  // this will be used to determine if the popup is open to the correct account.
+  const extensionState = popupStoreFacade.currentStore?.getState();
```
piyalbasu, line 74: I was looking for a solution where we didn't have to remember to always send the "active" public key to the background with every request as that would be easy to forget in the future.

This solution automatically appends the active public key to every request that goes to the message listener.

@shared/api/helpers/popupStoreFacade.ts
```diff
@@ -0,0 +1,9 @@
+/* 
+  This object can store data from `popup` without having to pass it as function params.
+  Generally, we prefer passing data as function params, but there are cases where we want to transmit
+  lots of data on every call. This is is a way to do that.
+*/
+
+export const popupStoreFacade = {
```
piyalbasu, line 7: By hanging the popup's redux store off of this object, the background always has access to the current state of the store. It can just call `popupStoreFacade.currentStore.getState()` to get the full state of the popup's redux store. 

I went over a few different options on how to do this:

One option was to send the store/parts of the store along with every request to the background. I moved away from this for the reasons mentioned above about having to manually include this on every request.

I also tried refactoring to move the entire redux store to the `@shared` folder, but that ended up becoming a huge refactor.

I settled on this is as the least invasive. @CassioMG @aristidesstaffieri Lmk what you think!

@shared/api/internal.ts
```diff
@@ -1221,6 +1221,10 @@ export const saveSettings = async ({
     console.error(e);
   }
 
+  if (response.error) {
+    throw new Error(response.error);
+  }
```
piyalbasu, line 1226: now that an account mismatch returns an error, we should make sure all of our services properly handle errors

@shared/api/internal.ts
```diff
@@ -1529,3 +1543,28 @@ export const simulateTransaction = async (args: {
...
+  });
+
+  if (response.error) {
+    throw new Error(response.error);
+  }
+
+  return {
+    isAccountMismatch: response.isAccountMismatch,
+    error: response.error,
+  };
```
piyalbasu, line 1569: this duck will be used to check account mismatch status and set the redux state accordingly

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -1818,8 +1831,19 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
 
+  const currentState = sessionStore.getState();
+  const publicKey = publicKeySelector(currentState);
+
+  if (
+    request.activePublicKey &&
+    request.activePublicKey !== publicKey &&
+    request.type !== SERVICE_TYPES.GET_IS_ACCOUNT_MISMATCH
+  ) {
+    return { error: "Public key does not match active public key" };
```
piyalbasu, line 1845: user will return early here in case of account mismatch; the request will not make it through to the messageResponder

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -342,6 +342,17 @@ export const TransactionDetails = ({
   const { scannedAsset: scannedDestAsset } = useScanAsset(destAssetToScan);
   const isDestAssetSuspicious = isAssetSuspicious(scannedDestAsset);
 
+  // user has refreshed the page and lost needed transaction data. Force them to start over
+  if (amount === "0") {
```
piyalbasu, line 346: slightly unrelated fix; kick the user out of send/swap flow if they refreshed in the middle

@shared/api/helpers/getDomainFromIssuer.ts
```diff
@@ -19,6 +19,7 @@ export const getDomainFromIssuer = async ({
   try {
     /* First, check our localStorage cache in Background to see if we've found this url before */
     ({ assetDomain } = await sendMessageToBackground({
+      activePublicKey: null,
```
piyalbasu, line 22: there are some instances where a user may not yet have a public key (for ex: during onboarding or before the `loadAccount` request resolves). We don't want to block these types of requests but we do want to make a conscious choice to omit the `activePublicKey` here. In these cases, let's use `null`

extension/src/popup/ducks/accountServices.ts
```diff
@@ -65,9 +65,11 @@ export const createAccount = createAsyncThunk<
 
 export const fundAccount = createAsyncThunk(
   "auth/fundAccount",
-  async (publicKey: string) => {
+  async ({ publicKey }: { publicKey: string }, { getState }) => {
+    const activePublicKey = publicKeySelector(getState() as AppState);
```
piyalbasu, line 69: instead of making every component pass the `activePublicKey` through, I thought it was a bit neater just to fetch this value at the `ducks` level since the redux thunk has access to the store

@shared/api/helpers/popupStoreFacade.ts
piyalbasu: Talked with @aristidesstaffieri and we came up with a better way to do this in: https://github.com/stellar/freighter/pull/1848/commits/47b4898c5d140b06046e5e51c53e7c9a31cb1093

aristidesstaffieri (approved):

extension/src/popup/ducks/accountServices.ts
CassioMG: good call

@shared/api/internal.ts
```diff
@@ -100,9 +101,16 @@ export const createAccount = async (
   return { allAccounts, publicKey, hasPrivateKey };
 };
 
-export const fundAccount = async (publicKey: string): Promise<void> => {
+export const fundAccount = async ({
+  activePublicKey,
+  publicKey,
```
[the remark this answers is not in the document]
CassioMG: awesome

@shared/api/helpers/getDomainFromIssuer.ts
CassioMG: are we omitting it here because the `activePublicKey` shouldn't matter for fetching a asset domain?

@shared/api/helpers/getIconUrlFromIssuer.ts
```diff
@@ -43,6 +43,7 @@ export const getIconUrlFromIssuer = async ({
   try {
     /* First, check our localStorage cache in Background to see if we've found this url before */
     ({ iconUrl } = await sendMessageToBackground({
+      activePublicKey: null,
```
CassioMG, line 46: same question here, I guess it also doesn't matter for fetching the icon

@shared/api/internal.ts
```diff
@@ -960,6 +1024,7 @@ export const handleSignedHwPayload = async ({
 export const signTransaction = async (): Promise<void> => {
   try {
     await sendMessageToBackground({
+      activePublicKey: null,
       type: SERVICE_TYPES.SIGN_TRANSACTION,
     });
```
CassioMG, line 1029: I'm just trying to understand it a little better, would you mind explaining why we should omit `activePublicKey` in `SIGN_TRANSACTION` but is forwarded on `SIGN_FREIGHTER_TRANSACTION`? Thanks!

CassioMG: I see the other Freighter API methods like `SIGN_BLOB` and `SIGN_AUTH_ENTRY` also don't take it probably under the same reason, but it's not 100% clear to me why that would be the case - thanks!

@shared/api/internal.ts
```diff
@@ -1358,28 +1490,41 @@ export const loadSettings = (): Promise<
     ExperimentalFeatures & { assetsLists: AssetsLists }
 > =>
   sendMessageToBackground({
+    activePublicKey: null,
     type: SERVICE_TYPES.LOAD_SETTINGS,
   });
```
CassioMG, line 1495: I guess user settings is account agnostic, is that why we're omitting it here?

CassioMG: ah, scrape that. I think it's because we don't need the `activePublicKey` to `read` data, we'll only need it to `change/update` data to ensure we're not messing things up with old active accounts. So since this function is a "reader" it should be fine to omit it. Is my thought correct?

extension/src/background/messageListener/popupMessageListener.ts
CassioMG: nice safeguard

extension/src/helpers/activePublicKeyMiddleware.ts
```diff
@@ -0,0 +1,36 @@
+import { Action, Middleware } from "redux";
+import {
+  getIsAccountMismatch,
+  publicKeySelector,
+} from "popup/ducks/accountServices";
+import { AppDispatch } from "popup/App";
+
+/* 
+  This middleware is design to check if there is a mismatch between the popup's active account and the
```
CassioMG, line 9:
```suggestion
  This middleware is designed to check if there is a mismatch between the popup's active account and the 
```

extension/src/popup/components/AccountMismatch/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+              </div>
+            </View.Content>
+          </div>
+          <LoadingBackground isActive isFullScreen isOpaque />,
+        </>,
+
+        document.querySelector("#modal-root")!,
+      )
+    : null;
+};
```
CassioMG, line 48:
```suggestion

    if (isAccountMismatch) {
      return createPortal(
        <>
          <div
            className="AccountMismatch__notification"
            data-testid="account-mismatch"
          >
            <View.Content hasNoTopPadding hasNoBottomPadding>
              <div className="AccountMismatch__notification__content">
                <div>
                  <Notification
                    title={t(
                      `Your session with the public key ${truncatedPublicKey(publicKey)} has ended. Please refresh the browser.`,
                    )}
                    variant="warning"
                  />
                </div>
              </div>
            </View.Content>
          </div>
          <LoadingBackground isActive isFullScreen isOpaque />,
        </>,

        document.querySelector("#modal-root")!,
      );
    };
    
    return null;
```

CassioMG (approved):
LGTM! Just left some minor comments

@shared/api/helpers/getDomainFromIssuer.ts
piyalbasu: Yup, exactly. Fetching an asset domain isn't anything specific to a public key

@shared/api/helpers/getIconUrlFromIssuer.ts
piyalbasu: Yup, same. Fetching an icon is not something specific to an account

@shared/api/internal.ts
piyalbasu: Oh actually - this is a great callout! I was under the impression that because we have an account picker on the Sign Transaction/Blob/Auth entry screen, this wouldn't be needed. But upon further inspection, this _is_ needed. I'm adding this now!

@shared/api/internal.ts
piyalbasu: Yup, that's correct. Really the thing is that `settings` are the same across all accounts. They are not divided by account, so loading settings will always be the same no matter the active public key

piyalbasu:
Closing in favor of PR #1892

PR #1853, opened 2025-02-18
separate connected apps by network and public key

### LINKED ISSUE #1652: Limit connected dapps to certain public keys
Right now when you connect a dapp to Freighter, the connection is global. This means if I have 3 addresses in Freighter, and I connect dapp.com, dapp.com can request and instantly receive any one of those 3 public keys on any network.

For now, "connection" just means that Freighter will transmit the public key without prompting the user. But in the future, we may want to give a connected dapps more rights and privileges. In that case, we will want to limit what accounts a dapp can connect to by having a Freighter user have to opt each one of their addresses in.

For example: If I connect public key G1 on Testnet to a dapp, the dapp shouldn't be connected to G2 or G1 on Mainnet. 

* We'll need to adjust our schema to silo connection by public key and network
* We'll need to show users what keys on a network are connected to each dapp in `Connected dapps`. How will we show this?
* Will we want to limit signing of tx's to just keypairs that are connected to the originating dapp?

piyalbasu:
Closes #1652 

Previously, if a user connected a dapp, they would be connected across all networks and public keys. Now, connecting a dapp only connects the active public key on the active network

Screenshot 2025-02-18 at 10 46 51 AM
Screenshot 2025-02-18 at 10 46 58 AM


extension/src/background/helpers/dataStorage.ts
```diff
@@ -290,6 +293,31 @@ export const addBlockaidAnnouncedIsEnabled = async () => {
...
 
+export const migrateAllowlistToKeyNetworkSchema = async () => {
+  const localStore = dataStorageAccess(browserLocalStorage);
+  const storageVersion = (await localStore.getItem(STORAGE_VERSION)) as string;
+
+  if (!storageVersion || semver.lt(storageVersion, "4.6.0")) {
+    const currentAllowlist = await localStore.getItem(ALLOWLIST_ID);
+
+    if (currentAllowlist) {
+      const lastUsedAccount = await localStore.getItem(LAST_USED_ACCOUNT);
```
piyalbasu, line 304: On migration, I didn't want to erase a user's current allowlist, but it also felt like a security risk to automatically opt all their Mainnet accounts into the existing allow list. As a compromise, I'm assigning any existing allowlist to the user's `lastUsedAccount` on Testnet

extension/src/background/helpers/dataStorage.ts
```diff
@@ -290,6 +293,39 @@ export const addBlockaidAnnouncedIsEnabled = async () => {
   }
 };
 
+export const migrateAllowlistToKeyNetworkSchema = async () => {
+  const localStore = dataStorageAccess(browserLocalStorage);
+  const storageVersion = (await localStore.getItem(STORAGE_VERSION)) as string;
+
+  if (!storageVersion || semver.lt(storageVersion, "4.6.0")) {
+    const currentAllowlist = await localStore.getItem(ALLOWLIST_ID);
```
piyalbasu, line 301: On migration, I didn't want to erase a user's current allowlist, but it also felt like a security risk to automatically opt all their Mainnet accounts into the existing allow list. As a compromise, I'm assigning any existing allowlist to the user's lastUsedAccount on Testnet

piyalbasu:
Screenshot 2025-02-18 at 11 31 16 AM

Also including the current network on the Grant Access screen

piyalbasu:
Closing this PR as it's diverged from the release branch. Re opened in #1882

PR #1872, opened 2025-02-25
[REFACTOR] Account balances data workflow

### LINKED ISSUE #1809: Refactor: Account balance hook
Today our account balances data fetching workflow uses a key in the transactionSubmission redux store and a thunk in order to supply account balances. Since we use account balances across many views, this has the side effect of forcing views to deal with stale loading flags and store state. We can improve our balances fetching workflows by using a data fetching hook in a similar way to history refactor https://github.com/stellar/freighter/pull/1698

aristidesstaffieri:
Closes #1809 

**What**
Refactors `getAccountBalances` from the `transactionSubmission` store into a standalone hook, this has lots of downstream effects that I will detail below.

Many views and other data fetching functions rely on the current store implementation of balances, so in order to refactor this into a hook, those flows also needed to be refactored.
In order to maintain/improve the data related UX, several views have been refactored to use "data hooks" instead of relying on the store balances.

**Data fetching changes**
Views now take more control over their data fetching flows by implementing data hooks, which are co-located with views in the source code. These hooks serve as a unified effect that encompasses all data that a view needs to represent under one lifecycle. In cases where we design for more lifecycles, we can add more hooks.

Keeping all related data needs in one hook encourages consuming views to not piece together different APIs into separate effects in order to construct a loading or error state.

The new data hooks expose a way to fetch data, types, and helpers to deal with the response types in a more type safe way from the consumer.

General data hooks - `extension/src/helpers/hooks`
View specific data hooks - `extension/src/views/<VIEW>/hooks`

**Why co-locate the data hooks?**
Co-locating data hooks with views enforces that these hooks are purpose built for the views they live next to, in a similar way to how we co-locate css specific to views.
There are more generic data fetching hooks that do not live co-located with views, and these hooks can be consumed by view data hooks when views need to build a hook that is composed of more than 1 data fetching routine.

**View changes**
Views that were impacted by the refactor now consume their data hooks and use state from the hooks to render different markup states, this should result in less views having disconnected effects that toggle internal view state derived from external routines. 

Future work should include more refactors to how we organize these views in order to make cleaner use of the state of the hooks. Patterns like `data?.<data key>` are hints that the views can be re-organized around the hook states in order to branch on different states(loading/success/error) and narrow the types from the hook state in each branch in order to have view code that can cleanly consume state without having to be defensive against the different possibilities.

**Type changes**
We previously used different representations of balances across different views(BalanceMap, Balances, Asset, TokenAsset, etc). We now use 1 representation of balances throughout our views, `AssetType[]`.
This type is a more strict union type of the different types of assets that are parts of our balances. 

**Why a list of assets instead of a map?**
Before we used a mix of balance types, sometime using the map and sometimes using the list. After reviewing I realized that in the majority of cases where we operate on 1 specific balance, we are iterating over balances before doing so. After refactoring routes this will become almost every case(with the exception of picking a balance from an asset in a transaction xdr). The benefit of the faster access in a map over iterating is minimal since balances are relatively short lists and having one common balance type across the application provides more consistency in balance processing/helpers.

This new type is supported by several helpers to help consuming code process balances, extract specific balances with type safety, and narrow a balance down to one specific type of asset. 

Code that consumes balances also needed to be re-worked in order to deal with the stricter types, often this involved narrowing types where we used defensive code in the previous version. This change allows the correct types to flow through the consuming code more deeply.

**Side effects**
Due to our move away from balances cached in the store, more views now fetch balances and other related data. This has the side effect of more loading states on pages that previously did not have them(but were also broken with certain user actions like page reloads). This side effect is temporary and will be alleviated by the next phase of the tech debt refactors which involves re-factoring our routes in order to not expose intermediary steps in flows as top level routes. When that happens, the inner steps of flows like send/swap will be able to use data from their parent components through props and these loading states will not be necessary. 

**TODOs**
In order to keep this scoped to only changes that were necessary to make due to the balances changes, there are several `TODOs` left in comments. Most of these point out already missing states that are now available through new state in data hooks. These can be addressed and removed when we refactor views to properly branch according to the types returned from their data hooks.


@shared/api/types/account-balance.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface LiquidityPoolShareAsset {
+  available: BigNumber;
+  total: BigNumber;
+  liquidityPoolId: string;
+  reserves: Horizon.HorizonApi.Reserve[];
+  limit: string;
+}
+
+export type AssetType =
```
aristidesstaffieri, line 47: this new union of the different asset types should better represent that actual different asset types, using a union here makes consumers of `balances` deal with the differences at the type level.

extension/src/helpers/request.ts
```diff
@@ -0,0 +1,23 @@
+import { Action, IdleState, RequestState, State } from "constants/request";
+
+export const initialState: IdleState = {
+  state: RequestState.IDLE,
+  data: null,
+  error: null,
+};
+
+export const reducer = <T, K>(
```
aristidesstaffieri, line 9: this reducer can be used by any "data hook" in order to support a consistent data lifecycle across our data hooks.

extension/src/helpers/stellar.ts
```diff
@@ -82,7 +88,7 @@ export const getAssetFromCanonical = (canonical: string) => {
 
 export const getCanonicalFromAsset = (
   assetCode: string,
-  assetIssuer: string,
+  assetIssuer?: string,
```
aristidesstaffieri, line 91: this was already being used without an issuer for xlm, just cleaning up the type.

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -233,27 +234,30 @@ export const AccountAssets = ({
           key: "",
         };
         let code = "";
-        if (rb.liquidityPoolId) {
+        if ("liquidityPoolId" in rb) {
```
aristidesstaffieri, line 237: We need to use the `in` syntax instead of checking for the property by access in order to let TS narrow the type. When we do this, the statement will understand that the type was narrowed down to whatever members of the union the assertion is true for.

extension/src/popup/components/sendPayment/SendSettings/Settings/hooks/useGetSettingsData.tsx
```diff
@@ -0,0 +1,283 @@
...
+      const payload = {
+        balances: balancesResult,
+        ...simResponse,
+      } as GetSettingsData;
+      // some flows still use transaction details from the store
+      const simulationResponse =
+        simResponse.payload && "simulationTransaction" in simResponse.payload
+          ? simResponse.payload?.simulationTransaction
+          : "";
+      reduxDispatch(saveTransactionFee(payload.recommendedFee));
```
aristidesstaffieri, line 262: we still need to update the store here until we do the work to refactor the routes in order to not depend on this state across views.

extension/src/popup/helpers/balance.ts
```diff
@@ -1,46 +1,103 @@
...
+import {
+  AssetType,
+  ClassicAsset,
+  LiquidityPoolShareAsset,
+  SorobanAsset,
+} from "@shared/api/types/account-balance";
 import { NetworkDetails } from "@shared/constants/stellar";
 import { isContractId } from "./soroban";
 
+export const isClassicBalance = (balance: AssetType): balance is ClassicAsset =>
```
aristidesstaffieri, line 15: these helpers are shorthand to help us narrow types in places where we are working with `AssetType`[] and need to operate on keys that only belong to some members of the union type.

@shared/api/types/backend-api.ts
```diff
@@ -0,0 +1,15 @@
+import { AssetBalance, NativeBalance, TokenBalance } from "./types";
+
+export interface BalanceMap {
+  [key: string]: AssetBalance | NativeBalance | TokenBalance;
+  native: NativeBalance;
+}
+
+export type Balances = BalanceMap | null;
+
+export interface AccountBalancesInterface {
```
aristidesstaffieri, line 10: Started moving types that are specific to API response to this file, more todo. 
View code now uses `AssetType`[] for balances.

@shared/api/types/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./types";
```
aristidesstaffieri, line 1: This is needed to gradually move away from the 1 big types file, in order to import types from there into the new type files, while maintaining the old imports.

extension/src/helpers/hooks/useGetAssetDomains.ts
```diff
@@ -1,65 +1,89 @@
...
+import { getNativeContractDetails } from "../../popup/helpers/searchAsset";
+import { useIsSoroswapEnabled, useIsSwap } from "../../popup/helpers/useIsSwap";
+import { getAssetDomain } from "../../popup/helpers/getAssetDomain";
+import {
+  AccountBalances,
+  isGetBalancesError,
+  useGetBalances,
+} from "./useGetBalances";
+
+export const isGetAssetDomainsError = (
```
aristidesstaffieri, line 25: These kind of helpers can be used by other hooks or views consuming this in order to narrow the type and have a type safe value in their branching logic.

extension/src/helpers/metrics.ts
```diff
@@ -161,3 +162,40 @@ export const emitMetric = async (name: string, body?: any) => {
   });
   await uploadMetrics();
 };
+
+export const storeBalanceMetricData = (
```
aristidesstaffieri, line 166: just moving this out of the store.

extension/src/helpers/hooks/useGetAssetDomains.ts
```diff
@@ -137,27 +154,17 @@ export const useFetchDomains = () => {
...
-    balances,
-    isManagingAssets,
-    isSorobanSuported,
-    isSwap,
-    isSoroswapEnabled,
-    assetSelect.isSource,
-    soroswapTokens,
-    networkDetails,
-  ]);
+      const payload = { domains, isManagingAssets, balances };
```
aristidesstaffieri, line 157: In every case where we need this data, we're in a top level route that would need to fetch balances in order to get domains. Instead of making those cases wrap this hook and get balances themselves, it gets balances and returns them here as well.

extension/src/popup/components/manageAssets/ManageAssetRowButton/index.tsx
```diff
@@ -130,12 +133,6 @@ export const ManageAssetRowButton = ({
       );
 
       if (submitFreighterTransaction.fulfilled.match(submitResp)) {
-        dispatch(
-          getAccountBalances({
-            publicKey,
-            networkDetails,
-          }),
-        );
```
piyalbasu, line 138: after we change a trustline, why do we not want to re-fetch balances? Our list of balances should be different after tx submission

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -327,89 +235,13 @@ export const TransactionDetails = ({
...
-        code: destAsset.code,
-        networkDetails,
-      });
-      setDestAssetIcons({
-        [getCanonicalFromAsset(destAsset.code, destAsset.issuer)]: iconURL,
-      });
-    })();
-  }, [destAsset.code, destAsset.issuer, networkDetails]);
-
+  // TODO: figure out which view actually needs this
```
piyalbasu, line 240: This is needed for `Confirm Send` so we can let a user know if they forgot to include a memo for a dest address that requires it

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -539,15 +371,56 @@ export const TransactionDetails = ({
...
+      details!.destinationBalances.balances,
+      dest,
+    );
+    if (
+      sourceBalance &&
+      details!.isSourceAssetSuspicious &&
+      "blockaidData" in sourceBalance
+    ) {
+      return sourceBalance.blockaidData;
+    }
```
piyalbasu, line 398: if `sourceBalance` is suspicious, we're terminating early and we're not collecting any info about about `destBalance`. This subtly different than the logic it's replacing which would pass `isSuspicious === true` info about both assets if both were suspicious

extension/src/popup/components/manageAssets/AssetVisibility/index.tsx
```diff
@@ -6,67 +6,76 @@ import { Loader } from "@stellar/design-system";
...
 
   const ManageAssetRowsWrapperRef = useRef<HTMLDivElement>(null);
-
-  const { assets, isManagingAssets } = useFetchDomains();
+  const { state: domainState, fetchData } = useGetAssetDomains(
+    publicKey,
+    networkDetails,
+    {
+      isMainnet: isMainnet(networkDetails),
+      showHidden: false,
```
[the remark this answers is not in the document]
aristidesstaffieri: hey no its not, thanks. I've updated it in 53fe6456e3762bee38153d1a67934547e35b4bd9

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
aristidesstaffieri: ok cool I left it in place where it was and this can likely become a prop once the send routes are collapsed down into 1 route.

extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
```diff
@@ -100,140 +127,51 @@ export const Settings = ({
...
+    publicKey,
+    networkDetails,
+    simulationMode,
+    transactionFee || recommendedFee,
+    {
+      isMainnet: isMainnet(networkDetails),
+      showHidden: false,
+      includeIcons: true,
+    },
+    {
```
piyalbasu, line 142: for a lot of these params, we're just passing data from `transactionDataSelector` to this hook. Would it maybe be a little cleaner just to call `useSelector(transactionDataSelector);` inside of `useGetSettingsData`?

aristidesstaffieri: in the longer term the hope is that this data wont need to be in that store at all, once the route refactors happen all of the args from the previous send steps will be available as props so we'll rely on the store less and less. Right now, I've left some store interactions in select hooks where it was impossible to not do it due to the relationship between the different route views but I don't think this will be necessary without it. Wdyt?

extension/src/helpers/hooks/useGetAssetDomains.ts
```diff
@@ -1,65 +1,76 @@
...
+import { getAssetDomain } from "../../popup/helpers/getAssetDomain";
+import { AccountBalances, useGetBalances } from "./useGetBalances";
+
+export interface AssetDomains {
+  balances: AccountBalances;
+  domains: ManageAssetCurrency[];
+  isManagingAssets: boolean;
+}
+
+export function useGetAssetDomains(
```
piyalbasu, line 27: NIT: one thing that kept tripping me up as I was reviewing is that `useGetAssetDomains` also always refetches and returns balances, which is not clear from the function name. Is it possible to call rename this to reflect that?

aristidesstaffieri: Sure I can rename it to `useGetAssetDomainsWithBalances`. I wrote a comment in the diff describing the thinking behind doing this

```
In every case where we need this data, we're in a top level route that would need to fetch balances in order to get domains. Instead of making those cases wrap this hook and get balances themselves, it gets balances and returns them here as well.
```

@shared/api/helpers/soroban.ts
```diff
@@ -578,3 +579,11 @@ export const getIsTokenSpec = async (contractId: string, serverUrl: string) => {
   const spec = await getContractSpec(contractId, serverUrl);
   return isTokenSpec(spec);
 };
+
+export const xlmToStroop = (lumens: BigNumber | string): BigNumber => {
```
piyalbasu, line 583: pretty NIT: but maybe this file should just be called `stellar.ts` now because `xlmToStroop` isn't specific to Soroban tx's

aristidesstaffieri: yea true, there is already that file in the general shared helpers so I moved it there
2617eef498ded727cfb8d24841bc1b77f94ee18f

extension/src/helpers/hooks/useGetAssetDomains.ts
aristidesstaffieri: renamed in 2617eef498ded727cfb8d24841bc1b77f94ee18f

piyalbasu: Cool - yeah, I think it's helpful to have a bit more context in the code itself

extension/src/popup/components/sendPayment/SendSettings/Settings/index.tsx
piyalbasu: Cool - yeah, I think that's fine for now!

piyalbasu (approved):

PR #1877, opened 2025-02-28
add UI that links out to Coinbase onramp

### LINKED ISSUE #1863: Integrate Coinbase onramp
Given the findings from Bruno and Piyal
https://docs.google.com/document/d/1PATc6zh5RvwQ3SDDYN462GeqmEUUulkB9feNBaPYb9Q/edit?tab=t.0#heading=h.i8g2bve1not7
https://docs.google.com/document/d/1cHYQRocIgV_RFcYD0cipa2Q3glAFfwG1y9Dt1ovM6uc/edit?tab=t.0

We appears to be leaning in the direction of integrating with Coinbase as an on ramp. Coinbase provides a URL-based solution that opens in a new tab. 

Acceptance Criteria: 

In Freighter, implement a button that links out to this experience and sets reasonable defaults like:
1. destination account
2. asset type

Use this designs to stylize the button that leads the user to this experience:
https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5129&t=PJprU2dPFHyPfAPB-0

NOTE: this design specifies Moonpay. Once confirmed, work with Charles to replace with a Coinbase button

piyalbasu:
Closes #1863 

Implements API call from this PR: https://github.com/stellar/freighter-backend/pull/212

This PR adds a button that fetches a JWT from Freighter BE to enable an authenticated connection to Coinbase's onramp UI. If a user has a Coinbase account, they will be able to use their connected payment methods. If not, they will be asked to enter debit card information and use Coinbase's Guest Checkout experience.

Note that in this PR, we simply link out to Coinbase and let the Coinbase handle the flow. We simply pass the destination address and the desired asset (XLM) along to Coinbase, rather than giving them a UI to pre-select their `asset` and `amount`. I've taken this approach for a few reasons:

1. We only care about XLM (right now), so we would just be showing a field for an `amount`. That really doesn't save the user much time as they will need to go through Coinbase's UX to complete the transaction anyway
2. This simpler integration allows Coinbase to make changes to both security and UX on their end without us having to worry about it.
3. A simpler integration allows us to deliver on this feature faster and in a way that will be very easy to port to mobile

NOTE: slow load times and RPC warnings in the demo video are due to me running this against a local instance of Freighter BE

https://github.com/user-attachments/assets/3c18373e-d8de-4906-901b-495a95e2b63d


extension/e2e-tests/addXlm.test.ts
```diff
@@ -0,0 +1,30 @@
+import { test, expect, expectPageToHaveScreenshot } from "./test-fixtures";
+import { loginAndFund } from "./helpers/login";
+
+test("should show add XLM page and open Coinbase", async ({
```
piyalbasu, line 4: this test will not pass until Freighter BE is deployed

extension/src/popup/views/AddXlm/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const handleOnrampClick = async () => {
+    const options = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ address: publicKey }),
+    };
+    const url = `${INDEXER_URL}/onramp/token`;
+    const response = await fetch(url, options);
```
aristidesstaffieri, line 34: Should there be a loading state for this action?

I think even though it opens a link, there is some unknown amount of time that this call needs to resolve between the time the user clicks the button and the time the link opens, when this takes longer than a user's perception of instant, it could feel like they clicked a button and nothing happened.

Can I suggest that we try moving towards our data hook style of fetching data in order to re-use the same patterns and primitives used in hooks like `useGetHistory`?

JakeUrban:
Why do we only care about XLM? USDC ramping is free.

JakeUrban:
I also thought we’d have the option to buy assets in the asset details page for both XLM and USDC?

piyalbasu:
> Why do we only care about XLM? USDC ramping is free.

I guess it's not that we don't care, it's that we've specifically designed screens for "Adding XLM", in the context of funding/creating accounts on Mainnet. Sounds there's a disconnect here on what the expectation of this ticket is

JakeUrban:
Agreed, the spec doc @brunomuler puts together mentions that we want our ramp provider to support both XLM and USDC, and at the bottom it has a asset selection mockup to select USDC or XLM after selecting a buy button.

Can you sync with Charles and Bruno to confirm what is in scope?

piyalbasu:
> Agreed, the spec doc @brunomuler put together mentions that we want our ramp provider to support both XLM and USDC, and at the bottom it has a asset selection mockup to select USDC or XLM after selecting a buy button.
> 
> Can you sync with Charles and Bruno to confirm what is in scope?

Will do. It's a fairly small addition to add USDC support now, so we can do that easily if needed

aristidesstaffieri (approved):

PR #1884, opened 2025-03-03
Feature/coinbase onramp usdc

piyalbasu:
Part 2 of a ticket for #1863 

I met with Charles and we walked through different scenarios where we'd want the user to be able to add funds. We settled on:

* On Mainnet, when a user has an unfunded account, the can click `Add XLM` and are taken to a screen where they can click `Buy With Coinbase`. Clicking this button will take them to Coinbase where they get dropped into a payment flow with XLM selected (this was already covered in #1877 )

* On Mainnet, when a user clicks into their XLM or USDC asset detail view, they'll have a BUY button. Clicking this button will open Coinbase with the asset pre-selected for them

* On Mainnet, When the user clicks the `...` dropdown, they will see a `Buy with Coinbase` button. When they click this, Coinbase will open without any asset selected. From here, the user can use Coinbase's UI to buy any supported asset they choose


piyalbasu:
https://github.com/user-attachments/assets/082dbb0a-67e2-4f64-a370-994b71ab9848

extension/src/helpers/hooks/useGetOnrampToken.tsx
```diff
@@ -1,15 +1,51 @@
...
 type SuccessReturnType = { token: string | null; error: string | null };
 
-function useGetOnrampToken(publicKey: string) {
+interface GetCoinBaseUrlParams {
+  token: string;
+  asset?: string;
+}
+
+const getCoinbaseUrl = ({ token, asset }: GetCoinBaseUrlParams) => {
+  const selectedAsset = asset ? `&defaultAsset=${asset}` : "";
```
aristidesstaffieri, line 16: will this work if `!asset`? Wondering if the coinbase page will let them pick an asset there or if this should actually not be an optional param.

piyalbasu: This is actually an [optional param](https://docs.cdp.coinbase.com/onramp/docs/api-onramp-initializing), so it's okay not to pass it in the case of `!asset` here. WIthout this param, a user gets a dropdown with the most popular asset on their selected network pre-selected. But they are able to change this in the Coinbase UI. We actually want the experience when the user clicks `Buy with Coinbase` because they haven't specified what asset they're trying to buy.

Screenshot 2025-03-04 at 1 33 13 PM

When they click the BUY button the XLM or USDC asset detail page, we do want to pass this `defaultAsset` query param as the user is specifying what asset they want.

aristidesstaffieri (approved):

PR #1850, opened 2025-02-12
[FIX] Add asset warning details

aristidesstaffieri:
Closes #1790 

What
Fixes alignment on Add Asset warning modal
Adds handling of custom tokens

Screenshot 2025-02-12 at 12 01 13 PM

PR #1889, opened 2025-03-04
[FIX] Add Asset Warnings

### LINKED ISSUE #1790: Tweak "add asset" API modal

aristidesstaffieri:
Closes #1790 

Reroll of #1850 

What
Fixes alignment on Add Asset warning modal
Adds handling of custom tokens

Screenshot 2025-02-12 at 12 01 13 PM


extension/src/popup/components/WarningMessages/styles.scss
```diff
@@ -385,10 +385,10 @@
   z-index: var(--z-index--scam-warning);
   display: flex;
   position: fixed;
-  height: fit-content;
   width: 100%;
-  bottom: -1.5rem;
   left: 0;
+  top: 0;
```
piyalbasu, line 390: I think this is a bug? Your original PR didn't have this and `top: 0` is causing the modal to align to the top of the screen in your e2e snapshots

aristidesstaffieri: yeah so actually I think I missed this in the original PR, without this we still get the black background that shows behind the element for the window so my thought was that in modals the element that is the modal should always fill the whole window, wdyt?

aristidesstaffieri:
@piyalbasu updated screenshots:

Screenshot 2025-03-06 at 12 52 38 PM


Screenshot 2025-03-06 at 12 51 54 PM

piyalbasu (approved):

extension/src/popup/components/Notification/index.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { Icon } from "@stellar/design-system";
+import classNames from "classnames";
+
+// import IconWarning from "../../assets/icon-warning.svg";
+// import IconShieldPlus from "../../assets/icon-shield-plus.svg";
```
piyalbasu, line 7: let's rm these comments

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -980,71 +999,72 @@ export const TokenWarning = ({
...
-                  )}
-                </Notification>
-              ) : (
-                <UnverifiedTokenNotification />
-              )}
-            </div>
+              <div data-testid="token-warning-notification">
+                {isVerifiedToken ? (
+                  <NotificationV2
+                    description={`This asset is part of the asset list(s): "${verifiedLists.join(", ")}". Freighter uses asset lists to check assets you interact with. You can define your own assets lists in Settings.`}
```
piyalbasu, line 1024: I know we're not using Translation atm, but should we add t("") here just for parity with the rest of the app

extension/src/popup/components/Notification/index.tsx
aristidesstaffieri: thanks, removed in 940424a18d93fae4cf19906ad2acc46515f5d582

extension/src/popup/components/WarningMessages/index.tsx
aristidesstaffieri: yup for sure, added in 940424a18d93fae4cf19906ad2acc46515f5d582

PR #1897, opened 2025-03-07
Feature/decline tx not on allowlist

piyalbasu:
Closes https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=98426935

To enhance security, instead of just warning a user that a requesting dapp is not on the allowlist, let's completely block signing with an informational message. 

As this may breaks some current dapp integrations, we will need to communicate this to the ecosystem before release


extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -188,15 +174,13 @@ export const freighterApiMessageListener = (
...
-            if (!isDomainListedAllowed) {
-              allowList.push(punycodedDomain);
-              localStore.setItem(ALLOWLIST_ID, allowList.join());
-            }
             resolve({
               contractId,
             });
           }
 
+          console.log("RESPONSE");
```
aristidesstaffieri, line 182: stray log?

piyalbasu:
Closing in favor of #1952

PR #1913, opened 2025-03-12
Feature/improve new account flow

piyalbasu:
Closes https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=95415217

This tightens up the flow for creating a new account from the password prompt screen. 

* Previously, the user was not made aware that this would wipe existing data from Freighter. It looked more like the user was simply adding a new account
* When the user does go through with this flow, we now make sure to wipe all data related to the previous account from the background


piyalbasu:
https://github.com/user-attachments/assets/95cfd628-cfdc-48d6-9b8a-ebc969006c88

aristidesstaffieri (approved):

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -431,8 +431,33 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
     return { publicKey };
   };
 
+  /* 
+  The user has lost their password and they want to reimport an account.
+  Remove all references to the previous account in localStore so the user has a clean slate to start over.
+  */
+  const _removePreviousAccount = async () => {
```
CassioMG, line 438: +1

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -431,8 +431,33 @@ export const popupMessageListener = (request: Request, sessionStore: Store) => {
...
+    await localStore.remove(APPLICATION_ID);
+    await localStore.remove(RECENT_ADDRESSES);
+    await localStore.remove(LAST_USED_ACCOUNT);
+    await localStore.remove(TOKEN_ID_LIST);
+
+    const keyIdList = await getKeyIdList();
+
+    for (let i = 0; i < keyIdList.length; i += 1) {
+      const k = keyIdList[i];
+      await localStore.remove(`stellarkeys:${k}`);
```
CassioMG, line 449: is this the list of account stellar addresses the user has created?

CassioMG: like in the format `stellarkeys:GCGP37AHTWZXS7SRWEP6KNXB4U4U3D473BL3WKAPGXBBDIOGGCA367YP`

piyalbasu: Yup, that's mostly correct:

These are the encrypted blobs that are created by `@stellar/typescript-wallet-sdk-km`. The format is `stellarkeys:{keyID}`. So, for ex, it might be `stellarkeys:0.987383812837182`

CassioMG: aah, got it. So it's actually encrypted there. Good to know, thanks for the clarification!

piyalbasu:
Close in favor of #1951

PR #1920, opened 2025-03-17
restore a default page for both xlm purchases and generic purchases

piyalbasu:
After further discussion with product/design, updating the Coinbase on ramp flow #1863 

Instead of jumping straight to Coinbase, we'll now show users a landing screen for adding funds from Coinbase. A query param will allow users to default to an XLM purchase experience if we link from the "Add XLM" button. Otherwise, they will just see a generic "Add Funds" screen where they will be able to choose what asset they'd like to purchase in Coinbase's UI

Figma: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2708-11212&m=dev


piyalbasu:
https://github.com/user-attachments/assets/529d317c-601d-49f1-b763-bc137a52e32e

aristidesstaffieri (approved):

CassioMG (approved):

PR #1922, opened 2025-03-19
pin @types/react to stable version

piyalbasu:
There was an issue found in the prod pipeline where the latest version of @types/react was unexpectedly causing a breaking change. This PR pins the version to a stable version we have previously tested and QA'ed


CassioMG (approved):
Thanks for fixing this!

PR #1930, opened 2025-03-26
Feature/copy paste phrase

### LINKED ISSUE #1072: Ability to paste secret recovery phrase

I want to be able to copy and paste the full secret recovery phrase, and not having to manually enter each words.

piyalbasu:
Closes #1072 

After talking with the Bruno and the security team, we're declining to add the "copy mnemonic phrase" at the moment. Bruno is working on a new product spec for onboarding and he is going to work a more secure way to copy the mnemonic phrase into that flow. Due to ongoing security vulnerabilities (demonic, StilachiRAT), we need to be very careful about how we bring this functionality back. Just presenting a "Copy mnemonic phrase" button with no warning/information is not a good idea

We can, however, allow users to paste their recovery phrase in the "Import Wallet" flow. After a user does so successfully, we will clear the clipboard for them to prevent the mnemonic phrase from living longer in the clipboard than is necessary or safe


extension/src/popup/basics/buttons/BackButton/index.tsx
```diff
@@ -29,7 +31,7 @@ export const BackButton = ({
 
   if (customButtonComponent) {
     const ClonedComponent = customButtonComponent;
-    return React.cloneElement(ClonedComponent, { onclick: handleClick });
+    return React.cloneElement(ClonedComponent, { onClick: handleClick });
```
piyalbasu, line 34: this is a fix for an unrelated issue. Due to an incomplete type, we had change the handler to `onclick`, but that does not actually work. This reverts back to `onClick` and fixes the type so we don't get an incorrect warning

aristidesstaffieri: woah crazy, nice catch.

aristidesstaffieri (approved):

PR #1932, opened 2025-03-27
hide dust payments in asset details history

### LINKED ISSUE #1888: Hide dust payments in asset details view

Reduces clutter in the asset details screen.

All payments less than 0.1 XLM should be filtered out.

This might be overly complex, but we could add a rule to hide all payments under a specified dollar value. If we filter out XLM payments less that 0.1, they just might start sending 0.0000001 USDC for example.

piyalbasu:
Closes #1888 

Hide dust payments in the Asset Details screen (if the user has that setting configured)


aristidesstaffieri (approved):

piyalbasu:
Closing in favor of #1949

PR #1950, opened 2025-04-03
[REFACTOR] Remove send/swap routes

### LINKED ISSUE #1938: Refactor send flow to only use 1 top level route
Right now, we use a route for every step of the send flow. We should refactor the send flow so that each step is a child of the send route in order to avoid users breaking the flow by reloading or restarting it in the middle.

Send uses 1 route, all other steps are children of this route

Use local storage to keep route state and re-route when to beginning when its missing.

### LINKED ISSUE #1947: Refactor swap flow to only use 1 top level route

aristidesstaffieri:
Closes #1938 
Closes #1947 

What
Refactors the send & swap workflows to no longer use route based steps and only use 1 route with child steps to go through the flow.

Why
In the current route based implementation, users can break the application by refreshing the route at any point through the flow because each step relies on data configured from previous steps. Using a single component mounted a single route fixes this and allows us to refactor the application away from keeping workflow data in a store because all workflow data and context can be kept in the parent component mounted at the route.

Not Included
This does not include the refactor to pull workflow data out of the transactionSubmission store, but it does unlock the path for that refactor. We can now pull out the send/swap settings and transactionXdr from the store and keep this state in the route level parent component in order to decouple the components from that store state, which is prone to problems due to being shared across views and needing to be reset by the different views using it.

This also does not include improvements to our loading states through the steps, but we can now move common data fetching(like balances) to the parent route level component to avoid fetching it for the steps that need it along the way. This should make the workflow faster.


extension/e2e-tests/sendPayment.test.ts-snapshots/send-payment-settings-chromium-darwin.png
aristidesstaffieri, line 1: unclear about why the simulator renders this button differently, it used to look like it was being hovered but now renders as it actually looks in a real browser but there is no change in the way this actually renders so updated the snapshot.

extension/src/popup/components/manageAssets/SelectAssetRows/index.tsx
```diff
@@ -138,9 +139,7 @@ export const SelectAssetRows = ({ assetRows, balances }: SelectAssetRowsProps) =
                   <div>
                     {isContract
                       ? getTokenBalanceFromCanonical(issuer)
-                      : formatAmount(
-                          getAccountBalance(issuer || "native"),
-                        )}{" "}
+                      : formatAmount(getAccountBalance(issuer))}{" "}
```
aristidesstaffieri, line 142: this was previously set incorrectly causing xlm to not be found by `getAccountBalance`

extension/e2e-tests/sendPayment.test.ts-snapshots/send-payment-settings-chromium-darwin.png
piyalbasu: Oh yeah - weird. FWIW,  I forced regenerating this screenshot and I got the same result

piyalbasu (approved):

PR #1952, opened 2025-04-05
if domain not allowed, completely block instead of warn

piyalbasu:
Re-roll of #1897 

Closes https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=98426935

To enhance security, instead of just warning a user that a requesting dapp is not on the allowlist, let's completely block signing with an informational message. 

As this may breaks some current dapp integrations, we will need to communicate this to the ecosystem before release


extension/e2e-tests/freighterApiIntegration.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      "docs.freighter.app is currently not connected to Freighter",
+    ),
+  ).toBeVisible();
+  await expectPageToHaveScreenshot({
+    page: popup,
+    screenshot: "domain-not-allowed.png",
+  });
+});
+
+test.skip("should sign message when allowed", async ({ page, extensionId }) => {
```
piyalbasu, line 189: note this test is not passing because these e2e tests caught that we're not properly passing the network in @stellar/freighter-api/signMessage

The fix is here: https://github.com/stellar/freighter/pull/1952/files#diff-a07d7b44912cd9de72ec5cbce0973a3ffa42d8f1d9d80096eb4de1efcd044300R141

I'll deploy a new version of freighter-api along with this release and that will make this test start working

piyalbasu:
In the latest commit, I'm adding a UX update:

When the sign (tx|message|auth entry) modal pops up, we give the user the chance to switch which account they are signing with. Because `allowList` is now segmented by public key, it's possible that the account they switch to is NOT connected to the dapp.

To combat that, I'm moving the `isDomainListedAllowed` check from freighterApiMessageListener (where it is only calculated once, using the current public key) to the UI, where it can be checked and re-checked every time the account changes.

Also, if the domain is not allowlisted, I don't completely block the UI so the user still has the option to change accounts to one that is allowlisted.

Screenshot 2025-04-07 at 3 49 35 PM

@stellar/freighter-api/src/signAuthEntry.ts
```diff
@@ -8,13 +12,22 @@ export const signAuthEntry = async (
...
     address?: string;
-  }
+  },
 ): Promise<
   { signedAuthEntry: Buffer | null; signerAddress: string } & {
     error?: FreighterApiError;
   }
 > => {
   if (isBrowser) {
+    const { isAllowed } = await requestAllowedStatus();
```
piyalbasu, line 22: at the freighter-api level, if the current public key is not connected to the dapp, give the user an opportunity to connect

extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -171,21 +155,7 @@ export const freighterApiMessageListener = (
...
-        publicKey,
-        networkDetails,
-      });
-      const isDomainListedAllowed = isSenderAllowed({
-        sender,
-        allowListSegment,
-      });
-
       const tokenInfo: TokenToAdd = {
-        isDomainListedAllowed,
```
piyalbasu, line 188: as mentioned in the commit description, we are moving this check to the UI where it can be calculated dynamically

aristidesstaffieri (approved):

PR #1954, opened 2025-04-08
[FEATURE] Adds support for icons from token lists

### LINKED ISSUE #1791: Token list refresh

aristidesstaffieri:
Closes #1791 

What
Adds an API level lookup for icon URLs from any active token lists in order to support icons from tokens that don't have a home domain( like SEP-41 tokens, USDL, etc. ).

Why
SEP-41 tokens currently have no support for icons but its possible for them to be added to community token lists with a corresponding icon URL.

Related:
Adds USDL to Soroswap's token list
https://github.com/soroswap/token-list/pull/55/files

Screenshot 2025-04-08 at 11 16 21 AM


extension/src/popup/__testHelpers__/index.tsx
```diff
@@ -254,3 +254,24 @@ export const mockAccountHistory = [
     type_i: 1,
   },
 ];
+
+export const validAssetList = {
+  name: "PiyalBasu Top 50",
```
CassioMG, line 259: good list name 💯

extension/src/popup/helpers/__tests__/getIconFromTokenLists.test.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+  it("should return an icon if an asset is in a token list by issuer", async () => {
+    const icon = await getIconFromTokenLists({
+      networkDetails: TESTNET_NETWORK_DETAILS,
+      id: VERIFIED_TOKEN_ISSUER,
+      assetsLists: DEFAULT_ASSETS_LISTS,
+    });
+    expect(icon).toEqual(EXPECTED_ICON_URL);
+  });
+  it("should return undefined if an asset is on on the token list", async () => {
```
CassioMG, line 32:
```suggestion
  it("should return undefined if an asset is not on the token list", async () => {
```

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -95,7 +95,7 @@ export const AssetIcon = ({
   }
 
   // Get icons for Soroban tokens
```
CassioMG, line 97:
```suggestion
  // Get icons for Soroban tokens which are not present in assetIcons list
```

@shared/api/helpers/getIconFromTokenList.ts
```diff
@@ -0,0 +1,45 @@
...
+  AssetListReponseItem,
+  AssetsLists,
+} from "@shared/constants/soroban/asset-list";
+import { NetworkDetails } from "@shared/constants/stellar";
+import { isContractId } from "./soroban";
+import { getCombinedAssetListData } from "./token-list";
+
+export const getIconFromTokenLists = async ({
+  networkDetails,
+  id,
```
CassioMG, line 11: suggestion: should we have separate `contractId` and `issuerKey` params and let the function return whatever it finds giving priority to contractId? So we don't need to call it twice and await twice like it's happening [here](https://github.com/stellar/freighter/pull/1954/files#diff-2be5811b6be2dbb65455ed40f311049026b12db67e54e87a508650f266b4b09eR1016-R1034).

Then this `getIconFromTokenLists` function could return something like `{ icon, canonicalAsset }` so we could use `canonicalAsset` as the icon key for `assetIcons`.

I think having  `contractId` and `issuerKey` params would also make the function declaration a bit easier to understand since we'd be removing the ambiguity.

What do you think?

CassioMG (approved):
LGTM! Left a suggestion regarding the function inputs and response

Thanks for moving/extracting stuff to @shared/api so it'll be easier to reuse on the mobile wallet! 👊

@shared/api/helpers/getIconFromTokenList.ts
aristidesstaffieri: I updated it to use your suggested pattern, I also pass in the code from the caller because its safer to take the code from the extension to make the canonical vs assuming the code from the token list will match. Wdyt?
e16ae4317324b14cf2884980c7ed0f86112c666b

CassioMG: thanks for the update, looks good to me!

PR #1968, opened 2025-04-15
[REFACTOR] Popup Message Listener

### LINKED ISSUE #1955: Split message listener request and response types
Improved correctness of our message passing pipeline between the background and popup

Split the [Response and Request types](https://github.com/stellar/freighter/blob/master/%40shared/api/types.ts#L24) and make the type a union of all request and response types
Plumb that through to each messageResponder in order to make it possible for responders to reject invalid requests.

Forgo this refactor and accept the loose typing in this layer.

aristidesstaffieri:
Closes #1955 

**What**
Refactors the popupMessageListener and related code to enable better type safety at the listener boundary.

**Why**
Our message listener uses one big type of optional values to model it's messages which makes it hard for consumers to rely on type safety when crafting a message. This change enables(but does not completely implement) type safety from the background/content script to the popup through the messaging layer by allowing individual message types to flow through the calls.

**Included**
Refactors `Request` -> `ServiceMessageRequest`, which is a union of all possible internal service message types.
Moves all handlers for internal messages previously found in the listener body to their own files, and pulls dependencies into handler arguments.
Moves internal message helpers previously found in the listener body to helper files.
Refactors helpers related to create account to take dependencies as arguments instead of relying on the scope.
Adds a new test for create-account, adds new test helpers/mocks for popup message tests.

**Not Included/TODO**
Does not refactor the external messaging layer
Does not include test for all untested handlers(added create-account test for reference)
Does not make popupMessageListener generic in order for callers to define a message type(depends on external messages refactor)
Does not refactor extensionMessaging helpers to be generic in order for callers to define a message type(and infer a response type). This depends on the external messaging refactor.


@shared/api/types/message-request.ts
```diff
@@ -0,0 +1,387 @@
+import { Transaction } from "stellar-sdk";
```
piyalbasu, line 1: super NIT: but we use camelCasing for our file names in this repo. TBH, I don't really care which one we use, as long as it's consistent. Since we already use camelCasing in this repo, I think we should probably keep the file naming style the same here

aristidesstaffieri: ah you're right, sorry old habit. I'll rename that.

extension/src/background/messageListener/handlers/import-hardware-wallet.ts
```diff
@@ -0,0 +1,39 @@
...
+  const { publicKey, hardwareWalletType, bipPath } = request;
+
+  await storeHardwareWalletAccount({
+    publicKey,
+    hardwareWalletType,
+    bipPath,
+    sessionStore,
+    localStore,
+  });
+  const hasPrivateKeySelector = buildHasPrivateKeySelector(localStore);
```
piyalbasu, line 31: NIT: should we favor passing an object so it will be easier to add params in the future if needed?
`buildHasPrivateKeySelector({ localStore });`

extension/src/background/messageListener/handlers/load-settings.ts
```diff
@@ -0,0 +1,53 @@
...
+export const loadSettings = async ({
+  localStore,
+}: {
+  localStore: DataStorageAccess;
+}) => {
+  await verifySorobanRpcUrls();
+
+  const isDataSharingAllowed =
+    (await localStore.getItem(DATA_SHARING_ID)) ?? true;
+  const featureFlags = await getFeatureFlags();
```
piyalbasu, line 28: similar to comment above - we'll probably want to pass `localStore` to all these `get` helpers

piyalbasu (approved):

extension/src/background/messageListener/handlers/load-settings.ts
aristidesstaffieri: I intentionally didn't refactor all of the helpers to take args yet to keep the scope of this PR as tight as possible, I did do the ones needed for create account in order to show an initial test example for a handler but if you're fine with the increased scope of the PR I can refactor the rest of the helpers.

extension/src/background/messageListener/handlers/import-hardware-wallet.ts
aristidesstaffieri: yeah for sure, will update this.

extension/src/background/messageListener/handlers/fund-account.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getNetworkDetails } from "background/helpers/account";
+
+export const fundAccount = async ({
+  request,
+}: {
+  request: FundAccountMessage;
+}) => {
+  const { publicKey } = request;
+
+  const { friendbotUrl } = await getNetworkDetails();
```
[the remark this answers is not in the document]
piyalbasu: Yeah, unless, I'm mistaken, this does `localStore.getItem(NETWORK_ID)` under the hood

danilosilvackl (approved):

@shared/api/types/message-request.ts
aristidesstaffieri: done in f0fa2f67df0c9f455b836f93315a1ab51b31ae8f

PR #1969, opened 2025-04-15
Feature/add asset unification

### LINKED ISSUE #1792: Improve UX for manually add assets
"curious if we've thought about trying to get rid of the "add an asset" vs "add manually" choice in this flow?
I've seen users get confused by this because they don't know when to search manually or how it's different from the regular search.
Under the hood, one does an asset search and the other does either a contract id or issuer lookup depending on if you enter a G or C key.
I think we can pretty much do the same thing while using one input by parsing the input into: asset code or contract id or public key and doing the search that matches the input type." - @aristidesstaffieri

piyalbasu:
Closes #1792

* We can now search for assets using stellar.expert and add a C address in the same input in Manage Assets.

* I maintained the `Manually Add Asset` screen because stellar.expert doesn't support custom networks or Futurenet. If we decide to deprecate support for this, we can remove this screen

* I also took this opportunity to restyle the `New Asset` warning to follow the same styles as other asset/token warnings. This warning was still using a very old design 

https://github.com/user-attachments/assets/47403244-26a8-4641-bff4-210a80df5b23


extension/README.md
```diff
@@ -12,7 +12,7 @@ this by adding an `.env` file at the path `extension/.env`.
 Inside this file, you can configure a backend by setting a value for the global
 variable `INDEXER_URL`. For example, to connect to the production backend, in
 your `.env` file, you can add the line
-`INDEXER_URL=https://freighter-backend.stellar.org/api/v1`.
+`INDEXER_URL=https://freighter-backend-prd.stellar.org/api/v1`.
```
piyalbasu, line 15: unrelated improvement to README

extension/src/popup/basics/layout/View/index.tsx
```diff
@@ -243,6 +248,7 @@ export const ViewInset: React.FC<ViewInsetProps> = ({
       hasTopBorder ? "View__inset--top-border" : "",
       hasScrollShadow ? "View__inset--scroll-shadows" : "",
       hasNoTopPadding ? "View__inset--no-top-padding" : "",
+      hasTopInput ? "View__inset--top-input" : "",
```
piyalbasu, line 251: this is a small styling fix for an issue I noticed: for screens where there is an input at the top of the screen (Manage assets, send payment), the input highlight was being cut off by the top header when the input was focused. This adds a small amount of space to around the input to accommodate for this highlight

extension/src/popup/components/WarningMessages/styles.scss
```diff
@@ -418,14 +332,23 @@
...
 
   &__code,
   &__domain {
     display: flex;
+    font-size: pxToRem(14px);
     justify-content: center;
   }
 
+  &__domain {
+    color: var(--sds-clr-gray-11);
```
piyalbasu, line 349: there was some discrepancies from the Figma in the styling for these warnings, so addressing while I'm here

extension/src/popup/components/manageAssets/SearchAsset/hooks/useAssetLookup.ts
```diff
@@ -0,0 +1,373 @@
+import { useReducer } from "react";
```
piyalbasu, line 1: this pulls out all the logic for looking up assets based on type into a reusable hook. This hook can be dropped into a shared lib for use in both extension and mobile

extension/src/popup/basics/layout/View/styles.scss
```diff
@@ -170,6 +170,12 @@
     padding-top: 0 !important;
   }
 
+  &__inset--top-input {
+    // this is a variant of no-top-padding
+    // it provides a ltitle room for a top input's focus glow
```
aristidesstaffieri, line 175: tiny typo: ltitle -> little

extension/src/popup/components/manageAssets/SearchAsset/hooks/useAssetLookup.ts
```diff
@@ -0,0 +1,373 @@
...
+        console.error(error);
+        dispatch({ type: "FETCH_DATA_ERROR", payload: DEFAULT_PAYLOAD });
+        return;
+      }
+
+      console.log(assetRows);
+
+      // Only show records that have a domain and domains that don't have just whitespace
+      // We omit these results as a safety precaution and to encourage asset issuers to add a domain to their asset
+      assetRows = assetRows.filter(
```
aristidesstaffieri, line 324: This shows up on the xlm SAC which I don't think we want, along with the new asset warning.
Screenshot 2025-04-16 at 9 42 33 AM

aristidesstaffieri:
When we add a SAC, we should check the the verified lists for that SACs issuer by getting it from the name field, Otherwise we see "asset is not verified warning" for assets that are on the list when adding them by their SAC address.

Screenshot 2025-04-16 at 9 46 42 AM

aristidesstaffieri:
> I maintained the Manually Add Asset screen because stellar.expert doesn't support custom networks or Futurenet. If we decide to deprecate support for this, we can remove this screen

Is this screen still exposed or just left in code so we can decide wether to keep it later?

extension/src/popup/components/manageAssets/SearchAsset/hooks/useAssetLookup.ts
piyalbasu: > This shows up on the xlm SAC which I don't think we want, along with the new asset warning. Screenshot 2025-04-16 at 9 42 33 AM

Ah thanks - this actually exposed a different issue: SDF's stellar.toml doesn't have CORS enabled, which is why we're getting this warning. I'm working with ops to resolve

piyalbasu:
> When we add a SAC, we should check the verified lists for that SACs issuer by getting it from the name field, Otherwise we see "asset is not verified warning" for assets that are on the list when adding them by their SAC address.
> 
> Screenshot 2025-04-16 at 9 46 42 AM

We do actually do this. I believe this is on testnet, right? One of the assets on stellar.expert's testnet asset list is not passing schema validation, so it's completely disregarded the asset list. (This should be improved, but this tends to only be an issue with stellar.expert's testnet asset list). If you try `CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75` on Mainnet, you'll see that USDC evaluates correct as a verified asset

piyalbasu:
> > I maintained the Manually Add Asset screen because stellar.expert doesn't support custom networks or Futurenet. If we decide to deprecate support for this, we can remove this screen
> 
> Is this screen still exposed or just left in code so we can decide wether to keep it later?

It's still exposed. It's what a user will be redirected to if they're not on mainnet or testnet

aristidesstaffieri:
> > When we add a SAC, we should check the verified lists for that SACs issuer by getting it from the name field, Otherwise we see "asset is not verified warning" for assets that are on the list when adding them by their SAC address.
> > Screenshot 2025-04-16 at 9 46 42 AM
> 
> We do actually do this. I believe this is on testnet, right? One of the assets on stellar.expert's testnet asset list is not passing schema validation, so it's completely disregarding the asset list. (This should be improved, but this tends to only be an issue with stellar.expert's testnet asset list). If you try `CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75` on Mainnet, you'll see that USDC evaluates correct as a verified asset

ah yup thanks I didn't notice that, this looks good.

aristidesstaffieri (approved):

PR #1996, opened 2025-04-24
[FIX] Correct destination for muxed addresses

aristidesstaffieri:
What
Restores correct destination for Muxed addresses in send flow.
Adds Muxed address destination test case.

Why
The base account was being used as the destination address.


extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -132,7 +133,7 @@ export const TransactionDetail = ({
                       <div>{t("To")}</div>
                       <div className="InfoRow__right">
                         <KeyIdenticon
-                          publicKey={to}
+                          publicKey={to_muxed || to}
```
aristidesstaffieri, line 136: This was missing in history, showing the muxed address as the "to" field in cases where the field is present.

piyalbasu (approved):

PR #1997, opened 2025-04-24
Bugfix/fix incorrect bal in swap

### LINKED ISSUE #1995: swap uses wrong tokens for balance checks
When a user goes to the swap screen, it shows incorrect asset values, and when I try to execute a swap, it checks those incorrect values, leading to an inability to swap assets that are actually available.

piyalbasu:
Closes #1995 

This fix makes sure to check both asset code and issuer when fetching and rendering balances in the asset selector in Send/Swap. It also makes sure to take this into account when determining if a trustline is active or not

### DIFF
--- a/extension/src/popup/helpers/balance.ts
+++ b/extension/src/popup/helpers/balance.ts
@@ -36,12 +36,20 @@ export const findAssetBalance = (
       "token" in balance && "issuer" in balance.token
         ? balance.token.issuer.key
         : "";
-    return balanceIssuer === asset.issuer;
+    const balanceCode =
+      "token" in balance && "code" in balance.token ? balance.token.code : "";
+    return balanceIssuer === asset.issuer && balanceCode === asset.code;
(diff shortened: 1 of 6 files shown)


extension/src/popup/helpers/balance.ts
```diff
@@ -36,12 +36,20 @@ export const findAssetBalance = (
...
-    return balanceIssuer === asset.issuer;
+    const balanceCode =
+      "token" in balance && "code" in balance.token ? balance.token.code : "";
+    return balanceIssuer === asset.issuer && balanceCode === asset.code;
   }) as Exclude<AssetType, LiquidityPoolShareAsset> | undefined;
 };
 
-export const getBalanceByIssuer = (issuer: string, balances: AssetType[]) =>
-  balances.find((balance) => {
+export const getBalanceByAsset = (
```
piyalbasu, line 45: this fixes an issue where there could be a collision if 2 assets had the same issuer

aristidesstaffieri (approved):

PR #1998, opened 2025-04-25
[BUG] resets submission in store on failed results

aristidesstaffieri:
What
Resets the submission values in the store when a user dismisses the SubmitFail component

Why
We were only resetting the store on successful submissions, so if you submitted a transaction that failed then the settings would be retained in the store and would be the initial state of the forms on the next time the user entered the send/swap flow.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from "react";
+import { useNavigate } from "react-router-dom";
 import { useSelector, useDispatch } from "react-redux";
 import { createPortal } from "react-dom";
 import get from "lodash/get";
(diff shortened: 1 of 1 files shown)


extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -298,7 +298,6 @@ export const SubmitSuccess = ({ viewDetails }: { viewDetails: () => void }) => {
           onClick={() => {
             dispatch(resetSubmission());
             navigateTo(ROUTES.account, navigate);
-            dispatch(resetSubmission());
```
aristidesstaffieri, line 301: this was being dispatched twice for some reason, just removed one.

extension/src/popup/components/sendPayment/SendConfirm/SubmitResult/index.tsx
```diff
@@ -45,7 +46,6 @@ import {
 import { Loading } from "popup/components/Loading";
 
 import "./styles.scss";
-import { useNavigate } from "react-router-dom";
```
aristidesstaffieri, line 48: just moved this to the top to keep the typical import ordering pattern we use.

PR #2019, opened 2025-04-28
Feature/hide claimable balance spam

### LINKED ISSUE #1866: Filter transactions with multiple claimable balance operations from account history

Claimable balances are largely spam, but they clutter user's account history.

We have tentative plans to move these transactions to a "pending transactions" screen in the future, but for right now it would be great to just declutter history by filtering them out.

Wait to do anything until we prioritize adding the "pending transactions" screen that allows you to claim these assets.

piyalbasu:
Closes #1866

Filters out `create claimable balance` tx's from history to reduce noise. In a follow up PR, we will move these to a `Pending` tab so they will still be accessible.

### DIFF
--- a/extension/src/popup/helpers/account.ts
+++ b/extension/src/popup/helpers/account.ts
@@ -90,7 +90,7 @@ export const getIsCreateClaimableBalanceSpam = (
   // transaction_attr does not exist in HorizonOperation type
   const op = operation as any;
   if (op.type === "create_claimable_balance") {
-    if (op?.transaction_attr?.operation_count > 10) {
+    if (op?.transaction_attr?.operation_count > 50) {
       return true;
     }
   }
--- a/extension/src/popup/helpers/account.ts
+++ b/extension/src/popup/helpers/account.ts
@@ -84,6 +84,20 @@ export const getIsDustPayment = (
   "amount" in operation &&
   new BigNumber(operation.amount).lte(new BigNumber(0.1));
 
+export const getIsCreateClaimableBalanceSpam = (
+  operation: HorizonOperation,
+) => {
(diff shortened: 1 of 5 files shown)


extension/src/popup/helpers/account.ts
```diff
@@ -84,6 +84,20 @@ export const getIsDustPayment = (
   "amount" in operation &&
   new BigNumber(operation.amount).lte(new BigNumber(0.1));
 
+export const getIsCreateClaimableBalanceSpam = (
+  operation: HorizonOperation,
+) => {
+  // transaction_attr does not exist in HorizonOperation type
+  const op = operation as any;
+  if (op.type === "create_claimable_balance") {
+    if (op?.transaction_attr?.operation_count > 50) {
```
piyalbasu, line 93: the 50 here is a bit arbitrary - I wanted to keep it low enough that it catches most spam, but high enough that it doesn't accidentally catch real tx's.

In my experience, the vast majority of this type of spam has had 100 operations

aristidesstaffieri: curious why we need to look at op count and can't just filter on type here?

piyalbasu: We're checking for op count bc there are legitimate use cases for `create claimable balance`. What we're trying to hide are the spammy claimable balance ops that try to "hide" a large number of operations inside. In practice, these tend to have a `create claimable balance` op at the top of the tx followed by 99 other ops

aristidesstaffieri: ah gotcha 👍

aristidesstaffieri (approved):

PR #2027, opened 2025-04-30
Moves account and settings loading state to router components

### LINKED ISSUE #1957: Improve and align loading states
Fixes our glitchy initial loading state and improves the UX of our loading states

Fix the glitchy loading state resulting from the settings loading state being separate from the individual state loading states.
Make all loaders the same implementation, size and design.
Implement missing loaders in critical flows.

aristidesstaffieri:
Closes #1957 

Motivation
We want to eliminate the "loading glitch" caused by the different loaders implemented through the react render tree.

**What**
Moves the responsibility to fetch account and settings data from the router outlet to individual route components.
Creates `useGetAppData` hook to be consumed by views/data hooks that rely on account or settings data.
Removes `Loader` in the router outlet in order to implement only 1 single loading state for any given view.
Removes public/private key wrappers in favor of in-view handling of the missing keys(because this logic is no longer handled above the rendering of each component in the dom tree).

**Important Details**
Routes that are top level router component and rely on account or settings data now can no longer use store selectors to access this data, they need to call `useGetAppData` in their data hooks in order to fetch the data that was previously fetched at the router outlet level. This gives those components access to the same data while allowing us to implement 1 single loading state.

`useGetAppData` allows the caller to pass in a parameter to accept the cached values for account and settings details, when this flag is passed in the hook will return the values from the store if they exist, otherwise will fetch them.
This allows us to not have to incur the cost of fetching these details over and over again when switching routes when we have previously fetched this data. It's important that we don't get cached data when switching accounts or networks(see Account component for example).

`useGetAppData` returns a discriminated union keyed by the property `type`. This allows the hook to replace the routing logic that used to live in the router outlet and allows the application to route users based on their `applicationState`.
It also exports types to support this pattern in hooks or views that consume this hook( [AppDataType](https://github.com/stellar/freighter/pull/2027/files#diff-25fe6581be36bdd2196d2d4cdb0c4f8f42f2fb9d559f09a207f9f6687a992a4bR22) & [NeedsReRoute](https://github.com/stellar/freighter/pull/2027/files#diff-25fe6581be36bdd2196d2d4cdb0c4f8f42f2fb9d559f09a207f9f6687a992a4bR26) ). Hooks or views the consume `useGetAppData` can use these type to also return a discriminated union or in views to narrow the return type down to the underlying resolved data types.

Components that are children of top level router components still access account and settings data through store selectors. This is ok because their parents have fetched this data but they can also be refactored in the future to get these as props for a tighter relationship.

Top level router components will now have logic like this to replace the previous logic implemented in the `<PublicKeyRoute>` and `<PrivateKeyRoute>` components.

```
if (accountData.data?.type === AppDataType.REROUTE) {
    if (accountData.data.shouldOpenTab) {
      openTab(newTabHref(accountData.data.routeTarget));
      window.close();
    }
    return (
      <Navigate
        to={`${accountData.data.routeTarget}${location.search}`}
        state={{ from: location }}
        replace
      />
    );
  }
  ```
  
**Why can't this be abstracted into a common component?**
React components can't optionally pass rendering to their children so in order to either render `<Navigate />` or keep processing the component body we need this logic to be inline in the component that needs it. 

One potential future option is to implement a component that renders `<Navigate />` or then calls a [render prop](https://react.dev/reference/react/Children#calling-a-render-prop-to-customize-rendering). This would allow us to abstract the logic to re-route components but would require a significant refactor of all top level route components.

Before:
https://github.com/user-attachments/assets/07bfbf87-fd69-4587-8324-9a636a8b2d3c

After:
https://github.com/user-attachments/assets/cc2a8bed-d506-4aca-b968-66883b4e99b7

### CODE REVIEW

File: extension/e2e-tests/helpers/sendPayment.ts [hunk 1]
```diff
@@ -1,6 +1,8 @@
+import test from "@playwright/test";
 import { expect, expectPageToHaveScreenshot } from "../test-fixtures";
 
 export const sendXlmPayment = async ({ page }) => {
+  test.setTimeout(180_000);
```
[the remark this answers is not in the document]
aristidesstaffieri: yeah it seemed to be in my case. I had noticed previously that this would sometimes timeout but it started happening pretty consistently and in different spots more recently.

extension/src/popup/components/manageAssets/AddAsset/hooks/useGetAddAssetData.tsx
```diff
@@ -0,0 +1,91 @@
+import { useReducer } from "react";
+
+import { RequestState } from "../../../../../constants/request";
```
piyalbasu, line 3: It may not be necessary for this PR - but it'd be nice to have an alias for these helpers so we don't have to do these ugly `../../` imports

aristidesstaffieri: actually yeah there is one already but I couldn't get the compiler to find it from here so let me look at our alias setup again to see if I can figure out why.

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -29,14 +25,20 @@ import {
...
 import "./styles.scss";
+import { useGetAddAssetData } from "./hooks/useGetAddAssetData";
+import { RequestState } from "helpers/hooks/fetchHookInterface";
+import { Loading } from "popup/components/Loading";
+import { NetworkDetails } from "@shared/constants/stellar";
+import { Notification } from "@stellar/design-system";
+import { openTab } from "popup/helpers/navigate";
+import { newTabHref } from "helpers/urls";
+import { AppDataType } from "helpers/hooks/useGetAppData";
+import { reRouteOnboarding } from "popup/helpers/route";
```
piyalbasu, line 41: this isn't a hard rule in our eslint config (though it probably should be) - but in the rest of the codebase, we generally keep external lib imports at the top and relative file paths at the bottom. I think you'd want to move the `design-system` import up and keep the rest of the imports above

```
import { ManageAssetRows, ManageAssetCurrency } from "../ManageAssetRows";
import { SearchInput, SearchCopy, SearchResults } from "../AssetResults";

import "./styles.scss";
```

aristidesstaffieri: ah yeah thanks I'll move these up

extension/src/popup/components/manageAssets/AddAsset/index.tsx
```diff
@@ -254,27 +267,36 @@ export const AddAsset = () => {
...
-        await handleIssuerLookup(contractId);
-      } else {
-        setVerifiedAssetRows([]);
-        setUnverifiedAssetRows([]);
-      }
-    }, 500),
+    debounce(
+      async (
+        { target: { value: contractId } },
+        publicKey: string,
```
piyalbasu, line 273: just curious - I see that for a lot of function calls, instead of inheriting values like `publicKey` from the lexical scope as it was before. What's the benefit of doing this?

aristidesstaffieri: The benefit is the type narrowing that we get by passing this in because it gets called in the body of the component where the return type from the data hook has already been narrowed to a resolved type. If we just got it from the scope then we would have to write a conditional around it to deal with the union type from the data hook since at that point in the body the type hasn't been narrowed yet.

extension/src/popup/components/manageAssets/SearchAsset/index.tsx
```diff
@@ -58,32 +61,36 @@ const ResultsHeader = () => {
 
 export const SearchAsset = () => {
```
piyalbasu, line 62: I noticed that if you refresh this page and then try to search, it throws an error due to a missing asset list

extension/src/helpers/hooks/useGetBalances.tsx
```diff
@@ -28,28 +28,27 @@ export interface AccountBalances {
...
+  includeIcons: boolean;
+}) {
   const [state, dispatch] = useReducer(
     reducer<AccountBalances, unknown>,
     initialState,
   );
   const { assetsLists } = useSelector(settingsSelector);
 
-  const fetchData = async (): Promise<AccountBalances | Error> => {
+  const fetchData = async (
```
piyalbasu, line 41: I wonder if this method should have a `useCache` option, similar to `useGetAppData`. I see that if I open Freighter, go to Manage Assets, and then try to search for an asset, I'm making 3 separate calls to refetch balances  🤔 

I know we've talked before about trying to make sure we have up-to-date data at all times, but I wonder if there are some places where it strategically makes sense to cache to avoid constantly re-fetching balances

aristidesstaffieri: yeah I agree, I'm actually doing this now [in this follow up PR](https://github.com/stellar/freighter/compare/refactor/loading-states-2...refactor/send-swap-data-2). 
I tried to keep the scope of this to just what was needed to fix the loading problem since its already a lot of code but I am actively adding caches for balances/icons/domains in these other flows. The caches will be hit from the data hooks optionally so that should give us the flexibility to speed these flows up while maintaining the loading lifecycle.

extension/src/popup/views/ReviewAuth/hooks/useGetReviewAuthData.tsx
```diff
@@ -0,0 +1,106 @@
...
+      const publicKey = appData.account.publicKey;
+      const allAccounts = appData.account.allAccounts;
+      const networkDetails = appData.settings.networkDetails;
+
+      // handle auto selecting the right account based on `accountToSign`
+      let currentAccount = allAccounts.find(
+        (account) => account.publicKey === publicKey,
+      );
+
+      allAccounts.forEach((account) => {
```
piyalbasu, line 62: Can we split this "account selector" logic in the Sign tx/msg/auth entry views into a helper?

extension/src/popup/views/AccountCreator/index.tsx
```diff
@@ -31,24 +26,21 @@ import "./styles.scss";
...
   const dispatch = useDispatch<AppDispatch>();
   const location = useLocation();
   const params = new URLSearchParams(location.search);
   const isRestartingOnboardingParam = params.get("isRestartingOnboarding");
+  const isOverWritingAccountParam = params.get("isOverWritingAccount");
   const isRestartingOnboarding = isRestartingOnboardingParam === "true";
-
-  const isShowingOverwriteWarning =
-    applicationState === APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED;
+  const isOverWritingAccount = isOverWritingAccountParam === "true";
```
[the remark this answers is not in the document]
piyalbasu: Yeah, but if you manually go to `/account-creator` (no query param), you will not see a warning message. You only see it if the app navigates you there with the appropriate query param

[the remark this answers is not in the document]
aristidesstaffieri: I see yeah this would be better off not being a route but I can add a data hook to this view to fetch settings to fix this. The downside is we'll fetch settings and add a loading state in every case to make the warning show up in this edge case.

piyalbasu:
This is generally looking pretty good. My main thoughts are:

1) The logic around re-routing happening in every component is a little brutal. I definitely see the difficulty you ran into with trying to make it reusable. Let's think about how we can improve this. I think it's more of a developer experience issue now more than anything

2) We should think about adding some more test cases before we go live with this. Looks like all the existing test cases are passing which is a great start, but we're touching basically all the logic in the app. There's definitely some uncovered code paths that may be affected

extension/src/popup/views/AccountCreator/index.tsx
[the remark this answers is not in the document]
piyalbasu: Yeah, I don't think it's a huge problem, but in the spirit of maintaining current app logic, this is probably the way to go. You should just be calling `loadAccount` to get `applicationState`, which doesn't call the BE, so it should be pretty fast

piyalbasu:
> > This is generally looking pretty good. My main thoughts are:
> > 
> > 1. The logic around re-routing happening in every component is a little brutal. I definitely see the difficulty you ran into with trying to make it reusable. Let's think about how we can improve this. I think it's more of a developer experience issue now more than anything
> > 2. We should think about adding some more test cases before we go live with this. Looks like all the existing test cases are passing which is a great start, but we're touching basically all the logic in the app. There's definitely some uncovered code paths that may be affected
> 
> For sure, here are my thoughts on this -
> 
> Thoughts on 1 - Routing has to happen in components so in order to achieve the goal of un-breaking the loading state while maintaining the functionality to route to account creator based on application state, we have to do it in the views with data derived from data hooks.
> 
> One way to get away from this could be to think about handling the abandoned account creation differently. I think instead of re-routing users this could be an error case from data hooks that could be rendered in views. This error case could give users an option to open account creator.
> 
> This is better for a few reasons - Users will better understand the state. Today the page just reroutes them to account creator without any indication of why that happened. We can abstract this better into an Error component that can be rendered in views, instead of having routing logic in every view. For example, we could have -
> 
> ```
> type DataError = "onboarding-abandoned" | ... other data related errors;
> ```
> 
> and views can handle errors similarly to how they already do, but with this added case.
> 
> Thoughts on 2 If you have any test cases in mind I can add them now. This basically touches the entire application so it's hard to say anything other having 100% test coverage would give us full confidence in the changes but I'm not sure we can achieve that as part of this PR.

Yeah, I like the idea for 1. One of the main reasons for re-routing based on application state was so if a user clicked the Freighter icon in the toolbar (which always goes to `/`), it appropriately sent the user to either onboarding or `/account`. So, maybe only need to handle this re-routing in the `/` route.

For additional test cases, I'd say we should have a few tests around the main functionality you've changed in this PR, which is re-routing and loading the current account/settings: 
* We should have a test to make sure you go to onboarding vs the app based on your application state (maybe just on the `/` route, as discussed above).
* We should also have a test test to make sure the unlock-account re-routing still works. 
* The logic for switching accounts is different now - probably a good idea to make sure this correctly loads the new balances (payments e2e tests should cover signing from different accounts already in the send/swap flow)
* Similarly, we should test to make sure when going through the sign tx flow, if you switch accounts, it correctly displays in the UI and signs with the correct account

These are the main ones I'm thinking about as I'm manually testing

aristidesstaffieri:
> > > This is generally looking pretty good. My main thoughts are:
> > > 
> > > 1. The logic around re-routing happening in every component is a little brutal. I definitely see the difficulty you ran into with trying to make it reusable. Let's think about how we can improve this. I think it's more of a developer experience issue now more than anything
> > > 2. We should think about adding some more test cases before we go live with this. Looks like all the existing test cases are passing which is a great start, but we're touching basically all the logic in the app. There's definitely some uncovered code paths that may be affected
> > 
> > 
> > For sure, here are my thoughts on this -
> > Thoughts on 1 - Routing has to happen in components so in order to achieve the goal of un-breaking the loading state while maintaining the functionality to route to account creator based on application state, we have to do it in the views with data derived from data hooks.
> > One way to get away from this could be to think about handling the abandoned account creation differently. I think instead of re-routing users this could be an error case from data hooks that could be rendered in views. This error case could give users an option to open account creator.
> > This is better for a few reasons - Users will better understand the state. Today the page just reroutes them to account creator without any indication of why that happened. We can abstract this better into an Error component that can be rendered in views, instead of having routing logic in every view. For example, we could have -
> > ```
> > type DataError = "onboarding-abandoned" | ... other data related errors;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and views can handle errors similarly to how they already do, but with this added case.
> > Thoughts on 2 If you have any test cases in mind I can add them now. This basically touches the entire application so it's hard to say anything other having 100% test coverage would give us full confidence in the changes but I'm not sure we can achieve that as part of this PR.
> 
> Yeah, I like the idea for 1. One of the main reasons for re-routing based on application state was so if a user clicked the Freighter icon in the toolbar (which always goes to `/`), it appropriately sent the user to either onboarding or `/account`. So, maybe only need to handle this re-routing in the `/` route.
> 
> For additional test cases, I'd say we should have a few tests around the main functionality you've changed in this PR, which is re-routing and loading the current account/settings:
> 
> * We should have a test to make sure you go to onboarding vs the app based on your application state (maybe just on the `/` route, as discussed above).
> * We should also have a test test to make sure the unlock-account re-routing still works.
> * The logic for switching accounts is different now - probably a good idea to make sure this correctly loads the new balances (payments e2e tests should cover signing from different accounts already in the send/swap flow)
> * Similarly, we should test to make sure when going through the sign tx flow, if you switch accounts, it correctly displays in the UI and signs with the correct account
> 
> These are the main ones I'm thinking about as I'm manually testing

Should we just remove the routing logic from everything but account view? Other data hooks will fail in the case of onboarding abandonment but this seems edge case enough to accept to me.

Sounds good I can work on these tests.

piyalbasu:
> Should we just remove the routing logic from everything but account view? Other data hooks will fail in the case of onboarding abandonment but this seems edge case enough to accept to me.
> 


So this would be jumping to the solution you mentioned where we just show a generic error that instructs them to go to `account-creator`? That sounds reasonable to me, but let's talk it over with product before we make a change like that

aristidesstaffieri:
> > Should we just remove the routing logic from everything but account view? Other data hooks will fail in the case of onboarding abandonment but this seems edge case enough to accept to me.
> 
> So this would be jumping to the solution you mentioned where we just show a generic error that instructs them to go to `account-creator`? That sounds reasonable to me, but let's talk it over with product before we make a change like that

I was actually just suggesting that we don't handle the app state at all in views that are not the account view. This would retain the behavior where when you click on the logo it routes you to onboarding if you abandoned it. Other views would just hit errors in their data hooks due to the missing settings data. It seems very rare that a real user would abandon onboarding and then manually navigate to anything other than onboarding and to me this seems ok to let them hit the error as a trade off for dropping that logic from every other view.

extension/src/popup/components/manageAssets/AddAsset/index.tsx
aristidesstaffieri: done in b7b42d484ff662838af76195296b4406773534a6

extension/src/popup/views/AccountCreator/index.tsx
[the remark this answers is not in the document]
aristidesstaffieri: I've added this in 6986f6d395e26e444e8f9f0e5b555cd41074d049

extension/src/popup/views/ReviewAuth/hooks/useGetReviewAuthData.tsx
aristidesstaffieri: helper added in 20a3f4e6a3ab4a13e6177307623591e46a23e678

piyalbasu:
> > > Should we just remove the routing logic from everything but account view? Other data hooks will fail in the case of onboarding abandonment but this seems edge case enough to accept to me.
> > 
> > 
> > So this would be jumping to the solution you mentioned where we just show a generic error that instructs them to go to `account-creator`? That sounds reasonable to me, but let's talk it over with product before we make a change like that
> 
> I was actually just suggesting that we don't handle the app state at all in views that are not the account view. This would retain the behavior where when you click on the logo it routes you to onboarding if you abandoned it. Other views would just hit errors in their data hooks due to the missing settings data. It seems very rare that a real user would abandon onboarding and then manually navigate to anything other than onboarding and to me this seems ok to let them hit the error as a trade off for dropping that logic from every other view.

I agree it's probably an edge case, but that doesn't seem like a good experience: just seeing an error with no explanation. My vote would be to continue to think about your first suggested solution (showing an informative error message in this case) and implement that when we're in agreement

aristidesstaffieri:
@piyalbasu I've added the following tests - 

> We should have a test to make sure you go to onboarding vs the app based on your application state (maybe just on the / route, as discussed above).
c573b263082138a6a86f679a08b0c41e774acf20

> We should also have a test test to make sure the unlock-account re-routing still works.
3795d4422b2e4ca3677034a51e0ab0dcb67dd189


> The logic for switching accounts is different now - probably a good idea to make sure this correctly loads the new balances (payments e2e tests should cover signing from different accounts already in the send/swap flow)
058c42d5aad82cd71f66f7204b87f8086cf1707f
One note I want to point out about this test is that since our test use real accounts on the real network, we can't rely on hardcoded balances to assert that the correct balance is rendered for the active account so my assertion is that the balance after changing accounts is different which should be resilient to changes to the account due to other tests changing the state or real activity on that account.

Similarly, we should test to make sure when going through the sign tx flow, if you switch accounts, it correctly displays in the UI and signs with the correct account
99a07e68d19f6a2ab227f5b4424d4de899206993

piyalbasu (approved):

Issue #1956, opened 2025-04-09
Refactor internal APIs to return strict types


aristidesstaffieri:
### What problem does your feature solve?
Improves the type safety of our internal APIs and their consumption

### What would you like to see?
Refactor all internal APIs to return strict types instead of `any`s

### What alternatives are there?
Accept the loose types for internal APIs

PR #2039, opened 2025-05-08
return the public key from popupMessageListener in all cases

### LINKED ISSUE #1843: freighterApi.requestAccess returns empty public key
From Discord: when FreighterApi.requestAccess() is called and the user hasn't signed into their wallet yet, after they sign in and approve access, requestAccess() returns a empty address. This only happens when the user hasn't signed in yet.

Steps to repo:

* log out of Freighter
* call freighterApi.requestAccess
* Enter password
* Click Connect
* Observe the an empty string is returned for the public key

Acceptance Criteria:
* The user's public key should be returned

piyalbasu:
Closes #1843 

Fixes issue where if a user is logged out when freighter-api requests access, logging in and granting access doesn't return the public key. The fix here follows the model we use in our other "sign" methods where we pass the address from popupMessageListener to freighterApiMessageListener using the responseQueue

### DIFF
--- a/extension/src/background/messageListener/handlers/grantAccess.ts
+++ b/extension/src/background/messageListener/handlers/grantAccess.ts
@@ -41,7 +41,7 @@ export const grantAccess = async ({
   });
 
   if (typeof response === "function") {
-    return response(url);
+    return response(url, publicKey);
   }
 
   return { error: "Access was denied" };
--- a/extension/e2e-tests/freighterApiIntegration.test.ts
+++ b/extension/e2e-tests/freighterApiIntegration.test.ts
@@ -321,3 +321,30 @@ test("should not add token when not allowed", async ({ page, extensionId }) => {
     screenshot: "domain-not-allowed-add-token.png",
   });
 });
+test("should get public key when logged out", async ({ page, extensionId }) => {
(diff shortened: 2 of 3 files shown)


extension/src/background/messageListener/handlers/grantAccess.ts
```diff
@@ -41,7 +41,7 @@ export const grantAccess = async ({
   });
 
   if (typeof response === "function") {
-    return response(url);
+    return response(url, publicKey);
```
piyalbasu, line 44: In my next PR, I'm going to do some work to properly type `response` params. Right now, it takes `response(message?: any, messageAddress?: any) `

aristidesstaffieri: thanks this was lined up to be [part of this work](https://github.com/stellar/freighter/issues/1956) but I haven't made it there yet.

aristidesstaffieri (approved):

extension/src/background/messageListener/handlers/grantAccess.ts
piyalbasu: Oh nice! I'll hold off then

PR #2042, opened 2025-05-09
[FEATURE] Cache slice

### LINKED ISSUE #2029: Optimize calls to get balances in send/swap flows

aristidesstaffieri:
Closes #2029 

**What**
Adds a new store slice - cache slice

**Why**
The cache slice is used to cache data that hooks can optionally take advantage of when it makes sense to rely on a cached value, like in later stages of the send/swap flows.

The key difference between this slice and our previous use of keys like `balances` in the transactionSubmission store is that we only cache data in this slice and not state related to the data fetching lifecycle. This allows views to be completely unaware of a cache hit or miss due to the cache lookup happening in the data hooks. Data hooks can be implemented to either take cache params or always try to use a cache key in order improve the time it takes to resolve their data. Wether we have a cache hit or miss, views still go through the same data fetching lifecycles(`IDLE -> LOADING -> SUCCESS|ERROR`).

**Current Cache Keys**

`balanceData` - This is a map of public key to `AccountBalancesInterface`. We can now use this for user balances or when fetching destination balances in flows like send/swap.
`icons` - A map of asset code to icon url. These are expensive to fetch in flows where we need an icon for every asset that a user has at once, like when you are selecting an asset in the send/swap flows.
`homeDomains` - A map of issuers to home domains. These are expensive to fetch in similar situations as the asset icons, we previously had to do a `server.loadAccount` for every issuer in order to resolve their home domains and this happened for every asset in flows like send/swap when a user is selecting an asset.
`tokenLists` - A list of token list responses. We fetch these in several steps of our key flows and some lists resolve faster than others, with some taking several seconds to resolve.

**When to use this cache**
I currently took advantage of this cache in the flows that were most affected by redundant data fetching, which are the send and swap flows.
I did not use the cache on views where a user is more likely to be negatively impacted by stale data like in the main account view.

**TODO**
Some cache keys could be good candidates for persisting longer than 1 session, like the token lists( and maybe home domains?). The cache slice can be updated to use internal APIs to persist data that we decide to cache longer in local storage.

### DIFF
--- a/extension/src/popup/components/manageAssetsLists/ModifyAssetList/index.tsx
+++ b/extension/src/popup/components/manageAssetsLists/ModifyAssetList/index.tsx
@@ -20,6 +20,7 @@ import { SubviewHeader } from "popup/components/SubviewHeader";
 import { View } from "popup/basics/layout/View";
 import { addAssetsList, modifyAssetsList } from "popup/ducks/settings";
 import { navigateTo } from "popup/helpers/navigate";
+import { saveTokenLists } from "popup/ducks/cache";
 import { DeleteModal } from "../DeleteModal";
 
 import "./styles.scss";
@@ -233,6 +234,7 @@ export const ModifyAssetList = ({
     }
 
     if (modifyAssetsList.fulfilled.match(modifyAssetsListResp)) {
+      dispatch(saveTokenLists([]));
       navigateTo(ROUTES.manageAssetsLists, navigate);
     }
   };
--- a/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
+++ b/extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
@@ -19,18 +19,13 @@ import {
   useScanTx,
 } from "popup/helpers/blockaid";
 import { BlockAidScanTxResult } from "@shared/api/types";
-import { AccountBalancesInterface } from "@shared/api/types/backend-api";
 import { getIconUrlFromIssuer } from "@shared/api/helpers/getIconUrlFromIssuer";
 import { stellarSdkServer } from "@shared/api/helpers/stellarSdkServer";
 import { findAssetBalance } from "popup/helpers/balance";
 import { getAssetFromCanonical, xlmToStroop } from "helpers/stellar";
-import { getAccountBalances, getAssetIcons } from "@shared/api/internal";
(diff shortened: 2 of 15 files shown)


extension/src/popup/components/manageAssetsLists/ModifyAssetList/index.tsx
```diff
@@ -233,6 +234,7 @@ export const ModifyAssetList = ({
     }
 
     if (modifyAssetsList.fulfilled.match(modifyAssetsListResp)) {
+      dispatch(saveTokenLists([]));
```
aristidesstaffieri, line 237: When a new token list is added by a user, we don't necessarily fetch the token list response so we need to clear the cache so that it can be populated with the correct token list responses on the next fetch.

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -646,6 +647,7 @@ export const TransactionDetails = ({
                   variant="secondary"
                   onClick={() => {
                     dispatch(resetSimulation());
+                    dispatch(resetSubmission());
```
aristidesstaffieri, line 650: this is unrelated but I noticed this should be resettings the submission state when the user cancels the flow.

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
```diff
@@ -201,7 +197,13 @@ function useGetTxDetailsData(
   );
 
   const { fetchData: fetchBalances } = useGetBalances(balanceOptions);
-  const { assetsLists } = useSelector(settingsSelector);
+  // tx details needs to show icons and filter hidden assets
```
aristidesstaffieri, line 200: This is a side effect of the work to block API calls for non active public keys. Some of the APIs that the getBalances hook uses go to the internal APIs so when we use them for destination balances they will fail due to that check.

I can step around this in this case but it would be great to rethink how that works. To me, it seems like there is no security benefit to blocking all internal API calls when the caller is not the active public key. All balance data is public so I would say that the only time this becomes relevant is at the time of signing a transaction. 

The downside of doing this is that we can't use any of our internal APIs to get data about any public key that is not the active public key in Freighter which we need to do for destination balances and potentially other features in the future(shadow accounts or view other wallets).

piyalbasu (approved):

PR #2043, opened 2025-05-09
Bugfix/sign auth entry return

### LINKED ISSUE #1960: signAuthEntry does not adhere to SEP43

According to sep43, signAuthEntry should return a string of the signed auth entry in base64 but today it returns a Buffer<UintArray> of the signed entry.

### What did you do?

Signed auth entry and got response

b64 response of the signed auth entry

### What did you see instead?
Buffer<UintArray> of the signed entry

piyalbasu:
Closes #1960 

Similar to our incremental upgrade to signMessage, I'm branching logic based on the API version. This ensures devs using older versions of freighter-api won't have their UX broken by this change.

In addition, I'm adding some better typing for the response queue. It's not the cleanest; I had to do some casting as I ran into some issues because of how different some of the responses are. This is at least a bit more informative than the `any` typing we were using before

### DIFF
--- a/@shared/api/external.ts
+++ b/@shared/api/external.ts
@@ -159,6 +159,7 @@ export const submitMessage = async (
 
 export const submitAuthEntry = async (
   entryXdr: string,
+  version: string,
   opts?: {
     address?: string;
     networkPassphrase?: string;
@@ -175,6 +176,7 @@ export const submitAuthEntry = async (
     response = await sendMessageToContentScript({
       entryXdr,
       accountToSign,
+      apiVersion: version,
       networkPassphrase: opts?.networkPassphrase,
       type: EXTERNAL_SERVICE_TYPES.SUBMIT_AUTH_ENTRY,
     });
--- a/@shared/api/types/message-request.ts
+++ b/@shared/api/types/message-request.ts
@@ -33,8 +33,18 @@ export interface EntryToSign {
   networkPassphrase?: string;
 }
 
-export type ResponseQueue = Array<
-  (message?: any, messageAddress?: any) => void
+export type RequestAccessResponse = string;
+export type SignAuthEntryResponse = Buffer | null;
+export type SignTransactionResponse = string;
+export type SignBlobResponse = Buffer | null;
+export type AddTokenResponse = boolean;
+export type SetAllowedStatusResponse = string;
+export type SignedHwPayloadResponse = string | Buffer<ArrayBufferLike>;
+export type RejectAccessResponse = undefined;
+export type RejectTransactionResponse = undefined;
+
+export type ResponseQueue<T> = Array<
+  (message: T, messageAddress?: string) => void
 >;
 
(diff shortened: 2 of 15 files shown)


extension/src/background/messageListener/freighterApiMessageListener.ts
```diff
@@ -472,9 +494,20 @@ export const freighterApiMessageListener = (
...
+          signedAuthEntry: SignAuthEntryResponse,
+          signerAddress?: string,
+        ) => {
           if (signedAuthEntry) {
-            resolve({ signedAuthEntry });
+            if (apiVersion && semver.gte(apiVersion, "4.2.0")) {
+              resolve({
+                signedAuthEntry:
+                  Buffer.from(signedAuthEntry).toString("base64"),
+                signerAddress,
```
piyalbasu, line 506: I also discovered an issue here where we were not returning the signer address.

@shared/api/external.ts
```diff
@@ -159,6 +159,7 @@ export const submitMessage = async (
 
 export const submitAuthEntry = async (
   entryXdr: string,
+  version: string,
```
aristidesstaffieri, line 162: since the second argument was already `opts`, will older clients clash with this new definition?

piyalbasu: @shared/api/external actually gets bundled into @stellar/freighter-api, so only users of the new api version will be using this function signature. I tested with old versions of freighter-api just to be sure

We actually did something similar when you updated the signMessage payload a few months back: https://github.com/stellar/freighter/pull/1755/files#diff-a07d7b44912cd9de72ec5cbce0973a3ffa42d8f1d9d80096eb4de1efcd044300R97

aristidesstaffieri (approved):

@shared/api/types/message-request.ts
```diff
@@ -33,8 +33,18 @@ export interface EntryToSign {
...
+export type SignAuthEntryResponse = Buffer | null;
+export type SignTransactionResponse = string;
+export type SignBlobResponse = Buffer | null;
+export type AddTokenResponse = boolean;
+export type SetAllowedStatusResponse = string;
+export type SignedHwPayloadResponse = string | Buffer<ArrayBufferLike>;
+export type RejectAccessResponse = undefined;
+export type RejectTransactionResponse = undefined;
+
+export type ResponseQueue<T> = Array<
```
aristidesstaffieri, line 46: Should these be promises that return the response values?

piyalbasu: in freighterApiMessage listener, we wrap the invocations of `response()` in a Promise, but the `response` itself is actually just synchronous function call, even though it contains the `resolve` from the wrapping Promise

collasanta:
@piyalbasu @aristidesstaffieri Hey guys, I noticed that after the 5.33 release, the sign message flow from my app stopped working. After signing it, I get a user reject message as a response. I tried downgrading the extension to 5.32, and it worked again.

PR #2044, opened 2025-05-09
make sure to close request access window when chaining api methods

piyalbasu:
A small bug I noticed while testing #1843 and #1960

In this next version of freighter-api (unreleased), to aid users who are trying to sign a tx, but haven't added the dapp to "connected apps", we will first display the `Grant Access` modal. If the user says yes, we will then show the Sign modal they were intending to sign.

I noticed a bug where there was sometimes a race condition where opting in to Grant Access (and subsequently dismissing that modal) would inadvertently trigger this logic in the following Sign modal before the user could sign the tx: https://github.com/stellar/freighter/blob/master/extension/src/background/messageListener/freighterApiMessageListener.ts#L329

This bugfix makes sure the Grant Access has properly closed before moving onto the show the Sign modal

### DIFF
--- a/@stellar/freighter-api/package.json
+++ b/@stellar/freighter-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stellar/freighter-api",
-  "version": "4.2.0",
+  "version": "4.1.0",
   "license": "Apache-2.0",
   "author": "Stellar Development Foundation <hello@stellar.org>",
   "description": "Utility functions to interact with Freighter extension",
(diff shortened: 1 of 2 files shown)


piyalbasu:
Demo of what the Grant Access -> Sign flow looks like:


https://github.com/user-attachments/assets/79eea932-f5d1-49be-976a-fff9a4c9d83b

aristidesstaffieri (approved):

PR #2049, opened 2025-05-12
Bugfix/dont hide send swap

### LINKED ISSUE #1964: Send/Swap should be available from asset detail in all cases
5.30.0

### What did you do?
Went to an asset's detail, did not see send/swap options because it was flagged as suspicious

send/swap

### What did you see instead?
no options

piyalbasu:
Closes #1964 

Don't disable send/swap buttons for suspicious assets

### DIFF
--- a/extension/src/popup/components/account/AssetDetail/index.tsx
+++ b/extension/src/popup/components/account/AssetDetail/index.tsx
@@ -242,65 +242,65 @@ export const AssetDetail = ({
               />
             </div>
           </div>
-          {isSuspicious ? null : (
-            <div className="AssetDetail__actions">
-              {selectedBalance?.total &&
-              new BigNumber(selectedBalance?.total).toNumber() > 0 ? (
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -242,65 +242,65 @@ export const AssetDetail = ({
...
+                    onClick={() => {
+                      handleOnrampClick();
+                    }}
+                    isLoading={isTokenRequestLoading}
+                  >
+                    {t("BUY")}
+                  </Button>
+                )}
+              </>
+            ) : (
```
aristidesstaffieri, line 291: Will this still show the swap option for soroban tokens where you have a balance of 0? It looks like we don't show swap for any soroban tokens when we do have a balance for the selected asset.

aristidesstaffieri (approved):

PR #2052, opened 2025-05-13
[FEATURE] Adds Contract Addresses to Recent Addresses

### LINKED ISSUE #1774: Recent addresses on "Send To" missing smart wallet (C addresses) destinations

aristidesstaffieri:
Closes #1774 

What
Adds contract addresses to recent addresses when they are part of a successfull send as the destination.
Fixes initial lookup for recent addresses on load of the Send To component.

Why
They are currently missing from recent addresses

### DIFF
--- a/extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
+++ b/extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
@@ -127,7 +127,7 @@ function useSendToData() {
     const { networkDetails } = appData.settings;
     const _isMainnet = isMainnet(networkDetails);
 
-    if (Object.keys(errors).length !== 0) {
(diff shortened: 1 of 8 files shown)


extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
```diff
@@ -127,7 +127,7 @@ function useSendToData() {
     const { networkDetails } = appData.settings;
     const _isMainnet = isMainnet(networkDetails);
 
-    if (Object.keys(errors).length !== 0) {
+    if (Object.keys(errors).length !== 0 && userInput) {
```
aristidesstaffieri, line 130: Since formik validates the empty form, this condition was triggered on load which meant that we never showed the recent addresses on load.

piyalbasu (approved):

PR #2063, opened 2025-05-15
upgrade stellar-sdk and update cpu and memory usage in send/swap flow

piyalbasu:
Related to: https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=110914471

Incorporating some small changes needed for the stellar-sdk update.

* renaming SorobanRpc to rpc
* renaming simulationResponse payload
* manually getting the resource data from the simulationResponse (more details in the file)


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/index.tsx
```diff
@@ -548,10 +548,28 @@ export const TransactionDetails = ({
                   <div>{t("Resource cost")} </div>
                   <div className="TransactionDetails__row__right">
                     <div className="TransactionDetails__row__right__item">
-                      {transactionSimulation.response.cost.cpuInsns} CPU
```
piyalbasu, line 551: stellar-sdk removed the `cost` object [here](https://github.com/stellar/js-stellar-sdk/pull/1084/files)

We should be getting it from `transactionData` by calling `simulationResponse.transactionData.build().resources().instructions`, but since the simulationResponse is serialized/deserialized when it comes back from the API, we lose access to the SimulationResponse prototype methods. This is a quick n dirty way to do this until we implement a better way to transmit this data

aristidesstaffieri: what about building a new `SorobanDataBuilder` instance so that we don't have to dig into the internals? I think generally the internal fields aren't a guaranteed API by the library so it seems risky to use them.

Example - https://github.com/stellar/js-stellar-sdk/blob/master/test/unit/server/soroban/simulate_transaction_test.js#L264

piyalbasu: Yeah, that's a good call about the fields not being guaranteed. It looks like we still need the `.build()` method. Maybe we should hold off on doing this until we update the upstreams systems rather than bandaiding this

aristidesstaffieri: It looks like the SorobanDataBuilder takes an xdr string or the class so it seems like we should be able to just pass in the transactionData, no?

https://github.com/stellar/js-stellar-base/blob/66f824e2bafbe18d16a6921109f75e15c80f174d/src/sorobandata_builder.js#L33

piyalbasu: Yeah, the problem is we don't have an XDR string of the `transactionData`. We get it back as a plain JS object from the endpoint. So, in that code path, it falls down at `sorobanData.toXDR()`

aristidesstaffieri: hmm yeah I see, well maybe its ok to use the internals. I expected there to be a way to instantiate the `SorobanDataBuilder` from the `transactionSimulation.response` but I don't see it either.

aristidesstaffieri: I think this probably just highlights that we should be serializing the `simulationResponse.transactionData` to xdr at the API instead of returning the `SimulateTransactionSuccessResponse` object.

aristidesstaffieri: https://github.com/stellar/freighter-backend/issues/256

aristidesstaffieri (approved):

PR #2070, opened 2025-05-20
use prod BE for all routes except Discover endpoint

piyalbasu:
Because the v2 BE only has the Discover endpoint, we need to rely on the v1 BE for all the endpoint except for Discover. Here we're injecting a second env var for the v2 URL

### DIFF
--- a/.github/workflows/submitBeta.yml
+++ b/.github/workflows/submitBeta.yml
@@ -1,6 +1,7 @@
 name: Beta Deployment
 env:
   INDEXER_URL: ${{ secrets.INDEXER_URL }}
+  INDEXER_V2_URL: ${{ secrets.INDEXER_BETA_URL }}
 on:
   workflow_dispatch:
     inputs:
--- a/.github/workflows/submitBeta.yml
+++ b/.github/workflows/submitBeta.yml
@@ -1,6 +1,6 @@
 name: Beta Deployment
 env:
-  INDEXER_URL: ${{ secrets.INDEXER_BETA_URL }}
+  INDEXER_URL: ${{ secrets.INDEXER_URL }}
 on:
   workflow_dispatch:
     inputs:
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -605,18 +605,23 @@ export const getTokenPrices = async (tokens: string[]) => {
 };
 
 export const getDiscoverData = async () => {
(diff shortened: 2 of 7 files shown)


@shared/api/internal.ts
```diff
@@ -605,18 +605,23 @@ export const getTokenPrices = async (tokens: string[]) => {
 };
 
 export const getDiscoverData = async () => {
-  const url = new URL(`${INDEXER_URL}/protocols`);
+  const url = new URL(
+    "https://freighter-backend-v2-stg.stellar.org/api/v1/protocols",
```
piyalbasu, line 609: hardcoded to staging URL. This will be changed to the prod URL before release

@shared/api/internal.ts
```diff
@@ -605,18 +605,23 @@ export const getTokenPrices = async (tokens: string[]) => {
...
   const parsedResponse = (await response.json()) as {
     data: {
-      description: string;
-      icon_url: string;
-      name: string;
-      website_url: string;
-      tags: string[];
-      is_blacklisted: boolean;
-    }[];
+      protocols: {
```
piyalbasu, line 614: it looks like the API payload changed slightly since this was merged. Updating here

.github/workflows/submitBeta.yml
```diff
@@ -1,6 +1,6 @@
 name: Beta Deployment
 env:
-  INDEXER_URL: ${{ secrets.INDEXER_BETA_URL }}
+  INDEXER_URL: ${{ secrets.INDEXER_URL }}
```
piyalbasu, line 3: Use the usual BE url since we still need it for all routes that are not for Discover

@shared/api/internal.ts
aditya1702: @piyalbasu Why cant we use the ${INDEXER_BETA_URL} to construct only protocols URL here? Why cant we have more than 1 secret?

.github/workflows/submitBeta.yml
aditya1702: Cant we also add `INDEXER_BETA_URL: $${{ secrets.INDEXER_BETA_URL }}` apart from the indexer url?

@shared/api/internal.ts
piyalbasu: Oh you're right - fixing this

aristidesstaffieri (approved):

PR #2073, opened 2025-05-21
Poc/shared lib

piyalbasu:


extension/src/popup/components/accountHistory/HistoryItem/index.tsx
```diff
@@ -178,64 +181,85 @@ export const HistoryItem = ({
...
+        Icon[historyItemData.actionIconString as keyof typeof Icon];
+
+      if (isCreateAccountOperation(type)) {
         // If you're not creating an external account then this means you're
         // receiving some XLM to create(fund) your own account
-        const _isRecipient = !isCreateExternalAccount;
-        const paymentDifference = _isRecipient ? "+" : "-";
-        const formattedAmount = `${paymentDifference}${formatAmount(
-          new BigNumber(startingBalance).toString(),
-        )} ${destAssetCode}`;
```
piyalbasu, line 192: we can get rid of all this formatting logic and just use what comes back from the shared lib

PR #2080, opened 2025-05-22
adding new url's for BE v2

### LINKED ISSUE #2071: Allow Freighter to use 2 BE URL's
Currently, the Freighter codebase relies on one env var for the BE url (INDEXER_URL) which points to Freighter BE v1. This is injected at build time. 

For the short-term, we will need a second env var that points to the Freighter BE v2 URL as some features are only available at v2 (for ex. Discover protocols).

We will need to:

* Make sure that Freighter BE v1 and Freighter BE v2 both have staging URL's that are publicly accessible
* Make sure that Freighter BE v1 and Freighter BE v2 both have prod URL's that are publicly accessible
* Work with ops to make sure all 4 of the endpoints are added to GH as Secrets
* Update prod and beta pipelines to inject both v1 and v2 URL's for the respective environments

piyalbasu:
Closes #2071 

Adds configurations for prod BE v2 URL's in GitHub Actions

### DIFF
--- a/.github/workflows/runTests.yml
+++ b/.github/workflows/runTests.yml
@@ -1,7 +1,7 @@
 name: Run Tests
 env:
   INDEXER_URL: ${{ secrets.INDEXER_URL }}
-  INDEXER_V2_URL: ${{ secrets.INDEXER_BETA_URL }}
+  INDEXER_V2_URL: ${{ secrets.INDEXER_V2_URL }}
 on: [pull_request]
 concurrency:
   group: ${{ github.workflow }}
--- a/.github/workflows/runTests.yml
+++ b/.github/workflows/runTests.yml
@@ -1,7 +1,7 @@
 name: Run Tests
 env:
   INDEXER_URL: ${{ secrets.INDEXER_URL }}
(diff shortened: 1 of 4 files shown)


.github/workflows/runTests.yml
```diff
@@ -1,7 +1,7 @@
 name: Run Tests
 env:
   INDEXER_URL: ${{ secrets.INDEXER_URL }}
-  INDEXER_V2_URL: ${{ secrets.INDEXER_BETA_URL }}
+  INDEXER_V2_URL: ${{ secrets.INDEXER_V2_URL }}
```
piyalbasu, line 4: Open to other thoughts here - but I feel like we'll probably want to use the prod (stable) deployment of BE v2 to run tests against

aristidesstaffieri: one downside to this is that we won't be able to write tests against new features until they're in the prd backend, which is maybe fine but something to note.

piyalbasu: Yeah, that's definitely true. Hm, on second thought, maybe it's better to test against the beta BE to help surface any issues there before we deploy

aristidesstaffieri (approved):

PR #2091, opened 2025-05-28
[BUG] avoids debouncing action when there is no user input

aristidesstaffieri:
What
Avoids debouncing action in our SendTo component when there is no user input.

Why
The action was being debounced in every case, which meant that on load the loading time had to wait for the length of the debounce to load recent addresses, even though there would never be a destination address to use to look up balances.
This makes recent addresses and initial loading time of the SendTo component instant, in the 1-5ms range.

https://github.com/user-attachments/assets/e4f87256-a7e4-4302-9ee2-a12b42b09138

### DIFF
--- a/extension/src/popup/components/sendPayment/SendSettings/Settings/hooks/useGetSettingsData.tsx
+++ b/extension/src/popup/components/sendPayment/SendSettings/Settings/hooks/useGetSettingsData.tsx
@@ -194,6 +194,7 @@ function useGetSettingsData(
         publicKey,
         balanceOptions.isMainnet,
         networkDetails,
+        true,
       );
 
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/sendPayment/SendSettings/Settings/hooks/useGetSettingsData.tsx
```diff
@@ -194,6 +194,7 @@ function useGetSettingsData(
         publicKey,
         balanceOptions.isMainnet,
         networkDetails,
+        true,
```
aristidesstaffieri, line 197: unrelated but missing use of the cache, this will lower the latency on the settings page.

PR #2098, opened 2025-06-03
if creating muxed account extract G address before creating operation

### LINKED ISSUE #2092: `destination is invalid` error when funding a non-existant muxed account
There's been an uptick of `destination is invalid` errors in Sentry: https://stellarorg.sentry.io/issues/6637476387/?project=5599259&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=13

Steps to repro:

* generate a keypair (don't fund it)
* create a muxed address from the public key
* try to send 1 xlm to this muxed acct to fund it
* Observe that Operation.createAccount throws an error on a muxed account

Evidently, Operation.createAccount doesn't accept a muxed account as a destination (though Operation.payment does). We need to decode the muxed account before passing to Operation.createAccount

piyalbasu:
Closes #2092 

We've seen an uptick of `destination is invalid` errors in Sentry. This is a fix for that issue - it makes sure that we extract the base address from the muxed account before we pass it to stellar-sdk's `Operation.createAccount` method as this method only supports G addresses. This is not an issue with `Operation.payment`.

I also confirmed that this is just an issue for muxed account and not for federated addresses

### DIFF
--- a/extension/e2e-tests/sendPayment.test.ts
+++ b/extension/e2e-tests/sendPayment.test.ts
@@ -201,7 +201,7 @@ test("Send doesn't throw error when creating muxed account", async ({
   extensionId,
 }) => {
   test.slow();
-  await login({ page, extensionId });
+  await loginAndFund({ page, extensionId });
   await page.getByTitle("Send Payment").click({ force: true });
 
   await expect(page.getByText("Send To")).toBeVisible();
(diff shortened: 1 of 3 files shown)


danilosilvackl (approved):
lgtm!

PR #2101, opened 2025-06-04
use the appropriate signing method when signing a message

piyalbasu:
Closes https://github.com/Creit-Tech/Stellar-Wallets-Kit/issues/67

Fixes an issue with the signMessage API method not properly signing and returning a message. The UI was just calling the wrong signing method here

### DIFF
--- a/extension/e2e-tests/freighterApiIntegration.test.ts
+++ b/extension/e2e-tests/freighterApiIntegration.test.ts
@@ -192,7 +192,7 @@ test("should not sign auth entry when not allowed", async ({
   });
 });
 
-test.skip("should sign message when allowed", async ({ page, extensionId }) => {
+test("should sign message when allowed", async ({ page, extensionId }) => {
   await loginToTestAccount({ page, extensionId });
   await allowDapp({ page });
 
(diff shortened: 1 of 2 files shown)


extension/e2e-tests/freighterApiIntegration.test.ts
```diff
@@ -192,7 +192,7 @@ test("should not sign auth entry when not allowed", async ({
   });
 });
 
-test.skip("should sign message when allowed", async ({ page, extensionId }) => {
```
piyalbasu, line 195: we had been skipping this test because we were waiting for a newer version of freighter-api to be deployed. We should have turned this on sooner, which would have helped catch this bug

danilosilvackl (approved):

extension/src/popup/views/SignMessage/index.tsx
```diff
@@ -77,7 +77,7 @@ export const SignMessage = () => {
     setIsPasswordRequired,
     verifyPasswordThenSign,
     hardwareWalletType,
-  } = useSetupSigningFlow(rejectTransaction, signTransaction, message.message);
```
danilosilvackl, line 80: just out of curiosity, the `signTransaction` method is still going to be used somewhere else right?

piyalbasu: Yup, it's still used in the Sign Transaction view: https://github.com/stellar/freighter/blob/master/extension/src/popup/views/SignTransaction/index.tsx#L115

danilosilvackl: noice, thank you!

PR #2111, opened 2025-06-09
[BUG] Handle errors to resolve account ID from fed address

aristidesstaffieri:
What
Handles errors to resolve the account ID from a federated address in the data hook for send-to view component.
Improves error messaging in cases where the toml is malformed
Explicitly checks for request state errors in send-to view

Why
In some cases the toml from federated addresses is malformed and not properly handled by our addresses resolver in `useSendToData`.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
+++ b/extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
@@ -2,6 +2,7 @@ import { useReducer } from "react";
 import { Federation } from "stellar-sdk";
 import { FormikErrors } from "formik";
 import debounce from "lodash/debounce";
+import * as Sentry from "@sentry/browser";
 
 import { initialState, isError, reducer } from "helpers/request";
 import { AccountBalances, useGetBalances } from "helpers/hooks/useGetBalances";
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
```diff
@@ -113,41 +119,47 @@ function useSendToData() {
...
-      } as ResolvedSendToData;
-      dispatch({ type: "FETCH_DATA_SUCCESS", payload });
-      return payload;
+        _isMainnet,
+      );
+    } catch (error) {
+      dispatch({ type: "FETCH_DATA_ERROR", payload: error });
+      Sentry.captureException(`Error send-to data: ${error}`);
+      return error;
     }
```
JakeUrban: This is a large try/catch block, is it possible to refactor this so that its clearer what can throw an error and what is safe?

JakeUrban (approved):
One comment but otherwise LGTM

extension/src/popup/components/sendPayment/SendTo/hooks/useSendToData.tsx
aristidesstaffieri: actually on second look it was hasty to add that try/catch, the only thing that can throw is the app data fetch which already handles exceptions and returns an error response that is handled in this block. The only error handling that should be needed to catch this case was added to `getAddressFromInput`.

I removed this try/catch in -
https://github.com/stellar/freighter/pull/2111/commits/e563426792ede15ec9e123b402a1e4a36972232b

PR #2129, opened 2025-06-24
SDS V3 Account and Nav Redesign

### LINKED ISSUE #1916: Account Screen Redesign
Update Account UI to reflect the design outlined in Figma below:

https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2547-36692&m=dev

NOTE: As we don't have a `Positions` or `Collectibles` view yet, this will need to be omitted until we have those features completed.

aristidesstaffieri:
Closes #1916 

Implements the re-design for the accounts page, related elements, and app navigation.

- Removes global navigation
- Adds new account view based navigation options
- Re-builds the "add wallet" workflow to match re-design
- Implements re-design for account view

Screenshot 2025-06-24 at 11 28 22 AM

Screenshot 2025-06-24 at 11 28 54 AM

Screenshot 2025-06-24 at 11 29 07 AM

Screenshot 2025-06-24 at 11 29 32 AM


sdfcharles:
@JakeUrban

1. for the account switcher, its a new view in case they have several wallets. if it was a dropdown, my thinking was that it would feel a bit cramped, esp if you did have 4-5+ wallets.
2. yeah happy to increase the font size on those to the next step up @aristidesstaffieri

aristidesstaffieri:
> 1. for the account switcher, its a new view in case they have several wallets. if it was a dropdown, my thinking was that it would feel a bit cramped, esp if you did have 4-5+ wallets.

@sdfcharles I think the comment is about the the carat icon being used in a case where it's not a dropdown anymore. I also noticed this and kept this in so that we can match what the mobile app looks like but I do think normally a carat indicates a dropdown or like an in page modal/pane vs a navigation to a new view.

PR #2131, opened 2025-06-25
Asset detail redesign

aristidesstaffieri:
Implements the asset detail redesign

- Adds token price and delta to asset detail
- Implements new design for current value and balance in asset detail

Note: The asset detail history still reflects the existing history design and will be upgraded with the work to implement the history redesign.

Screenshot 2025-06-25 at 9 37 32 AM

Screenshot 2025-06-25 at 9 37 48 AM


extension/src/popup/components/__tests__/AssetDetail.test.tsx
```diff
@@ -259,132 +260,4 @@ describe("AssetDetail", () => {
       "100 BAZ",
     );
   });
-  it("should display Send and Swap buttons", async () => {
```
aristidesstaffieri, line 262: These buttons are no longer part of the asset detail view.

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -180,137 +150,84 @@ export const AssetDetail = ({
...
+                  ),
+                )}`
+              : null}
+          </div>
+          {assetPrice && assetPrice.percentagePriceChange24h ? (
             <div
-              className={`AssetDetail__total__copy ${
-                isSuspicious ? "AssetDetail__total__copy--isSuspicious" : ""
-              }`}
+              className={`AssetDetail__delta ${getPriceDeltaColor(
```
danilosilvackl, line 203: since we talked about the designs on the retro, I will leave this comment here, but we don't need to stress it out right now.

mobile implementation diff
- we use the arrows down and up to represent the diff on the delta

danilosilvackl (approved):
lgtm!

extension/src/popup/components/account/AssetDetail/index.tsx
aristidesstaffieri: ah thanks @danilosilvackl 
I assume then that we want to do the same for the main Account page but will double check with design

aristidesstaffieri: @danilosilvackl it sounds like design actually wants to change mobile to follow the extension pattern in this case so leaving this for now Thanks for pointing out the inconsistency!

JakeUrban:
Looks great too. Some comments / questions:

- There is a fair amount of spacing between the heading and the token icon, more than I've seen on any other view within freighter. Not a big deal but something I noticed
- Do we update the price / 24hr change / values in this view like we do on the account view?

aristidesstaffieri:
> Looks great too. Some comments / questions:
> 
> * There is a fair amount of spacing between the heading and the token icon, more than I've seen on any other view within freighter. Not a big deal but something I noticed

hmm let me double check the spacing used in the mock vs what I used here and will report back

> * Do we update the price / 24hr change / values in this view like we do on the account view?

Ah yeah good call it is not updated on the asset detail view after initial load, should it be?

PR #2146, opened 2025-07-16
[BUG] wasm upload render

### LINKED ISSUE #2106: Render wasm uploads better
We're trying to render the contents of the wasm, but we can probably just display the hash.

aristidesstaffieri:
Closes #2106 

Uses the wasm's hash as a display value when signing contract upload transactions

Screenshot 2025-07-16 at 3 08 16 PM


JakeUrban (approved):

JakeUrban:
I don't think we need to support rendering the wasm as text in our UI, that sounds more like a developer tool.

PR #2148, opened 2025-07-17
[CHORE] upgrades stellar-hd-wallet dep to remove pbkdf2

### LINKED ISSUE #2132: Upgrade bip39
https://github.com/stellar/freighter/security/dependabot/263
https://github.com/stellar/freighter/security/dependabot/262

bip39 upgraded to upgrade pbkdf2

N/A

aristidesstaffieri:
Closes #2132 

Addresses the following:
https://github.com/stellar/freighter/security/dependabot/263
https://github.com/stellar/freighter/security/dependabot/262

### DIFF
--- a/extension/e2e-tests/helpers/login.ts
+++ b/extension/e2e-tests/helpers/login.ts
@@ -1,5 +1,7 @@
-import { generateMnemonic } from "stellar-hd-wallet";
+import StellarHDWallet from "stellar-hd-wallet";
 import { expect } from "../test-fixtures";
+
+const { generateMnemonic } = StellarHDWallet;
 
 export const PASSWORD = "My-password123";
 
--- a/extension/e2e-tests/onboarding.test.ts
+++ b/extension/e2e-tests/onboarding.test.ts
@@ -1,7 +1,9 @@
 import { shuffle } from "lodash";
-import { generateMnemonic } from "stellar-hd-wallet";
+import StellarHDWallet from "stellar-hd-wallet";
 import { test, expect, expectPageToHaveScreenshot } from "./test-fixtures";
 import { loginToTestAccount, PASSWORD } from "./helpers/login";
+
+const { generateMnemonic } = StellarHDWallet;
 
 test.beforeEach(async ({ page, extensionId }) => {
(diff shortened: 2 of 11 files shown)


extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
```diff
@@ -157,31 +157,52 @@ const getBuiltTx = async (
...
-    publicKey,
-  );
-  const transaction = new TransactionBuilder(sourceAccount, {
-    fee: xlmToStroop(fee).toFixed(),
-    networkPassphrase: networkDetails.networkPassphrase,
-  })
-    .addOperation(operation)
-    .setTimeout(transactionTimeout);
+  try {
+    const operation = getOperation(
```
JakeUrban, line 161: This change seems independent of the stellar-hd-wallet upgrade -- I'm fine with merging it in the same PR but I'm curious why a try/catch is necessary when constructing a transaction?

JakeUrban (approved):

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
aristidesstaffieri: I just cherry picked[ this commit ](https://github.com/stellar/freighter/commit/468acee3b09ac08481e8acdb0bbbc9598b86fd29)into this branch in order to include it into the patch release, I thought it would be ok to skip the PR ceremony in this case but happy to separate it. The reason to add this is to debug the "destination is invalid" error we've been seeing which can be traced back to creating this op.

I suspect what's happening is either - 
the user is making a payment to an account which is unfunded but they are not sending xlm
the user is making a path payment to an unfunded account.

Our logs don't have enough context to tell today so this should add the right info to identify this issue.

I think either way we have a product choice to make here as to if we prefer allowing the user to create the account inline or let them fail because of the account being unfunded.

PR #2149, opened 2025-07-17
[QA] 5.34.0 beta.2

aristidesstaffieri:
QA Fixes for 5.34.0 beta #2

Adds memo title for memo button when memo is set
displays federated address through send flow
tweaks alignment and padding of asset icons and indicator
Uses current set destination as a default value in the input for send flow's set destination step
Adds invalid state error for "amount too high" invalid states in send & swap

### DIFF
--- a/extension/src/popup/components/InternalTransaction/SubmitFail/styles.scss
+++ b/extension/src/popup/components/InternalTransaction/SubmitFail/styles.scss
@@ -1,81 +1,143 @@
-@use "../../../styles/utils.scss" as *;
-
-.SendingTransaction {
-  display: flex;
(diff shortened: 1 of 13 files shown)


extension/src/popup/components/InternalTransaction/SubmitFail/styles.scss
```diff
@@ -1,81 +1,143 @@
-@use "../../../styles/utils.scss" as *;
```
aristidesstaffieri, line 1: Changes to this file are unrelated to the QA but I noticed the styles were ported incorrectly for this component when I moved it.

PR #2153, opened 2025-07-22
[FEATURE] Grant Access Redesign

### LINKED ISSUE #2150: Grant Access Redesign

aristidesstaffieri:
Closes #2150 
Implements redesign for grant access modal

Screens:
Screenshot 2025-07-22 at 10 18 59 AM

Screenshot 2025-07-22 at 10 09 53 AM

Screenshot 2025-07-22 at 10 10 29 AM

### DIFF
--- a/extension/src/popup/views/GrantAccess/hooks/useGetGrantAccessData.ts
+++ b/extension/src/popup/views/GrantAccess/hooks/useGetGrantAccessData.ts
@@ -1,5 +1,4 @@
 import { useReducer } from "react";
-import Blockaid from "@blockaid/client";
 
 import { initialState, isError, reducer } from "../../../../helpers/request";
 import {


aristidesstaffieri:
One difference between the extension and mobile implementations is that the extension will not have a "dapp name" field. This comes from the wallet connect protocol so right now we use the domain as the title instead in the extension. 

We could consider special casing certain protocols or domains to include a title if we want to be more consistent here.

danilosilvackl (approved):

JakeUrban:
@sdfcharles how do you feel about the vertical spacing / alignment? Should we vertically center this content or keep it at the top?

extension/src/popup/helpers/blockaid.ts
```diff
@@ -17,6 +17,24 @@ import { METRIC_NAMES } from "popup/constants/metricsNames";
...
+    "A transaction draining the user’s native currency to a malicious address.",
+  seed_farming: "A request for the user to enter the seed phrase.",
+  malicious_network_interaction:
+    "A malicious network interaction with a known Command and Control (CNC) server owned by an attacking group.",
+  malicious_sdk: "A known piece of malicious code is embedded within the site.",
+  investment_scam:
+    "A request for the user to deposit money into a scam investment system.",
+  other: "A malicious behavior was detected by the Blockaid network.",
+};
+
```
CassioMG, line 37: this is great

sdfcharles:
> @sdfcharles how do you feel about the vertical spacing / alignment? Should we vertically center this content or keep it at the top?

let's keep it at the top and buttons at the bottom 👍

sdfcharles:
> One difference between the extension and mobile implementations is that the extension will not have a "dapp name" field. This comes from the wallet connect protocol so right now we use the domain as the title instead in the extension.
> 
> We could consider special casing certain protocols or domains to include a title if we want to be more consistent here.

yes we should do this with -- we can start with the ones on our discover list at the very least

cc @JakeUrban

PR #2171, opened 2025-07-31
[BUG] Handle unexpected errors in all sequential migrations

aristidesstaffieri:
Description:
Users can have storage with bad or unexpected values, migrations that rely on these values and make assumptions on their schema can throw and result in subsequent migrations now running for that user. 

This adds error handling around all sequential migrations with default storage setters in the error handler which should result in all migrations running for all users.

### DIFF
--- a/extension/src/background/helpers/dataStorage.ts
+++ b/extension/src/background/helpers/dataStorage.ts
@@ -107,16 +107,9 @@ export const migrateTokenIdList = async () => {
   const storageVersion = (await localStore.getItem(STORAGE_VERSION)) as string;
 
   if (shouldRunMigration({ storageVersion, migrationVersion: "1.0.0" })) {
-    try {
-      if (Array.isArray(tokenIdsByKey)) {
-        const newTokenList = {
-          [NETWORKS.FUTURENET]: tokenIdsByKey,
-        };
-        await localStore.setItem(TOKEN_ID_LIST, newTokenList);
-      }
-    } catch (error) {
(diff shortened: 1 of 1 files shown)


extension/src/background/helpers/dataStorage.ts
```diff
@@ -107,9 +107,16 @@ export const migrateTokenIdList = async () => {
...
+        const newTokenList = {
+          [NETWORKS.FUTURENET]: tokenIdsByKey,
+        };
+        await localStore.setItem(TOKEN_ID_LIST, newTokenList);
+      }
+    } catch (error) {
       const newTokenList = {
-        [NETWORKS.FUTURENET]: tokenIdsByKey,
+        [NETWORKS.FUTURENET]: [],
       };
```
JakeUrban: What error are we expecting to catch here?

JakeUrban: I also thought users with the migration issues had gotten past these versions fine, whats the rationale for updating their logic?

JakeUrban (approved):

extension/src/background/helpers/dataStorage.ts
aristidesstaffieri: yeah true this one doesn't have any logic that throws so I can undo this one. 

These did run but the thinking is that generally all of the sequential migrations should set a default value and continue the migration chain in case a user has a storage value that is unexpected. Future storage bugs could cause past migrations to fail in unlikely cases but since it happened this time I thought it would be best to guard against future cases.

PR #2184, opened 2025-08-06
[BUG] race condition when using useNetworkFees hook

### LINKED ISSUE #2183: useNetworkFees race condition
5.33.x

### What did you do?
Quickly try to add an asset after loading the asset rows from the manage assets flow.

Change trust op submitted in a transaction.

### What did you see instead?
insufficient fee error because the fee was not populated

aristidesstaffieri:
Closes #2183 

Problem:
The useNetworkFees hook performs an async call to get network fees from horizon.
Currently this hook does provide a way to properly wait on the promise to resolve, so all components that use this hook rely on timing to use the resolved fees which causes a race condition for some flows(mainly present when adding a new asset).

Solution:
This adds the default base fee in the hook and exposes the getter explicitly to be called by other hooks and async functions.
ManageAssets now calls the getter and waits for a fee response before building the changeTrust transaction which eliminates the race condition.

### DIFF
--- a/extension/e2e-tests/helpers/sendPayment.ts
+++ b/extension/e2e-tests/helpers/sendPayment.ts
@@ -27,10 +27,9 @@ export const sendXlmPayment = async ({ page }) => {
   await expect(page.getByTestId("SendSettingsTransactionFee")).toHaveText(
     /[0-9]/,
   );
-  // 100 XLM is the default, so likely a sign the fee was not set properly from Horizon
-  await expect(
-    page.getByTestId("SendSettingsTransactionFee"),
-  ).not.toContainText("100 XLM");
+  await expect(page.getByTestId("SendSettingsTransactionFee")).toContainText(
+    "100 XLM",
+  );
   await expectPageToHaveScreenshot(
     {
       page,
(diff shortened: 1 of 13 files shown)


extension/e2e-tests/helpers/sendPayment.ts
```diff
@@ -27,10 +27,9 @@ export const sendXlmPayment = async ({ page }) => {
   await expect(page.getByTestId("SendSettingsTransactionFee")).toHaveText(
     /[0-9]/,
   );
-  // 100 XLM is the default, so likely a sign the fee was not set properly from Horizon
-  await expect(
-    page.getByTestId("SendSettingsTransactionFee"),
-  ).not.toContainText("100 XLM");
+  await expect(page.getByTestId("SendSettingsTransactionFee")).toContainText(
```
aristidesstaffieri, line 30: this was just passing by coincidence before since `no value !== 100`

extension/src/popup/components/sendPayment/SendConfirm/TransactionDetails/hooks/useGetTxDetailsData.tsx
```diff
@@ -184,9 +184,12 @@ const getBuiltTx = async (
 
     return transaction;
   } catch (error) {
```
aristidesstaffieri, line 186: this change is unrelated but since we are going to release this patch it made sense to include a tweak to this log in order to get better details from this error case.

JakeUrban (approved):

PR #2188, opened 2025-08-07
[FEATURE] Redesign notifications, preferences, and recovery phrase

### LINKED ISSUE #1614: Redesign Preferences
Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5440&node-type=frame&m=dev

### LINKED ISSUE #1615: Redesign Reveal Recovery Phrase
Figma: https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-(SDS-v3)?node-id=1127-5353&node-type=frame&m=dev

### LINKED ISSUE #2118: Notification / Alerts Redesign
Redesign eng work for the new style of notifications and alerts.

Examples can be found in the V3 redesign 
https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=2757-1285&p=f&t=PgMPQ9Ck26E27NqF-0

aristidesstaffieri:
Closes #2118 
Closes #1614 
Closes #1615 

Screenshot 2025-08-07 at 11 56 02 AM

Screenshot 2025-08-07 at 12 02 49 PM


JakeUrban:
The "localhost is not currently connected to freighter" is a new error I haven't seen. Is this only shown in local development settings?

aristidesstaffieri:
> The "localhost is not currently connected to freighter" is a new error I haven't seen. Is this only shown in local development settings?

its actually not new, we've had a "domain not connected warning" for all signing modals where it shows that warning and disables submission if the domain hasn't been granted access to Freighter by the user.

JakeUrban:
Got it, so in practice users should never see this warning because being granted access is a requirement now?

aristidesstaffieri:
> Got it, so in practice users should never see this warning because being granted access is a requirement now?

Right yeah as long as dapps implement a good flow to `request access -> interact with other APIs` then users should not see this. Users can disconnect dapps themselves from the Freighter UI and some dapps may not reconnect as smoothly as others so I think it's possible to run into it but I don't commonly see it.

aristidesstaffieri:
> If dapps don't implement the `request access -> interact with other APIs` then I'm pretty sure requests to interact with the other APIs will be rejected. Didn't we recently change to this behavior? Previously we allowed dapps to interact directly with APIs and if they hadn't requested access they would see an indicator about interacting with the dapp for the first time.

Yeah so what happens currently is that you will see the "domain not connected" warning and the confirm button will be disabled.

It's been a while since this change but yeah I think you're right that we used to allow the interaction still and then at some point the connection state became more of a hard blocker for using the other APIs.

Issue #2201, opened 2025-08-19
Support new send navigational flow


aristidesstaffieri:
### What problem does your feature solve?
Send flow should support new flow architecture

### What would you like to see?
Send flow should have a new navigational flow.
Send Amount -> Set Destination OR Set Asset -> Review * Submit

### What alternatives are there?
Use current flow

PR #2206, opened 2025-08-22
[BUG] remeasure swap input mount

aristidesstaffieri:
Problem:
The input widths need to be re-calculated as the input value changes. During the swap flow, if we set an input value and then take a navigation and come back, the correct width is not measured because formik does not update values on mount and the formik lifecycle will not trigger a re-render until a user interacts with the input. If we use an effect to measure the width, we miss the opportunity to measure on mount because the effect happens before the formik has triggered a reflow.

Solution:
We need to use a callback ref in order to trigger the measuring logic when there is a re-render of the specific element reference that we need to measure.

This is not needed in the send flow because the navigational pattern today does not allow for an initial value on mount. When we implement [the new nav flow](https://github.com/stellar/freighter/issues/2201) on send we may need to take the same approach.

https://github.com/user-attachments/assets/049f9599-26aa-407e-ac72-e71a8c90c569

### DIFF
--- a/extension/src/popup/components/swap/SwapAmount/index.tsx
+++ b/extension/src/popup/components/swap/SwapAmount/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useLayoutEffect, useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Navigate } from "react-router-dom";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -125,11 +125,21 @@ export const SwapAmount = ({
   const cryptoInputRef = useRef<HTMLInputElement>(null);
   const usdInputRef = useRef<HTMLInputElement>(null);
 
-  const cryptoSpanRef = useRef<HTMLSpanElement>(null);
   const [inputWidthCrypto, setInputWidthCrypto] = useState(0);
+  const setCryptoSpan = (el: HTMLSpanElement | null) => {
+    if (el) {
+      const width = el.offsetWidth + 4;
+      setInputWidthCrypto(Math.max(DEFAULT_INPUT_WIDTH, width));
```
aristidesstaffieri, line 132: we still want the input field to have some width even when the input has been cleared in order to allow for a re-focus of the input field when there is no value present.

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -125,11 +125,21 @@ export const SwapAmount = ({
...
+    }
+  };
 
-  const fiatSpanRef = useRef<HTMLSpanElement>(null);
   const [inputWidthFiat, setInputWidthFiat] = useState(0);
+  const setFiatSpan = (el: HTMLSpanElement | null) => {
+    if (el) {
+      const width = el.offsetWidth + 2;
```
aristidesstaffieri, line 139: The fiat measurement will use slightly less padding due it having a visual prefix($).

PR #2232, opened 2025-09-04
bump sdk to latest

piyalbasu:
To keep parity with freighter-backend, upgrading stellar-sdk to the latest.

This doesn't appear to have any effect on the app, but is just to keep the versions tidy and in the event that our tests don't cover some edge case that could be resolved by this latest version


extension/src/popup/views/Account/hooks/useGetAccountHistoryData.tsx
```diff
@@ -0,0 +1,83 @@
+import { useReducer } from "react";
```
aristidesstaffieri: is this intended?

PR #2239, opened 2025-09-08
upgrade to yarn 2 and use resolutions to block vulnerable package versions

### LINKED ISSUE #2240: Upgrade Yarn and add resolutions for compromised npm packages
* Upgrade to Yarn 4+
* Use Yarn 4 resolutions to make sure we do not install compromised npm packages

piyalbasu:
Closes #2240 

In light of several low-level packages being compromised, we are forcing yarn to skip the affected versions.

This PR involves 2 parts:
* Adding the overrides (resolutions in yarn) for the affected packages we were using
* Upgrading to Yarn 2 because resolutions in Yarn 1 didn't allow us to *only* replace affected versions - it would do a blanket replacement of all versions of the affected package, which would cause dependency mismatches
* There were a couple of small issues that were a byproduct of the Yarn upgrade (noted in the code)


@shared/api/package.json
```diff
@@ -8,8 +8,8 @@
     "@stellar/js-xdr": "3.1.2",
     "bignumber.js": "9.3.0",
     "prettier": "3.5.3",
-    "stellar-sdk": "yarn:@stellar/stellar-sdk@14.1.1",
-    "stellar-sdk-next": "yarn:@stellar/stellar-sdk@14.1.1",
+    "stellar-sdk": "npm:@stellar/stellar-sdk@14.1.1",
+    "stellar-sdk-next": "npm:@stellar/stellar-sdk@14.1.1",
```
piyalbasu, line 12: this is the new parlance for aliases in yarn 4

extension/package.json
```diff
@@ -27,7 +27,7 @@
     "@testing-library/react": "16.3.0",
     "@testing-library/user-event": "14.6.1",
     "@types/chrome": "0.0.287",
-    "@types/history": "5.0.0",
+    "@types/history": "4.7.11",
```
piyalbasu: this weirdly started throwing an error - not sure what exactly happened, but there was no @types/history on npm. 4.7.11 was the latest version

extension/package.json
```diff
@@ -70,7 +70,6 @@
     "react": "19.0.0",
     "react-copy-to-clipboard": "5.1.0",
     "react-dom": "19.0.0",
-    "react-i18next": "15.4.1",
```
piyalbasu, line 73: I needed to move to this to the root level. Our Jest setup config suddenly wasn't able to find it otherwise

package.json
```diff
@@ -128,5 +129,20 @@
       "webpack>watchpack>watchpack-chokidar2>chokidar>fsevents": true
     }
   },
-  "packageManager": "yarn@1.22.22+sha256.c17d3797fb9a9115bf375e31bfd30058cac6bc9c3b8807a3d8cb2094794b51ca"
+  "resolutions": {
```
piyalbasu, line 132: This is the real meat of the PR: I'm replacing the affected version of the package with either:

a) if the maintainer had released a new "clean" version, jump to that version
b) if they hadn't, use the last "clean" version available

.yarnrc.yml
```diff
@@ -1 +1,5 @@
 enableScripts: false
+
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.9.4.cjs
```
piyalbasu: So with the latest version of Yarn, there's a pattern where some devs commit the actual Yarn binary to the project's repo. This provides a way to build the project without having to install yarn (which is a little more complex these days) and makes sure every dev is on the correct version of yarn. This also guards against a dev accidentally installing a nefarious version of yarn (which, given the type of vuln this PR addresses is maybe something we'd like to adopt)

piyalbasu: This also vastly simplifies our GHA setup as we would need to run a slightly convoluted Yarn install for every single pipeline without this

.github/workflows/deployFreighterApiBeta.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: release/${{ github.event.inputs.release }}
+      - name: Enable Corepack
```
piyalbasu, line 22: The install process for latest versions of Yarn, so we have to run this cmd now rather than running `npm -g yarn`

https://yarnpkg.com/migration/guide#migration-steps

extension/package.json
aristidesstaffieri: `This is a stub types definition. history provides its own type definitions, so you do not need this installed.`

It seems like we don't need this package at all if we're using the `history` package, which now that I look it seems like we don't use that at all so this can just be dropped.

aristidesstaffieri (approved):

extension/package.json
piyalbasu: Ah yes, good call. Removed!

PR #2243, opened 2025-09-10
ensure invoke host function tx shows contract parameters

### LINKED ISSUE #2235: Transaction signing screen has lost contract spec annotations
It looks like in a recent release of Freighter the screen that shows the transaction and the invocation in it has lost the display of information derived from contract specs.

In this screenshot I'm signing a transaction and being shown what contract function will be invoked. However, the parameters have no annotations anymore.

Note that at the top of the screenshot is the auth parameters, and the auth parameters still have contract spec annotations, but the invoke at the bottom does not. The invoke used to have annotations as of a couple weeks ago, but seem to have disappeared recently.

Image

Note that the fact that the auth parameters have annotations is a bug, and that bug is separately captured in:
- https://github.com/stellar/freighter/issues/2196

piyalbasu:
Closes #2235 

Ensures we properly decode the contract id from the xdr and then use it parse the contract parameters

### DIFF
--- a/extension/src/popup/components/__tests__/Operations.test.tsx
+++ b/extension/src/popup/components/__tests__/Operations.test.tsx
@@ -30,7 +30,7 @@ describe("Operations KeyVal", () => {
       });
     });
 
-    it("renders transfer operations", async () => {
+    it("renders transfer operations if contract spec is available", async () => {
       const CONTRACT =
         "CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE";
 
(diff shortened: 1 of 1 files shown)


extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -509,7 +512,7 @@ export const KeyValueInvokeHostFn = ({
               </>
             );
           }
-          const contractId = StrKey.encodeContract(address.contractId() as any);
```
piyalbasu, line 512: as I was fixing the issue, I noticed we are doing this in a few places to decode a contract ID. This appears to work, but only if you cast `.contractId()` as `any`. This is a code smell that could potentially bite us later on. I replaced all occurrences of this with the `addressToString` helper so we can standardize how we do this in the codebase

extension/src/popup/components/signTransaction/Operations/index.tsx
```diff
@@ -804,9 +804,7 @@ export const Operations = ({
 
             case xdr.HostFunctionType.hostFunctionTypeInvokeContract(): {
               const invocation = hostfn.invokeContract();
-              const contractId = StrKey.encodeContract(
-                invocation.contractAddress().toXDR(),
-              );
```
piyalbasu, line 809: this was the actually bug. This actually does not return the invocation's contract address. It seems to returns a random contract ID. Replacing this with the `addressToString` helper

extension/src/popup/helpers/soroban.ts
```diff
@@ -155,7 +155,8 @@ export const addressToString = (address: xdr.ScAddress) => {
   if (address.switch().name === "scAddressTypeAccount") {
     return StrKey.encodeEd25519PublicKey(address.accountId().ed25519());
   }
-  return StrKey.encodeContract(address.contractId() as any);
+
+  return Address.fromScAddress(address).toString();
```
piyalbasu, line 159: this is a type safe way to get contract id from ScAddress

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
aristidesstaffieri: the type for this broke after an sdk upgrade but nothing changed about the function, nice to not cast it though 👍

extension/src/popup/components/signTransaction/Operations/index.tsx
aristidesstaffieri: 🥇

extension/src/popup/components/__tests__/Operations.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      expect(parameterKeys[0]).toHaveTextContent("from");
+      expect(parameterKeys[1]).toHaveTextContent("to");
+      expect(parameterKeys[2]).toHaveTextContent("amount");
+
+      expect(parameterValues).toHaveLength(3);
+      expect(parameterValues[0]).toHaveTextContent(TEST_PUBLIC_KEY);
+      expect(parameterValues[1]).toHaveTextContent(TEST_PUBLIC_KEY);
+      expect(parameterValues[2]).toHaveTextContent("100");
+    });
+  });
```
aristidesstaffieri, line 195: should we also add a test case for when we cannot fetch the contract spec?

piyalbasu: Good call - I'll add!

piyalbasu: Added here: https://github.com/stellar/freighter/pull/2243/commits/08e9852a05d5c07f713202f6af25727c7d2900c8

aristidesstaffieri (approved):

PR #2244, opened 2025-09-11
Bugfix/rm auth param names

### LINKED ISSUE #2196: Incorrect details annotated on contract auths
When signing transactions that contain contract invocations with auths I've noticed that Freighter associates the root level auth with the invocation, annotating the auth with the field names from the contract interface specification.

However, if the annotations of an auth's parameters with the contract spec is accurate, it is only by accident. The contract spec describes the parameters of contract functions, and their invocations, but not the parameters of an auth.

The parameters may be in many cases the same parameter as the contract invocation, but they may not be. There's nothing that requires that they are.

For example, the following transaction invokes with a `from` `Address` and a `id` `u64`, but the authorisation parameters are a `token` `Address` and the `id` `u64`. The auths screen incorrectly describes the first auth parameter as "from" when it is not the from address.

Invocation screen:

Image

Auths screen:

Image

The above screenshots are from this transaction:
https://stellar.expert/explorer/public/tx/eca9efdd6a9d9e35d09cc43c89fb5bc2f30a05563216dfc0dd5f9c00563c18af

piyalbasu:
Closes #2196 

Authorizations should not show contract invocation parameters

Screenshot 2025-09-11 at 11 01 01 AM

### DIFF
--- a/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
+++ b/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
@@ -47,7 +47,7 @@ export const KeyValueList = ({
       className="Operations__pair--value"
       data-testid="OperationKeyVal__value"
     >
-      <span className={"Operations__pair--value-text"}>{operationValue}</span>
+      <span className="Operations__pair--value-text">{operationValue}</span>
     </div>
   </div>
 );
@@ -217,18 +217,7 @@ export const KeyValueLine = ({
     );
   }
   return (
-    <>
-      <KeyValueList operationKey={t("Asset Code")} operationValue={line.code} />
-      <KeyValueList
-        operationKey={t("Asset Issuer")}
-        operationValue={
-          <CopyValue
-            value={line.issuer}
-            displayValue={truncateString(line.issuer)}
-          />
-        }
-      />{" "}
-    </>
+    <KeyValueList operationKey={t("Asset Code")} operationValue={line.code} />
   );
 };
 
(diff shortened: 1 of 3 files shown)


extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -405,12 +407,12 @@ export const KeyValueInvokeHostFnArgs = ({
       }
     }
 
-    if (contractId && fnName) {
+    if (contractId && fnName && !isAuthEntry) {
```
piyalbasu, line 410: don't fetch the spec if we're populating the authorization parameters

extension/src/popup/components/__tests__/AuthEntry.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+  it("renders auth entries for create contract v2", async () => {
+    const assetCode = "KHL1";
+    const assetType = new xdr.AlphaNum4({
+      assetCode: Buffer.from(assetCode),
+      issuer: Keypair.fromPublicKey(TEST_PUBLIC_KEY).xdrAccountId(),
+    });
+
+    const args = new xdr.CreateContractArgsV2({
```
aristidesstaffieri, line 97: wdyt about adding coverage for CreateContractArgsV1 as well as the other executable types?

piyalbasu: Yup - added in https://github.com/stellar/freighter/pull/2244/commits/e3c414a24a2c9fdd89f33bad3881d631e9d8b626

Added the other `SorobanAuthorizedFunction` types: `sorobanAuthorizedFunctionTypeContractFn` and `sorobanAuthorizedFunctionTypeCreateContractHostFn`

PR #2246, opened 2025-09-11
add issuer for changeTrust op

### LINKED ISSUE #2194: Can't identify asset being trusted in change trust operation
When being asked to execute transactions containing a change trust option, none of the views show who the issuer is of the asset. It's not possible as a user to for certain identify the asset being trusted without knowing who the issuer is.

These are screenshots of a transaction conctaining a change trust operation:

Image

Image

piyalbasu:
Closes #2194 

* Make sure to show asset issuer on changeTrust ops
* I also noticed that the `Copied` tooltip wasn't showing when I clicked on the copy icon. Fixing that while I'm here (this fix may not need to be ported to mobile

Screenshot 2025-09-11 at 1 49 30 PM

### DIFF
--- a/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
+++ b/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
@@ -35,11 +35,9 @@ import "./styles.scss";
 export const KeyValueList = ({
   operationKey,
   operationValue,
-  isOverflowAllowed = false,
 }: {
   operationKey: string;
   operationValue: string | number | React.ReactNode;
-  isOverflowAllowed?: boolean; // flag is used to show `Copied` tooltip
 }) => (
   <div className="Operations__pair" data-testid="OperationKeyVal">
     <div className="Operations__pair--key" data-testid="OperationKeyVal__key">
@@ -49,11 +47,7 @@ export const KeyValueList = ({
       className="Operations__pair--value"
       data-testid="OperationKeyVal__value"
     >
-      <span
-        className={`Operations__pair--value-text ${isOverflowAllowed ? "Operations__pair--value-text--overflow-allowed" : ""}`}
-      >
-        {operationValue}
-      </span>
+      <span className={"Operations__pair--value-text"}>{operationValue}</span>
     </div>
   </div>
 );
@@ -233,7 +227,6 @@ export const KeyValueLine = ({
             displayValue={truncateString(line.issuer)}
           />
         }
-        isOverflowAllowed
       />{" "}
     </>
   );
@@ -510,7 +503,6 @@ export const KeyValueInvokeHostFn = ({
                       displayValue={truncateString(salt, 8)}
                     />
                   }
-                  isOverflowAllowed
                 />
                 <KeyValueList
                   operationKey={t("Executable Type")}
@@ -528,7 +520,6 @@ export const KeyValueInvokeHostFn = ({
                         )}
                       />
                     }
-                    isOverflowAllowed
                   />
                 )}
               </>
@@ -553,7 +544,6 @@ export const KeyValueInvokeHostFn = ({
                     displayValue={truncateString(salt, 8)}
                   />
                 }
-                isOverflowAllowed
               />
               <KeyValueList
                 operationKey={t("Executable Type")}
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -35,9 +35,11 @@ import "./styles.scss";
 export const KeyValueList = ({
   operationKey,
   operationValue,
+  isOverflowAllowed = false,
 }: {
   operationKey: string;
   operationValue: string | number | React.ReactNode;
+  isOverflowAllowed?: boolean; // flag is used to show `Copied` tooltip
```
piyalbasu: So the reason the `copied` tooltip wasn't showing up was because the style `Operations__pair--value-text` has `overflow:hidden`. I'm not quite sure what case this solves, so rather than possibly break styling, I'm allowing this component to be configured for both `overflow: hidden` and `overflow:visible`.

Since this is really only used when we pass `CopyValue`, I toyed with the idea of programmatically adding the style based on the component that was being passed. But, ultimately, that felt a little hacky and felt more deterministic to just make this a configurable flag

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -492,6 +510,7 @@ export const KeyValueInvokeHostFn = ({
                       displayValue={truncateString(salt, 8)}
                     />
                   }
+                  isOverflowAllowed
```
piyalbasu: adding this flag every time we pass `CopyValue` as the `operationValue`

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
```diff
@@ -217,7 +223,19 @@ export const KeyValueLine = ({
     );
   }
   return (
-    <KeyValueList operationKey={t("Asset Code")} operationValue={line.code} />
+    <>
+      <KeyValueList operationKey={t("Asset Code")} operationValue={line.code} />
+      <KeyValueList
```
piyalbasu, line 222: adding the issuer here

extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx
aristidesstaffieri: We could try adding e2e tests that prove that these columns won't overflow their containers without this prop if you want to remove it completely.

It looks like it's a fail safe for the column overflowing but maybe in practice all cases of values handle their own formatting/truncation.

piyalbasu: I thought about this a bit more and it looks like `overflow:hidden` is used to guard against an unexpected long string to prevent the UI from doing this:

Screenshot 2025-09-12 at 12 57 48 PM

So, ultimately, I do think we need to keep `overflow:hidden` whenever possible to prevent this. I added a more flexible CSS fix that disables `overflow:hidden` for the brief moment a child is trying to show the Copied tooltip. This should help us make sure we don't break the layout, but also show the Copied tooltip without having to worry about setting a configurable flag

https://github.com/stellar/freighter/pull/2246/commits/7be46f8160a4d4a025a0340bf6785c71a0e89f8a

A more complete fix would be to open a Copied tooltip in a React portal to prevent these weird CSS collisions. But that's a bigger architectural change in SDS that we'll need to discuss with that crew

piyalbasu: In terms of testing, I'd really like to add tests for every kind of op plus strings of random length to have a more holistic view of how our UI's hold up. I've been playing around trying to get AI to generate more tests for us, but thus far, every time cursor creates a test for me, it fails 😞 I'll keep working on this

aristidesstaffieri: ok sounds good, I think also we should maybe sync with @sdfcharles about the differences in copy implementations from mobile <-> extension. Maybe this problem goes away if we decide to move to a notification style copy feedback like in mobile.

extension/src/popup/components/signTransaction/Operations/styles.scss
```diff
@@ -70,6 +70,10 @@
           white-space: nowrap;
           min-width: 0;
           max-width: 100%;
+
+          &:has(.Floater.trigger.trigger--active) {
+            overflow: visible;
+          }
```
piyalbasu, line 76: temporarily disable overflow:hidden while showing the `Copied` tooltip

PR #2252, opened 2025-09-16
cache account balances and poll for updates

### LINKED ISSUE #2248: Poll and cache account balances
In order to prevent fetching /account-balances every single time we load the Account view, we can cache the value so we don't need to make this blocking call on every page load. To get data updates we can do the following:

* in the background of Account, poll /account-balances for updates
* opportunistically fetch /account-balances when we suspect a change has occurred (after completing send/swap, adding an asset)

Open questions:
* How should we animate in deltas to account balance updates
* How often should we poll? Do we need to be concerned with rate limiting or overloading the BE?

piyalbasu:
Closes #2248 

1. We no longer make fetching a user's balance a prerequisite for loading the Account view. We fetch it once on app load, and then rely on the cache for subsequent Account view loads. For example, if you open the app, load the Account view, navigate to Settings, and then navigate back to Account, we will only have loaded account balances from the BE once. While not wholly impressive at the moment (this really only saves me ~750ms on subsequent loads of the Account view), this sets a pattern for making some of these once synchronous calls more asynchronous. In the following PR's, I'll be moving other requests to this pattern and, cumulatively, these will add up to real gainz.

2. To help ensure users don't get stuck with stale information, when a user adds/removes a trustline or sends/swaps, we will opportunistically fetch account balances so when they return to the Account view, the user will see their latest action reflected accurately

3. Also, when a user changes accounts or changes networks, I am forcing a re-fetch of balances. My hypothesis is that users are more willing to accept a slightly longer load time for changing an account and ensuring they have up-to-date data since this is likely they're not doing too frequently

### DIFF
--- a/extension/src/popup/ducks/cache.ts
+++ b/extension/src/popup/ducks/cache.ts
@@ -14,6 +14,11 @@ interface SaveBalancesPayload {
   networkDetails: NetworkDetails;
 }
 
+interface ClearBalancesPayload {
+  publicKey: PublicKey;
+  networkDetails: NetworkDetails;
+}
+
 interface SaveIconsPayload {
   icons: Record<AssetCode, IconUrl>;
 }
@@ -57,6 +62,11 @@ const cacheSlice = createSlice({
         },
       };
     },
+    clearBalancesForAccount(state, action: { payload: ClearBalancesPayload }) {
+      delete state.balanceData[action.payload.networkDetails.network][
+        action.payload.publicKey
+      ];
+    },
     saveIconsForBalances(state, action: { payload: SaveIconsPayload }) {
       state.icons = {
         ...state.icons,
@@ -93,4 +103,5 @@ export const {
   saveIconsForBalances,
   saveDomainForIssuer,
   saveTokenLists,
+  clearBalancesForAccount,
 } = cacheSlice.actions;
--- a/extension/src/popup/views/Account/hooks/useGetAccountData.tsx
+++ b/extension/src/popup/views/Account/hooks/useGetAccountData.tsx
@@ -1,5 +1,4 @@
 import { useEffect, useReducer, useState } from "react";
-import { useSelector } from "react-redux";
 
 import { RequestState } from "constants/request";
 import { initialState, isError, reducer } from "helpers/request";
@@ -20,7 +19,6 @@ import { useDispatch } from "react-redux";
 import { AppDispatch } from "popup/App";
 import { makeAccountActive } from "popup/ducks/accountServices";
 import { changeNetwork } from "popup/ducks/settings";
-import { balancesSelector } from "popup/ducks/cache";
 
 export const getTokenPrices = async ({
   balances,
@@ -69,8 +67,6 @@ function useGetAccountData(options: {
   const { fetchData: fetchBalances } = useGetBalances(options);
   const { fetchData: fetchHistory } = useGetHistory();
 
-  const cachedBalances = useSelector(balancesSelector);
-
   const fetchData = async ({
     useAppDataCache = true,
     updatedAppData,
@@ -108,16 +104,11 @@ function useGetAccountData(options: {
       const allowList = appData.settings.allowList;
       const isMainnetNetwork = isMainnet(networkDetails);
 
-      const hasBalanceCache =
-        cachedBalances &&
-        Object.keys(cachedBalances[networkDetails.network]?.[publicKey] || {})
-          .length > 0;
-
       const balancesResult = await fetchBalances(
         publicKey,
         isMainnetNetwork,
         networkDetails,
-        hasBalanceCache && !shouldForceBalancesRefresh,
+        !shouldForceBalancesRefresh,
       );
 
       const history = await fetchHistory(publicKey, networkDetails);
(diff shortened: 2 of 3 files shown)


piyalbasu:
Polling account balances every 30s for changes that happen outside of the app:

https://github.com/user-attachments/assets/f6962e4c-bbf6-4f29-bdd6-e4451b278a4a

@sdfcharles I'm following the precedent set by token prices and not animating when there's a delta. Let's talk more about this if we'd like to reintroduce the idea of animating changes.

Also, I think it'd be a good idea to add a "refresh" button so user's can short circuit the 30s wait and choose to re-fetch balances early. Wdyt?

piyalbasu:
When taking actions that will affect the user's account balance, we'll re-fetch the balances for the user before returning them to the Account view

https://github.com/user-attachments/assets/e02258a7-2fdc-447e-9fca-9157ef2a8659

extension/e2e-tests/sendPayment.test.ts
```diff
@@ -9,6 +9,9 @@ import {
   stubTokenPrices,
 } from "./helpers/stubs";
 
+const MUXED_ACCOUNT_ADDRESS =
+  "MCCRNOIMODZT7HVQA3NM46YMLOBAXMSMHK3XXIN62CTWIPPP3J2E4AAAAAAAAAAAAEIAM";
```
piyalbasu, line 13: I accidentally funded the muxed test acct, so creating a new one here

extension/src/popup/ducks/cache.ts
```diff
@@ -47,7 +51,10 @@ const cacheSlice = createSlice({
     saveBalancesForAccount(state, action: { payload: SaveBalancesPayload }) {
       state.balanceData = {
         ...state.balanceData,
-        [action.payload.publicKey]: action.payload.balances,
+        [action.payload.networkDetails.network]: {
```
piyalbasu, line 59: balances should be organized by network so we show the correct balance when a user switches network

extension/src/popup/views/Account/index.tsx
```diff
@@ -217,7 +221,12 @@ export const Account = () => {
             <NotFundedMessage
               canUseFriendbot={!!resolvedData!.networkDetails.friendbotUrl}
               publicKey={resolvedData?.publicKey || ""}
-              reloadBalances={fetchData}
+              reloadBalances={() =>
+                fetchData({
+                  useAppDataCache: true,
+                  shouldForceBalancesRefresh: true,
```
piyalbasu, line 227: funding with Friendbot should force a refresh

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -98,11 +107,19 @@ function useGetAccountData(options: {
       const networkDetails = appData.settings.networkDetails;
       const allowList = appData.settings.allowList;
       const isMainnetNetwork = isMainnet(networkDetails);
+
+      const hasBalanceCache =
```
aristidesstaffieri: it seems like there is a similar check to this inside of `fetchBalances` also, do we need both?

aristidesstaffieri:
I'm not sure if this would be considered in scope for this PR but I think if you [checked the icon cache here instead of the token list cache](https://github.com/stellar/freighter/blob/master/extension/src/helpers/hooks/useGetBalances.tsx#L101) you can avoid fetching the tomls on subsequent navs as well.

extension/src/popup/views/Wallets/index.tsx
```diff
@@ -399,6 +406,12 @@ export const Wallets = () => {
                     isSelected={isSelected}
                     onClick={async (publicKey) => {
                       await dispatch(makeAccountActive(publicKey));
+                      await fetchBalances(
```
aristidesstaffieri: Making this call here creates a lag or a frozen feeling because there is no visible loading state and nothing happens until after this promise resolves. 

Is there a way to push this responsibility to the Account view or to represent the loading state?

piyalbasu:
> I'm not sure if this would be considered in scope for this PR but I think if you [checked the icon cache here instead of the token list cache](https://github.com/stellar/freighter/blob/master/extension/src/helpers/hooks/useGetBalances.tsx#L101) you can avoid fetching the tomls on subsequent navs as well.

Yup - that's on my list of things to do. I haven't created tix for all of the issues yet, but I'm tackling basically everything in this doc: https://docs.google.com/document/d/1c-EF_kbXAInmmv5V5H1aYo4kT7ZhB5Fw7P7dlrfU4y8/edit?tab=t.0

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
piyalbasu: Ah good catch - yup, removed here: https://github.com/stellar/freighter/pull/2252/commits/f564c2fd2cde6d1676385afbae064f5a6ba339f5

extension/src/popup/views/Wallets/index.tsx
piyalbasu: Good callout. I tweaked this in the next commit so Account will take on the responsibility of fetching the balances and showing a loader

https://github.com/stellar/freighter/pull/2252/commits/f564c2fd2cde6d1676385afbae064f5a6ba339f5

aristidesstaffieri: nice workaround!

aristidesstaffieri (approved):

PR #2269, opened 2025-09-25
gracefully degrade on errors from Blockaid

### LINKED ISSUE #2262: TypeError: Cannot read properties of undefined (reading 'status') using freighter wallet on localhost
Link to Slack convo: https://stellarfoundation.slack.com/archives/C06MNK4TC22/p1758041001010589

User is running into issue where, when connected to a custom network, trying to connect Freighter to a dapp is showing `TypeError: Cannot read properties of undefined (reading 'status')` on the grant-access view

piyalbasu:
closes #2262 

Fixes 2 bugs where some slightly off typing always expected a Blockaid response object when there were instances where there would not be one: 

1. The UI didn't handle errors from Blockaid well on Mainnet/Testnet/Futurenet. Some slightly off typing allowed us to receive an error but still attempt to show a Blockaid warning. Users received a blocking "Failed to fetch your account data" error

2. On custom networks, the lack of a Blockaid object was causing a TypeError. The code didn't handle a missing Blockaid missing response.

ALSO, it looks like we weren't scanning dapps on custom network, but upon further consideration, I don't think we need to skip dapp scanning on custom networks. The dapp that the user is interacting with could still be malicious and it is possible to run a local network in public mode (which connects to pubnet), so we should likely still try to warn users here

### DIFF
--- a/extension/src/popup/views/__tests__/GrantAccess.test.tsx
+++ b/extension/src/popup/views/__tests__/GrantAccess.test.tsx
@@ -9,7 +9,6 @@ import { Wrapper, mockAccounts } from "../../__testHelpers__";
 import {
   TESTNET_NETWORK_DETAILS,
   DEFAULT_NETWORKS,
-  NetworkDetails,
   MAINNET_NETWORK_DETAILS,
 } from "@shared/constants/stellar";
 import { GrantAccess } from "../GrantAccess";


extension/src/popup/helpers/blockaid.ts
```diff
@@ -40,14 +40,9 @@ export const useScanSite = () => {
   const [error, setError] = useState(null as string | null);
   const [isLoading, setLoading] = useState(true);
 
-  const scanSite = async (url: string, networkDetails: NetworkDetails) => {
+  const scanSite = async (url: string) => {
     setLoading(true);
     try {
-      if (isCustomNetwork(networkDetails)) {
```
piyalbasu, line 46: On the off-chance a user is running quickstart in local mode, let's still try to warn them if needed: https://developers.stellar.org/docs/tools/quickstart

extension/src/popup/views/GrantAccess/hooks/useGetGrantAccessData.ts
```diff
@@ -44,9 +44,13 @@ function useGetGrantAccessData(url: string) {
         return appData;
       }
 
-      const scanData = await scanSite(url, appData.settings.networkDetails);
-      if (!scanData && !isCustomNetwork(appData.settings.networkDetails)) {
```
piyalbasu, line 48: this caused a typeError on a custom network in all instances - we previously weren't scanning for custom networks, thus an empty `scanData` was ALWAYS coming back but we weren't throwing an error here to catch it.

For non-custom networks, a user could get into this state if scanSite threw an error - in which case a user would see a blocking error

piyalbasu:
Screenshot 2025-09-25 at 12 53 52 PM
Screenshot 2025-09-25 at 12 50 05 PM

aristidesstaffieri (commented):
👏 this lgtm!

PR #2272, opened 2025-09-26
handle missing scan-tx result; add disabled state for Confirm Anyway

### LINKED ISSUE #2217: 400 Errors from blockaid should be hanled better
Bad simulation requests result in an error to parse the blockaid result payload. 

### What did you do?
https://stellarorg.sentry.io/issues/6838338718/?alert_rule_id=15034055&alert_type=issue&notification_uuid=f6eb982a-5a23-4ead-a74a-4c897849f0ea&project=5599259&referrer=slack

correct handling of this error case

### What did you see instead?
error trying to destructure the response

piyalbasu:
Closes #2217 

This PR cleans up 2 things in the Sign Transaction view:

1) When the there was no "scan result" from Blockaid (either because the scan-tx endpoint errored or because we're on a custom network), the UI crashed. The UI now handles a missing scan result

2) While testing, I noticed that the "Confirm Anyway" button that appears when a tx is expected to fail had no disabled state. This button is disabled if the dapp is not on Freighter's allowlist. I added some styling for this (graying out the font and adding a `not-allowed` cursor)

Screenshot 2025-09-26 at 12 58 10 PM

### DIFF
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -341,10 +341,12 @@ export const SignTransaction = () => {
                     </span>
                   </div>
                 </div>
-                <BlockaidTxScanLabel
-                  scanResult={scanResult!}
-                  onClick={() => setActivePaneIndex(1)}
-                />
+                {scanResult && (
+                  <BlockaidTxScanLabel
+                    scanResult={scanResult}
+                    onClick={() => setActivePaneIndex(1)}
+                  />
+                )}
                 {!isDomainListedAllowed && (
                   <DomainNotAllowedWarningMessage domain={domain} />
                 )}
--- a/extension/src/popup/views/__tests__/SignTransaction.test.tsx
+++ b/extension/src/popup/views/__tests__/SignTransaction.test.tsx
@@ -624,4 +624,466 @@ describe("SignTransactions", () => {
     );
     await waitFor(() => screen.getByTestId("InsufficientBalanceWarning"));
   });
+  it("renders blockaid scan label when tx expected to fail", async () => {
+    let currentSignTxDataMock = {
+      state: {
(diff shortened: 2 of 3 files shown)


extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -341,10 +341,12 @@ export const SignTransaction = () => {
                     </span>
                   </div>
                 </div>
-                <BlockaidTxScanLabel
-                  scanResult={scanResult!}
-                  onClick={() => setActivePaneIndex(1)}
-                />
+                {scanResult && (
```
piyalbasu, line 344: this value can, in fact, be null

extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -624,4 +624,466 @@ describe("SignTransactions", () => {
...
+              networkPassphrase: "Test SDF Network ; September 2015",
+              networkName: "Test Net",
+            },
+          },
+        }}
+      >
+        <SignTransaction />
+      </Wrapper>,
+    );
+    await waitFor(() => screen.getByTestId("SignTransaction"));
```
aristidesstaffieri, line 740: wdyt about also asserting for the other elements that should be affected by the blockaid scan, like the button being disabled?

aristidesstaffieri (approved):

extension/src/popup/views/__tests__/SignTransaction.test.tsx
piyalbasu: The confirm button should not be disabled due to a blockaid scan (only isMemoRequired and isDomainAllowed affect the ability to sign), but I can add assertions to make sure we're showing the "Confirm Anyway" button instead of the regular "Confirm" button

Issue #2258, opened 2025-09-22
Cache asset lists and load icons async


piyalbasu:
When loading the app, a big blocking bottleneck is having to download all asset lists before showing balances. Let's cache the asset lists and then move to a more async flow for icons

PR #2273, opened 2025-09-26
Feature/move history fetch to bg

### LINKED ISSUE #2254: Move account-history fetch to background task
Currently, when a user loads the Account view, it waits to load account-balances, account-history, and token icons before letting a user see their balance.

Account history is only really needed for a subview (Asset details); there is no need to hold up the UX to wait for this request to resolve.

Acceptance Criteria:
* fetching /account-history should not block UI rendering on Account
* if a user manages to click Asset Details before the /account-history response has resolved, let the user see a loading state while we wait

Question:
* should we cache /account-history so we don't need to make this request again if a user goes to view their history

piyalbasu:
Closes #2254 

This removes retrieving /account-history from the blocking path on Account load. This call can be one of the larger payloads to retrieve in a long line of requests. On my test account, this makes the Account view load about 500ms for me 🎉. As with the /account-balances refactor, these are all cumulative changes to chip away at sluggish loading times. Once i complete #2258, I'll post a video showing a before and after view of streamlining all of these calls

### DIFF
--- a/extension/src/popup/views/Account/index.tsx
+++ b/extension/src/popup/views/Account/index.tsx
@@ -1,9 +1,8 @@
-import React, { useRef, useState, useEffect } from "react";
+import React, { useState, useEffect } from "react";
 import { Navigate, useLocation } from "react-router-dom";
 import { useSelector } from "react-redux";
 import { Notification } from "@stellar/design-system";
 import { useTranslation } from "react-i18next";
-import { isEqual } from "lodash";
 
 import {
   settingsSorobanSupportedSelector,
@@ -27,7 +26,6 @@ import { getTotalUsd } from "popup/helpers/balance";
 import { NetworkDetails } from "@shared/constants/stellar";
 import { reRouteOnboarding } from "popup/helpers/route";
 import { AppDataType } from "helpers/hooks/useGetAppData";
-import { AccountBalances } from "helpers/hooks/useGetBalances";
 
 import { useGetAccountData, RequestState } from "./hooks/useGetAccountData";
 import { useGetAccountHistoryData } from "./hooks/useGetAccountHistoryData";
@@ -68,16 +66,10 @@ export const Account = () => {
       ? accountData.data?.balances
       : null;
 
-  const previousAccountBalancesRef = useRef<AccountBalances | null>(null);
-
   useEffect(() => {
     const getData = async () => {
-      /* Only fetch history data if the account balances have changed or if the previous account balances are null */
-      if (
-        accountBalances &&
-        !isEqual(accountBalances, previousAccountBalancesRef.current)
-      ) {
-        previousAccountBalancesRef.current = accountBalances;
+      if (accountBalances) {
+        // tie refresh history data to account balances requests
         await fetchHistoryData({ balances: accountBalances });
       }
     };


extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -296,30 +356,24 @@ export const AssetDetail = ({
...
-                    networkDetails={networkDetails}
-                    setActiveHistoryDetailId={() =>
-                      setActiveAssetId(operation.id)
-                    }
-                  />
-                ))}
-              </>
-            </div>
-          ) : (
+          {assetOperations === null ? (
```
piyalbasu: if we are in the rare case where a user tries to view their Asset Details, but history has not finished loading yet, this allows us to just show a <Loader /> until that request finishes

extension/src/popup/ducks/cache.ts
```diff
@@ -27,20 +35,30 @@ type SaveDomainPayload = Record<PublicKey, HomeDomain>;
...
   };
   icons: Record<AssetCode, IconUrl>;
   homeDomains: Record<PublicKey, HomeDomain>;
   tokenLists: AssetListResponse[];
+  tokenDetails: {
+    [contractId: string]: TokenDetailsResponse;
+  };
+  historyData: {
+    [network: string]: Record<PublicKey, HistoryResponse>;
+  };
```
piyalbasu, line 55: let's cache tokenDetails and historyData. tokenDetails are unlikely to change, but we can refresh historyData as needed

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -112,29 +108,17 @@ function useGetAccountData(options: {
         !shouldForceBalancesRefresh,
       );
 
-      const history = await fetchHistory(publicKey, networkDetails);
```
piyalbasu, line 117: removing history fetch from the blocking path when loading Account

extension/src/popup/views/Account/index.tsx
```diff
@@ -56,6 +62,29 @@ export const Account = () => {
...
+
+  useEffect(() => {
+    const getData = async () => {
+      /* Only fetch history data if the account balances have changed or if the previous account balances are null */
+      if (
+        accountBalances &&
+        !isEqual(accountBalances, previousAccountBalancesRef.current)
+      ) {
+        previousAccountBalancesRef.current = accountBalances;
+        await fetchHistoryData({ balances: accountBalances });
```
piyalbasu, line 73: we can tie historyData refresh to updates to account balances. Since a user can only get to history screens from Account, they will always have to go through this update. This helps ensure users always have updated history

extension/src/popup/components/account/AssetDetail/index.tsx
piyalbasu: https://github.com/user-attachments/assets/f4d522ab-f8e1-43ec-a01a-0746122041eb

Simulation of what a slow loading history would look like

extension/src/helpers/hooks/useTokenDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }): Promise<TokenDetailsResponse | Error> => {
+    dispatch({ type: "FETCH_DATA_START" });
+    try {
+      /*
+        Unlike the other cache hooks, this hook can be called multiple times within one render.
+        For example, when constructing the history rows, this hook can be called many times as we iterate over the history items. 
+        If we have cached token details earlier in the loop, we won't have access to the update redux state until the next render.
+        To workaround this, we will also check the redux state manually here rather than waiting for the next render to 
+        update the selector hook for us.
+      */
```
piyalbasu, line 43: Example: In my account history, I have 10 transfers using the same custom token. The getHistoryData hook will run once on first render and then call this method 10 times as we iterate over the history items. However, because we're still in the first render, the useSelector hook won't have access to the updated redux state yet. This will result in `cachedTokenDetails` being empty for all 10 iterations, resulting in 10 separate API calls to get the token details.

extension/src/helpers/hooks/useTokenDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+    networkDetails: NetworkDetails;
+  }): Promise<TokenDetailsResponse | Error> => {
+    dispatch({ type: "FETCH_DATA_START" });
+    try {
+      /*
+        Unlike the other cache hooks, this hook can be called multiple times within one render.
+        For example, when constructing the history rows, this hook can be called many times as we iterate over the history items. 
+        If we have cached token details earlier in the loop, we won't have access to the update redux state until the next render.
+        To workaround this, we will also check the redux state manually here rather than waiting for the next render to 
+        update the selector hook for us.
```
piyalbasu, line 42: Example: In my account history, I have 10 transfers using the same custom token. The getHistoryData hook will run once on first render and then call this method 10 times as we iterate over the history items. However, because we're still in the first render, the useSelector hook won't have access to the updated redux state yet. This will result in `cachedTokenDetails` being empty for all 10 iterations, resulting in 10 separate API calls to get the token details.

extension/e2e-tests/loadAccount.test.ts
```diff
@@ -454,6 +454,6 @@ test("Clears cache and fetches balances if it's been 2 minutes since the last ba
 
   // make sure we fetch the new balance quickly rather than waiting for the 30 second interval
   await expect(page.getByTestId("asset-amount")).toHaveText("999,111", {
-    timeout: 500,
+    timeout: 3000,
```
piyalbasu, line 457: this test is a bit flakey - I believe just waiting a little longer for the UI to re-render should solve this

extension/src/helpers/hooks/useGetHistory.tsx
```diff
@@ -1,26 +1,40 @@
...
   ): Promise<HistoryResponse | Error> => {
     dispatch({ type: "FETCH_DATA_START" });
     try {
-      const data = await getAccountHistory(publicKey, networkDetails);
+      const cachedHistoryData =
+        cachedHistory[networkDetails.network]?.[publicKey];
+      const data =
+        useCache && cachedHistoryData
+          ? cachedHistoryData
+          : await getAccountHistory(publicKey, networkDetails);
```
piyalbasu, line 33: following a similar pattern to account balance caching

extension/src/helpers/hooks/useTokenDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+      */
+      const cachedTokenDetailsData =
+        cachedTokenDetails[contractId] ||
+        store.getState().cache.tokenDetails[contractId];
+
+      const data =
+        useCache && cachedTokenDetailsData
+          ? cachedTokenDetailsData
+          : await getTokenDetails({ contractId, publicKey, networkDetails });
+      if (data) {
```
aristidesstaffieri: maybe this if statement only exists to appease the type checker but it seems like if there is no data then we will stay in the `FETCH_DATA_START` indefinitely. Should this throw or in some other way this possibility?

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -40,12 +40,81 @@ import {
...
+  networkDetails: NetworkDetails;
+  activeAssetId: string;
+  setActiveAssetId: (id: string) => void;
+  setActiveOperation: (operation: OperationDataRow) => void;
+  isSorobanAsset: boolean;
+}) => {
+  const { t } = useTranslation();
+  const { isHideDustEnabled } = useSelector(settingsSelector);
+
+  const sortedAssetOperations = assetOperations.filter((operation) => {
```
aristidesstaffieri: super nit but this filters instead of sorts so maybe it can be called `filteredAssetOperations`?

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -40,12 +40,81 @@ import {
...
+    if (operation.metadata.isDustPayment && isHideDustEnabled) {
+      return false;
+    }
+
+    return true;
+  });
+
+  useEffect(() => {
```
aristidesstaffieri: it looks like this state can be derived from props entirely. Can this be a memoized value instead(will cause less renders)

```
const activeOperation = useMemo(() => {
  return sortedAssetOperations.find((op) => op.id === activeAssetId);
}, [activeAssetId, sortedAssetOperations]);
```

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -40,12 +40,81 @@ import {
...
+    const activeOperation = sortedAssetOperations.find(
+      (op) => op.id === activeAssetId,
+    );
+    setActiveOperation(activeOperation as OperationDataRow);
+  }, [activeAssetId, setActiveOperation, sortedAssetOperations]);
+
+  if (!assetOperations && !isSorobanAsset) {
```
aristidesstaffieri: we already check for `sortedAssetOperations.length` in the body so it seems redundant to also check for `!assetOperations` given that its not a nullable prop.

extension/src/popup/ducks/cache.ts
```diff
@@ -37,13 +47,21 @@ interface InitialState {
   icons: Record<AssetCode, IconUrl>;
   homeDomains: Record<PublicKey, HomeDomain>;
   tokenLists: AssetListResponse[];
+  tokenDetails: {
```
aristidesstaffieri, line 50: should token details also be keyed by network in the cache?

extension/src/popup/views/Account/index.tsx
aristidesstaffieri: users can still access history without hitting the account page first through the URL(in full screen mode) or by hitting refresh on the extension popup. I think this is ok though, it seems like the only downside is no polling updates for history if you don't hit accounts first.

extension/src/helpers/hooks/useTokenDetails.tsx
piyalbasu: ah yup, good call

extension/src/popup/components/account/AssetDetail/index.tsx
piyalbasu: Yeah, good catch - most of this code already existed in this file, I just moved it around. But happy to make that change

extension/src/popup/ducks/cache.ts
piyalbasu: My understanding is that the C... address takes network into account when generating the address, so a C address can only live on one network. If you know differently, let me know!

aristidesstaffieri: ah I see thanks, yeah it does look like the network is part of the address hash so this should be ok.

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -129,27 +171,33 @@ export const AssetDetail = ({
     assetIssuer,
   });
 
-  if (!assetOperations && !isSorobanAsset) {
```
piyalbasu, line 132: I'm removing this `!isSorobanAsset` check. This was added a long time ago here: https://github.com/stellar/freighter/pull/738

This was supposed to be blocking showing history for custom tokens, but wasn't actually working. We can see history for all classic and custom tokens

extension/src/popup/components/account/AssetDetail/index.tsx
piyalbasu: Thanks for the comments here. I actually went back and re-thought what I was doing and came up with a more streamlined approach for this. Check it out here: https://github.com/stellar/freighter/pull/2273/commits/f0bb85f49bdcf425de1cc4bb51cd0b36a3e2d545

extension/src/popup/views/Account/index.tsx
aristidesstaffieri: Im also wondering if it makes sense to fetch history on a cadence regardless of balance changes. Let's say that an account is getting their transactions sponsored, their balance may not change as their history grows and they would have a stale history as long as their popup session is going.

PR #2275, opened 2025-09-29
add cache for balances to ensure we do a fresh lookup when needed

piyalbasu:
Related to #2248 

While doing some testing, I noticed a way to make the UX of polling balances a little better. Currently, we load account balances once on app open, and then poll for balance updates every 30 seconds. But, once a user opens the app, if they are clicking around in Freighter and never settle on the Account view for the requisite 30 seconds, they will never poll for a balance update. They could potentially have stale data for an extended period.

To rectify this, I've add a `updatedAt` key to the cached account balances. If a user hasn't refreshed their account balances for 2 min, we can force them to do a fresh fetch via a non-blocking call. 

https://github.com/user-attachments/assets/40c10b87-8f75-4f87-9c5a-2956ef97174e

### DIFF
--- a/extension/src/popup/ducks/cache.ts
+++ b/extension/src/popup/ducks/cache.ts
@@ -12,7 +12,6 @@ interface SaveBalancesPayload {
   publicKey: PublicKey;
   balances: AccountBalancesInterface;
   networkDetails: NetworkDetails;
-  usedCache: boolean;
 }
 
 interface ClearBalancesPayload {
@@ -58,18 +57,16 @@ const cacheSlice = createSlice({
       state.tokenLists = [];
     },
     saveBalancesForAccount(state, action: { payload: SaveBalancesPayload }) {
-      if (!action.payload.usedCache) {
-        state.balanceData = {
-          ...state.balanceData,
-          [action.payload.networkDetails.network]: {
-            ...state.balanceData[action.payload.networkDetails.network],
-            [action.payload.publicKey]: {
-              ...action.payload.balances,
-              updatedAt: Date.now(),
-            },
+      state.balanceData = {
+        ...state.balanceData,
+        [action.payload.networkDetails.network]: {
+          ...state.balanceData[action.payload.networkDetails.network],
+          [action.payload.publicKey]: {
+            ...action.payload.balances,
+            updatedAt: Date.now(),
           },
-        };
-      }
+        },
+      };
     },
     clearBalancesForAccount(state, action: { payload: ClearBalancesPayload }) {
       delete state.balanceData[action.payload.networkDetails.network][
--- a/extension/src/helpers/hooks/useGetBalances.tsx
+++ b/extension/src/helpers/hooks/useGetBalances.tsx
@@ -118,14 +118,17 @@ function useGetBalances(options: {
         reduxDispatch(saveIconsForBalances({ icons }));
       }
 
-      reduxDispatch(
-        saveBalancesForAccount({
-          publicKey,
-          balances: accountBalances,
-          networkDetails,
-          usedCache: useCache && !!cachedBalanceData,
-        }),
-      );
+      if (!(useCache && !!cachedBalanceData)) {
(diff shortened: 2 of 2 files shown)


extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -216,24 +218,70 @@ function useGetAccountData(options: {
...
     const interval = setInterval(async () => {
-      const publicKey = resolvedData.publicKey;
-      const networkDetails = resolvedData.networkDetails;
-      const balancesResult = await fetchBalances(
-        publicKey,
-        _isMainnet,
-        networkDetails,
-        false,
-      );
+      try {
```
piyalbasu, line 225: another small adjustment: let's fail quietly if we throw an error on update. A user's session should not crash after we've already loaded some data - let's let them continue with potentially stale data

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -216,24 +222,70 @@ function useGetAccountData(options: {
...
+        captureException(
+          `Error refreshing cache balances on Account - ${error}`,
+        );
+      }
+    };
+
+    if (
+      cachedBalances[networkDetails.network]?.[publicKey]?.updatedAt &&
+      cachedBalances[networkDetails.network]?.[publicKey]?.updatedAt <
+        Date.now() - 120000
```
piyalbasu, line 281: if it's been 2 minutes or more since a user has refreshed, we might have some stale data. Let's refresh once immediately without having to wait 30 seconds

extension/e2e-tests/loadAccount.test.ts
```diff
@@ -372,3 +372,94 @@ test("Can't change settings on a stale window", async ({
...
+      },
+      isFunded: true,
+      subentryCount: 0,
+      error: {
+        horizon: null,
+        soroban: null,
+      },
+    };
+
+    console.log("fulfilling");
```
aristidesstaffieri: leftover debug log?

extension/src/popup/ducks/cache.ts
```diff
@@ -54,13 +58,18 @@ const cacheSlice = createSlice({
...
     },
     saveBalancesForAccount(state, action: { payload: SaveBalancesPayload }) {
-      state.balanceData = {
-        ...state.balanceData,
-        [action.payload.networkDetails.network]: {
-          ...state.balanceData[action.payload.networkDetails.network],
-          [action.payload.publicKey]: action.payload.balances,
-        },
-      };
+      if (!action.payload.usedCache) {
```
aristidesstaffieri: why would the cache need to check this flag? It seems like there's not point in calling this action without this flag.

piyalbasu: It's because `saveBalancesForAccount` gets called every time we call [fetchBalances](https://github.com/stellar/freighter/pull/2275/files#diff-ddf4e0b14fa1f48f0a6835e9e5a6430812dc6c4deda9595ad5a4ec08fcb42588R126), whether we've fetched fresh data from the API or we've just pulled the balance data from the cache. We should _only_ update `updateAt` if the data has been retrieved from the API to get a better view of if the user's data is stale or not

piyalbasu: Maybe it's clearer if we only call `saveBalancesForAccount` when necessary rather than `saveBalancesForAccount` handling the cache vs not-cached logic here?

aristidesstaffieri: yeah I think so, updating the cache only when it makes sense is the responsibility of the caller imo.

piyalbasu: Updated here: https://github.com/stellar/freighter/pull/2275/commits/b4e62917a9fcc44eb5723d4e7705532c3ea801a9

aristidesstaffieri (approved):

PR #2286, opened 2025-10-01
Bugfix/update custom token balance

piyalbasu:
Related to #2248 

After adding a custom token, it should refresh the balance for the user so the UI updates properly

### DIFF
--- a/extension/e2e-tests/addAsset.test.ts
+++ b/extension/e2e-tests/addAsset.test.ts
@@ -9,8 +9,12 @@ import {
 } from "./helpers/stubs";
 import { truncateString } from "../src/helpers/stellar";
 
-test("Adding unverified Soroban token", async ({ page, extensionId }) => {
+test("Adding and removing unverified Soroban token", async ({
+  page,
+  extensionId,
+}) => {
   test.slow();
+  await stubTokenDetails(page);
   await loginToTestAccount({ page, extensionId });
 
   await page.getByTestId("account-options-dropdown").click();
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/ToggleTokenInternal/index.tsx
```diff
@@ -57,7 +62,23 @@ export const ToggleTokenInternal = ({
...
+    if (isError<AccountBalances>(balancesResult)) {
+      // we don't want to throw an error if balances fail to fetch as this doesn't affect the tx submission
+      // let's simply log the error and continue - the user will need to refresh the Account page or wait for polling to refresh the balances
+      captureException(
+        `Failed to fetch balances after ${!asset.isTrustlineActive ? "add" : "remove"} token - ${JSON.stringify(
+          balancesResult.message,
+        )} ${networkDetails.network}`,
+      );
+    }
+    onCancel();
```
piyalbasu, line 81: one small change here - I noticed when adding a trustline, we just dismiss the slideup modal, but on custom tokens, we were nav'ing back to Account. I'm matching the trustline UX and just dismissing the modal

extension/e2e-tests/addAsset.test.ts
```diff
@@ -48,6 +48,39 @@ test("Adding unverified Soroban token", async ({ page, extensionId }) => {
...
+    "E2E Token",
+  );
+  await expect(page.getByTestId("ToggleToken__asset-add-remove")).toHaveText(
+    "Add Token",
+  );
+  await page.getByRole("button", { name: "Confirm" }).click();
+  await expect(
+    page.getByTestId("ManageAssetRowButton__ellipsis-E2E"),
+  ).toBeVisible();
+
```
aristidesstaffieri, line 66: should we assert that the new balance is visible on the account page before removing it?

Also, if this test cleans up state by removing the token and it's not actually part of the test, then maybe that code should go into an `afterAll` block.

piyalbasu: Yeah, I can add a test to make sure it's on the balances page

For your second point, we're actually testing that remove works (and that it properly removes the token from the UI). Custom tokens don't persist, so there isn't really a need to clean up state. So, I think it makes sense to leave this part of the test here

aristidesstaffieri: oh gotcha, then maybe we can just rename the test to reflect that.

piyalbasu: Addressed here: https://github.com/stellar/freighter/pull/2286/commits/c8c632a397484e9a03fbf9c07a65aeab4b02b885

PR #2291, opened 2025-10-02
Bugfix/fix avail balance

### LINKED ISSUE #2290: Fix available balance calculation

5.35.0

### What did you do?

* Create an account with 2.9 XLM
* Add a trustline
* Try to swap to that trustline
* Observe that Freighter says you have 0 XLM available

I should have had 1.4 XLM avail (minus recommended fee)

### What did you see instead?
0 XLM available

piyalbasu:
Closes #2290 

Fixes 2 issues with calculating available balance:

1) we were incorrectly defaulting to 0 when we were trying to account for negative numbers (caused by a high fee). We are now using the same logic as freighter-mobile

2a) we were re-calculating minimum balance even though Freighter BE returns that value for us. Removing that extra logic

2b) for custom networks that just retrieve balances from Horizon or Quickstart, we have a [transformer](https://github.com/stellar/freighter/blob/master/%40shared/helpers/stellar.ts#L124) that does this already

### DIFF
--- a/@shared/api/types/account-balance.ts
+++ b/@shared/api/types/account-balance.ts
@@ -11,6 +11,7 @@ export interface NativeAsset {
   total: BigNumber;
   buyingLiabilities: string;
   sellingLiabilities: string;
+  minimumBalance: string;
   blockaidData: Blockaid.TokenScanResponse;
 }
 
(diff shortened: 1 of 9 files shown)


aristidesstaffieri (approved):
nice good catch 👏

PR #2300, opened 2025-10-08
check for updated appdata before showing password modal

### LINKED ISSUE #2245: [BUG] Entering incorrect password during expired session when confirming transaction does not throw "incorrect password" error
Reproduction:

- Log onto Freighter extension without needing to enter your password
- initiate a Send
- Enter wrong password
- Pressing "enter" doesn't also trigger "Confirm Send"
- If you enter the wrong password, you get a "transaction failed" error, but it doesn't tell you why the transaction failed.

Intended Behavior:
- Entering an incorrect password should throw an error
- Entering an incorrect password should NOT allow you to proceed, and the "Confirm Send" button should not be clickable.

If your session is still active, it wouldn't ask you for your password here. So it must've been that it was active when you opened but quickly expired

piyalbasu:
Closes #2245 

This fixes 2 related bugs in the send/swap flow:

1. If a user entered the wrong password at the "verify password" stage, the UI didn't confirm the password and it would attempt to sign and submit with an incorrect password. This would of course fail, but just would give a generic "tx failed" msg. Now, the UI properly surfaces an incorrect password message and allows a user to retry

2. If a user's session expired after Freighter had already opened and loaded account details, once the user got to the send tx step, the app would try to sign/submit. This would also fail because the Background script didn't have the password. Now, the UI checks loadAccount once again before showing/hiding the verify modal to make sure everything is in sync

https://github.com/user-attachments/assets/2cddd991-0d70-49d4-af8a-772515ae3d35

### DIFF
--- a/extension/src/popup/components/InternalTransaction/SubmitTransaction/hooks/useGetSubmitAccountData.tsx
+++ b/extension/src/popup/components/InternalTransaction/SubmitTransaction/hooks/useGetSubmitAccountData.tsx
@@ -8,9 +8,6 @@ import { AppDataType, NeedsReRoute } from "helpers/hooks/useGetAppData";
 import { loadAccount } from "@shared/api/internal";
 import { AppDispatch } from "popup/App";
 import { saveAccount } from "popup/ducks/accountServices";
-import { APPLICATION_STATE } from "@shared/constants/applicationState";
-import { ROUTES } from "popup/constants/routes";
-import { POPUP_WIDTH } from "constants/dimensions";
 
 interface ResolvedSubmitAccountData {
   type: AppDataType.RESOLVED;
@@ -32,22 +29,6 @@ function useGetSubmitAccountData() {
     try {
       const account = await loadAccount();
       reduxDispatch(saveAccount(account));
-
-      if (
-        !account.publicKey ||
-        account.applicationState === APPLICATION_STATE.APPLICATION_STARTED
-      ) {
-        const hasOnboarded =
-          account.applicationState ===
-          APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED;
-        const payload = {
-          type: "re-route",
-          routeTarget: hasOnboarded ? ROUTES.unlockAccount : ROUTES.welcome,
-          shouldOpenTab: window.innerWidth === POPUP_WIDTH && !hasOnboarded,
-        } as NeedsReRoute;
-        dispatch({ type: "FETCH_DATA_SUCCESS", payload });
-        return payload;
-      }
 
       const payload = {
         type: AppDataType.RESOLVED,


extension/src/popup/components/InternalTransaction/SubmitTransaction/index.tsx
```diff
@@ -75,22 +79,48 @@ export const SendingTransaction = ({
...
     const getData = async () => {
       await fetchData({
         isSwap,
       });
     };
-    if (!isVerifyAccountModalOpen) {
+    if (
+      submitAccountState.state === RequestState.SUCCESS &&
+      submitAccountState.data.type === AppDataType.RESOLVED &&
+      hasPrivateKey
```
piyalbasu, line 94: let's make this more deterministic; instead of basing this trigger off of the modal being opened or closed, let's base it off of having re-loaded account info and checking if we have the private key needed

extension/src/popup/components/InternalTransaction/SubmitTransaction/index.tsx
```diff
@@ -75,22 +79,48 @@ export const SendingTransaction = ({
...
+    };
+    getData();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    if (
+      submitAccountState.state === RequestState.SUCCESS &&
+      submitAccountState.data.type === AppDataType.RESOLVED &&
+      !submitAccountState.data.hasPrivateKey
```
piyalbasu, line 113: re-check the hasPrivateKey status and open the modal if needed

extension/src/popup/components/InternalTransaction/SubmitTransaction/index.tsx
```diff
@@ -75,22 +79,48 @@ export const SendingTransaction = ({
...
+    }
+  }, [submitAccountState]);
 
   const handleConfirm = async (password: string) => {
-    await dispatch(confirmPassword(password));
-    setIsVerifyAccountModalOpen(false);
+    const confirmPasswordResp = await dispatch(confirmPassword(password));
+    if (confirmPassword.fulfilled.match(confirmPasswordResp)) {
+      setIsVerifyAccountModalOpen(false);
+    }
```
piyalbasu, line 123: don't dismiss the modal until we've properly confirmed

extension/src/popup/components/InternalTransaction/SubmitTransaction/index.tsx
```diff
@@ -104,6 +134,13 @@ export const SendingTransaction = ({
   const dstAssetIcons =
     destinationAsset !== "native" ? { [destinationAsset]: dstAssetIcon } : {};
 
+  if (
+    submitAccountState.state == RequestState.IDLE ||
+    submitAccountState.state == RequestState.LOADING
+  ) {
+    return <Loading />;
```
aristidesstaffieri, line 141: Is this loader mean to be centered aligned? It seems like it should be wrapped in a `<View.Content>`

piyalbasu: Ah, this component comes with a built-in `<View.Content>`, so it's not necessary here

Freeze-frame on the very brief loading state:

Screenshot 2025-10-08 at 5 38 07 PM

aristidesstaffieri: oh got it, thanks!

PR #2308, opened 2025-10-15
Feature/move icons to own hook

### LINKED ISSUE #2258: Cache asset lists and load icons async

piyalbasu:
Closes #2258 

There are 2 main bottlenecks for `useGetBalances({ withIcons: true })`:

1) on first load of the app, a user has to fetch all their token lists before they can start to fetch their icons. 
2) we must fetch these tokens lists, fetch their icons (either from cache, asset list, or issuer) before we can show the user any of their balance info

The goal of this PR is to eliminate both of those for the Account view. Here, I'm doing something similar to what I've done with the previous [account history pr](#2273 ); I'm moving this icons fetch to a separate hooks that runs after account balances are fetched.

This hook first attempts to retrieve all previously fetched icons from the background's long-lived cache (without loading the token lists at all). This retrieves the icons very quickly and immediately returns them to the UI. For all remaining icons, we go through the familiar process of checking asset lists and the asset issuer. Because this now happens async, this process ca run while the user already has a usable UI. 

I am not removing `useGetBalances({ withIcons: true })` in other views because this method actually runs quite quickly once the redux store has been hydrated. The main issue is on app first load. If a user does happen to refresh the app on a page that uses `useGetBalances({ withIcons: true })`, I added a background cache lookup to make it run even faster. (It would be quite an edge case for a user to not have any icons saved in this scenario)

### DIFF
--- a/@shared/api/helpers/getIconUrlFromIssuer.ts
+++ b/@shared/api/helpers/getIconUrlFromIssuer.ts
@@ -96,6 +96,7 @@ export const getIconUrlFromIssuer = async ({
           iconUrl,
           type: SERVICE_TYPES.CACHE_ASSET_ICON,
         });
+        break;
       }
     }
   }
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -1102,6 +1102,7 @@ export const getAssetIcons = async ({
         if (cachedIcon === null) {
           // if we've tried to fetch this icon before and it wasn't found, we marked it as null
           // don't bother trying to fetch it again
+          // this null value is only stored in Redux, so we will re-try on next app reload
           continue;
         }
 


@shared/api/helpers/getIconFromTokenList.ts
```diff
@@ -36,6 +39,7 @@ export const getIconFromTokenLists = async ({
           issuerId &&
           record.issuer &&
           record.issuer === issuerId &&
+          record.code === code &&
```
piyalbasu, line 42: this is a small bug I found while testing - we need to check the asset issuer AND asset code because one G address could be the issuer for multiple assets

@shared/api/helpers/getIconFromTokenList.ts
```diff
@@ -46,6 +50,15 @@ export const getIconFromTokenLists = async ({
     }
   }
 
+  if (verifiedToken?.icon) {
```
piyalbasu, line 53: if we find an icon, let's cache it in the background so in the future, we don't need to consult the token list at all; we'll already have the url in our background

@shared/api/helpers/getIconUrlFromIssuer.ts
```diff
@@ -84,7 +84,8 @@ export const getIconUrlFromIssuer = async ({
 
   if (toml.CURRENCIES) {
     /* If we find some currencies listed, check to see if they have the currency we're looking for listed */
-    toml.CURRENCIES.every(async ({ code: currencyCode, issuer, image }) => {
+    for (const currency of toml.CURRENCIES) {
```
piyalbasu, line 87: another small bug I found while testing - Array.every doesn't properly await promises. Changing this to a for loop

@shared/api/internal.ts
```diff
@@ -1034,24 +1037,42 @@ export const getTokenDetails = async ({
...
+  try {
+    ({ icons } = await sendMessageToBackground({
+      activePublicKey,
+      type: SERVICE_TYPES.GET_CACHED_ASSET_ICON_LIST,
+    }));
+  } catch (e) {
+    return { icons };
+  }
+  return { icons };
+};
```
piyalbasu, line 1055: I'm adding a method to grab the entire cached icon list. Previously, we were consulting this list one asset at a time

@shared/api/internal.ts
```diff
@@ -1034,24 +1037,54 @@ export const getTokenDetails = async ({
...
-  const assetIcons = {} as { [code: string]: string };
+  const assetIcons = {} as { [code: string]: string | null };
+  const skipLookup = !assetsListsData || !networkDetails;
 
   if (balances) {
-    let icon = "";
     const balanceValues = Object.values(balances);
 
     for (let i = 0; i < balanceValues.length; i++) {
+      let icon = "";
```
piyalbasu, line 1087: another small bugfix. This just _happened_ to work, but this `icon` needs to be reset on every loop

extension/src/helpers/hooks/useGetBalances.tsx
```diff
@@ -99,6 +100,14 @@ function useGetBalances(options: {
       } as AccountBalances;
 
       if (options.includeIcons) {
+        let cachedIconsFromCache = cachedIcons;
+        if (!Object.keys(cachedIcons).length) {
```
piyalbasu, line 104: if we haven't hydrated redux yet (first load), use background storage's asset cache as the cache

extension/src/popup/components/account/AccountAssets/index.tsx
```diff
@@ -36,129 +37,141 @@ export const SorobanTokenIcon = ({ noMargin }: { noMargin?: boolean }) => (
...
+}
+
+const shouldAssetIconSkipUpdate = (
+  prevProps: AssetIconProps,
+  nextProps: AssetIconProps,
+) =>
+  isEqual(prevProps.assetIcons, nextProps.assetIcons) &&
+  prevProps.isSuspicious === nextProps.isSuspicious;
+
+export const AssetIcon = memo(
```
piyalbasu, line 59: memozing here prevents unnecessary renders/flickers when asset balances poll

extension/src/popup/views/Account/hooks/useGetIcons.tsx
```diff
@@ -0,0 +1,144 @@
...
+      if (Object.keys(cachedIcons).length) {
+        icons = await getAssetIcons({
+          balances:
+            cachedBalances[networkDetails.network]?.[publicKey]?.balances,
+          cachedIcons,
+        });
+
+        payload.icons = icons;
+        reduxDispatch(saveIconsForBalances({ icons }));
+        dispatch({ type: "FETCH_DATA_SUCCESS", payload });
```
piyalbasu: dispatch any found icons in the cache so we can show them

extension/src/popup/views/Account/hooks/useGetIcons.tsx
```diff
@@ -0,0 +1,144 @@
...
+      const updatedIcons = await getAssetIcons({
+        balances: assetsWithoutIcons,
+        networkDetails,
+        assetsListsData,
+        cachedIcons: {},
+      });
+      payload.icons = { ...assetsWithIcons, ...updatedIcons };
+      reduxDispatch(saveIconsForBalances({ icons: updatedIcons }));
+      reduxDispatch(saveTokenLists(assetsListsData));
+      dispatch({ type: "FETCH_DATA_SUCCESS", payload });
```
piyalbasu, line 71: if we find more icons after lookup, dispatch them here

@shared/api/internal.ts
```diff
@@ -1096,7 +1144,7 @@ export const retryAssetIcon = async ({
 }: {
   key: string;
   code: string;
-  assetIcons: { [code: string]: string };
+  assetIcons: { [code: string]: string | null };
```
piyalbasu, line 1148: there's a lot of these type updates in this PR - this just allows us to skip an unnecessary asset issuer lookup if previously in the user's session we weren't able to find the icon

@shared/api/internal.ts
```diff
@@ -1078,9 +1121,14 @@ export const getAssetIcons = async ({
           if (tokenListIcon.icon && tokenListIcon.canonicalAsset) {
             icon = tokenListIcon.icon;
             canonical = tokenListIcon.canonicalAsset;
+          } else {
+            // if we still don't have the icon, we try to get it from the issuer
+            icon = await getIconUrlFromIssuer({ key, code, networkDetails });
```
piyalbasu, line 1127: small adjustment here - look up the icon issuer last because fetching from the fetched token list should be faster. Let's save the most expensive call for last and only if needed

piyalbasu:
This video illustrates the UX:

The first load of Account shows what the loading experience is for a new user who had not previously cached any icons.

The second load shows what happens when a user has already cached their icons

https://github.com/user-attachments/assets/d1f7c4b0-2e82-44c9-97c2-8fb1df91a763

@shared/api/internal.ts
```diff
@@ -1066,8 +1099,18 @@ export const getAssetIcons = async ({
           continue;
         }
 
-        icon = await getIconUrlFromIssuer({ key, code, networkDetails });
+        if (cachedIcon === null) {
```
aristidesstaffieri, line 1102: So in the case that an image URL didn't resolve temporarily for whatever reasons(network or something), then this cache will prevent the icon from ever being used in the future?

Maybe this cache should have a ttl?

@shared/api/helpers/getIconFromTokenList.ts
aristidesstaffieri: wdyt about adding a test that asserts that two assets from the same issuer get the correct icon?

@shared/api/helpers/getIconUrlFromIssuer.ts
aristidesstaffieri: should this break out of the loop once the icon has been found?

@shared/api/helpers/getIconFromTokenList.ts
piyalbasu: That's actually how I found this bug :) The test is right here: https://github.com/stellar/freighter/pull/2308/files#diff-9943e9b6f3ea7a02af1d671f5ea2a1b9381361230431f521219ea5f36702961cR324

@shared/api/internal.ts
piyalbasu: Ah, so something important here (that I can document better) is this `null` value is only used in Redux. We only ever store an icon in the "hard cache" in the background when we've successfully found an icon. So, if there were to be a networking issue, a user would re-fetch the next time they reload the app

@shared/api/helpers/getIconFromTokenList.ts
aristidesstaffieri: ah nice, ty

aristidesstaffieri (approved):
👏

PR #2310, opened 2025-10-15
skip blockaid scan on first fetch of account-balances

### LINKED ISSUE #2309: Fetch Blockaid scan status async
Currently, fetching /account-balances from Freighter BE on Mainnet delivers varied responses times; sometimes as long as several seconds. Why is that? And why does this not appear to be an issue on Testnet?

It looks like our built-in call to Blockaid  to attach scanned status on Mainnet makes response times vary quite a bit. Let's fetch just the balances first to get the UI loaded and then get the scan status appended later on

piyalbasu:
Closes #2309 

While testing perf improvements, I noticed something interesting: I get highly variable response times for `/account-balances` on Mainnet, but it's fairly consistent (and fast) on Testnet. Upon further investigation, it looks like adding the Blockaid status in [Freighter BE](https://github.com/stellar/freighter-backend/blob/main/src/route/index.ts#L446) is actually to blame here. Blockaid scans can sometimes a long time (adding up to several seconds sometimes), even when the list of assets remain static.

My suggestion here is to treat blockaid scans like we're treating other ancillary information like account history and icons: let's remove it from the critical path in Account and simply append it afterwards.

In this PR, the initial call to `/account-balances` includes a `is_scan_skipped` flag that will return just account balances. Once the UI is already rendered, we will then make the more expensive call to `/account-balances` that includes our Blockaid statuses.

Why make a 2nd call to `/account-balances` rather than just remove the automatic scanning in `/account-balances` and add the Blockaid results ourselves?

1. For backwards compatibility with older versions of Freighter and for Freighter Mobile, I want to only change `/account-balances` behavior by intentionally passing a flag. This approach allows that

2. I think it's still valuable to have an all-in-one response that contains both balances + blockaid scan results. For ex, when we poll for balance updates, it's useful to get all the info in one shot. In this scenario, we're much less concerned with a slightly longer response time as the UI is already usable. 

3. This continues to centralize the logic for merging Blockaid scan results with account-balances in Freighter BE instead of duplicating it across client UI's

Companion BE PR: https://github.com/stellar/freighter-backend/pull/274

### DIFF
--- a/extension/src/popup/views/Account/hooks/useGetAccountData.tsx
+++ b/extension/src/popup/views/Account/hooks/useGetAccountData.tsx
@@ -104,12 +104,13 @@ function useGetAccountData(options: {
       const allowList = appData.settings.allowList;
       const isMainnetNetwork = isMainnet(networkDetails);
 
+      // let's fetch *just* the balances (without Blockaid scan results) to quickly be able to show the user their balances
       const balancesResult = await fetchBalances(
         publicKey,
         isMainnetNetwork,
         networkDetails,
         !shouldForceBalancesRefresh,
-        true,
+        true, // skip the Blockaid scan,
       );
 
       if (isError<AccountBalances>(balancesResult)) {
@@ -140,13 +141,14 @@ function useGetAccountData(options: {
       dispatch({ type: "FETCH_DATA_SUCCESS", payload });
 
       if (isMainnetNetwork) {
+        // now that the UI has renderered, on Mainnet, let's make an additional call to fetch the balances with the Blockaid scan results included
         try {
           const balancesResult = await fetchBalances(
             publicKey,
             isMainnetNetwork,
             networkDetails,
             false,
-            false,
+            false, // don't skip the Blockaid scan,
           );
 
           const scannedPayload = {
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -530,11 +530,11 @@ export const migrateAccounts = async ({
 export const getAccountIndexerBalances = async ({
   publicKey,
   networkDetails,
-  isScanSkipped,
+  shouldSkipScan,
 }: {
   publicKey: string;
   networkDetails: NetworkDetails;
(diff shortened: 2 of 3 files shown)


piyalbasu:
In this video, I'm connected to the Freighter BE dev instance. I do a few quick close and opens to give a taste of what the loading time generally looks like with the improvement from this PR. I am loading balances without the Blockaid scan and then immediately making an async call to get the Blockaid scan results. Doing some tests, I'm getting the initial `/account-balances` to load in around 500-900ms without the Blockaid warning while the subsequent call with the Blockaid results can take anywhere from 1.5 - 2.5s!

One UX concession we are making here is that the Blockaid warning icon pops in after the UI has already loaded. (You can see this at the end of the video). @sdfcharles Is this a concession you're willing to make for these perf gainz? IMO, users holding malicious assets is probably a small subset of our users


https://github.com/user-attachments/assets/5cb2265a-b3ce-4db4-b2f3-008595cab436

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -135,6 +138,27 @@ function useGetAccountData(options: {
       }
 
       dispatch({ type: "FETCH_DATA_SUCCESS", payload });
+
+      if (isMainnetNetwork) {
+        try {
+          const balancesResult = await fetchBalances(
```
aristidesstaffieri, line 146: So it seems like the reason to call fetch balances again is for convenience. We only need to scan the assets at this point but then you would have to merge them into the balances or change the schema that expects it that way in the view. It's probably worth a comment explaining this since it's unexpected to call the same endpoint twice in a row, wdyt?.

piyalbasu: Good call! I'll add.

And yes, this is mostly for convenience: we have to make an API call anyway for the scan, so may as well as return the fully formed balances with scan results

PR #2313, opened 2025-10-20
default to empty object if request fails

### LINKED ISSUE #2311: Handle account mismatch errors with fetching account balances
Related to: https://stellarorg.sentry.io/issues/6949357132/?alert_rule_id=15034055&alert_type=issue&notification_uuid=e27e5b9e-f50c-4902-b41c-9163e64616e6&project=5599259&referrer=slack

I was able to repro this by expanding Freighter into fullscreen mode, leaving this open, and then opening the extension and switching the account. When we do this, per a recommendation from the Security team, we disable the fullscreen window from fetching data from the `background`.

But, whenever Freighter polls balances, it needs to fetch the user's list of `hidden assets` from `background`. This call (correctly) returns an error in the case and does not return the user's hidden assets config. The bug that Sentry is picking up comes from the `formatBalances` helper that toggles assets on/off. This code does not handle the missing config gracefully.

Acceptance Criteria: 
* the `formatBalances` should gracefully handle the missing config

piyalbasu:
Closes #2311 

This is related to the alert that has been popping up in our alerts channel for the last couple of weeks since I increased the Sentry coverage. This fixes a bug where if a user has the app open in fullscreen mode, goes into the extension and changes the account, polling in the fullscreen window errors because it's trying to retrieve information from Freighter's Background from a non-active account. The security team requested that we implement this to prevent non-active account windows from being able to access Freighter's Background.

This is expected behavior and does not surface any error to the user. Polling just silently stops working. However, this does trigger a TypeError in the code that Sentry correctly reports. This occasionally blasts our alerts channel because of this Sentry alert. My fix here merely fixes the TypeError. Fixing this will make sure if we receive future alerts, we can be confident that there is actually an issue and not this specific TypeError that is being swallowed here.

At one time, this type of account mismatch would automatically trigger a warning notification letting the user know they needed to refresh their window. However, this was based on Redux middleware and since we've moved away from routing requests to Background through Redux, we need to reconsider how we would let the user know that they should refresh their window.

More information: https://stellarfoundation.slack.com/archives/C06MNK4TC22/p1760635333843599

### DIFF
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -1979,7 +1979,7 @@ export const getHiddenAssets = async ({
     activePublicKey,
   });
 
-  return { hiddenAssets: response.hiddenAssets, error: response.error };
+  return { hiddenAssets: response.hiddenAssets || {}, error: response.error };
 };
 
 export const changeAssetVisibility = async ({


@shared/api/internal.ts
```diff
@@ -1979,7 +1979,7 @@ export const getHiddenAssets = async ({
     activePublicKey,
   });
 
-  return { hiddenAssets: response.hiddenAssets, error: response.error };
+  return { hiddenAssets: response.hiddenAssets || {}, error: response.error };
```
piyalbasu, line 1982: There isn't a great way to test this as the error is caught and thus does not surface anything to the user 🤔

aristidesstaffieri: seems safe enough to test by watching the rate of alerts imo

PR #2320, opened 2025-10-23
scroll on long strings; pretty print json

### LINKED ISSUE #2225: Text overflow in sign message display
5.34.1

### What did you do?
Used a stringified object as my input

in line breaking of text for all structures

### What did you see instead?
one long line of text, overflowing the container and causing a horizontal scroll

piyalbasu:
Closes #2225

This PR does 3 things:

1) When a user receives a long string in `Sign Message`, it should not break out of the container; it now scrolls
2) When a user receives json in `Sign Message` (like from near intents), it should pretty print
3) Fixes and enables Freighter-api integration tests

Some examples of different sign message payload:

Long string:

https://github.com/user-attachments/assets/f0dac152-ada3-4c64-9d7f-9d694065d5f1

JSON:

https://github.com/user-attachments/assets/fb9d025a-960c-45e0-9fb0-f23ac497147a

@sdfcharles Let me know what you think about the JSON implementation. I swapped in SDF colors to this lib to make it consistent with the design system. Happy to make any tweaks. Once we agree on a look, I can make a ticket for the mobile team to implement a similar solution

Maybe the font size can be reduced? I kept it the same as the rest of the UI for consistency, for now


extension/e2e-tests/freighterApiIntegration.test.ts
```diff
@@ -196,24 +208,54 @@ test.skip("should not sign auth entry when not allowed", async ({
 
   const popup = await popupPromise;
 
-  await expect(popup.getByText("Confirm Authorization").first()).toBeVisible();
-  await expect(
-    popup.getByText(
-      "docs.freighter.app is not currently connected to Freighter",
-    ),
-  ).toBeVisible();
+  await expect(popup.getByText("Connection Request")).toBeVisible();
```
piyalbasu, line 211: I need to open a ticket for this: currently sign auth entry and sign message show the `Connection Request` modal when not allowed while Sign Tx and Add Token show the actual signing/adding modal with a message that says "This dapp is not connected". We need to make this consistent

extension/src/popup/components/signMessage/index.tsx
```diff
@@ -1,15 +1,39 @@
...
+  try {
+    JSON.parse(str);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
 export const Message = (props: BlobProps) => {
+  console.log(typeof props.message);
```
aristidesstaffieri: leftover debug log

extension/src/popup/components/signMessage/index.tsx
```diff
@@ -1,15 +1,39 @@
...
+  const isJson = isJsonString(props.message);
   return (
     <div className="SignMessageData">
-      <div className="Message">{props.message}</div>
+      <div className="Message">
+        <div className="Message__Prefix">{props.prefix}</div>
+        {isJson ? (
+          <JSONPretty
+            json={props.message}
+            onJSONPrettyError={(e) => console.error(e)}
```
aristidesstaffieri: should we log this remotely?
Also is there any possible cleanup that we should do if this fails like show the un-pretty json?

piyalbasu: I'll return just the json as a string in an error case

piyalbasu: Ah, the lib actually already returns the json as string automatically in the error case. It actually uses the same `isJson` logic that I'm using here under the hood, so it's unlikely we ever hit it. I'm adding a Sentry log here just in case

piyalbasu:
UI updates in the latest commit:


https://github.com/user-attachments/assets/e78d66bf-24c9-4318-a871-121f489456bc

Screenshot 2025-10-24 at 4 45 50 PM

PR #2323, opened 2025-10-24
re-searching so should abort any in flight API requests

### LINKED ISSUE #2191: Asset search is not cancel-aware
5.33.9

### What did you do?
Search for an asset code(xlm) in manage-assets.
Then remove one character, trigger re-search.

A loading state, followed by the new search results.

### What did you see instead?
A flicker of the old search results, then the new results.
This happens because `useAssetLookup` does not implement cancelation on any of its data getters.
fetchTokenData and `addBlockaidScanResults` will be fired on a debounced keystroke event, but the promises can resolve after the search has changes. This hook should implement request cancelation in order to avoid flashes of stale content.

piyalbasu:
Closes #2191 

Implements a similar solution in Freighter mobile - now, when searching an asset and then quickly changing your search query, the UI will appropriately cancel any in flight API requests to prevent weird UI collisions

https://github.com/user-attachments/assets/e8e965d6-a02d-4bb7-8528-9db41a639110

### DIFF
--- a/extension/src/popup/components/__tests__/SearchAsset.test.tsx
+++ b/extension/src/popup/components/__tests__/SearchAsset.test.tsx
@@ -157,8 +157,10 @@ describe("SearchAsset", () => {
       expect(fetchSpy).toHaveBeenCalledTimes(4);
 
       // check that we only have results for our last search
-      expect(screen.getByTestId("search-asset-input")).not.toHaveValue("USDC");
-      expect(screen.getByTestId("search-asset-input")).toHaveValue("XLM");
+      expect(screen.getByTestId("ManageAssetCode")).not.toHaveTextContent(
+        "USDC",
+      );
+      expect(screen.getByTestId("ManageAssetCode")).toHaveTextContent("XLM");
     });
   });
 });


extension/src/popup/components/__tests__/SearchAsset.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <SearchAsset />
+      </Wrapper>,
+    );
+    await waitFor(() => {
+      expect(screen.getByTestId("AppHeaderPageTitle")).toHaveTextContent(
+        "Choose Asset",
+      );
+    });
+  });
+  it("should cancel the request when search is changed", async () => {
```
aristidesstaffieri, line 52: it would be awesome to have a test that asserts that the correct search results are displayed after a cancelation, wdyt?

piyalbasu: Good idea - added here: https://github.com/stellar/freighter/pull/2323/commits/a9986c3f6b2222e007a44f4b94859bbbfb0f2f6f

PR #2352, opened 2025-10-29
add ledger support for new trustline flow

### LINKED ISSUE #2337: [Bug] Freighter-specific operations fail on Ledger
5.35.3

### What did you do?
0. I connected my ledger to Freighter (extension), clicked "add wallet" and "connect a hardware device". Then I added my Ledger Nano.
1. I went to "Manage Assets" and tried to add a trustline
2. I went to "Swap" and tried to swap an asset
3. I tried to Send funds from my Ledger Nano S (updated to latest version)
4. I tried to sign a Soroban transaction on swap.xbull.app

1. A completed trustline addition
2. A completed swap operation
3. A completed Send transaction
4. A completed Soroban operation (in this case, adding a trustline)

### What did you see instead?

For 1 & 2:

Image

3. UNKNOWN_ERROR (0x6511)

4.  Image

piyalbasu:
Closes #2337 

I'm still investigating why Keeks and Bruno were having issues with *all* Ledger tx's in Freighter, but while investigating I did notice that Ledger support had not been implemented for the new trustline tx flow.

This is a critical feature for Ledger users; they currently can't add trustlines through Freighter. The only workaround is creating changeTrust op themselves and signing the tx. Because of that, I think it's pertinent to get a quick fix out ASAP and then handle any redesigns later on.

To accomplish, I took the existing hardware signature flow and added it to the change trust flow. It was a bit like forcing a circle into a square - I think this warrants some more thought around how we handle this flow. We haven't revisited the UX for this in years. @sdfcharles Let me know your thoughts on this!

https://github.com/user-attachments/assets/886845a0-f8cc-4766-96ca-3b73a1c07c3f

### DIFF
--- a/extension/src/popup/helpers/__tests__/hardwareConnect.test.ts
+++ b/extension/src/popup/helpers/__tests__/hardwareConnect.test.ts
@@ -1,0 +1,111 @@
+import TransportWebHID from "@ledgerhq/hw-transport-webhid";
+import * as StellarSDK from "stellar-sdk";
+import {
+  connectToLedgerTransport,
+  createWalletConnection,
+  getWalletPublicKey,
+  hardwareSign,
+  hardwareSignAuth,
+} from "popup/helpers/hardwareConnect";
+import { WalletType } from "@shared/constants/hardwareWallet";
+import { TESTNET_NETWORK_DETAILS } from "@shared/constants/stellar";
+
+jest.mock("@ledgerhq/hw-transport-webhid", () => {
+  return {
+    list: jest.fn().mockResolvedValue([
+      {
+        close: jest.fn().mockImplementation(() => Promise.resolve()),
+      },
+    ]),
+    create: jest.fn().mockResolvedValue({
+      close: jest.fn(),
+    }),
+    request: jest.fn().mockResolvedValue({
+      close: jest.fn(),
+    }),
+  };
+});
+
+jest.mock("@ledgerhq/hw-app-str", () => {
+  return jest.fn().mockImplementation(() => {
+    return {
+      getPublicKey: (param: string) =>
+        Promise.resolve({
+          rawPublicKey: Buffer.from(param),
+        }),
+      signTransaction: () => Promise.resolve({ signature: "signTransaction" }),
+      signHash: () => Promise.resolve({ signature: "signHash" }),
+      signSorobanAuthorization: () =>
+        Promise.resolve({ signature: "signSorobanAuthorization" }),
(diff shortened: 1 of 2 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrust.tsx
```diff
@@ -104,9 +132,17 @@ function useGetChangeTrust() {
     }
   };
 
+  useEffect(() => {
+    if (preparedTransaction) {
+      submitHwData({ publicKey, networkDetails });
```
piyalbasu, line 137: Not very straightforward 😢 In order to re-use `state`, we listen for the `preparedTransaction` from the hwSign flow and then submit once we have it

extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useResetChangeTrustData.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const resetChangeTrustData = async ({
+    isHardwareWallet,
+  }: {
+    isHardwareWallet: boolean;
+  }) => {
+    dispatch({ type: "FETCH_DATA_START" });
+
+    reduxDispatch(resetSubmission());
+    const balancesResult = await fetchBalances(
```
piyalbasu: I previously had just been re-fetching balances at the end of `fetchData` from `useGetChangeTrust`, but I realized that this was causing parent components to re-render before we got to the "Success|Fail" screen. I'm now calling this when the user clicks `Done` at the end of their tx. I moved this into its own hook so I could have a loading state to let the user know there's an API call happening

extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrust.tsx
aristidesstaffieri: so this works because only the hw flow uses the store correct? 

Should we clean up the store state when this hook unmounts?

aristidesstaffieri: oh I see `resetChangeTrustData` doing that now actually in the success/fail steps.

PR #2351, opened 2025-10-29
[FEATURE] adds send and swap buttons to asset detail view

aristidesstaffieri:
Closes #2347 

Adds back the send/swap buttons on the asset detail views.

Screenshot 2025-10-29 at 10 59 08 AM

PR #2353, opened 2025-10-30
[FEATURE] new send/swap navigation flow

### LINKED ISSUE #2201: Support new send navigational flow

aristidesstaffieri:
Closes #2201 

What
Updates the navigation flow in the send and swap flows to match design updates.
Adds query parameters to set the `asset` in the send flow, and to set the `source_asset/destination_asset` in the swap flow.
Adds new components - SelectionTile, AssetTile, AddressTile. Used in send and swap flows.
Updates tests to match new flow expectations.

Unblocks: https://github.com/stellar/freighter/pull/2351

Walkthrough for both flows:

https://github.com/user-attachments/assets/b6d216fc-2ac2-491c-b6ad-aadeb4745c33

### DIFF
--- a/extension/src/popup/components/sendPayment/AddressTile/index.tsx
+++ b/extension/src/popup/components/sendPayment/AddressTile/index.tsx
@@ -6,6 +6,8 @@ import { IdenticonImg } from "popup/components/identicons/IdenticonImg";
 import { truncatedFedAddress, truncatedPublicKey } from "helpers/stellar";
 import { SelectionTile } from "popup/components/SelectionTile";
 
+import "./styles.scss";
+
 interface AddressTileProps {
   address: string;
   federationAddress?: string;
@@ -36,7 +38,7 @@ export const AddressTile = ({
 
   return (
     <SelectionTile
-      icon={<Icon.Plus stroke="#707070" />}
+      icon={<Icon.Plus className="TileIcon" />}
       primaryText={t("Send to")}
       secondaryText={t("Choose Recipient")}
       onClick={onClick}
--- a/extension/src/popup/components/sendPayment/AddressTile/styles.scss
+++ b/extension/src/popup/components/sendPayment/AddressTile/styles.scss
@@ -1,0 +1,3 @@
+.TileIcon {
+  stroke: var(--sds-clr-gray-09);
+}


extension/src/popup/components/AssetTile/__tests__/index.test.tsx
```diff
@@ -0,0 +1,288 @@
+import React from "react";
+import { render, fireEvent, screen } from "@testing-library/react";
+import { AssetTile } from "../index";
+
+// Mock the dependencies
+jest.mock("popup/components/account/AccountAssets", () => ({
+  AssetIcon: ({ code, icon }: { code: string; icon: string | null }) => (
```
piyalbasu: is there a compelling reason to not use the real component and test against it? I'm not sure that testing that this mock component can render provides us that much value

extension/src/popup/components/AssetTile/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <AssetIcon
+            assetIcons={
+              asset.canonical !== "native"
+                ? { [asset.canonical]: assetIcon }
+                : {}
+            }
+            code={asset.code}
+            issuerKey={asset.issuer}
+            icon={assetIcon}
+            isSuspicious={false}
```
piyalbasu: We should pass the actual Blockaid designation here

aristidesstaffieri: we actually don't scan assets at any spot where this is used, but I can make this take the arg and have the components pass down their own values.

extension/src/popup/components/sendPayment/SendDestinationAsset/index.tsx
```diff
@@ -111,25 +112,15 @@ export const SendDestinationAsset = ({
...
-            <Button isRounded size="sm" variant="tertiary">
-              <Icon.ChevronRight />
-            </Button>
-          </div>
+          <AddressTile
+            address={destination}
+            federationAddress={federationAddress}
+            onClick={goToDestination}
+          />
           <TokenList
```
piyalbasu, line 124: This is unrelated to your change set, but while testng I noticed on `Account` an asset with no USD value is represented as `--` while in the asset list, it's just blank space. If it's a simple fix, can we address here?

Account screen:
Screenshot 2025-10-31 at 2 56 48 PM

Swap select screen:
Screenshot 2025-10-31 at 2 56 57 PM

extension/src/popup/components/SelectionTile/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  testId?: string;
+}
+
+export const SelectionTile = ({
+  icon,
+  primaryText,
+  secondaryText,
+  onClick,
+  isEmpty = false,
+  useIconWrapper = true,
```
piyalbasu: NIT: can we call this `shouldUseIconWrapper` or something similar since the `use-` prefix is reserved for React hooks

extension/src/popup/components/sendPayment/SendDestinationAsset/index.tsx
aristidesstaffieri: yeah it seems like it should also have the dashes so I can add those.

extension/src/popup/components/sendPayment/AddressTile/__tests__/index.test.tsx
```diff
@@ -0,0 +1,214 @@
+import React from "react";
+import { render, fireEvent, screen } from "@testing-library/react";
+import { AddressTile } from "../index";
+
+// Mock the dependencies
+jest.mock("popup/components/identicons/IdenticonImg", () => ({
+  IdenticonImg: ({ publicKey }: { publicKey: string }) => (
+    <div data-testid="identicon-img" data-public-key={publicKey}>
```
piyalbasu: I imagine you're using this custom component to help test the value that's being passed to `IdenticonImg`. I think it might be more useful to just add an `alt` tag that contains the passed `public key` on the real `IdenticonImg` component. 

The thing that gives me pause is that `IdenticonImg` returns an img and this component returns a `div`

extension/src/popup/components/sendPayment/AddressTile/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }
+        onClick={onClick}
+        testId="address-tile"
+      />
+    );
+  }
+
+  return (
+    <SelectionTile
+      icon={<Icon.Plus stroke="#707070" />}
```
piyalbasu: is it possible to add the styling using the SDS var? It will help guard against SDS changes in the future

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -81,7 +85,28 @@ export const SendPayment = () => {
...
   });
 
-  const [activeStep, setActiveStep] = React.useState(STEPS.DESTINATION);
+  const [activeStep, setActiveStep] = React.useState(STEPS.AMOUNT);
+
+  // Handle query params and set defaults on mount
+  React.useEffect(() => {
+    const params = new URLSearchParams(location.search);
+    const destinationParam = params.get("destination");
+    const assetParam = params.get("asset");
```
piyalbasu, line 95: It'd be nice to add some validation here. Since these params are being added in-app, it will likely be rare for an unsanitized value to be passed here, but I did notice 

- An invalid `destination` renders weird and could lead to errors further down the flow:
Screenshot 2025-10-31 at 3 25 03 PM

- an invalid `asset` causes a breaking error:
Screenshot 2025-10-31 at 3 25 41 PM

extension/src/popup/views/Swap/index.tsx
```diff
@@ -33,7 +34,26 @@ export const Swap = () => {
   useEffect(() => {
     dispatch(resetSimulation());
     dispatch(resetSubmission());
-  }, [dispatch]);
+
+    // Handle query params and set defaults on mount
+    const params = new URLSearchParams(location.search);
+    const sourceAssetParam = params.get("source_asset");
+    const destinationAssetParam = params.get("destination_asset");
```
piyalbasu, line 42: Same comment as above in regards to query params

extension/src/popup/components/AssetTile/__tests__/index.test.tsx
aristidesstaffieri: it was just easier to mock it because it reaches into store state and these tests weren't testing the internals of `AssetIcon`, but I updated the tests to include the store state so it doesn't need to be mocked.
147e2a3040726f49a4bbd8e5a5243be8bac67701

extension/src/popup/components/sendPayment/AddressTile/index.tsx
aristidesstaffieri: I don't think so, how would you set the var from the JS side?

extension/src/popup/components/AssetTile/index.tsx
aristidesstaffieri: Added in 6832b73236cfe59fce67606a883a70ccab943931

extension/src/popup/components/SelectionTile/index.tsx
aristidesstaffieri: updated in 6832b73236cfe59fce67606a883a70ccab943931

extension/src/popup/components/sendPayment/SendDestinationAsset/index.tsx
aristidesstaffieri: added in ca892bca9b22bc40ba40f5203dbc798f11828c69

extension/src/popup/components/sendPayment/AddressTile/__tests__/index.test.tsx
aristidesstaffieri: I updated the tests to use the real IdenticonImg in c37c00ef9eed2264c64a99a88dd5d957e82ca726

extension/src/popup/views/SendPayment/index.tsx
aristidesstaffieri: good call, I added query param validation and fallback in 009e2322f67527a69d1a4833a3b03f51b59f991c

extension/src/popup/views/Swap/index.tsx
aristidesstaffieri: also added in 009e2322f67527a69d1a4833a3b03f51b59f991c

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -555,87 +553,36 @@ export const SwapAmount = ({
                     {t("Set Max")}
                   </Button>
                 </div>
-                <div
-                  className="SwapAsset__EditSrcAsset"
+                <AssetTile
+                  isSuspicious={false}
```
CassioMG, line 557: is hardcoding `false` here intended?

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -555,87 +553,36 @@ export const SwapAmount = ({
...
-                    >
-                      <Icon.ChevronRight />
-                    </Button>
-                  )}
-                </div>
+                  emptyLabel="Send"
+                  testId="swap-src-asset-tile"
+                />
+                <AssetTile
+                  isSuspicious={false}
```
CassioMG, line 570: same

extension/src/popup/views/SendPayment/index.tsx
```diff
@@ -81,7 +86,42 @@ export const SendPayment = () => {
     isMainnet: isMainnet(networkDetails),
   });
 
-  const [activeStep, setActiveStep] = React.useState(STEPS.DESTINATION);
+  const [activeStep, setActiveStep] = React.useState(STEPS.AMOUNT);
+
+  // Handle query params and set defaults on mount
+  React.useEffect(() => {
```
CassioMG:
```suggestion
  useEffect(() => {
```

extension/src/popup/components/swap/SwapAmount/index.tsx
aristidesstaffieri: it is yeah, or at least it was what existed here so far. We don't scan in this flow so this has been set to false. This PR only moves that logic because it abstracts the common dom elements.

extension/src/popup/components/swap/SwapAmount/index.tsx
aristidesstaffieri: answered on the other comment about this.

extension/src/popup/components/sendPayment/AddressTile/index.tsx
piyalbasu: I think you can just add a class name with the styling you want instead of setting `stroke`, no?

aristidesstaffieri: ah apologies I didn't realize they got passed through, updated in b01e0712dec2991b7eb97978d2953ba3c437e099

PR #2361, opened 2025-10-31
[CHORE] git process updates

### LINKED ISSUE #1805: `Bump versions PR` job not working
When we deploy the extension, there's a job that that bumps the versions to the appropriate version before deploying to the stores. After deployment succeeds, the pipeline would automatically open a PR that would request to merge these version changes back to the repo. It was helpful to make sure our versions stay up to date in the repo.

Right now, it's opening a blank PR to the repo after deployment, which gets automatically closed by GitHub. Let's make sure this job correctly opens the PR to merge these version changes back to the repo.

*BONUS POINT*
In the past, when this PR was working, the tests would not automatically run when the PR was programmatically opened. Because of this, we'd be unable to merge this PR unless we nudged the CI by committing an empty commit to this branch. Is it possible to solve this issue while we're fixing this?

aristidesstaffieri:
Closes #1805 

Removes version bump PR step from submit production job.
Adds version check and update for release branches in test run step, adds script to bump versions in release branches.
Moves the add translations hook to run on pre-commit instead of pre-push, removes translation commit.

### DIFF
--- a/.github/workflows/submitProduction.yml
+++ b/.github/workflows/submitProduction.yml
@@ -5,11 +5,6 @@ env:
   HUSKY: 0
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: New Version
-        default: "1.0.0"
-        required: true
 jobs:
   bump-version:
     name: Bump Package Version and Submit Extension
@@ -27,6 +22,12 @@ jobs:
           gh run watch ${{ github.run_id }}
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Get version from package.json
+        id: package_version
+        run: |
+          VERSION=$(grep -o '"version": "[^"]*"' extension/package.json | head -1 | sed 's/"version": "\(.*\)"/\1/')
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Version from package.json: $VERSION"
       - name: Enable Corepack
         run: corepack enable
       - name: Build extension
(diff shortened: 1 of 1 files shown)


.github/workflows/submitProduction.yml
```diff
@@ -27,31 +27,6 @@ jobs:
           gh run watch ${{ github.run_id }}
```
aristidesstaffieri, line 22: @piyalbasu I think we could remove the step that posts this to #releases channel in slack since theres another bot doing that for the whole org.

CassioMG: good call

CassioMG (approved):
LGTM! Should we update the Release Process docs once this PR is merged? It appears we have a [google doc](https://docs.google.com/document/d/1UmKnTAZ1yuyhOxIheLRJEo9bn9IjVgtglnL2DCMrg_k/edit?tab=t.0#heading=h.6d6vj1jknakf) version and a [confluence](https://stellarorg.atlassian.net/wiki/spaces/PM/pages/4352606225/Freighter+Release+Process) version, not sure if we are maintaining only one of them or both

.github/workflows/submitProduction.yml
piyalbasu: Yup, let's remove the Slack post in this pipeline. Could you also remove the similar Slack post in the Beta pipeline?

aristidesstaffieri: I dont see betas getting published to #releases, should we keep this for betas in our channel?

aristidesstaffieri:
> LGTM! Should we update the Release Process docs once this PR is merged? It appears we have a [google doc](https://docs.google.com/document/d/1UmKnTAZ1yuyhOxIheLRJEo9bn9IjVgtglnL2DCMrg_k/edit?tab=t.0#heading=h.6d6vj1jknakf) version and a [confluence](https://stellarorg.atlassian.net/wiki/spaces/PM/pages/4352606225/Freighter+Release+Process) version, not sure if we are maintaining only one of them or both

hmm yeah I don't know which one is the real one now, but I can update the process wherever. @piyalbasu do you have a preference on which docs we keep?

.github/workflows/submitProduction.yml
```diff
@@ -27,31 +27,6 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v4
```
piyalbasu, line 24: So with your new bash script, we (basically) don't need the `github.event.inputs.version` param anymore. The only place we're still using it now is when we create the GH release and upload the build. If you're able to update those steps to just use the value from the package.json, we would be able to remove the `on > workflow_dispatch > inputs` field from this pipeline completely 🎉

aristidesstaffieri: nice, good idea. I made this change in 543112d96c1d4629b087301169f629a7b15d0005

PR #2363, opened 2025-11-05
Feature/use ledger key for home domains

### LINKED ISSUE #2259: Load home-domains using RPC
We load home domains for several UX views. Let's use RPC to make this call faster and do this async so it doesn't block UI loading

piyalbasu:
Closes #2259 

_Overview_
This PR swaps out all instances where we fetched home domains using Horizon for the new endpoint in Freighter BE v2 where we fetch accounts by using ledger keys. This allows us to fetch multiple accounts in one RPC call 🎉 The general strategy I'm employing here is to fetch multiple domains at once rather than fetching them one at a time as needed as we iterate over a list. This methodology can be seen in a couple of different spots: when fetching icons for balances and when fetching icons for history rows.

As we shift these lookups to our own BE (rather than Horizon), we need to be cognizant of the number of requests users are making (both for speed and to limit 429's).

_Implementation Details_

* I've created a generic `getLedgerKeyAccounts` helper so we can easily fetch any account information (not just home domains) from ledger keys in the future. It's useful for quickly getting an account's xlm balance, for example

* `getIconUrlFromIssuer` used to lookup the home domain, toml, and then finally the icon url in one method. It still does that, but it now allows you to pass in a `homeDomain` and skip the home domain lookup entirely. This is useful if you know you need icons for 5 different assets and you want to look up all 5 home domains in one API call rather than letting `getIconUrlFromIssuer` make one request for a home domain each

### DIFF
--- a/@shared/api/helpers/getLedgerKeyAccounts.ts
+++ b/@shared/api/helpers/getLedgerKeyAccounts.ts
@@ -25,11 +25,19 @@ export const getLedgerKeyAccounts = async ({
     const url = new URL(`${INDEXER_V2_URL}/ledger-key/accounts`);
     url.searchParams.append("network", networkDetails.network);
     const response = await fetch(url, options);
+    if (!response.ok) {
+      const _err = JSON.stringify(response);
+      captureException(
+        `Failed to fetch ledger key accounts - ${response.status}: ${response.statusText}`,
+      );
+
+      throw new Error(_err);
+    }
     const { data } = (await response.json()) as { data: LedgerKeyAccounts };
 
     fetchedAccounts = data.ledger_key_accounts;
   } catch (e) {
-    captureException(`Error fetching asset domains: ${e}`);
+    captureException(`Error fetching ledger key accounts: ${e}`);
   }
 
   return fetchedAccounts;
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -1156,20 +1156,25 @@ export const getAssetIcons = async ({
       networkDetails,
     });
 
+    const iconsToFetch = [] as Promise<void>[];
     for (const { key, code } of domainsToFetch) {
       const canonical = getCanonicalFromAsset(code, key);
       if (assetDomains[key]) {
-        const icon = await getIconUrlFromIssuer({
+        const fetchIcon = getIconUrlFromIssuer({
           key,
           code,
           networkDetails,
           homeDomain: assetDomains[key],
+        }).then((icon) => {
+          assetIcons[canonical] = icon || null;
         });
-        assetIcons[canonical] = icon || null;
-        break;
+
+        iconsToFetch.push(fetchIcon);
+      } else {
+        assetIcons[canonical] = null;
       }
-      assetIcons[canonical] = null;
     }
+    await Promise.all(iconsToFetch);
   }
   return assetIcons;
 };
@@ -1230,6 +1235,7 @@ export const getAssetDomains = async ({
       assetDomains[key] = value.home_domain;
     });
   } catch (e) {
+    captureException(`Error constructing asset domains: ${e}`);
     return {};
   }
 


extension/src/helpers/hooks/useGetAssetDomainsWithBalances.ts
```diff
@@ -116,18 +116,9 @@ export function useGetAssetDomainsWithBalances(getBalancesOptions: {
           const cachedHomeDomain = homeDomains[issuer.key];
           if (useCache && cachedHomeDomain) {
             domain = cachedHomeDomain;
-          } else {
+          } else if (cachedHomeDomain !== null) {
```
piyalbasu, line 120: using a similar strategy here that I used with icon url's; if we tried to look it up and didn't find it the first time, don't try again for the rest of the session. We'll retry next to the user re-opens the app

extension/src/popup/components/account/AssetDetail/index.tsx
```diff
@@ -136,7 +138,7 @@ export const AssetDetail = ({
     accountBalances.balances,
   ) as Exclude<AssetType, LiquidityPoolShareAsset>;
 
-  const icons = accountBalances.icons || {};
+  const icons = assetIcons || {};
```
piyalbasu, line 150: I included this quick fix for asset icons I noticed while QA'ing the home domains in this view; we weren't loading icons for non-XLM assets. This was a result of the icons refactor I did in this release branch

extension/src/popup/views/Account/hooks/useGetAccountHistoryData.tsx
```diff
@@ -58,9 +71,18 @@ function useGetAccountHistoryData() {
           networkDetails: networkDetails,
           publicKey,
           fetchTokenDetails,
+          icons: cachedIcons,
```
piyalbasu, line 74: For account history icons, previously we would need to make a new lookup for every icon (regardless of if we had fetched it earlier in the iteration) because selectors only update on a new render. It wasn't possible to cache using redux

Now, I'm also passing along a temporary obj that caches newly fetched icons as we iterate. If we fetched (for ex) a USDC icon once, we don't need to look up the home domain's toml again the next time we come across the need for a USDC icon further down the loop.

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -613,8 +732,18 @@ const createHistorySections = async (
...
   }) => Promise<TokenDetailsResponse | Error>,
-) =>
-  await operations.reduce(
+  homeDomains: { [assetIssuer: string]: string | null },
+) => {
+  /* 
+    To prevent multiple requests for home domains as we build each row, 
+    we iterate through the operations and collect the asset issuers that need home domains in a single request.
+  */
+  const fetchedHomeDomains = await getHomeDomainsForOperations(
```
piyalbasu, line 741: When building account history rows, we used to iterate over operations while building rows and (if needed) make a request for home domains along the way in order to fetch the icon url. This led to a scenario where we could make up to 100 separate requests for a home domain for account history.

@shared/api/internal.ts
```diff
@@ -1146,6 +1149,28 @@ export const getAssetIcons = async ({
...
+  if (domainsToFetch.length > 0 && networkDetails) {
+    const assetDomains = await getAssetDomains({
+      assetIssuerDomainsToFetch: domainsToFetch.map(({ key }) => key),
+      networkDetails,
+    });
+
+    for (const { key, code } of domainsToFetch) {
+      const canonical = getCanonicalFromAsset(code, key);
+      if (assetDomains[key]) {
+        const icon = await getIconUrlFromIssuer({
```
aristidesstaffieri: can we fetch these in parallel?

@shared/api/internal.ts
```diff
@@ -1146,6 +1149,28 @@ export const getAssetIcons = async ({
...
+          key,
+          code,
+          networkDetails,
+          homeDomain: assetDomains[key],
+        });
+        assetIcons[canonical] = icon || null;
+        break;
```
aristidesstaffieri: why break here? Doesn't this just break after the first iteration?

@shared/api/helpers/getLedgerKeyAccounts.ts
```diff
@@ -0,0 +1,36 @@
...
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({
+        public_keys: accountList,
+      }),
+    };
+    const url = new URL(`${INDEXER_V2_URL}/ledger-key/accounts`);
+    url.searchParams.append("network", networkDetails.network);
+    const response = await fetch(url, options);
+    const { data } = (await response.json()) as { data: LedgerKeyAccounts };
```
aristidesstaffieri, line 36: this should check `response.ok` before parsing the json to avoid a runtime error in the event of an API error.

@shared/api/internal.ts
```diff
@@ -1178,31 +1203,36 @@ export const retryAssetIcon = async ({
...
+  try {
+    const fetchedAccounts = await getLedgerKeyAccounts({
+      accountList: filteredAssetIssuerDomainsToFetch,
+      networkDetails,
+    });
+    Object.entries(fetchedAccounts).forEach(([key, value]) => {
+      assetDomains[key] = value.home_domain;
+    });
+  } catch (e) {
+    return {};
```
aristidesstaffieri, line 1239: should we have any visibility into this error?

@shared/api/internal.ts
piyalbasu: I'll make this more clear on the next commit!

PR #2369, opened 2025-11-06
only fetch asset list data if needed

### LINKED ISSUE #2260: Don't block UI for account-balances and asset list data in Sign Tx
We wait to load user balances and asset list data before we can show the Sign TX screen.

Find a way to make these calls optional or faster

piyalbasu:
Closes #2260 

__What__

Sign Transaction is currently taking around 2 seconds to load. Not super long - but should likely be a bit quicker given how simple the UI is.  I made a few small adjustments here:

1. Only fetch asset list data if needed (if we're dealing with asset diffs/changeTrust AND if we weren't able to already fetch the icon from the background cache

2. Render the UI without icons first and then go back and fill in the icon data after

3. Fetch balances without the Blockaid scan to get the user's available XLM balance quicker

__Why__

This gives the user a UI to look at a bit more quickly. They may need to wait an extra amount of time for the icon to finish loading if they have never encountered the asset before

NOTE: I took a look at rendering the UI with out the Blockaid scan and then going back for it after the first `dispatch`, but the `scan-tx` is pretty fast - it was actually making the UI look a bit more jumpy because it was populating so quickly after the first render. Now we're just waiting for the unscanned account balances and the scan-tx result before rendering. On a fast connection, that's taking about 500-750ms, which is fairly quick.

BEFORE: 

https://github.com/user-attachments/assets/938469c2-0ee7-4fec-8c25-e41a3d9c8238

AFTER:

https://github.com/user-attachments/assets/b5e33de3-1122-4224-807b-f647bb0a9872

EXAMPLE OF LOADING WITHOUT ICON CACHE FOLLOWED BY A LOAD WITH THE CACHE:

https://github.com/user-attachments/assets/270f467f-f668-4bce-8419-6ceb11078894

### DIFF
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -30,7 +30,6 @@ import {
 } from "helpers/urls";
 import { emitMetric } from "helpers/metrics";
 import {
-  getCanonicalFromAsset,
   getTransactionInfo,
   isFederationAddress,
   isMuxedAccount,
@@ -525,13 +524,11 @@ const AssetDiffs = ({ assetDiffs, icons }: AssetDiffsProps) => {
       default: {
         const code = "code" in diff.asset ? diff.asset.code! : "";
         const issuer = "issuer" in diff.asset ? diff.asset.issuer! : "";
-        const canonical = getCanonicalFromAsset(code, issuer);
-        const icon = icons[canonical];
         return (
           <div className="SignTransaction__AssetDiffRow">
             <div className="SignTransaction__AssetDiffRow__Asset">
               <AssetIcon
-                assetIcons={code !== "XLM" ? { [canonical]: icon } : {}}
+                assetIcons={code !== "XLM" ? icons : {}}
                 code={code}
                 issuerKey={issuer}
               />
@@ -572,12 +569,10 @@ export const Trustline = ({ operations, icons }: TrustlineProps) => {
     const renderTrustlineAsset = (line: Asset | LiquidityPoolAsset) => {
       if ("code" in line) {
         const { code, issuer } = line;
-        const canonical = getCanonicalFromAsset(code, issuer);
-        const icon = icons[canonical];
         return (
           <>
             <AssetIcon
-              assetIcons={code !== "XLM" ? { [canonical]: icon } : {}}
+              assetIcons={code !== "XLM" ? icons : {}}
               code={code}
               issuerKey={issuer}
             />


extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -89,20 +90,48 @@ function useGetSignTxData(
         publicKey,
         isMainnetNetwork,
         networkDetails,
+        false,
+        true,
```
piyalbasu, line 94: fetch the balances without the scan

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -89,20 +90,48 @@ function useGetSignTxData(
...
+        allAccounts,
+        publicKey,
+        accountToSign,
+        setActiveAccount: (account: string) =>
+          reduxDispatch(makeAccountActive(account)),
+      });
+
+      if (!currentAccount) {
+        setAccountNotFound(true);
+      }
```
piyalbasu, line 108: moving this info up higher in the order so we can save icon work for last

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -101,14 +100,14 @@ export const SignTransaction = () => {
...
-  const { state: scanTxState, fetchData } = useGetSignTxData(
+  const { state: signTxState, fetchData } = useGetSignTxData(
     {
       xdr: transactionXdr,
       url,
     },
     {
       showHidden: false,
-      includeIcons: true,
+      includeIcons: false,
```
piyalbasu, line 110: no need to proactively fetch balance icons (as this can cause needlessly fetching all the asset list data). We'll selectively fetch icons depending on if we have asset diffs or a change trust op in the tx

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -191,21 +190,21 @@ export const SignTransaction = () => {
   const isSubmitDisabled = isMemoRequired || !isDomainListedAllowed;
 
   if (
-    scanTxState.state === RequestState.IDLE ||
-    scanTxState.state === RequestState.LOADING
+    signTxState.state === RequestState.IDLE ||
+    signTxState.state === RequestState.LOADING
```
piyalbasu, line 194: renaming this var to be more indicative of what we fetched

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -525,13 +524,11 @@ const AssetDiffs = ({ assetDiffs, icons }: AssetDiffsProps) => {
...
         const code = "code" in diff.asset ? diff.asset.code! : "";
         const issuer = "issuer" in diff.asset ? diff.asset.issuer! : "";
-        const canonical = getCanonicalFromAsset(code, issuer);
-        const icon = icons[canonical];
         return (
           <div className="SignTransaction__AssetDiffRow">
             <div className="SignTransaction__AssetDiffRow__Asset">
               <AssetIcon
-                assetIcons={code !== "XLM" ? { [canonical]: icon } : {}}
+                assetIcons={code !== "XLM" ? icons : {}}
```
piyalbasu, line 531: a small adjustment here:

previously, until icons were loaded, the `assetIcons` param was `{"G123": undefined}`, which would cause `<AssetIcon` to show the "not found" image placeholder. If we pass it an empty object, (which `icons` will be until fetched), it will properly show the loading state

aristidesstaffieri: ah nice, this fallback really makes it better.

aristidesstaffieri (approved):

PR #2373, opened 2025-11-09
Feature/cache token prices

### LINKED ISSUE #2372: Cache token prices
Currently, on the Wallets view, we need to fetch token-prices for each Mainnet wallet before we load the UI. A user can have an unlimited number of wallets, which can lead to this view taking a long time to load as we wait for every token-prices call to resolve.

Let's optimize this is new 2 ways:

1) Batch load token-prices so a user can see a few of their wallet totals while we wait for the rest to resolve

2) Cache the token-prices so we don't need to reload toke-prices we just fetched

piyalbasu:
Closes #2372 

_What_

1. On the Wallets view, fetch token-prices in batches rather than waiting for all token-prices for all wallets before rendering a UI. We will load 6 at a time as that is currently what fits in the extension's popup view without scrolling. We will render the first 6 wallets and then load the UI so a user can interact with their UI while we continue to load the rest of the token-prices

2. Cache token-prices to reduce the number of calls we have to make to the BE for this info. This will make subsequent views of the Wallets view, as well as the initial load for views that also use token-prices (like Account, and Send/Swap asset select) much faster. We will employ a similar strategy to what we've with caching account-balances

3. Fetch wallet balances using the `should_skip_scan=true` flag, which will make the balances calls resolve faster

_Why_

We've gotten some complaints that the Wallet select page sometimes doesn't load. My suspicion is these users had many wallets and were waiting a long time for both account balances and token-prices to load

_Examples_

Loading token-prices for all wallets and then navigating around, using the cached token-prices to reduce load time:

https://github.com/user-attachments/assets/2e108f05-fd3e-4640-80bc-db5490b77c1c

Loading token-prices on a throttled network to show the loading state for wallets that are still in process:

https://github.com/user-attachments/assets/9d14935f-e343-4ba1-9d81-d7daf1c39461

### DIFF
--- a/extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
+++ b/extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
@@ -149,7 +149,7 @@ function useGetWalletsData() {
     } catch (error) {
       dispatch({
         type: "FETCH_DATA_ERROR",
-        payload: { error, isFetchingTokenPrices: false },
+        payload: error,
       });
       captureException(`Error loading wallets data - ${error}`);
       return error;


extension/src/helpers/hooks/useGetBalances.tsx
```diff
@@ -84,8 +84,10 @@ function useGetBalances(options: {
     try {
       const cachedBalanceData =
         cachedBalances[networkDetails.network]?.[publicKey];
+      const isCacheValid =
+        cachedBalanceData && cachedBalanceData.updatedAt > Date.now() - 180000; // 3 minutes
```
piyalbasu: let's use a similar cache-busting technique for both account-balances and token-prices to make our lives simpler

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -14,31 +13,11 @@ import {
...
 import { NetworkDetails } from "@shared/constants/stellar";
 import { APPLICATION_STATE } from "@shared/constants/applicationState";
-import { useDispatch, useSelector } from "react-redux";
+import { useDispatch } from "react-redux";
 import { AppDispatch } from "popup/App";
 import { makeAccountActive } from "popup/ducks/accountServices";
 import { changeNetwork } from "popup/ducks/settings";
-import { balancesSelector } from "popup/ducks/cache";
-
-export const getTokenPrices = async ({
```
piyalbasu, line 23: this whole block has been moved to the useGetTokenPrices hook

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -255,49 +241,6 @@ function useGetAccountData(options: {
     return () => clearInterval(interval);
   }, [_isMainnet, state.data, fetchBalances]);
 
-  useEffect(() => {
-    // if it's been 2 minutes since the last balance update, force update
```
piyalbasu, line 259: rather than employ this method of checking for a stale cache - let's make things simpler and do that inside the `fetchBalances` helper. This will allow any view to clear the cache, not just Account

extension/src/helpers/hooks/useGetBalances.tsx
```diff
@@ -84,8 +84,10 @@ function useGetBalances(options: {
     try {
       const cachedBalanceData =
         cachedBalances[networkDetails.network]?.[publicKey];
+      const isBalancesCacheValid = isCacheValid(cachedBalanceData);
```
piyalbasu, line 87: Introducing a helper to clear cache for cached items that we will want to refresh mid-session, unlike things like token lists which can live for the whole length of the user's session

extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
```diff
@@ -22,10 +23,14 @@ export interface ResolvedData {
     [key: string]: string;
   };
   applicationState: APPLICATION_STATE;
+  isFetchingTokenPrices: boolean;
 }
 
 type GetWalletsData = NeedsReRoute | ResolvedData;
 
+// the number of wallets that fit in the extension viewport at once
+const BATCH_SIZE = 6;
```
aristidesstaffieri, line 32: I'm surprised that this makes any difference. Did we profile the change in relation to batching? The prices are stored at rest when we get them and the lookup gets them all at once so I would expect the difference between a handful and even hundreds of prices to be pretty much the same.

extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
```diff
@@ -60,42 +127,23 @@ function useGetWalletsData() {
...
+            accounts: batch,
+            networkDetails,
+            isMainnetNetwork,
+            useCache,
+            payload,
+          });
+        }
       }
 
+      payload.isFetchingTokenPrices = false;
```
aristidesstaffieri, line 146: should this also be set to false in the catch block to avoid a forever state in case of failures in the batch calls?

extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
piyalbasu: Yeah, the rub here is that you need to fetch the account balances for each wallet in addition to grabbing the token-prices. The token-prices endpoint (on a fast connection) is just 26ms. But grabbing the balances (with scan) was something like 500-600ms. So we were looking at something like a half second per wallet. Extrapolated to 20 wallets (which is possible, but maybe not common), that's 10 seconds - even longer on a slower connection

By not getting the blockaid scan, now fetching the account-balance plus token-price for each wallet is like 200ms. For 20 wallets, it's around 4 seconds on a fast connection. Much improved - but it's still good practice to guard against the worst case. By batching into 6, the users max wait time is ever only around 1.2s (on fast internet). With caching, this number comes down even more

A potential additional improvement is for token-prices to fetch balances for you so we don't need to make separate calls for balances and token-prices, but that still doesn't fix the issue of a user with a ton of wallets still waiting for these calls to resolve.

aristidesstaffieri: got it, sounds like the bottleneck is not really prices but they get batched due to their relationship to the balances(which are the slow part).

sgtm

extension/src/popup/views/Wallets/hooks/useGetWalletsData.tsx
```diff
@@ -147,7 +147,10 @@ function useGetWalletsData() {
       dispatch({ type: "FETCH_DATA_SUCCESS", payload });
       return payload;
     } catch (error) {
-      dispatch({ type: "FETCH_DATA_ERROR", payload: error });
+      dispatch({
+        type: "FETCH_DATA_ERROR",
+        payload: { error, isFetchingTokenPrices: false },
```
CassioMG: does the payload here accept both formats?

`payload: error`
and 
`payload: { error, isFetchingTokenPrices: false }`

piyalbasu: The type does allow this, but on second thought, the `FETCH_DATA_ERROR` type will trigger this warning notification: https://github.com/stellar/freighter/blob/master/extension/src/popup/views/Wallets/index.tsx#L346

So, `isFetchingTokenPrices` doesn't actually do anything this in this situation. @aristidesstaffieri I'm going to roll this change back because of that.

And also just a note, this catch block only gets triggered if, for some reason, we're not able to fetch any of the user's wallets (the call to `batchedFetchBalances` itself fails). In the event just one of the wallet fetches fails, we handle that gracefully in `batchedFetchBalances` and just show the token prices for wallets we were able to

PR #2374, opened 2025-11-10
[BUG] SAC token management improvements

### LINKED ISSUE #2220: Add SACs should result in new trustline
If you search for an asset by contract address and you add it to your wallet, we will add it is a custom token. This means that we will track the ID manually and not add a trustline.
SACs should be treated as classic assets and a trustline should be added in this case.

### What did you do?
Add any classic asset by contract ID

Add trustline internal transaction dialog

### What did you see instead?
Add as custom token dialog.

aristidesstaffieri:
Closes #2220 

What
Adds SAC detection during the add/remove token flows. This results in SACs being treated as their classic counterparts(add or remove trustline) in this flow.

Why
When searching for an asset by its SAC address, you are not able to change trust on the token because it was being treated as a custom token.

### DIFF
--- a/extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrustData.tsx
+++ b/extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrustData.tsx
@@ -53,13 +53,14 @@ function useGetChangeTrustData({
     try {
       const payload = { flaggedKeys: {} } as ChangeTrustData;
 
-      const isSac = isAssetSac(
-        asset.code,
-        asset.issuer,
-        asset.contract,
-        networkDetails.networkPassphrase,
+      const isSac = isAssetSac({
+        asset: {
+          code: asset.code,
+          issuer: asset.issuer,
+          contract: asset.contract,
+        },
         networkDetails,
-      );
+      });
       if (!asset.contract || isSac) {
         const server = stellarSdkServer(
           networkDetails.networkUrl,
--- a/extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
+++ b/extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
@@ -1,6 +1,5 @@
 import React, { useMemo, useState } from "react";
 import { createPortal } from "react-dom";
-import { Networks } from "stellar-sdk";
 import { useSelector } from "react-redux";
 import { useTranslation } from "react-i18next";
 
@@ -81,13 +80,14 @@ export const ManageAssetRows = ({
 
   const shouldChangeTrust = useMemo(() => {
     if (selectedAsset && selectedAsset.contract) {
-      return isAssetSac(
-        selectedAsset.code,
-        selectedAsset.issuer,
-        selectedAsset.contract,
-        networkDetails.networkPassphrase,
+      return isAssetSac({
+        asset: {
+          code: selectedAsset.code,
+          issuer: selectedAsset.issuer,
+          contract: selectedAsset.contract,
+        },
         networkDetails,
-      );
+      });
     }
     return true;
   }, [selectedAsset, networkDetails]);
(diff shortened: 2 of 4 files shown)


extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
```diff
@@ -79,6 +79,19 @@ export const ManageAssetRows = ({
...
+    if (selectedAsset && selectedAsset.contract) {
+      return isAssetSac(
+        selectedAsset.code,
+        selectedAsset.issuer,
+        selectedAsset.contract,
+        networkDetails.networkPassphrase,
+        networkDetails,
+      );
+    }
+    return true;
```
aristidesstaffieri, line 92: this is only valid when an asset is selected, so the correct default is `true` by this conditional.

extension/src/helpers/__tests__/useGetAssetDomainsWithBalances.test.tsx
```diff
@@ -27,7 +27,21 @@ jest.mock("@shared/api/helpers/token-list", () => ({
   getCombinedAssetListData: jest.fn().mockResolvedValue([]),
 }));
 jest.mock("popup/helpers/account", () => ({
-  ...jest.requireActual("popup/helpers/account"),
+  sortBalances: (balances: any) => {
```
piyalbasu, line 30: Curious why we need to overwrite the original implementation? This looks pretty similar to the original imp but does not take into account LP's

extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrustData.tsx
```diff
@@ -52,7 +53,14 @@ function useGetChangeTrustData({
     try {
       const payload = { flaggedKeys: {} } as ChangeTrustData;
 
-      if (!asset.contract) {
+      const isSac = isAssetSac(
```
piyalbasu: NIT: can we use an object here instead of individual params? A long list of params can get cumbersome if/when we need to add more down the road. 

Also wondering if this can just be `isAssetSac({ asset, networkDetails })`

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
piyalbasu: Mind putting that in as a comment? It'll be helpful for the next dev who comes across this

extension/src/helpers/__tests__/useGetAssetDomainsWithBalances.test.tsx
aristidesstaffieri: our existing mock doesn't actually seem to be including the rest of the file with `requireActual` so it was failing to fetch balances due to missing `sortBalances` function. I mocked it because it's not part of this unit test but Im not sure why the import in that file doesn't see the mock. 🤔

extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/hooks/useChangeTrustData.tsx
aristidesstaffieri: updated in ca3584cd4e94d5f1bed4127ab77bca857d3a8ac4

extension/src/helpers/__tests__/useGetAssetDomainsWithBalances.test.tsx
piyalbasu: Oh that is weird. I wonder if there's some competing mocks colliding or something

piyalbasu (approved):

extension/src/popup/components/manageAssets/ManageAssetRows/index.tsx
CassioMG: +1

PR #2381, opened 2025-11-13
load backend settings async on Account view

### LINKED ISSUE #2380: Make backend settings load async on the Account view
Currently, before we render the UI in the Account view, we load the user's settings. Most of these settings are saved locally in local storage and resolve quite quickly. However, we also load a couple of settings from Freighter BE (feature flags and user notification). While these are 2 small API requests, they do delay UI rendering significantly

_Acceptance Criteria_

On the Account view, load these BE settings async so we can render the UI quickly

piyalbasu:
Closes #2380 

_What_

Load BE settings async (rather than loading them alongside user settings before we render) on the Account view

_Why_

Currently, before we render the UI in the Account view, we load the user's settings. Most of these settings are saved locally in local storage and resolve quite quickly. However, we also load a couple of settings from Freighter BE (feature flags and user notification). While these are 2 small API requests, they do delay UI rendering significantly

_TODO_

* We need to re-enable RPC health check so we can automatically load an RPC warning if the RPC is down 
* Move feature flags to Amplitude. The problem with having feature flags set in the Freighter BE is we need to do a BE release if we want to make an update. As we saw with the recent RPC outage, this isn't always possible if we're working on something else with our backend infra

_Videos_
Because these notifications now load async, I added an animation to make these pop in a bit less jarring. @sdfcharles Let me know what you think!

https://github.com/user-attachments/assets/5c849a92-3a37-4c5d-98bc-0f28d0ea439f

And now that we don't need to wait for these API calls to finish, loading the Account view now takes about 0.5 seconds (!!!!!!!)

https://github.com/user-attachments/assets/c70b6a8b-c4a8-40b1-911d-fb9ff0bd9e05

### DIFF
--- a/extension/src/popup/views/__tests__/Account.test.tsx
+++ b/extension/src/popup/views/__tests__/Account.test.tsx
@@ -310,7 +310,6 @@ describe("Account view", () => {
   });
 
   it("loads accounts", async () => {
-    console.log("loads accounts");
     render(
       <Wrapper
         routes={[ROUTES.account]}
@@ -347,9 +346,6 @@ describe("Account view", () => {
     ).toHaveLength(0);
   });
   it("should show user notification if user notification is enabled", async () => {
-    console.log(
-      "should show user notification if user notification is enabled",
-    );
     jest.spyOn(ApiInternal, "loadBackendSettings").mockImplementationOnce(() =>
       Promise.resolve({
         isSorobanPublicEnabled: true,


extension/src/helpers/hooks/useGetAppData.tsx
```diff
@@ -80,12 +88,20 @@ function useGetAppData() {
         return payload;
       }
 
+      let backendSettings = {};
+
+      if (useBackendSettings) {
+        backendSettings = await loadBackendSettings();
+        reduxDispatch(saveBackendSettingsAction(backendSettings));
+      }
```
piyalbasu, line 96: currently, I'm only making this change in the Account view as this is the most frequently reloaded page. In other views, it will still use the old model of loading ALL settings before we render as that's a much more niche scenario

extension/src/helpers/hooks/useGetAppData.tsx
```diff
@@ -43,7 +48,10 @@ function useGetAppData() {
   const currentAccount = useSelector(accountSelector);
   const currentSettings = useSelector(settingsSelector);
 
-  const fetchData = async (useCache = true): Promise<AppData | Error> => {
+  const fetchData = async (
+    useCache = true,
+    useBackendSettings = true,
```
aristidesstaffieri, line 53: wondering if this should default to false since these settings are exclusively used on the account page?

piyalbasu: I think what you're asking is if Account should be setting a flag since it's the only view skipping this call. Is that correct? That is what's happening here.

The idea here is by default, fetch the backend settings now. If you want to _skip_ this, you specifically set the flag to false. The only caller that sets this to `false` is the Account view. All other views don't pass a param and it defaults to fetching the BE settings now (which is the original implementation)

aristidesstaffieri: well my question was really - can we skip fetching backend settings in all other views altogether? Afaict only the accounts view actually shows a notification using the settings.

piyalbasu: Oh okay I see. Today, we technically could. Really the main danger here is the `settings` redux slice not accurately reflecting the state of the BE if you reloaded the app on a non-Account view (we do check `isSorobanSuported` in a couple of other views).

I guess my gut reaction is: I'm not sure I see a compelling reason to remove this from other views and potentially get ourselves into a situation where `settings` are partially loaded on app load. I'm leaning towards keeping this future-proof.

But i could be convinced otherwise!

PR #2389, opened 2025-11-18
Add error handling for Search Asset results

### LINKED ISSUE #2339: Token List does not load on GitHub 429 Error (infinite Loading state)
Freighter Extension v5.35.3
Chrome Version 141.0.7390.108 (Official Build) (arm64)

Try to search for many different tokens. At some point the tokenList.json fetched from GitHub will return a 429 (Too Many Requests)

Image

Image

The search does not timeout and does not show an error. If keeps on an infinite loop

1. Either timeout and show an error to the user after X seconds
2. Ignore the tokenList.json and return data from the search, without the soroswap https://raw.githubusercontent.com/soroswap/token-list/main/tokenList.json response (?)

leofelix077:
Closes #2339 

429 is hard to reproduce, as it depends on GitHub returning the 429 or not , so the network error was mocked to be cancelled to simulate the error state and early return + Unit tests added for the scenarios

https://github.com/user-attachments/assets/9d486c3b-9896-4dd8-aaca-e05b60d160aa

https://github.com/user-attachments/assets/f99856e6-77df-4661-92c3-d1f65ce732e6

### DIFF
--- a/extension/src/popup/views/SignTransaction/hooks/__tests__/useGetSignTxData.test.tsx
+++ b/extension/src/popup/views/SignTransaction/hooks/__tests__/useGetSignTxData.test.tsx
@@ -55,9 +55,16 @@ jest
   .spyOn(AccountHelpers, "signFlowAccountSelector")
   .mockReturnValue(mockAccounts[0]);
 
-const GetCombinedAssetListDataSpy = jest
-  .spyOn(TokenListHelpers, "getCombinedAssetListData")
-  .mockResolvedValue([]);
+jest.spyOn(TokenListHelpers, "getCombinedAssetListData").mockResolvedValue([
+  {
+    name: "Test Asset List",
+    description: "Test description",
+    network: "testnet",
+    version: "1.0.0",
+    provider: "test",
+    assets: [],
+  },
+]);
 
 jest.mock("react-router-dom", () => ({
   ...jest.requireActual("react-router-dom"),
@@ -80,7 +87,16 @@ describe("useGetSignTxData", () => {
     balances: mockBalances.balances,
   };
 
-  const tokenListData = [{ id: "example-token-list" }];
+  const tokenListData = [
+    {
+      name: "Example Asset List",
+      description: "Example description",
+      network: "testnet",
+      version: "1.0.0",
+      provider: "example",
+      assets: [],
+    },
+  ];
   const canonicalKey = "XLM:GABCDEF";
   const cachedIcons = {
     [canonicalKey]: "https://cached/icon/url.png",
@@ -176,7 +192,7 @@ describe("useGetSignTxData", () => {
       await result.current.fetchData();
     });
 
-    expect(GetCombinedAssetListDataSpy).toHaveBeenCalled();
+    // Verify that asset lists data was used (icons populated indicates asset lists data length > 0)
     // @ts-ignore
     expect(result.current.state.data?.icons).toEqual({
       [TEST_CANONICAL]: "https://example.com/icon.png",
@@ -219,7 +235,7 @@ describe("useGetSignTxData", () => {
       await result.current.fetchData();
     });
 
-    expect(GetCombinedAssetListDataSpy).not.toHaveBeenCalled();
+    // Verify that asset lists data was not used (no icons indicates asset lists data length = 0 or not fetched)
     // @ts-ignore
     expect(result.current.state.data?.icons).toEqual({});
     // @ts-ignore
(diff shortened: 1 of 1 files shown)


piyalbasu:
@leofelix077 Thanks for looking at this. I think a more complete fix for this would be to use cached asset list data rather than continuing to re-fetch asset list data on every search query and handling inevitable 429 errors.

Looks like the call to lookup asset lists originates at `useAssetLookup`. There, you should be able to grab the cached asset list from `popup/ducks/cache` and pass it down to the `getAssetLists` helper. Would you mind taking a look at that?

.github/workflows/deployFreighterApiBeta.yml
```diff
@@ -28,12 +29,12 @@ jobs:
       - name: Version npm package
         run: |
           cd @stellar/freighter-api
-          yarn version --new-version ${{ github.event.inputs.release }}-beta.${{ github.event.inputs.version }}
```
leofelix077: Same from https://github.com/stellar/freighter/pull/2400 -> Was initially branched from main. Didn't want to merge main again into 5.36.0 as it's on late stages for release

The code files are related to the PR, but not the ones on @stellar shared packages or package.json changes

extension/package.json
```diff
@@ -37,7 +37,6 @@
     "@types/react-redux": "7.1.34",
     "@types/react-router-dom": "5.3.3",
     "@types/redux": "3.6.31",
-    "@types/testing-library__jest-dom": "6.0.0",
```
leofelix077, line 40: deprecated was causing tsc issues

extension/src/popup/helpers/searchAsset.ts
```diff
@@ -113,16 +127,12 @@ export const getVerifiedTokens = async ({
...
-      for (const record of list) {
-        const regex = new RegExp(contractId, "i");
-        if (record.contract && record.contract.match(regex)) {
-          verifiedToken = record;
-          verifiedLists.push(data.name);
-          break;
-        }
-      }
+    const matchingRecord = data.assets?.find(
+      (record) => record.contract === contractId,
```
piyalbasu: the above ignores case for the contract ID while the new line explicitly expects the same exact string. We should preserve this to guard against cases where (for ex) a user forgets to capitalize a letter of a contract ID. A very edge case - but unless there's a strong reason to remove this, let's maintain it

extension/src/popup/helpers/__tests__/searchAsset.test.js
```diff
@@ -85,6 +143,6 @@ describe("searchAsset", () => {
     expect(assets).toStrictEqual([]);
 
     // error for missing `name` and error for additional key `title`
-    expect(errors).toHaveLength(2);
+    expect(errors).toHaveLength(3);
```
piyalbasu, line 146: Looks like the 3rd error is for the extra key `feedback`. Mind adding that to the comment on ln 145?

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -152,23 +157,25 @@ function useGetSignTxData(
...
             const code = diff.asset.code;
             let canonical = getCanonicalFromAsset(code, key);
-            if (cachedIcons[canonical]) {
-              icons[canonical] = cachedIcons[canonical];
+            const cachedIcon = cachedIcons[canonical];
+            if (cachedIcon) {
+              icons[canonical] = cachedIcon;
             } else {
-              let icon = await getIconUrlFromIssuer({
+              let icon: string | null = await getIconUrlFromIssuer({
```
piyalbasu: why change the type here? `getIcornUrlFromIssuer` can only return a string

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -130,10 +131,14 @@ function useGetSignTxData(
       dispatch({ type: "FETCH_DATA_SUCCESS", payload: firstRenderPayload });
 
       // Add all icons needed for tx assets
-      const icons = {} as { [code: string]: string | null };
-      let assetsListsData: AssetListResponse[] = [];
+      const icons = {} as { [code: string]: string };
+      // Initialize with cached lists, but we'll fetch fresh data when needed for icon lookups
+      let assetsListsData: AssetListResponse[] = cachedTokenLists || [];
+      let hasFetchedAssetLists = false;
```
piyalbasu: should this be `hasFetchedAssetLists = assetsListsData.length > 0`? Looks like you could already have the assetLists if `cachedTokenLists` is populated

@shared/api/helpers/token-list.ts
```diff
@@ -53,10 +53,17 @@ export const schemaValidatedAssetList = async (
...
   networkDetails,
   assetsLists,
+  cachedAssetLists,
 }: {
   networkDetails: NetworkDetails;
   assetsLists: AssetsLists;
+  cachedAssetLists?: AssetListResponse[];
 }) => {
+  // If cached asset lists are provided and not empty, use them instead of fetching
+  if (cachedAssetLists?.length) {
```
piyalbasu, line 63: can we add a test for this?

leofelix077: thought I had replied to your comments before, just saw it now that I didn't. but added some tests now

extension/src/popup/helpers/__tests__/searchAsset.test.js
leofelix077: added the comment ✅

leofelix077:
@piyalbasu forgot to reply earlier when I did the adjustments from your comments, but it should all be updated now/fixed the last minor points

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
leofelix077: yep yep. we can use it directly

PR #2394, opened 2025-11-20
Dismissable App Promo Banner

leofelix077:
I added it as a storage setting, so it's shown only once per user, and depends on the extension store, so only shown on the pop-up version, but depending on the direction we want to go, can add it as a "session only" banner without memory if it has been dismissed or not and/or make it available on both the full-screen and the popup versions

⚠️ - Not currently available for full-screen, just the mockup before implementing the store restrictions:
Screenshot 2025-11-19 at 16 27 10

Screenshot 2025-11-19 at 16 27 14

https://github.com/user-attachments/assets/82c27c49-b931-4bfa-8a56-c75dad678e54

### DIFF
--- a/extension/e2e-tests/addAsset.test.ts
+++ b/extension/e2e-tests/addAsset.test.ts
@@ -192,9 +192,7 @@ test("Adding classic asset on Testnet", async ({ page, extensionId }) => {
   await page.getByTestId("BackButton").click();
   await expect(page.getByText("Your assets")).toBeVisible();
   await expect(page.getByTestId("ManageAssetCode")).toHaveText("USDC");
-  await expect(page.getByTestId("ManageAssetDomain")).toHaveText(
-    "Stellar Network", // It shows Stellar Network after adding the asset on 5.36.0 release
-  );
+  await expect(page.getByTestId("ManageAssetDomain")).toHaveText("centre.io");
   await page.getByTestId("ManageAssetRowButton__ellipsis-USDC").click();
   await page.getByText("Remove asset").click();
   await expect(


extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useState, useEffect } from "react";
+import { Icon } from "@stellar/design-system";
+import { Text } from "@stellar/design-system";
```
piyalbasu: NIT: could you combine these imports into 1 line?

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from "background/helpers/dataStorageAccess";
+import { openTab } from "popup/helpers/navigate";
+import FreighterLogo from "popup/assets/logo-freighter-shadow.png";
+
+import "./styles.scss";
+
+const STORAGE_KEY = "mobileAppBannerDismissed";
+
+// Only create store if browserLocalStorage is available
+const localStore = browserLocalStorage
```
aristidesstaffieri: we tend to contain interactions with storage to the background script by implementing store methods or helpers(I know its not necessary but it is a pattern in this repo). Can we continue this pattern here?

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  ? dataStorageAccess(browserLocalStorage)
+  : null;
```
piyalbasu: This technically isn't wrong - but we generally follow a pattern of keeping interactions with the extension's data store in the `background` folder. This allows us to keep UI concerns and data storage concerns separate. That way, it's easier to keep track of what is making changes to our stored data (this is important because we keep sensitive data, like encrypted private keys, in our data store. We want to make sure requests to write data to our store are all handled in the same place to prevent collisions)

You can use the example of `addTokenId` as a model for how to send request to write to the data store:

- the UI calls `addTokenId`, which sends a message to the background script: https://github.com/stellar/freighter/blob/master/%40shared/api/internal.ts#L1704

- The background script picks up the request: https://github.com/stellar/freighter/blob/master/extension/src/background/messageListener/popupMessageListener.ts#L395

- The background script hands off the request to a handler that writes data to the store: https://github.com/stellar/freighter/blob/master/extension/src/background/messageListener/handlers/addTokenId.ts

Does that make sense? Happy to hop on a call and talk through it if not!

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: oh yes. I'm used to this being done automatically on save. didn't even notice it was in two lines

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: oh yes. it does. thanks for explaining. it's good for me to get a better hang on how things are done on the extension. it's very helpful!

Will take a look by myself first to try and understand it. if I get block I can ping you for sure. thanks a lot

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: will do. hopping on @piyalbasu 's comment as well. will check on how it's being done on the other places and follow the pattern

extension/src/background/messageListener/handlers/dismissMobileAppBanner.ts
```diff
@@ -0,0 +1,11 @@
+import { DataStorageAccess } from "background/helpers/dataStorageAccess";
+
+const STORAGE_KEY = "mobileAppBannerDismissed";
```
piyalbasu: You can actually store this in `/extension/src/constants/localStorageTypes`. Then, you can reuse this const in `getMobileAppBannerDismissed`, as well

extension/src/background/messageListener/handlers/dismissMobileAppBanner.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const dismissMobileAppBanner = async ({
+  localStore,
+}: {
+  localStore: DataStorageAccess;
+}): Promise<void> => {
+  await localStore.setItem(STORAGE_KEY, true);
```
piyalbasu: it'd be nice to have this function return `{isDismissed: localStore.getItem(STORAGE_KEY)}`. Then, in the UI, you can use the return value from this function to set the UI state. This makes sure that the value was correctly set in `localStore` and we don't accidentally have a mismatch in the UI state and the `localStore`

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  dismissMobileAppBanner,
+} from "@shared/api/internal";
+
+import { openTab } from "popup/helpers/navigate";
+import FreighterLogo from "popup/assets/logo-freighter-shadow.png";
+
+import "./styles.scss";
+
+// Check if browser storage is available (extension context, not fullscreen)
+const isStorageAvailable = (): boolean => {
```
piyalbasu: I think this check is a holdover from the last version of this PR, right? You shouldn't need to check this anymore. Whether you're in extension mode or in fullscreen mode, you should have access to the background storage now.

You should be able to safely remove this

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [isDismissed, setIsDismissed] = useState<boolean | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const checkDismissedStatus = async () => {
+      try {
+        const dismissed = await getMobileAppBannerDismissed();
+        setIsDismissed(dismissed);
+      } catch (error) {
+        console.error("Error checking banner dismissal status:", error);
```
piyalbasu: let's use `captureException` here to report this error to Sentry so we can have visibility if there's an issue

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }, []);
+
+  const handleDismiss = async (e: React.MouseEvent) => {
+    e.stopPropagation();
+    try {
+      await dismissMobileAppBanner();
+      setIsDismissed(true);
+    } catch (error) {
+      console.error("Error dismissing banner:", error);
```
piyalbasu: similar to above comment - let's use `captureException`

extension/src/popup/components/account/MobileAppBanner/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      setIsLoading(false);
+    }
+  }, []);
+
+  const handleDismiss = async (e: React.MouseEvent) => {
+    e.stopPropagation();
+    try {
+      await dismissMobileAppBanner();
+      setIsDismissed(true);
```
piyalbasu: Related to this comment: https://github.com/stellar/freighter/pull/2394/files#r2571903580

`dismissMobileAppeBanner` should return `{isDismissed: true}` and you can use that value in `setIsDismissed` to make sure there isn't a possible mismatch between the UI and the background storage

extension/src/background/messageListener/handlers/dismissMobileAppBanner.ts
leofelix077: moved to the local consts now. thanks 💪

extension/src/background/messageListener/handlers/dismissMobileAppBanner.ts
leofelix077: oh good one. adjusted now to have the return as part of it

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: oh yes. it was a leftover. removed it

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: moved it all to be captureException for sentry

extension/src/popup/components/account/MobileAppBanner/index.tsx
leofelix077: adjusted as well

extension/e2e-tests/addAsset.test.ts
```diff
@@ -192,7 +192,9 @@ test("Adding classic asset on Testnet", async ({ page, extensionId }) => {
   await page.getByTestId("BackButton").click();
   await expect(page.getByText("Your assets")).toBeVisible();
   await expect(page.getByTestId("ManageAssetCode")).toHaveText("USDC");
-  await expect(page.getByTestId("ManageAssetDomain")).toHaveText("centre.io");
+  await expect(page.getByTestId("ManageAssetDomain")).toHaveText(
```
leofelix077: this test was failing. also on 5.37.0 branch. in Prod after adding an asset, on the Manage Assets page (both on Pubnet and testnet) it shows as Stellar Network. not sure if a bug or intended, but updated heree @piyalbasu. if it's a bug, let me know, can create a separate ticket for it

leofelix077: Not sure why it's mismatching from local e2e to ci:e2e

piyalbasu: Ah, I believe your local env may not be connecting to freighter-backend v2 to fetch home domains. In your `.env` file, make sure you have `INDEXER_V2_URL=https://freighter-backend-v2.stellar.org/api/v1`

PR #2400, opened 2025-11-21
Add memo-required flows for Dapp + Normal send

leofelix077:
https://github.com/user-attachments/assets/9780ccd0-d68c-46c5-830e-9c229d14aa91

https://github.com/user-attachments/assets/b3cfc7f6-26ce-48cb-acc6-5b4273301625

https://github.com/user-attachments/assets/b80f9a61-30d7-46d9-acee-f1d1174c1ea0


extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -334,7 +351,7 @@ export const SignTransaction = () => {
                   />
                   <div className="SignTransaction__TitleRow__Detail">
                     <span className="SignTransaction__TitleRow__Title">
-                      Confirm Transaction
+                      {t("Confirm transaction")}
```
leofelix077: I noticed many strings are set as loose texts without proper translations. I was confused for a bit if it was a problem with my browser or not, before noticing it. I updated some of them, at least on the files I touched during this ticket, but created a separate issue https://github.com/stellar/freighter/issues/2401 to go through all the files and make sure the extension has proper translation texts

.github/workflows/deployFreighterApiBeta.yml
```diff
@@ -28,12 +29,12 @@ jobs:
       - name: Version npm package
         run: |
           cd @stellar/freighter-api
-          yarn version --new-version ${{ github.event.inputs.release }}-beta.${{ github.event.inputs.version }}
```
leofelix077: unrelated changes to this. Just didn't wanna merge main into 5.36.0 again, as it's on late stages of testing

extension/e2e-tests/onboarding.test.ts
```diff
@@ -369,7 +369,8 @@ test("Incorrect mnemonic phrase", async ({ page }) => {
   const shuffledWords = shuffle(words);
 
   for (let i = 0; i < shuffledWords.length; i++) {
-    await page.getByLabel(shuffledWords[i]).check({ force: true });
+    // Use nth() to handle duplicate labels by selecting the first matching element
+    await page.getByLabel(shuffledWords[i]).first().check({ force: true });
```
piyalbasu: Hm, does this work in the case of duped words? I would think from this code it would never select the 2nd instance of a word so if the phrase was "foo bar bar", it would end up just selecting the first instance of "bar" and it would try to validate "foo bar", which would fail

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,271 @@
...
+    };
+
+    fetchMemoRequiredAccounts();
+  }, [shouldValidateMemo, activePublicKey]);
+
+  /**
+   * Effect to parse XDR and set initial memo validation state
+   * Runs when XDR, network, or validation settings change
+   */
+  useEffect(() => {
```
piyalbasu, line 174: I think you should be able to consolidate these next 2 `useEffect` calls - they seem duplicative. You should be able to just call `checkIsMemoRequired  after you build the xdr and set `hasMemo`

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,271 @@
...
+    // Reset validation state when XDR changes to ensure fresh validation
+    setIsMemoMissing(true);
+    setIsValidatingMemo(false);
+
+    if (!shouldValidateMemo) {
```
piyalbasu: NIt: I think we can also combine these next 2 if statements - `if (!shouldValidateMemo || (!xdr || !networkDetails))` since they do the same thing

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -386,13 +407,26 @@ export const SignTransaction = () => {
...
+                      <span>Memo</span>
+                    </div>
+                    <div className="SignTransaction__Metadata__Value">
+                      <span>
+                        {decodedMemo && decodedMemo.value
+                          ? decodedMemo.value
+                          : t("None")}
+                      </span>
+                    </div>
+                  </div>
```
piyalbasu, line 407: I think this makes sense - but the Figma design actually doesn't include the memo here. It shows it on the Transaction Details view. Let's check with @sdfcharles to see what he thinks about adding the memo here

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -579,9 +579,13 @@ export const SendAmount = ({
...
               memo={transactionData.memo || ""}
               onClose={() => setIsEditingMemo(false)}
-              onSubmit={({ memo }: { memo: string }) => {
+              onSubmit={async ({ memo }: { memo: string }) => {
                 dispatch(saveMemo(memo));
                 setIsEditingMemo(false);
+                // Regenerate transaction XDR with new memo (now reads memo from Redux state inside fetchData)
+                await fetchSimulationData();
+                // Reopen review sheet after memo is saved and XDR is regenerated
+                setIsReviewingTx(true);
```
piyalbasu: Hm, why automatically open this review sheet here? As a user, I would expect this to come up only after I click `Review Send`. It's also a bit jarring because editing the fee doesn't open the review sheet - it's just editing the memo.

I think we should maintain the original UX of the user needing to click `Review Send` to get to the review sheet

piyalbasu: Also, I'm noticing a bug where if I'm sending to a memo required address, if I enter a memo, I don't see the memo warning (great!), but if I go back and clear the memo, I can get past the memo warning

piyalbasu:
Let's also add some tests for this functionality

extension/e2e-tests/onboarding.test.ts
leofelix077: unintended change

extension/src/popup/views/SignTransaction/index.tsx
leofelix077: I keep forgetting to check the Figma files for the extension. I had the app in my mind and tried to follow the same pattern. will check there more often for the next implementations, but in case we can keep/add to it, let me know @sdfcharles

extension/src/popup/components/sendPayment/SendAmount/index.tsx
leofelix077: oh this was intended for it to open after pressing the "Add memo" on the review flow, to send the user back to the review page, not in all cases. thanks for checking it. it shouldn't open the review when adding the memo from the sendPayment page. only when adding during the review. will adjust it

extension/src/popup/helpers/useValidateTransactionMemo.ts
leofelix077: yep. I took it mostly 'as-is' from the app, but can adjust it to merge it into a single one (also on the app, later)https://github.com/stellar/freighter-mobile/issues/599. created a separate ticket for it

extension/src/popup/helpers/useValidateTransactionMemo.ts
leofelix077: this one I split more for readability, but can join them into one, as there arent many clauses

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,252 @@
...
+   * Only validates on mainnet when the feature is enabled in preferences
+   * Can be bypassed for testing/CI by setting CI=true or IS_PLAYWRIGHT=true
+   */
+  const shouldValidateMemo = useMemo(() => {
+    // Check for test/CI environment variables
+    // Note: In browser extensions, process.env may not be available at runtime
+    // We check both process.env (for build-time vars) and window (for runtime vars)
+    const isCI = process.env.CI === "true" || (window as any).CI === "true";
+    const isPlaywright =
+      process.env.IS_PLAYWRIGHT === "true" ||
```
leofelix077, line 127: not sure about this approach, because on the app we don't validate memo on testnet, but for the e2e they run on the testnet, so could either always validate it, or override it similar to here with an env flag / during e2e tests

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,252 @@
...
+      (window as any).IS_PLAYWRIGHT === "true";
+    const bypassMainnetCheck = isCI || isPlaywright;
+
+    const shouldValidate = !!(
+      isMemoValidationEnabled &&
+      (isMainnet(networkDetails) || bypassMainnetCheck)
+    );
+
+    // Debug logging in development
+    if (!shouldValidate && isMemoValidationEnabled) {
```
piyalbasu: Let's leave this out of the codebase unless there's a compelling reason to keep it

extension/src/popup/helpers/useValidateTransactionMemo.ts
piyalbasu: I think this a pretty reasonable workaround! If you did want to make this simpler, we could fund an account on Mainnet for you to run your tests against. Let me know if you want to do that. Otherwise, I think this is fine

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,252 @@
...
+
+        try {
+          // Check cache first (memo-required accounts from API)
+          const isMemoRequiredFromCache = checkMemoRequiredFromCache(
+            transaction!,
+            memoRequiredAccounts,
+          );
+
+          // Also check SDK as fallback
+          const isMemoRequiredFromSDK = await checkMemoRequiredFromStellarSDK(
```
piyalbasu, line 222: NIT: it isn't necessary to check the SDK if `isMemoRequiredFromCache ` comes back as true

extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,252 @@
...
+  networkUrl: string,
+  networkPassphrase: string,
+): Promise<boolean> => {
+  const server = stellarSdkServer(networkUrl, networkPassphrase);
+
+  try {
+    await server.checkMemoRequired(transaction as any);
+    return false;
+  } catch (e: any) {
+    if ("accountId" in e) {
```
piyalbasu, line 76: might be helpful to add a comment here explaining why this means memo required

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -141,11 +143,24 @@ export const SignTransaction = () => {
...
-  const isMemoRequired = flaggedKeyValues.some(
+  const isMemoRequiredFromFlaggedKeys = flaggedKeyValues.some(
     ({ tags }) => tags.includes(TRANSACTION_WARNING.memoRequired) && !memo,
   );
 
+  // Combine both validation methods - use hook result if available, fallback to flaggedKeys
+  // isRequiredMemoMissing matches mobile app naming convention
+  const isRequiredMemoMissing = isValidatingMemo
+    ? isMemoRequiredFromFlaggedKeys
+    : isMemoMissingFromHook || isMemoRequiredFromFlaggedKeys;
```
piyalbasu: what's the benefit of re-scanning the tx using `isMemoMissingFromHook`? Is this not effectively doing the same thing `isMemoRequiredFromFlaggedKeys` did?

extension/src/popup/helpers/useValidateTransactionMemo.ts
leofelix077: I think for now, not to have to rewrite all the tests, not being such a critical issue, I'd leave it like this, but I created a separate follow-up ticket for it https://github.com/stellar/freighter/issues/2449 so it's not forgotten

extension/src/popup/helpers/useValidateTransactionMemo.ts
leofelix077: oops

extension/src/popup/views/SignTransaction/index.tsx
leofelix077: oh yes. removed the duplicated check

PR #2405, opened 2025-11-21
Feature/collectibles home tab

### LINKED ISSUE #2325: [Collectibles] Home tab UI
Let's port this from mobile.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6322-191995&t=i6DQHvjsIvsFy5wX-1):

Image

piyalbasu:
Closes #2325 

This PR adds a few pieces of showing Collectibles in the UI:

1. It fetches collectibles from the `/collectibles` endpoint; it then filters by `owner`; then it fetches token metadata for each collectible
2. It adds a Collectibles Tab to the Account view; this view gracefully handles errors, missing images, and images with incorrect resolution

https://github.com/user-attachments/assets/a30e0278-44e1-4e28-b795-a587e6f2df06

Missing Image Placeholder:
Screenshot 2025-11-21 at 6 16 46 PM

Empty state:
Screenshot 2025-11-21 at 6 38 51 PM

### DIFF
--- a/extension/e2e-tests/loadAccount.test.ts
+++ b/extension/e2e-tests/loadAccount.test.ts
@@ -712,28 +712,27 @@ test("Loads collectibles data with successful metadata", async ({
     });
   }
 
-  expect(imgs[0]).toHaveAttribute(
-    "src",
+  expect(await imgs[0].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/events/2023/5/NeToOQbYtaJILHMnkigEAsA6ckKYe2GAA4ppAOSp.jpg",
   );
-  expect(imgs[1]).toHaveAttribute(
-    "src",
+
+  expect(await imgs[1].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/2024/06/02/pepe-the-bot_ml4cWknXFrF3K3U1.jpeg",
   );
-  expect(imgs[2]).toHaveAttribute(
-    "src",
+
+  expect(await imgs[2].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/events/2023/8/5kFeYwNfhpUST3TsSoLxm7FaGY1ljwLRgfZ5gQnV.jpg",
   );
-  expect(imgs[3]).toHaveAttribute(
-    "src",
+
+  expect(await imgs[3].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/events/2025/7/Hdqv6YNVErVCmYlwobFVYfS5BiH19ferUgQova7Z.webp",
   );
-  expect(imgs[4]).toHaveAttribute(
-    "src",
+
+  expect(await imgs[4].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/events/2025/7/MkaASwOL8VA3I5B2iIfCcNGT29vGBp4YZIJgmjzq.jpg",
   );
-  expect(imgs[5]).toHaveAttribute(
-    "src",
+
+  expect(await imgs[5].getAttribute("src")).toBe(
     "https://nftcalendar.io/storage/uploads/events/2025/3/oUfeUrSj3KcVnjColyfnS5ICYuqzDbiuqQP4qLIz.png",
   );
 });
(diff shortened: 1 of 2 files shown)


piyalbasu:
@sdfcharles Let me know how you think we should handle non-square images. I'm scaling the image down and filling the background

Screenshot 2025-11-21 at 7 01 50 PM

@shared/api/helpers/getCollectibles.ts
```diff
@@ -0,0 +1,153 @@
...
+    const data = (await response.json()) as {
+      name?: string;
+      description?: string;
+      external_url?: string;
+      image?: string;
+      attributes?: {
+        traitType?: string;
+        value?: string | number;
+      }[];
+    };
```
CassioMG:
```suggestion
    const data = (await response.json()) as CollectibleMetadata;
```

@shared/api/types/types.ts
```diff
@@ -403,3 +403,88 @@ export interface LedgerKeyAccounts {
...
+  };
+}
+
+export interface CollectiblesResponse {
+  collections: CollectionResponse[];
+}
+
+export interface Collectibles {
+  collections: Collection[];
+}
```
CassioMG, line 501: looks like we have the same interface [declared here](https://github.com/stellar/freighter/pull/2405/files#diff-c3fdcb6e15400745bccd94a024719a85c43214534df2590690dfc2f3feb43a9fR9-R11)

@shared/api/types/types.ts
```diff
@@ -403,3 +403,88 @@ export interface LedgerKeyAccounts {
...
+  image?: string;
+  externalUrl?: string;
+  attributes?: {
+    traitType?: string;
+    value?: string | number;
+  }[];
+};
+
+export interface CollectionResponse {
+  collection: {
```
CassioMG:
```suggestion
  collection?: {
```

CassioMG:
> @sdfcharles Let me know how you think we should handle non-square images. I'm scaling the image down and filling the background
> 

@piyalbasu @sdfcharles Good point, thanks for raising this. I think on mobile we are doing the opposite: expanding the image to fill the container. The rationale is that users can always tap on the "View" button to view the whole image externally. This also assumes most collectibles will have a more-or-less squared format so we don't cut a lot of the image. But if we start seeing lots of rectangle collectibles like the one you used in this PR then we should probably revisit this 🤔

CassioMG (approved):
Great work on this PR! Love the smooth animation 👏

@shared/api/helpers/getCollectibles.ts
piyalbasu: Ah, I actually can't cast this as `CollectibleMetadata` because `CollectibleMetadata` expects camelCasing while the API returns snake_case. I can create a `CollectibleMetadataReponse` interface like I have with the other API responses and use that here instead

extension/src/popup/views/Account/contexts/activeTabContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { createContext, useState } from "react";
+
+export const TabsList = ["tokens", "collectibles"] as const;
```
aristidesstaffieri, line 3: Does the v2 API still only support collectibles for pubnet? I believe the mobile app only enables the collectibles tab on pubnet today.

piyalbasu: Good question - we actually do support collectibles on every network now. In this PR, I'm appending the network so we can get collectibles from Mainnet/Testnet/Futurenet.

But I should make sure we don't show this tab on custom networks. I'll add this!

I opened a ticket for Freighter Mobile to enable Testnet: https://github.com/stellar/freighter-mobile/issues/593

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -128,14 +129,19 @@ function useGetAccountData(options: {
...
-
-      payload.collectibles = collectiblesResult;
+      if (!isCustomNetwork(networkDetails)) {
+        const collectiblesResult = await fetchCollectibles({
+          publicKey,
+          networkDetails,
+          contracts: [
+            {
+              id: "CCTYMI5ME6NFJC675P2CHNVG467YQJQ5E4TWP5RAPYYNKWK7DIUUDENN",
+              token_ids: ["1", "2"],
```
CassioMG: is this hardcode data here intended?

piyalbasu: Ah, thanks for catching - I just needed this for testing. Removing

PR #2417, opened 2025-11-27
Add CAP-67/Mux Address support

### LINKED ISSUE #2315: Add Memo/CAP-67 support for Soroban Transactions/Mux adresses
Latest

### What did you do?
tried to make a transfer using a soroban operation

Success

### What did you see instead?
Failed, dude to adding a memo which is not supported.

leofelix077:
Closes #2315 

https://github.com/user-attachments/assets/330b10d0-30cc-4636-9c7c-0dd9c4cc23d3

https://github.com/user-attachments/assets/7b40f050-2e3f-4825-8bcc-2e61593018b1

https://github.com/user-attachments/assets/b7b8beec-da24-4849-8b0a-53b0041bdc41

https://github.com/user-attachments/assets/58acd19c-7517-4117-87c1-52313f76bf05

https://github.com/user-attachments/assets/58f76c9e-ff22-4cf8-851c-2fa6fb435af0


extension/src/helpers/muxedAddress.ts
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * A contract supports muxed addresses if the transfer function has a `to_muxed` parameter.
+ * This is the CAP-0067 pattern where the transfer signature is:
+ * transfer(from: Address, to_muxed: MuxedAddress, amount: i128)
+ *
+ * @param contractId - The contract ID to check
+ * @param networkDetails - Network details
+ * @returns Promise resolving to true if contract supports muxed addresses. false otherwise
+ */
+export async function checkContractMuxedSupport(
```
piyalbasu: NIT: since this is returning a boolean, would you mind using  the `is_` naming convention - something like `checkIsMuxedSupported`?

extension/src/helpers/muxedAddress.ts
```diff
@@ -0,0 +1,205 @@
...
+    const required = transferDef.properties?.args?.required;
+
+    // Check if 'to_muxed' parameter exists in properties (CAP-0067 pattern)
+    // Use 'in' operator to check if key exists, regardless of value
+    if (argsProperties && "to_muxed" in argsProperties) {
+      return true;
+    }
+
+    // Also check if to_muxed is in the required array (even if property doesn't exist)
+    if (Array.isArray(required) && required.includes("to_muxed")) {
```
piyalbasu, line 68: Just curious - where does this requirement come from? Looking at CAP-0067, I didn't see this, but maybe I missed it

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -350,6 +443,19 @@ export const SendAmount = ({
...
         <div className="SendAmount">
+          {isMuxedAddressWithoutMemoSupport && (
+            <div className="SendAmount__warning-banner">
+              <Notification
+                variant="error"
+                icon={<Icon.AlertCircle />}
+                title={t("Muxed address not supported")}
+              >
+                {t(
+                  "This token does not support muxed address (M-) as a target destination. Proceeding with this transaction could result in loss of the token.",
```
piyalbasu: NIT: We actually can't proceed with this transaction because the `Review Send` button is disabled. Should we remove the sentence `Proceeding with this transaction could result in loss of the token.`? Wdyt?

extension/src/popup/components/sendPayment/styles.scss
```diff
@@ -65,6 +65,10 @@
     }
   }
 
+  &__warning-banner {
+    padding: 16px;
```
piyalbasu: NIT: we're still in the process making this change everywhere, but we're trying to use the `pxToRem` helper for all px values. So, this style would become `padding: pxToRem(16px)`

extension/src/popup/helpers/soroban.ts
```diff
@@ -26,8 +26,96 @@ import { findAssetBalance, isSorobanBalance } from "./balance";
...
+ * - Contract address for Soroban tokens
+ * - "SYMBOL:CONTRACTID" for Soroban tokens with symbol
+ *
+ * @param tokenId - The token identifier string
+ * @returns Contract ID if the token is a Soroban token, undefined otherwise
+ */
+export const getContractIdFromTokenId = (
+  tokenId: string,
+): string | undefined => {
+  if (tokenId === "native") {
```
piyalbasu, line 98: native actually has a contract id and we have a helper to retrieve it:
```
import {
  getNativeContractDetails,
} from "popup/helpers/searchAsset";
```

extension/src/helpers/muxedAddress.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  // Disable memo for all M addresses (memo is encoded in the address)
+  if (isMuxedAccount(targetAddress)) {
+    return {
+      isMemoDisabled: true,
+      memoDisabledMessage: t("Memo is disabled for this transaction"),
+    };
+  }
+
+  if (!contractId || !networkDetails) {
```
piyalbasu: just curious - why are we checking `networkDetails` in relation to memo? It seems unrelated to the rest of the logic here

extension/src/helpers/muxedAddress.ts
```diff
@@ -0,0 +1,205 @@
...
+ * @param params Parameters for determining muxed destination
+ * @returns The final destination address (may be muxed if memo was provided and contract supports it)
+ */
+export function determineMuxedDestination(
+  params: DetermineMuxedDestinationParams,
+): string {
+  const { recipientAddress, transactionMemo, contractSupportsMuxed } = params;
+
+  const isRecipientGAddress =
+    isValidStellarAddress(recipientAddress) && !isContractId(recipientAddress);
```
piyalbasu: `isValidStellarAddress` will actually return true for a M address or a federated address - so this actually isn't necessarily a G address

extension/src/helpers/stellar.ts
```diff
@@ -107,7 +115,208 @@ export const formatDomain = (domain: string) => {
...
+ * Determines if two Stellar addresses refer to the same account
+ * Only considers direct match between Ed25519 public keys
+ * Doesn't consider federation addresses or contract IDs to be the same account
+ * For muxed accounts, compares their base G addresses
+ *
+ * @param address1 First address to compare
+ * @param address2 Second address to compare
+ * @returns True if addresses refer to the same account
+ */
+export const isSameAccount = (address1: string, address2: string): boolean => {
```
piyalbasu: is this helper being used? I didn't see it being called anywhere

extension/src/popup/components/InternalTransaction/EditMemo/index.tsx
```diff
@@ -39,19 +49,27 @@ export const EditMemo = ({ memo, onClose, onSubmit }: EditMemoProps) => {
...
-                    What is this transaction for? (optional)
-                  </div>
+                  {disabled && disabledMessage && (
+                    <div className="EditMemo__description EditMemo__description--warning">
+                      {disabledMessage}
+                    </div>
+                  )}
+                  {!disabled && (
+                    <div className="EditMemo__description">
+                      What is this transaction for? (optional)
```
piyalbasu: NIT: let's translate this

extension/src/popup/components/sendPayment/SendAmount/hooks/useSimulateTxData.tsx
```diff
@@ -380,18 +387,60 @@ function useSimulateTxData({
...
+      // For Soroban transfers, check if contract supports muxed and determine final destination
+      let finalDestination = destination;
+      let sorobanMemo = memo;
+      if (simParams.type === "soroban") {
+        try {
+          const contractSupportsMuxed = await checkContractMuxedSupport({
+            contractId: tokenAddress,
+            networkDetails,
+          });
+          finalDestination = determineMuxedDestination({
```
piyalbasu: I think this can be in the `else` statement because we don't need to determine the muxed destination unless `contractSupportsMuxed `, right?

extension/src/popup/components/sendPayment/SendAmount/index.tsx
```diff
@@ -111,6 +122,87 @@ export const SendAmount = ({
...
+      sendAmountData.data.type === AppDataType.REROUTE ||
+      !("userBalances" in sendAmountData.data)
+    ) {
+      return undefined;
+    }
+    const sendData = sendAmountData.data;
+    const assetBalance = findAssetBalance(
+      sendData.userBalances.balances,
+      getAssetFromCanonical(asset),
+    );
```
piyalbasu: this doesn't seem necessarily wrong to me - but curious why we can't just split the `asset` on `:` and get `assetPart[1]` to get the contract id?

piyalbasu: You could use your `getContractIdFromTokenId` helper here

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -423,8 +494,15 @@ export const getRowDataByOpType = async (
   }
 
   if (isPayment) {
+    // Extract destination from XDR to get muxed address if present
+    const actualDestination = await extractDestinationFromXDR(
```
piyalbasu, line 499: let's add some tests to make sure we're properly rendering the history row

extension/src/helpers/muxedAddress.ts
leofelix077: this is coming from the contract-spec returned object. on the tests we did for the app, the param definition is returning as undefined, so we check for the required and to_muxed on the transfer, to check if the properties mention the support for muxed address or not

```
 "transfer": {
                "properties": {
                    "args": {
                        "additionalProperties": false,
                        "properties": {
                            "from": {
                                "$ref": "#/definitions/Address"
                            },
                            "to_muxed": {
                                "$ref": "#/definitions/undefined"
                            },
                            "amount": {
                                "$ref": "#/definitions/I128"
                            }
                        },
                        "type": "object",
                        "required": [
                            "from",
                            "to_muxed",
                            "amount"
                        ]
                    }
                },
```

extension/src/popup/components/sendPayment/SendAmount/index.tsx
leofelix077: I left like this to keep the same message as the app, but we could remove in both. @sdfcharles  wdyt?

extension/src/popup/components/sendPayment/styles.scss
leofelix077: ah good to know. thanks! adjusting here as well

extension/src/helpers/muxedAddress.ts
leofelix077: oh many of these were taken directly from the app. there the network details as marked as optional, so just a guard check for linting reasons, that should never be reached. but removed here now, as it's not needed on the extension

extension/src/popup/components/sendPayment/SendAmount/index.tsx
leofelix077: true. on the app we also just split it

extension/src/helpers/stellar.ts
leofelix077: oh no. I took a lot from the app code, but forgot to cleanup the unused parts

extension/e2e-tests/sendPayment.test.ts
```diff
@@ -640,208 +640,354 @@ test("Send token payment to C address", async ({ page, extensionId }) => {
   expect(accountBalancesRequestWasMade).toBeTruthy();
 });
 
-test("SendPayment persists amount and asset when navigating to choose address", async ({
+// Classic token to G address -> Normal (regression test)
```
piyalbasu, line 894: looks like there was a bad merge here and overwrote some tests that were here. I'd suggest using `git rebase` instead of merging the release branch. It will give you a cleaner commit history and help you catch issues like this

extension/src/popup/components/InternalTransaction/EditMemo/index.tsx
```diff
@@ -16,86 +16,95 @@ interface EditMemoProps {
...
       <div className="EditMemo">
         <Card>
-          <p>{t("Memo")}</p>
-          <Formik
-            initialValues={initialValues}
-            onSubmit={handleSubmit}
-            enableReinitialize
-          >
-            {renderForm}
+          <p>Memo</p>
```
piyalbasu: missing translation

extension/src/popup/metrics/access.ts
```diff
@@ -14,47 +14,58 @@ import { registerHandler, emitMetric, MetricsData } from "helpers/metrics";
...
-});
-registerHandler<AppState>(addToken.fulfilled, () => {
-  const metricsData: MetricsData = JSON.parse(
-    localStorage.getItem(METRICS_DATA) || "{}",
-  );
-  emitMetric(METRIC_NAMES.addToken, {
-    accountType: metricsData.accountType,
+// Defer registration to avoid circular dependency issues during module initialization
+// Register handlers after a microtask to ensure all modules are fully loaded
+Promise.resolve().then(() => {
```
piyalbasu, line 19: hm, how was this presenting itself when you ran your code? I haven't seen anyone else run into this issue, so I'm wondering if this is actually an issue in the codebase or with your setup

leofelix077: can you try running in your machine from scratch? not directly related to this PR, but I noticed sometimes after building and running `yarn start`, when accessing localhost:9000, it would just load infinitely. Also going on the extension side for the "load unpacked" it was throwing `background services error `, without ever loading the home screen

Adding these changes seem to have stopped the issue, that's why I added it here

PR #2420, opened 2025-12-01
[BUG] fixes settings state persistence for asset selection in send and swap flows

### LINKED ISSUE #2419: settings should persist all options( send & swap )
5.36

### What did you do?
Choose USDC.
Choose amount
Choose address -

Now the asset changes to XLM

asset selection retained

### What did you see instead?
asset selection wiped

aristidesstaffieri:
Closes #2419 

Updates the state management of settings related to the send and swap transactions to persist through configuration.

### DIFF
--- a/extension/src/popup/components/sendPayment/SendAmount/index.tsx
+++ b/extension/src/popup/components/sendPayment/SendAmount/index.tsx
@@ -26,6 +26,7 @@ import {
   transactionSubmissionSelector,
   saveAmount,
   saveAsset,
+  saveIsToken,
   saveMemo,
   saveTransactionFee,
   saveTransactionTimeout,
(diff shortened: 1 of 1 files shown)


piyalbasu:
I think I noticed one instance where we need to prevent overwriting the selected asset. If you change your asset to (for ex) USD, go back to the amount screen, then click the Asset Select tile again, but this time, don't change the asset, just click the x, it takes you back to Asset Amount and your asset is `native` again.

Looks like this `goToChooseAssetAction` helper auto sets you to `native` before you select a new asset

https://github.com/stellar/freighter/blob/682583ad591e1fde2ab89b789932b77079122da8/extension/src/popup/components/sendPayment/SendAmount/index.tsx#L268

Issue #2327, opened 2025-10-25
[Collectibles] Add Collectible UI


CassioMG:
Let's port this from mobile.

Note: we can ignore the "Where do I find these?" purple text for now.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6323-198441&t=i6DQHvjsIvsFy5wX-1):

<img width="1569" height="1250" alt="Image" src="https://github.com/user-attachments/assets/4f4b22c8-a8e8-4fb2-9074-1a9168fa8ea7" />

PR #2423, opened 2025-12-02
Feature/add collectibles

### LINKED ISSUE #2328: [Collectibles] Add functionality + localstorage
Let's port this from mobile.

This task is about implementing the add collectible functionality. We should use localstorage to persist basic data for the collectible similar to what we currently do with Sep-41 tokens while we don't have the indexer in place.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6323-198441&t=i6DQHvjsIvsFy5wX-1):

Image

piyalbasu:
Closes #2328 and #2327

This PR combines 2 tickets: implementing the UI for adding a collectible and storing/retrieving this added collectible data from extension storage.

* the UI for adding a collectible validates form fields and then adds the collectible if a proper contract address and token id are passed
* once proper params are entered, this data is saved in extension storage for that user's public key on their selected network
* when retrieving the data, Freighter will make a call to `/collectibles` to retrieve metadata and then cache that data to prevent unnecessary API calls during the session

https://github.com/user-attachments/assets/d6ece5a9-13f4-40b2-8a3c-9f9eee1dc44d


extension/public/static/manifest/v3.json
```diff
@@ -34,6 +34,6 @@
     "48": "images/icon48.png",
     "128": "images/icon128.png"
   },
-  "permissions": ["storage", "alarms"],
+  "permissions": ["storage", "alarms", "clipboardRead"],
```
piyalbasu, line 37: this is necessary to allow pasting into the input fields in the "Add Collectible" view

extension/src/background/messageListener/handlers/addCollectible.ts
```diff
@@ -0,0 +1,48 @@
...
+  localStore,
+}: {
+  request: AddCollectibleMessage;
+  localStore: DataStorageAccess;
+}) => {
+  const { network, publicKey, collectibleAddress, collectibleTokenId } =
+    request;
+
+  const collectibles = (await localStore.getItem(COLLECTIBLES_ID)) || {};
+  const networkCollectibles = collectibles[network] || {};
```
piyalbasu, line 17: when collectibles are saved, we delineate by both network and public key so (for ex) a collectible a user adds on G1 on testnet doesn't get loaded on G1 for mainnet. Similarly, a collectible added to G1 doesn't get loaded on G2

extension/src/background/messageListener/handlers/addCollectible.ts
```diff
@@ -0,0 +1,48 @@
...
+  const accountCollectibles: CollectibleContract[] =
+    networkCollectibles[publicKey] || [];
+
+  // does collectible contract already exist?
+  const collectibleContract = accountCollectibles.find(
+    (contract) => contract.id === collectibleAddress,
+  );
+  if (collectibleContract?.tokenIds.includes(collectibleTokenId)) {
+    return { error: "Collectible contract already exists" };
+  }
```
piyalbasu, line 28: let a user know if they're trying to add a collectible they've already added

extension/src/helpers/hooks/useGetCollectibles.tsx
```diff
@@ -22,49 +26,95 @@ const preloadImages = async (images: string[]) => {
   }
 };
 
-function useGetCollectibles() {
+function useGetCollectibles({ useCache = true }: { useCache?: boolean }) {
```
piyalbasu, line 29: on the Account view, we'll use the cache after the initial load to prevent re-fetching the same collectibles over and over again.

In the Add Collectibles view, we won't use the cache to force the app to fetch the newly added collectible

extension/src/helpers/hooks/useGetCollectibles.tsx
```diff
@@ -22,49 +26,95 @@ const preloadImages = async (images: string[]) => {
...
+
+    if (useCache && cachedCollectionsData) {
+      const payload = {
+        collections: cachedCollectionsData,
+      };
+      dispatch({ type: "FETCH_DATA_SUCCESS", payload });
+      return payload;
+    }
+
+    const storedCollectibles = await getCollectibles({
```
piyalbasu, line 56: we can retrieve stored collectibles directly from storage here and pass them on to the `/collectibles` endpoint

extension/src/popup/components/Tabs/index.tsx
```diff
@@ -1,33 +0,0 @@
...
-import classnames from "classnames";
-
-import "./styles.scss";
-
-interface TabsProps {
-  tabs: string[];
-  renderTab: (tab: string) => ReactNode;
-}
-
-export const Tabs = (props: TabsProps) => {
```
piyalbasu, line 11: this was an unused file with similar naming to the Tokens/Collectibles tab. Removing for clarity

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -1,38 +1,170 @@
-import React, { useContext } from "react";
+import React, { useEffect, useContext, useMemo, useState } from "react";
```
piyalbasu: these changes add the the modal for tokens/collectibles:

Screenshot 2025-12-02 at 4 12 08 PM

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -1,38 +1,170 @@
...
 
-  if (isCustomNetwork(networkDetails)) {
-    return null;
-  }
+  const isTokensTab = activeTab === 0;
+  const isCollectiblesTab = activeTab === 1;
+
+  useEffect(() => {
+    const queryParamTab = queryParams.get("tab");
+    const index = TabsList.findIndex((tab) => tab === queryParamTab);
```
piyalbasu: Adding a query param to automatically open the `collectibles` tab. This is useful to automatically show users their new collectible after they've successfully added one

extension/src/popup/views/Account/hooks/useGetAccountData.tsx
```diff
@@ -50,7 +50,9 @@ function useGetAccountData(options: {
   const { fetchData: fetchAppData } = useGetAppData();
   const { fetchData: fetchBalances } = useGetBalances(options);
   const { fetchData: fetchTokenPrices } = useGetTokenPrices();
-  const { fetchData: fetchCollectibles } = useGetCollectibles();
+  const { fetchData: fetchCollectibles } = useGetCollectibles({
+    useCache: true,
```
piyalbasu, line 54: use the cache after initially loading the collectibles as we expect the user to visit the Account view multiple times. No need to keep requesting the `/collectibles` endpoint if nothing changed

@shared/api/helpers/fetchCollectibles.ts
```diff
@@ -70,7 +70,7 @@ const fetchCollectibleMetadata = async (tokenUri: string) => {
  * });
  * ```
  */
-export const getCollectibles = async ({
+export const fetchCollectibles = async ({
```
piyalbasu, line 73: renaming this helper because I'm using `getCollectibles` to refer to getting collectibles from storage. This naming makes it more clear we're making an external fetch for this info

extension/src/background/messageListener/handlers/addCollectible.ts
```diff
@@ -0,0 +1,48 @@
...
+import { COLLECTIBLES_ID } from "constants/localStorageTypes";
+
+export const addCollectible = async ({
+  request,
+  localStore,
+}: {
+  request: AddCollectibleMessage;
+  localStore: DataStorageAccess;
+}) => {
+  const { network, publicKey, collectibleAddress, collectibleTokenId } =
```
CassioMG: nit: I'd suggest renaming `collectibleAddress` to either `collectionAddress`, `contractAddress` or `collectibleContractAddress` to make it clear it's the contract or collection address. I'm afraid `collectibleAddress` could sound a bit confusing since a same address can have multiple collectibles. Wdyt?

extension/src/background/messageListener/handlers/addCollectible.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  await localStore.setItem(COLLECTIBLES_ID, {
+    ...collectibles,
+    [network]: {
+      ...networkCollectibles,
+      [publicKey]: accountCollectibles,
+    },
+  });
+
+  return { collectiblesList: await localStore.getItem(COLLECTIBLES_ID) };
```
CassioMG: we could probably get rid of 1 `await` call here in case we extract the below to a const and just return it on line 47:
```
{
    ...collectibles,
    [network]: {
      ...networkCollectibles,
      [publicKey]: accountCollectibles,
    },
  }
```

@shared/api/helpers/fetchCollectibles.ts
CassioMG: +1

@shared/api/types/message-request.ts
```diff
@@ -352,6 +352,20 @@ export interface GetHiddenAssetsMessage extends BaseMessage {
   type: SERVICE_TYPES.GET_HIDDEN_ASSETS;
 }
 
+export interface AddCollectibleMessage extends BaseMessage {
+  type: SERVICE_TYPES.ADD_COLLECTIBLE;
+  network: string;
+  publicKey: string;
+  collectibleAddress: string;
```
CassioMG: same suggestion from the other comment to rename the `collectibleAddress` param

extension/public/static/manifest/v3.json
CassioMG: just curious, why is it needed for the "Add Collectible" input but not for the "Send" address input?

extension/src/helpers/hooks/useGetCollectibles.tsx
CassioMG: nice call

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -1,38 +1,170 @@
...
+          </div>
+          <div className="AccountTabs__modal__item__title">
+            {t("Add a token")}
+          </div>
+        </div>
+      </Link>
+      <Link to={ROUTES.manageAssets}>
+        <div className="AccountTabs__modal__item">
+          <div className="AccountTabs__modal__item__icon">
+            <Icon.EyeOff />
```
CassioMG: is this icon correct or just a placeholder?

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -1,38 +1,170 @@
...
+  const queryParams = useMemo(
+    () => new URLSearchParams(location.search),
+    [location.search],
+  );
 
-  if (isCustomNetwork(networkDetails)) {
-    return null;
-  }
+  const isTokensTab = activeTab === 0;
+  const isCollectiblesTab = activeTab === 1;
```
CassioMG: could we use enums here instead of the hardcoded numbers?

extension/src/popup/components/account/AccountTabs/index.tsx
CassioMG: nice call - would you mind adding a comment to the code explaining why we have it? Maybe we could also extract it to a hook to keep the view component slim/dumb

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -1,38 +1,170 @@
...
+    }
+  };
 
   return (
     <div className="AccountTabs">
-      {TabsList.map((tab, index) => {
-        return (
+      <div className="AccountTabs__tabs">
+        {TabsList.map((tab, index) => {
+          if (tab === "collectibles" && isCustomNetwork(networkDetails)) {
```
CassioMG: should we move "collectibles" to an enum here?

extension/src/popup/views/ManageCollectibles/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+interface FormValues {
+  collectibleAddress: string;
+  collectibleTokenId: string;
+}
+const initialValues: FormValues = {
+  collectibleAddress: "",
+  collectibleTokenId: "",
+};
+
+export const ManageCollectibles = () => {
```
CassioMG: how do we feel about renaming this component to `AddCollectible`? I think that would be a little more assertive since that's the only functionality this component has - Wdyt?

extension/src/popup/views/ManageCollectibles/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+        setAddCollectibleError(response.error);
+        return;
+      }
+
+      setAddCollectibleError("");
+
+      // refetch collectibles to update the UI before we navigate away
+      await fetchCollectibles({ publicKey, networkDetails });
+
+      navigateTo(ROUTES.account, navigate, `?tab=${TabsList[1]}`);
```
CassioMG: we could replace the `1` index with an enum here

extension/src/popup/views/ManageCollectibles/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+      initialValues={initialValues}
+      onSubmit={handleSubmit}
+      validationSchema={validateManageCollectiblesSchema}
+    >
+      {({ setFieldValue, touched, errors, dirty, isValid, isSubmitting }) => (
+        <Form className="ManageCollectibles" data-testid="ManageCollectibles">
+          <SubviewHeader title={t("Add Collectible")} />
+          <View.Content hasNoTopPadding>
+            <FormRows>
+              <Field name="collectibleAddress">
```
CassioMG: same suggestion from the other comment on renaming collectibleAddress

extension/src/popup/views/AddCollectibles/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+                    data-testid="collectible-address-wrapper"
+                  >
+                    <Input
+                      type="text"
+                      fieldSize="md"
+                      autoFocus
+                      autoComplete="off"
+                      data-testid="collectibleAddress"
+                      id="collectibleAdress"
+                      placeholder={t("Collectible address")}
```
CassioMG: @sdfcharles I've just noticed we have an inconsistency for the input placeholder here, on [mobile Figma](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=5923-142230&t=UAsAlVbNvThZmyfd-1) it says "Collection address" but on the [extension Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6323-193826&t=Nwu8WoPnCXGpNzJW-1) it says "Token address" and here on the PR it's using "Collectible address"

Image

Image



Since on the Collectible details UI (below) we list "Collection" and "Token ID" fields I'd vote for using "Collection address" as placeholder for consistency, but I'm not against using other names if you guys think it would look better

Image

Thanks!

extension/public/static/manifest/v3.json
piyalbasu: We actually don't currently have the paste button on the `Send` address input. Looks like this was added to the figma after initial implementation. I created a ticket here: https://github.com/stellar/freighter/issues/2434

CassioMG: aah, so this permission is only needed for actual paste buttons but not for the "cmd + p" keyboard command - TDIL

extension/src/popup/components/account/AccountTabs/index.tsx
piyalbasu: Good catch - I need to update this!

extension/src/popup/components/account/AccountTabs/index.tsx
piyalbasu: Good idea 👍

extension/src/popup/views/ManageCollectibles/index.tsx
piyalbasu: Sure - I was matching the naming of `ManageAssets|ManageAssetslists` in this repo (and to differentiate from `addCollectible` helpers, but maybe `AddCollectibles` is more clear

extension/src/popup/views/AddCollectibles/index.tsx
piyalbasu: Ah yes, I did notice this inconsistency and went ahead and matched what was in mobile. But we should change this in the figma, as well

CassioMG: @piyalbasu on mobile it's "**Collection** address" but on this PR the placeholder still says "**Collectible** address" - thanks!

extension/src/popup/views/AddCollectibles/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+      navigateTo(ROUTES.account, navigate, `?tab=${TabsList.COLLECTIBLES}`);
+    } catch (error) {
+      setAddCollectibleError("Unable to add collectible");
+      captureException(`Failed to add collectible: ${JSON.stringify(error)}`);
+    }
+  };
+
+  const validateAddCollectiblesSchema = YupObject().shape({
+    collectibleContractAddress: YupString()
+      .required(t("Collectible address is required"))
```
CassioMG: @piyalbasu I think this one also needs to be updated

piyalbasu: Oh gotcha - yes, you're correct. Updating that now to match mobile

piyalbasu: Happy to make this change, but we should think about our naming here. As a user, I still don't quite understand what the difference between a collectible and a collection is. The form says `Add Collectible`, but it's asking me to enter the `Collection address`. I find that to be a bit confusing

@sdfcharles @Keeeeeeeks Let's maybe think about this (or tell me why I'm wrong and why this actually makes sense 😄 )

PR #2424, opened 2025-12-03
Bugfix/sign for address

### LINKED ISSUE #2411: Signing with address not working
There are issues with signing with a user selected address. If, using freighter-api, a user passes the optional `address` param to request a sig from a specific public key the following happens:

1. Sign Tx: Freighter is set to `G1`. Dapp requests sig for `G2`. Freighter shows `G1` and signs with `G1`. The requested address is ignored
2. Sign Auth: Freighter is set to `G1`. Dapp requests sig for `G2`. Freighter shows `G1` and signs with `G2`. Correct behavior but wrong UI
3. Sign Msg: Freighter is set to `G1`. Dapp requests sig for `G2`. Freighter shows `G1` and signs with `G2`. Correct behavior but wrong UI

piyalbasu:
Closes #2411 

Fixes 2 issues regarding signing a tx/msg/auth entry with freighter-api:

1. When passing an address to the `sign transaction`, Freighter wasn't switching to the appropriate address
2. For all signing methods, the UI wasn't reflecting the address that had been requested (even if Freighter was signing with the correct address)

### DIFF
--- a/@shared/api/types/types.ts
+++ b/@shared/api/types/types.ts
@@ -152,10 +152,6 @@ export interface ExternalRequestAuthEntry extends ExternalRequestBase {
   entryXdr: string;
 }
 
-export interface ExternalRequestAllowedStatus extends ExternalRequestBase {
-  publicKey?: string;
-}
-
 export type ExternalRequest =
   | ExternalRequestToken
   | ExternalRequestTx
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -114,8 +114,6 @@ export const SignTransaction = () => {
     accountToSign,
   );
 
-  console.log(accountToSign);
-
   const {
     isConfirming,
     isPasswordRequired,


@shared/api/external.ts
```diff
@@ -68,34 +68,19 @@ export const submitToken = async (args: {
...
-  Previously, it accepted optional second and third string parameters.
-  This logic maintains backwards compatibility for older versions
-  */
-  if (typeof opts === "object") {
-    _accountToSign = opts.accountToSign;
-    networkPassphrase = opts.networkPassphrase;
-  } else {
-    network = opts;
-    _accountToSign = accountToSign;
-  }
```
piyalbasu, line 98: there was a little indirection in the type to provide backwards compatibility for an older version that made it easy to miss the bug here. v1.3.0 is 3 years old now, so I think we can safely drop support for it.

The bug was caused by `opts.address` not being set as `_accountToSign`

extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -111,8 +113,6 @@ export const SignAuthEntry = () => {
     });
   }
 
-  const publicKey = signAuthEntryData.data?.publicKey!;
```
piyalbasu, line 114: I'm applying this same fix across all signing UI's:

When a dapp requests a signature from a specific G address, Freighter UI opens the signing popup and (if it has the address), will switch to that address. This happens _after_ `signAuthEntryData.data` is already fetched, so the UI reflects whatever G address Freighter was already configured to.

By using the selector instead, the public key will update whenever the account is changed

@shared/api/external.ts
```diff
@@ -68,34 +68,19 @@ export const submitToken = async (args: {
...
   transactionXdr: string,
-  opts?:
-    | string
-    | {
-        accountToSign?: string;
-        networkPassphrase?: string;
-      },
-  accountToSign?: string,
+  opts?: {
+    address?: string;
```
CassioMG, line 72: Ah, I see now that all the sign APIs use `address` but this function used `accountToSign` so there was probably a mismatch here

@shared/api/external.ts
CassioMG: nice clean-up!

CassioMG (approved):
LGTM!

@shared/api/external.ts
piyalbasu: Yup - exactly that

PR #2435, opened 2025-12-04
Handle Blockaid unable to scan state

leofelix077:
[suspicious-add.webm](https://github.com/user-attachments/assets/b6ec11d5-845f-42fe-a87d-7c1c21bac64d)

[safe-add.webm](https://github.com/user-attachments/assets/a99d768a-5226-4a01-9255-fb114abb15d9)

[suspicious-swap.webm](https://github.com/user-attachments/assets/e84464eb-01ba-4d3a-8b6b-b43699b42a03)

[safe-send.webm](https://github.com/user-attachments/assets/bfa976a6-14ed-4746-8692-a782af184ba0)

[unable-add.webm](https://github.com/user-attachments/assets/13fdce10-7b08-401d-acfa-eaa2e64dd98a)

[malicious-send.webm](https://github.com/user-attachments/assets/571f3100-bdeb-4d80-a37b-0d223e8cbbee)

[safe-swap.webm](https://github.com/user-attachments/assets/e4e378b8-4e4d-4f9f-b38a-3003214d45ab)

[malicious-swap.webm](https://github.com/user-attachments/assets/53cdd0e0-9c50-4642-9439-c09b4b425e34)

[malicious-memo-ignore.webm](https://github.com/user-attachments/assets/2dfdfddc-1db5-4bee-bbb5-94218697c8af)

[unable-send.webm](https://github.com/user-attachments/assets/a9528c8c-1ff8-47de-a935-afc8d16bfe42)

[suspicious-send.webm](https://github.com/user-attachments/assets/0832b44b-7a24-469c-b2b7-896986c2892a)

[malicious-add.webm](https://github.com/user-attachments/assets/c7a9beeb-f4e8-4e00-b7c6-23e849dbeace)

[unable-memo-ignore.webm](https://github.com/user-attachments/assets/b0ac3a41-cfd5-4a22-b6bd-009ec0d74b34)

[unable-swap.webm](https://github.com/user-attachments/assets/6b300c32-dcc2-4f22-971b-38896d7dad2c)

[suspicious-memo-ignore.webm](https://github.com/user-attachments/assets/e105ebfc-655f-4243-aa97-96db4fca9fed)

### DIFF
--- a/extension/src/popup/components/InternalTransaction/ReviewTransaction/styles.scss
+++ b/extension/src/popup/components/InternalTransaction/ReviewTransaction/styles.scss
@@ -110,8 +110,8 @@
       width: 100%;
       border: none;
       background: none;
-      color: #a0a0a0;
-      font-family: "Inter", sans-serif;
+      color: var(--sds-clr-gray-09);
+      font-family: var(--sds-ff-base);
       font-weight: 600;
       font-style: normal;
       font-size: pxToRem(14px);
@@ -122,7 +122,7 @@
       text-align: center;
 
       &--default {
-        color: #a0a0a0;
+        color: var(--sds-clr-gray-09);
       }
 
       &--error {
--- a/extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/styles.scss
+++ b/extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/styles.scss
@@ -349,7 +349,7 @@
     border: none;
     background: transparent;
     color: var(--sds-clr-gray-09);
-    font-family: "Inter", sans-serif;
+    font-family: var(--sds-ff-base);
     font-weight: 600;
     font-style: normal;
     font-size: pxToRem(14px);
--- a/extension/src/popup/views/SignTransaction/index.tsx
+++ b/extension/src/popup/views/SignTransaction/index.tsx
@@ -363,6 +363,38 @@ export const SignTransaction = () => {
       ? scanResult.simulation.assets_diffs?.[publicKey]
       : undefined;
 
+  /**
+   * Renders at most one warning banner based on priority:
+   * 1. Domain  – destination domain not in allowlist
+   * 2. Blockaid – malicious / suspicious / unable-to-scan result
+   * 3. Memo    – required but missing
+   */
+  const renderBanner = () => {
+    // 1. Domain: site not in allowlist
+    if (!isDomainListedAllowed) {
+      return <DomainNotAllowedWarningMessage domain={domain} />;
+    }
+
+    // 2. Blockaid: malicious / suspicious / unable-to-scan
+    if (showBlockAidDetails) {
+      return (
+        <div className="SignTransaction__BlockaidDetails">
+          <BlockaidTxScanLabel
+            scanResult={scanResult}
+            onClick={() => setActivePaneIndex(1)}
+          />
+        </div>
+      );
+    }
+
+    // 3. Memo required
+    if (isMemoRequired) {
+      return <MemoRequiredLabel onClick={() => setActivePaneIndex(3)} />;
+    }
+
+    return null;
+  };
+
   return isPasswordRequired ? (
     <VerifyAccount
       isApproval
@@ -395,20 +427,7 @@ export const SignTransaction = () => {
                     </span>
                   </div>
                 </div>
-                {showBlockAidDetails && (
-                  <div className="SignTransaction__BlockaidDetails">
-                    <BlockaidTxScanLabel
-                      scanResult={scanResult}
-                      onClick={() => setActivePaneIndex(1)}
-                    />
-                  </div>
-                )}
-                {!isDomainListedAllowed && (
-                  <DomainNotAllowedWarningMessage domain={domain} />
-                )}
-                {isMemoRequired && (
-                  <MemoRequiredLabel onClick={() => setActivePaneIndex(3)} />
-                )}
+                {renderBanner()}
                 {assetDiffs && (
                   <AssetDiffs
                     icons={signTxState.data?.icons || {}}


extension/e2e-tests/blockaidScan.safe.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      };
+      await route.fulfill({ json });
+    });
+
+    await page.route("**/paths**", async (route) => {
+      const url = new URL(route.request().url());
+      const sourceAsset = url.searchParams.get("source_asset_code");
+      const destAsset = url.searchParams.get("destination_asset_code");
+
+      const createAssetObject = (assetCode: string | null, issuer: string) => {
```
piyalbasu: let's move this to the helpers folder so we can reuse this

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
```diff
@@ -96,6 +108,55 @@ export const ReviewTx = ({
     ? truncatedFedAddress(federationAddress)
     : truncatedPublicKey(destination);
 
+  // Scan source asset if not XLM
+  const srcAssetAddress =
+    srcAsset !== "native" && asset.issuer ? asset.issuer : null;
+  const { scannedAsset: srcAssetScanResult } = useScanAsset(
```
piyalbasu: is this what we're doing on mobile? We actually chose not to scan assets on transaction review because scanning the tx will tell you if either asset is susupicious/malicious

leofelix077: on the app we use separate pages for each case

e.g.:

Swap Page -> Scan source and destination assets
Transaction Page -> Scan only transaction
Add token -> Scan token being added
Dapp connect -> Scan site being added 

As here it's all the same page, we currently scan it all, but only show based on the context. if swap, then determine if should show based on source and destination assets. if a send then only based on the transaction directly without checking for the individual tokens

piyalbasu: You don't need to differentiate between different kinds of transactions, though. Whether it's a swap or a payment, Blockaid will scan all the assets involved. You don't need to scan each piece individually for a swap. If Blockaid finds any asset involved is malicious/suspicious (whether it's the asset being sent or the asset being received), it will surface it in the error.

Ultimately, it's not clear to me why we're rewriting the business logic behind how to scan a swap in this PR

extension/src/popup/helpers/blockaid.ts
```diff
@@ -213,20 +230,219 @@ export const useScanAsset = (address: string) => {
...
+ * In production, blockaid override state is ignored
+ */
+export const useIsAssetSuspicious = () => {
+  const [blockaidOverrideState, setBlockaidOverrideState] = useState<
+    string | null
+  >(null);
+
+  useEffect(() => {
+    if (!isDev) {
+      return;
```
piyalbasu: for each of these hooks: why do these need to be `useEffect`s? `isDev` isn't a side effect, it's deterministically set at build time. The value won't change

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
piyalbasu: I guess another way to phrase my question is: why isn't it enough to (for ex) show an unable to scan warning if `payload.scanResult` is set to `null` in this [hook](https://github.com/stellar/freighter/blob/7e689fa7ca8f685eb6364c0407f1df57c90f0c3c/extension/src/popup/components/swap/SwapAmount/hooks/useSimulateSwapData.tsx#L219)? We're adding a fair amount of complexity and new logic here when we're really just trying to show an error if the call to Blockaid fails.

But maybe there's a good reason that I'm just not seeing yet!

leofelix077: you are right. it was not showing multiple rows, but what I meant is to keep it somewhat similar to how the app is doing it. wdyt?


Screenshot 2025-12-18 at 10 35 15

Screenshot 2025-12-18 at 15 55 13

extension/src/popup/helpers/blockaid.ts
leofelix077: it was just a precaution not to let the override run outside the isDev env. but the checks are already being done inside the functions being called. the redundancy shouldn't be needed. done some refactor and removed it

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
piyalbasu: Ah got it. I wasn't aware that mobile did it this way. Let's loop in @sdfcharles to confirm what we should do here. We previously made the decision to only scan the tx, not individual assets. I'd advocate to not add this now:

1. As mentioned above, scanning the tx will tell you if an asset is suspicious/malicious. We won't be able to make a nice looking itemized list like in the figma - but we will have the information. The additional API calls add unnecessary overhead

2. A user will actually already know if these assets are suspicious or not. When we call `/account-balances`, Freighter BE comes back with the Blockaid scan results for each asset. Since the user already owns these assets, they will already know if Blockaid has marked them as suspicious/malicious. An extra scan here doesn't really provide much value. (It is true that the `/account-balances` doesn't let a user know if the scan failed, but we can address that in another ticket)

Currently, a swap to a malicious asset looks like this (not very pretty, but this _should_ be a rare screen to see):

Screenshot 2025-12-18 at 3 50 25 PM

leofelix077: ah fair enough. then I can just revert this part to keep it simpler as it was before. will just wait a bit for @sdfcharles before that

sdfcharles: > Ah got it. I wasn't aware that mobile did it this way. Let's loop in @sdfcharles to confirm what we should do here. We previously made the decision to only scan the tx, not individual assets. I'd advocate to not add this now:
> 
> 1. As mentioned above, scanning the tx will tell you if an asset is suspicious/malicious. We won't be able to make a nice looking itemized list like in the figma - but we will have the information. The additional API calls add unnecessary overhead
> 2. A user will actually already know if these assets are suspicious or not. When we call `/account-balances`, Freighter BE comes back with the Blockaid scan results for each asset. Since the user already owns these assets, they will already know if Blockaid has marked them as suspicious/malicious. An extra scan here doesn't really provide much value. (It is true that the `/account-balances` doesn't let a user know if the scan failed, but we can address that in another ticket)
> 
> Currently, a swap to a malicious asset looks like this (not very pretty, but this _should_ be a rare screen to see):
> 
> Screenshot 2025-12-18 at 3 50 25 PM

@piyalbasu @leofelix077 hmm yeah i think it makes sense not having to scan multiple times, one scan on the txn should be sufficient!

leofelix077: @piyalbasu  simplified the checks now

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
```diff
@@ -96,6 +142,37 @@ export const ReviewTx = ({
...
+
+  // Determine malicious/suspicious from scan result (not dev-only)
+  const isMalicious =
+    txScanResult?.validation &&
+    "result_type" in txScanResult.validation &&
+    txScanResult.validation.result_type === "Malicious";
+  const isSuspicious =
+    txScanResult?.validation &&
+    "result_type" in txScanResult.validation &&
+    txScanResult.validation.result_type === "Warning";
```
piyalbasu: It looks like `isMalicious` and `isSuspicious` grab the designation from `validation.result_type`. But `getTransactionWarningType` does the same thing. They seem duplicative of each other.

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -710,10 +659,112 @@ export const BlockaidTxScanLabel = ({
...
   onClick,
 }: {
-  scanResult: BlockAidScanTxResult;
+  scanResult: BlockAidScanTxResult | null | undefined;
   onClick: () => void;
 }) => {
   const { t } = useTranslation();
+  const shouldTreatAsUnableToScan = useShouldTreatTxAsUnableToScan();
+  const isTxSuspiciousCheck = useIsTxSuspicious();
+  const [blockaidState, setBlockaidState] = useState<string | null>(null);
```
piyalbasu: I would rename these vars to `[blockaidOverrideState, setBlockaidOverrideState` so they're more descriptive of what the values are being used for

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -710,10 +659,112 @@ export const BlockaidTxScanLabel = ({
...
+    getBlockaidOverrideState()
+      .then(setBlockaidState)
+      .catch(() => setBlockaidState(null));
+  }, [isDev]);
+
+  // Extract complex conditions for readability
+  const isUnableToScan = shouldTreatAsUnableToScan(scanResult);
+  const hasScanResult = !!scanResult;
+  const isMalicious =
+    isDev &&
```
piyalbasu: My understanding is that `isMalicious` and `isSuspicious` are used to replace the real Blockaid message with the overridden message. Lmk if that's incorrect

1. I think the var names should reflect that that is what they're being used for (something like `isMaliciousOverride`)

2. Why do we need to check `blockaidstate` and `isDev` here? It looks like `isTxSuspiciousCheck` checks these under the hood

extension/src/popup/components/WarningMessages/index.tsx
```diff
@@ -791,109 +842,196 @@ export const BlockaidTxScanLabel = ({
...
+      isMalicious = true;
+    } else if (result_type === "Warning" || result_type === "Spam") {
+      isSuspicious = true;
+    }
+  }
+
+  return { warnings, isMalicious, isSuspicious };
+};
+
 export const BlockAidTxScanExpanded = ({
```
piyalbasu: It looks like this component now handles both tx scans and asset scans. If that's correct, let's rename this to something more generic, like `BlockAidScanExpanded`

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
```diff
@@ -34,6 +36,50 @@ import { useValidateTransactionMemo } from "popup/helpers/useValidateTransaction
 
 import "./styles.scss";
 
+type WarningType =
```
piyalbasu: Should we reuse the `SecurityLevel` enum instead? They're pretty similar

extension/src/popup/views/Debug.tsx
```diff
@@ -1,17 +1,165 @@
...
+  };
+
   return (
-    <div>
-      <h1>{t("Debug")}</h1>
-      <p>{t("Use this page for development helpers")}</p>
-      <ul>
-        <li>{t("dummy entry")}</li>
-      </ul>
+    <div className="Debug">
```
piyalbasu, line 73: it would be great to add some directions on how to do this in the README for future devs

extension/playwright.config.ts
```diff
@@ -34,6 +34,8 @@ export default defineConfig({
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: "on-first-retry",
+    /* Record videos for debugging and post-run review */
+    video: "on",
```
piyalbasu: I think you were just using this for local testing, right? Let's remove this if that's the case

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
```diff
@@ -252,6 +324,394 @@ export const ReviewTx = ({
     onConfirm();
   };
 
+  const renderSendAsset = () => {
```
piyalbasu: Small improvement might be to move `renderSendAsset`, `renderSendDestination`, `renderActionButtons`, and the pane logic to a separate file. We're starting to have a lot of different UI logic in this one file

leofelix077:
@piyalbasu @sdfcharles  deleted my last comment. Just realized it was showing the same. I saw the video and thought the animation was sliding back to pane 0, but it isn't, so not a bug for the malicious case. it's in all cases like this

blockaid pane on index 0 and 2

after initial continue user can navigate back and forth from the review <> blockaid pane

it's following closer to how the app behaves as well. but we can see on the sync

extension/src/popup/views/GrantAccess/index.tsx
```diff
@@ -101,15 +113,42 @@ export const GrantAccess = () => {
...
+    // Override state takes precedence
+    isMalicious = blockaidOverrideState === SecurityLevel.MALICIOUS;
+    isSuspicious = blockaidOverrideState === SecurityLevel.SUSPICIOUS;
+    isUnableToScan = blockaidOverrideState === SecurityLevel.UNABLE_TO_SCAN;
+  } else {
+    // Use actual scan results
+    isUnableToScan = !scanData || scanData.status === undefined;
+    isMalicious = scanData?.status === "hit" && scanData.is_malicious;
+    // Treat non-malicious hits as suspicious (site has some concerns but not definitively malicious)
+    isSuspicious = scanData?.status === "hit" && !scanData.is_malicious;
```
piyalbasu: this logic is incorrect - benign sites will have `status === "hit"` and `isMalicious` will be false. This effectively shows an `isSuspicious` warning for all sites that are not malicious

Check out the types for `scanData`. There isn't a concept of `isSuspicious`. It's either malicious or not. We don't need to worry about suspicious

extension/src/popup/views/SignMessage/index.tsx
```diff
@@ -84,9 +91,33 @@ export const SignMessage = () => {
...
+  const [blockaidOverrideState, setBlockaidOverrideState] = useState<
+    string | null
+  >(null);
+
   useEffect(() => {
     const getData = async () => {
       await fetchData();
+      // Scan the site for security issues
+      try {
+        const result = await scanSite(url);
```
piyalbasu: Let's follow convention and do this scanSite logic in the associated hook for this file

piyalbasu: In addition, I would follow the convention we use in `extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx` and don't block the UI from loading while we're loading the scanSite data

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -132,6 +135,53 @@ export const SignTransaction = () => {
...
+    getBlockaidOverrideState()
+      .then(setBlockaidOverrideState)
+      .catch(() => setBlockaidOverrideState(null));
+  }, []);
+
+  useEffect(() => {
+    const scan = async () => {
+      if (domain) {
+        try {
+          const result = await scanSite(domain);
```
piyalbasu: same comment as above - let's do this in the hook. Also, make sure that waiting for `scanSite` doesn't block the UI from loading

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -132,6 +135,53 @@ export const SignTransaction = () => {
...
+  let isSiteUnableToScan = false;
+
+  if (blockaidOverrideState) {
+    isSiteMalicious = blockaidOverrideState === SecurityLevel.MALICIOUS;
+    isSiteSuspicious = blockaidOverrideState === SecurityLevel.SUSPICIOUS;
+    isSiteUnableToScan = blockaidOverrideState === SecurityLevel.UNABLE_TO_SCAN;
+  } else {
+    isSiteUnableToScan = !scanData || scanData.status === undefined;
+    isSiteMalicious = scanData?.status === "hit" && scanData.is_malicious;
+    isSiteSuspicious = scanData?.status === "hit" && !scanData.is_malicious;
```
piyalbasu: a lot of this logic looks similar to what you're doing in SignMessage. Let's create a helper so we're not duplicating code

extension/src/popup/views/SignTransaction/index.tsx
piyalbasu: We'll also need to do the same in `SignAuthEntry`

extension/public/static/manifest/v3.json
```diff
@@ -34,6 +34,6 @@
     "48": "images/icon48.png",
     "128": "images/icon128.png"
   },
-  "permissions": ["storage", "alarms"],
+  "permissions": ["storage", "alarms", "clipboardRead"],
```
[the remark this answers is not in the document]
CassioMG: I think @piyalbasu raised some points about clipboard permission recently in case that is applicable here

.vscode/mcp.json
```diff
@@ -1,14 +0,0 @@
-{
```
piyalbasu, line 1: I think this is the result of a bad rebase

extension/public/static/manifest/v3.json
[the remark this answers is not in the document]
piyalbasu: Yeah, this looks like the result of a bad rebase

@shared/api/internal.ts
```diff
@@ -1660,6 +1661,50 @@ export const saveExperimentalFeatures = async ({
...
+export const saveBlockaidOverrideState = async ({
+  overriddenBlockaidResponse,
+}: {
+  overriddenBlockaidResponse: string | null;
+}): Promise<{ overriddenBlockaidResponse: string | null }> => {
+  const response = (await sendMessageToBackground({
+    activePublicKey: null,
+    overriddenBlockaidResponse,
+    type: SERVICE_TYPES.SAVE_BLOCKAID_DEBUG_OVERRIDE,
+  })) as unknown as {
```
piyalbasu: wjy do we need to cast to `unknown` here? That can sometimes be a code smell that means there's something wrong with the type

extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -370,6 +375,12 @@ export const popupMessageListener = (
     case SERVICE_TYPES.LOAD_BACKEND_SETTINGS: {
       return loadBackendSettings({ localStore });
     }
+    case SERVICE_TYPES.SAVE_BLOCKAID_DEBUG_OVERRIDE: {
+      return saveBlockaidOverrideState({
+        request: request as SaveBlockaidDebugOverrideMessage,
```
piyalbasu: we shouldn't need to cast this as `SaveBlockaidDebugOverrideMessage`. It should be able to figure out the type on its own

extension/src/popup/components/InternalTransaction/ReviewTransaction/components/ActionButtons.tsx
```diff
@@ -0,0 +1,189 @@
...
+  paneConfig: {
+    reviewIndex: number;
+  };
+  isSubmitDisabled: boolean;
+  dstAsset?: {
+    canonical: string;
+    amount: string;
+  };
+  dest: any;
+  asset: any;
```
piyalbasu: let's add actual types for `dest` and `asset`

extension/src/popup/components/InternalTransaction/ReviewTransaction/components/SendAsset.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { NetworkDetails } from "@shared/constants/stellar";
+import { isMainnet } from "helpers/stellar";
+import { AssetIcon } from "popup/components/account/AccountAssets";
+import { CollectibleInfoImage } from "popup/components/account/CollectibleInfo";
+
+interface SendAssetProps {
+  isCollectible: boolean;
+  collectibleData: any;
+  assetIcons: { [key: string]: string | null };
+  asset: any;
```
piyalbasu: we should have types for `collectibleData` and `asset`

extension/src/popup/components/InternalTransaction/ReviewTransaction/components/SendDestination.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { IdenticonImg } from "popup/components/identicons/IdenticonImg";
+
+interface SendDestinationProps {
+  dstAsset?: {
+    icon: string | null;
+    canonical: string;
+    priceUsd: string | null;
+    amount: string;
+  };
+  dest: any;
```
piyalbasu: type for `dest`

extension/playwright.config.ts
```diff
@@ -39,6 +39,10 @@ export default defineConfig({
         IS_PLAYWRIGHT: "true",
       },
     },
+    viewport: {
+      width: 1280,
+      height: 720,
+    },
```
piyalbasu: i think this is the result of a bad rebase

extension/src/popup/components/InternalTransaction/ReviewTransaction/styles.scss
```diff
@@ -90,6 +91,44 @@
...
+    }
+
+    .ReviewTx__TextAction {
+      margin-top: pxToRem(12px);
+      width: 100%;
+      border: none;
+      background: none;
+      color: #a0a0a0;
+      font-family: "Inter", sans-serif;
+      font-weight: 600;
```
piyalbasu, line 115: we should use SDS values for the `color` and `font-family` here

extension/src/popup/components/InternalTransaction/ReviewTransaction/styles.scss
```diff
@@ -90,6 +91,44 @@
...
+      font-style: normal;
+      font-size: pxToRem(14px);
+      line-height: pxToRem(20px);
+      letter-spacing: 0;
+      padding: pxToRem(8px) 0;
+      cursor: pointer;
+      text-align: center;
+
+      &--default {
+        color: #a0a0a0;
```
piyalbasu: same - use SDS var

extension/src/popup/components/manageAssets/ManageAssetRows/ChangeTrustInternal/styles.scss
```diff
@@ -335,4 +342,38 @@
...
     }
   }
+
+  .ReviewTx__TextAction {
+    margin-top: pxToRem(12px);
+    width: 100%;
+    border: none;
+    background: transparent;
+    color: var(--sds-clr-gray-09);
+    font-family: "Inter", sans-serif;
```
piyalbasu: use SDS value here

PR #2451, opened 2025-12-10
Feature/collectible detail

### LINKED ISSUE #2326: [Collectibles] Collectible Details UI
Let's port this from mobile.

You should be able to use our [hardcoded dummyCollections](https://github.com/stellar/freighter-mobile/blob/main/src/ducks/collectibles.ts#L115) while building the UI so you don't necessarily need to connect with our backend right away.

Note1: the "Hidde collectible" feature should be implemented in this [other task](https://github.com/stellar/freighter/issues/2331).
Note2: we can skip "Report as spam" for now.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6322-192553&t=i6DQHvjsIvsFy5wX-1):

Image

piyalbasu:
Closes #2326 

Adds Collectible Details UI

* When clicking on a collectible, it will open up the details of that collectible (from the redux store) into a modal, similar to how we deal with asset details
* Clicking `refresh metadata` will re-fetch just the metadata for only the selected collectible. The new metadata is then added to redux to make sure our redux store stays in sync while minimizing the amount of API's calls we need to make

Screenshot 2025-12-10 at 1 13 12 PM
Screenshot 2025-12-10 at 1 13 19 PM

Example of changing metadata for a collectible:

https://github.com/user-attachments/assets/7fde0ef5-1b1b-4ab2-a181-efc9191caa47

### DIFF
--- a/extension/e2e-tests/addAsset.test.ts
+++ b/extension/e2e-tests/addAsset.test.ts
@@ -202,9 +202,7 @@ test("Adding classic asset on Testnet", async ({ page, extensionId }) => {
   await page.getByRole("button", { name: "Confirm" }).click();
   await page.getByText("Done").click();
   await expect(
-    page.getByText(
-      "You have no assets added. Get started by adding an asset below.",
-    ),
+    page.getByText("You have no assets added. Get started by adding an asset."),
   ).toBeVisible();
 });
 test.afterAll(async ({ page, extensionId }) => {


@shared/api/helpers/fetchCollectibles.ts
```diff
@@ -125,6 +136,8 @@ export const fetchCollectibles = async ({
                     collectible.token_uri,
                   );
                   return {
+                    collectionAddress: collection.address,
+                    collectionName: collection.name,
```
piyalbasu, line 139: adding these keys to the collectible object to make displaying details a little simpler. Mobile follows a similar data model

extension/src/popup/basics/shadcn/Popover.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from "react";
```
piyalbasu, line 1: implementing Popover from shadcn as a test run 🎉

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+              </div>
+            </Popover>
+          }
+        />
+        {state.state === RequestState.LOADING ? (
+          <Loading />
+        ) : (
+          <View.Content>
+            <div className="CollectibleDetail__content">
+              {collectible.metadata?.image && (
```
piyalbasu, line 139: we don't have any control of collectible metadata (as it's hosted by the NFT creator), so I'm using defensive checks for all the pieces of data we want to show. I'm working under the assumption some (or all) of these fields may not have been configured properly

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+                    >
+                      {t("View on stellar.expert")}
+                    </div>
+                  </div>
+                  <div className="CollectibleDetail__header__right-button__popover-content__item">
+                    <Icon.EyeOff className="CollectibleDetail__header__right-button__popover-content__item__icon" />
+                    <div className="CollectibleDetail__header__right-button__popover-content__item__label">
+                      {t("Hide collectible")}
+                    </div>
+                  </div>
```
piyalbasu, line 128: this button functionality will be implemented in the next PR

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const stellarExpertUrl = getStellarExpertUrl(networkDetails);
+
+  const handleRefreshMetadata = async () => {
+    await fetchCollectibleMetadata({
+      collectionAddress: selectedCollectible.collectionAddress,
+      tokenId: selectedCollectible.tokenId,
+      tokenUri: collectible.tokenUri || "",
+    });
+    setIsPopoverOpen(false);
```
piyalbasu, line 76: TODO: we don't have an error state designed for what to do if refreshing metadata fails. For now, this fails silently but we should provide the user with some feedback if it failed @sdfcharles

extension/src/popup/views/Account/index.tsx
```diff
@@ -290,7 +292,11 @@ export const Account = () => {
               ),
               <div data-testid="account-collectibles">
                 <AccountCollectibles
-                  collections={resolvedData?.collectibles?.collections || []}
+                  collections={
+                    collections[resolvedData?.networkDetails?.network || ""]?.[
+                      resolvedData?.publicKey || ""
+                    ] || []
+                  }
```
piyalbasu, line 299: use our value from redux so when we refresh metadata, the change is reflected in all views

extension/src/popup/basics/shadcn/Popover.tsx
piyalbasu: Picked up directly from here: https://ui.shadcn.com/docs/components/popover

extension/e2e-tests/helpers/stubs.ts
```diff
@@ -441,12 +443,26 @@ export const stubCollectibles = async (page: Page) => {
...
+      json.image =
+        "https://nftcalendar.io/storage/uploads/2024/06/02/pepe-the-bot_ml4cWknXFrF3K3U1.jpeg";
+      json.attributes = [
+        {
+          trait_type: "Background",
+          value: "Green (updated)",
+        },
+      ];
+    }
+    console.log(tokenMetadataCount);
```
aristidesstaffieri: leftover debug log

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+  };
+
+  return (
+    <div className="CollectibleDetail" data-testid="CollectibleDetail">
+      <View>
+        <SubviewHeader
+          title={name}
+          customBackAction={handleItemClose}
+          rightButton={
+            <Popover open={isPopoverOpen} onOpenChange={setIsPopoverOpen}>
```
aristidesstaffieri, line 86: A couple of things I noticed - 
The detail view should be a bottom slider but the Popover is for modals, should this [use the sheet](https://ui.shadcn.com/docs/components/sheet#usage) instead?
Our bottom sheets have the `x` icon for the dismiss action, here we use the back arrow.

piyalbasu: Ah yes, thanks for bringing that up. I did see that the mobile app uses the bottom sheet for this view, but I opted to re-use the style we use for the asset detail subview for consistency. 



@sdfcharles What do you think? I can implement the shadcn bottom sheet, but I'd imagine we'd want to do this for the asset detail subview, as well

piyalbasu: After chatting with Charles, we decided to move forward with implementing shadcn's sheet. It is a sizable change so I opened a new PR for this work to make it easier to review: #2463

PR #2458, opened 2025-12-12
[CHORE] adds platform to events emitted

### LINKED ISSUE #2199: Annotate Amplitude with project ID
Adds project ID in order to differentiate events in the console

all Freighter client metrics emitted with a project ID, such that Freighter Extension events are clearly distinguishable from Freighter iOS and Freighter Android events.

None

aristidesstaffieri:
Closes #2199 

Adds platform field to metrics, in order to be able to key events by platform.


CassioMG:
@aristidesstaffieri can we also assign a Platform `user` property in addition to the `event` property?

On mobile we are setting both so we can easily filter users as well

CassioMG:
> ok got it, it looks like mobile [uses the sdk here ](https://github.com/stellar/freighter-mobile/blob/14f91c4e82d310f4d4db8812e461e450a31b0e10/src/services/analytics/core.ts#L43)in order to automatically add these properties. We should add separate task for this, the extension does not use the sdk at all right now or identify users so its a bigger line of work and unrelated to separating events by platform.
> 
> I made this one to track that refactor -> #2464

Ah ok, got it that makes sense to me. Thanks for creating the ticket 👍

PR #2471, opened 2025-12-17
Add Cap-67/Mux address support

### LINKED ISSUE #2315: Add Memo/CAP-67 support for Soroban Transactions/Mux adresses

leofelix077:
Closes #2315 

https://github.com/user-attachments/assets/330b10d0-30cc-4636-9c7c-0dd9c4cc23d3

https://github.com/user-attachments/assets/7b40f050-2e3f-4825-8bcc-2e61593018b1

https://github.com/user-attachments/assets/b7b8beec-da24-4849-8b0a-53b0041bdc41

https://github.com/user-attachments/assets/58acd19c-7517-4117-87c1-52313f76bf05

https://github.com/user-attachments/assets/58f76c9e-ff22-4cf8-851c-2fa6fb435af0

### DIFF
--- a/extension/src/helpers/muxedAddress.ts
+++ b/extension/src/helpers/muxedAddress.ts
@@ -8,7 +8,6 @@ import {
   isMuxedAccount,
   isValidStellarAddress,
   createMuxedAccount,
-  getBaseAccount,
   isFederationAddress,
 } from "helpers/stellar";
 import { StrKey } from "stellar-sdk";
@@ -114,8 +113,8 @@ export function getMemoDisabledState(
   }
 
   // For tokens without Soroban mux support, disable memo
-  // If contractSupportsMuxed is null, we're still checking, so don't disable yet
-  // Only disable when explicitly false
+  // If contractSupportsMuxed is null, we're still checking, so disable memo until check completes
+  // Only allow when explicitly true
   if (contractSupportsMuxed === false) {
     return {
       isMemoDisabled: true,
@@ -123,10 +122,12 @@ export function getMemoDisabledState(
     };
   }
 
-  // If contractSupportsMuxed is null, we're still checking - allow memo for now
-  // (will be disabled once check completes if contractSupportsMuxed === false)
+  // If contractSupportsMuxed is null, we're still checking - disable memo until check completes
   if (contractSupportsMuxed === null) {
-    return { isMemoDisabled: false, memoDisabledMessage: undefined };
+    return {
+      isMemoDisabled: true,
+      memoDisabledMessage: t("Memo is not supported for this operation"),
+    };
   }
 
   if (isContractId(targetAddress)) {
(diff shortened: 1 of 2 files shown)


extension/src/popup/metrics/access.ts
```diff
@@ -14,47 +14,51 @@ import { registerHandler, emitMetric, MetricsData } from "helpers/metrics";
...
-});
-registerHandler<AppState>(addToken.fulfilled, () => {
-  const metricsData: MetricsData = JSON.parse(
-    localStorage.getItem(METRICS_DATA) || "{}",
-  );
-  emitMetric(METRIC_NAMES.addToken, {
-    accountType: metricsData.accountType,
+// Defer registration to avoid circular dependency issues during module initialization
+// Register handlers after a microtask to ensure all modules are fully loaded
+Promise.resolve().then(() => {
```
leofelix077: Posting the last open comment here from the previous PR

> hm, how was this presenting itself when you ran your code? I haven't seen anyone else run into this issue, so I'm wondering if this is actually an issue in the codebase or with your setup

can you try running in your machine from scratch? not directly related to this PR, but I noticed sometimes after building and running yarn start, when accessing localhost:9000, it would just load infinitely. Also going on the extension side for the "load unpacked" it was throwing background services error , without ever loading the home screen

Adding these changes seem to have stopped the issue, that's why I added it here

piyalbasu: When you're building your code, what cmd are you using? If you use `yarn build:extension:production && yarn start:extension` you will experience what you're describing: `localhost:9000` spinning forever. 

If you use `yarn build:extension && yarn start:extension`, it will work. I wonder if that's the issue you're running into?

extension/src/helpers/muxedAddress.ts
```diff
@@ -0,0 +1,228 @@
...
+  isFederationAddress,
+} from "helpers/stellar";
+import { StrKey } from "stellar-sdk";
+
+export interface CheckContractMuxedSupportParams {
+  contractId: string;
+  networkDetails: NetworkDetails;
+}
+
+/**
```
piyalbasu, line 20: let's add tests for the helpers in this file. We're covering a lot of different scenarios so it'd be good to have some coverage here

extension/src/popup/metrics/access.ts
leofelix077: let me revert these changes and see if I can reproduce it. it doesn't happen always, but sometimes it just goes to the main screen loading forever. no content loads. I just built it normal `yarn build && yarn start` from root. does building the api also affects it? (it's included on the base `:build` command)

leofelix077: maybe it was an old unpacked version loaded, without some of the message handlers mismatching with the full screen version and I didn't realize. not sure. rebuilt it a few times from scratch and it seems to be fine as well 🤷

piyalbasu: >does building the api also affects it? (it's included on the base :build command)

Hm, no this shouldn't affect anything 🤔 

>  it seems to be fine as well
Interesting. Okay, let's keep an eye on it and see if it becomes an issue again. The only reason I pushed back is because this part of the code has been in the repo for 5 years and this is the first time a dev has run into this issue. It made me think it may have more to do with your environment rather than the application code itself

piyalbasu (approved):

PR #2473, opened 2025-12-20
Add collectibles to send flow

### LINKED ISSUE #2329: [Collectibles] Send Collectible
Let's por this from mobile.

Users/accounts should be able to send collectibles between each other.

This is task encompasses both the `Send UI + functionality`.

The History UI is slated on this other task.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6322-189147&t=i6DQHvjsIvsFy5wX-1):
Image

piyalbasu:
Closes #2329 

This adds collectibles to Freighter's internal send flow. Users can send a collectible by either using the Send button on the Account view or they can use the Send button from the detail view.  This makes sure to handle the UI for both collectibles that have metadata properly configured and those that do not.

I spent a good amount of time thinking through how we can add this flow to the app in a way that makes sense. My 2 main options were to either 1) make a new, separate view that implements its own standalone flow (similar to how Swap is separate from SendPayment) or 2) add a configuration in the existing SendPayment flow that toggles on/off some collectibles elements when needed.

I decided to do the latter for a few reasons. Ultimately, sending a collectible is so similar to sending a payment, there was going to be a lot of duplication making a separate flow. In both flows,

*  you're sending a "thing" to an external destination that needs to be validated. (The one difference is there is no amount for a collectible)
* you need to configure fees (via tx simulation) and memo

In addition, because of CAP-67, memo validation is about to get a bit more complex. I'd like to be able to easily keep memo validation in sync between tokens/collectibles. Finally, sending collectibles doesn't really have its own standalone flow, begin to end. You can only send a collectible by starting a token payment flow. The UX flow is a branch off the existing token Send flow.

The downside is we have a few spots with `isCollectible` checks, but I tried to keep it as minimal and as clean as possible

VIDEO EXAMPLES:
Sending a collectible for a collectible with metadata

https://github.com/user-attachments/assets/2c868734-5cd8-4f39-8d6d-016e88964516

Sending a collectible for a collectible without metadata

https://github.com/user-attachments/assets/1733d042-9c90-4c76-8548-976e9bd975f1

An empty state for an account with no collectibles. This mirrors the empty state for assets:
Screenshot 2025-12-19 at 5 44 06 PM

### DIFF
--- a/extension/src/popup/views/Send/hooks/__tests__/useSendQueryParams.test.tsx
+++ b/extension/src/popup/views/Send/hooks/__tests__/useSendQueryParams.test.tsx
@@ -1,0 +1,613 @@
+import React from "react";
+import { Provider } from "react-redux";
+import { useLocation } from "react-router-dom";
+import { renderHook } from "@testing-library/react";
+import { StrKey } from "stellar-sdk";
+import {
+  makeDummyStore,
+  TEST_PUBLIC_KEY,
+  mockCollectibles,
+} from "popup/__testHelpers__";
+import {
+  MAINNET_NETWORK_DETAILS,
+  TESTNET_NETWORK_DETAILS,
+} from "@shared/constants/stellar";
+import { useSendQueryParams } from "../useSendQueryParams";
+import {
+  saveDestination,
+  saveAsset,
+  saveCollectibleData,
+  saveIsCollectible,
+  saveFederationAddress,
+} from "popup/ducks/transactionSubmission";
+import { initialState as transactionSubmissionInitialState } from "popup/ducks/transactionSubmission";
+import * as StellarHelpers from "@shared/helpers/stellar";
+import * as SorobanHelpers from "@shared/api/helpers/soroban";
+
+jest.mock("react-router-dom", () => ({
+  ...jest.requireActual("react-router-dom"),
+  useLocation: jest.fn(),
+}));
+
+jest.mock("@shared/helpers/stellar", () => ({
+  ...jest.requireActual("@shared/helpers/stellar"),
+  getAssetFromCanonical: jest.fn(),
+}));
+
+jest.mock("@shared/api/helpers/soroban", () => ({
+  ...jest.requireActual("@shared/api/helpers/soroban"),
+  isContractId: jest.fn(),
+}));
+
+describe("useSendQueryParams", () => {
+  const mockUseLocation = useLocation as jest.Mock;
+
+  const validEd25519PublicKey =
+    "GBTYAFHGNZSTE4VBWZYAGB3SRGJEPTI5I4Y22KZ4JTVAN56LESB6JZOF";
+  const validContractId =
+    "CAS3J7GYLGXMF6TDJBBYYSE3HW6BBSMLNUQ34T6TZMYMW2EVH34XOWMA";
+  const validAsset =
+    "USDC:GCK3D3V2XNLLKRFGFFFDEJXA4O2J4X36HET2FE446AV3M4U7DPHO3PEM";
+  const collectionAddress =
+    "CAS3J7GYLGXMF6TDJBBYYSE3HW6BBSMLNUQ34T6TZMYMW2EVH34XOWMA";
+  const collectibleTokenId = "1";
+
+  const defaultState = {
+    auth: {
+      publicKey: TEST_PUBLIC_KEY,
+      allAccounts: [],
+      applicationState: "MNEMONIC_PHRASE_CONFIRMED" as const,
+    },
+    settings: {
+      networkDetails: MAINNET_NETWORK_DETAILS,
+      networksList: [],
+    },
+    transactionSubmission: transactionSubmissionInitialState,
+    cache: {
+      collections: {
+        [MAINNET_NETWORK_DETAILS.network]: {
+          [TEST_PUBLIC_KEY]: mockCollectibles,
+        },
+      },
+    },
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockUseLocation.mockReturnValue({
+      pathname: "/send",
+      search: "",
+      state: null,
+    });
+    jest.spyOn(StrKey, "isValidEd25519PublicKey").mockImplementation((key) => {
+      return key === validEd25519PublicKey;
+    });
+    (SorobanHelpers.isContractId as jest.Mock).mockImplementation((id) => {
+      return id === validContractId;
+    });
+    (StellarHelpers.getAssetFromCanonical as jest.Mock).mockImplementation(
+      (canonical) => {
+        if (canonical === validAsset) {
+          return {
+            code: "USDC",
+            issuer: "GCK3D3V2XNLLKRFGFFFDEJXA4O2J4X36HET2FE446AV3M4U7DPHO3PEM",
+          };
+        }
+        throw new Error("Invalid asset");
+      },
+    );
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  const renderHookWithStore = (state = defaultState) => {
+    const store = makeDummyStore(state);
+    const dispatchSpy = jest.spyOn(store, "dispatch");
+    const Wrapper = ({ children }: { children: React.ReactNode }) => (
+      <Provider store={store}>{children}</Provider>
+    );
+    const hookResult = renderHook(() => useSendQueryParams(), {
+      wrapper: Wrapper,
+    });
+    return { hookResult, store, dispatchSpy };
+  };
+
+  describe("Collectible query params", () => {
+    it("should handle collectible params when collectible is found", () => {
+      mockUseLocation.mockReturnValue({
+        pathname: "/send",
+        search: `?collection_address=${collectionAddress}&collectible_token_id=${collectibleTokenId}`,
+        state: null,
+      });
+
+      const { dispatchSpy } = renderHookWithStore();
+
+      expect(dispatchSpy).toHaveBeenCalledWith(saveIsCollectible(true));
+      expect(dispatchSpy).toHaveBeenCalledWith(
+        saveCollectibleData({
+          collectionAddress,
+          tokenId: Number(collectibleTokenId),
+          name: "Stellar Frog 1",
+          collectionName: "Stellar Frogs",
+          image:
+            "https://nftcalendar.io/storage/uploads/events/2023/5/NeToOQbYtaJILHMnkigEAsA6ckKYe2GAA4ppAOSp.jpg",
+        }),
+      );
+      // Should not call saveDestination or saveAsset when collectible is found
+      expect(dispatchSpy).not.toHaveBeenCalledWith(
+        saveDestination(expect.anything()),
+      );
+      expect(dispatchSpy).not.toHaveBeenCalledWith(
+        saveAsset(expect.anything()),
+      );
+    });
+
+    it("should handle collectible params when collectible is not found", () => {
+      mockUseLocation.mockReturnValue({
+        pathname: "/send",
+        search: `?collection_address=INVALID_ADDRESS&collectible_token_id=999`,
+        state: null,
+      });
+
+      const { dispatchSpy } = renderHookWithStore();
+
+      // Should not set collectible data if not found
+      expect(dispatchSpy).not.toHaveBeenCalledWith(saveIsCollectible(true));
+      expect(dispatchSpy).not.toHaveBeenCalledWith(
+        saveCollectibleData(expect.anything()),
+      );
+    });
+
+    it("should handle collectible params with missing metadata", () => {
+      const collectionsWithMissingMetadata = [
+        {
+          collection: {
+            address: collectionAddress,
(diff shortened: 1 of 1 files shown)


@shared/api/internal.ts
```diff
@@ -1966,6 +1966,71 @@ export const modifyAssetsList = async ({
   return { assetsLists: response.assetsLists, error: response.error };
 };
 
+export const simulateSendCollectible = async (args: {
```
piyalbasu, line 1969: this mirrors the code `simulateTokenTransfer`

extension/e2e-tests/sendCollectible.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  await stubTokenDetails(page);
+  await stubAccountBalances(page);
+  await stubAccountHistory(page);
+  await stubTokenPrices(page);
+  await stubScanDapp(context);
+  await stubCollectibles(page);
+  await stubSimulateSendCollectible(page);
+  await stubScanTx(page);
+  await stubSubmitTx(page);
+  await stubFeeStats(page);
```
piyalbasu, line 31: these tests completely stub out interactions with Horizon and the network, so it should be fast and reliable 🎉 . This is a model we can start using for other tests that submit tx's, which are often flakey due to real world network conditions

extension/src/popup/components/InternalTransaction/ReviewTransaction/index.tsx
```diff
@@ -17,7 +17,7 @@ import {
   truncatedPublicKey,
 } from "helpers/stellar";
 
-import { SimulateTxData } from "popup/components/sendPayment/SendAmount/hooks/useSimulateTxData";
+import { SimulateTxData } from "popup/components/send/SendAmount/hooks/useSimulateTxData";
```
piyalbasu, line 24: Renaming `SendPayment` to `Send` as this flow can now send things that are not assets

extension/src/popup/components/account/AccountTabs/index.tsx
```diff
@@ -93,37 +174,10 @@ export const AccountTabs = () => {
...
-              key={tab}
-              onClick={() => {
-                setActiveTab(tab);
-              }}
-            >
-              {tabLabels[tab]}
-            </div>
-          );
-        })}
+        <TabButtons />
```
piyalbasu, line 180: We'll reuse these in the Send flow

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -76,6 +93,21 @@ export const CollectibleDetail = ({
     setIsPopoverOpen(false);
   };
 
+  const handleSendCollectible = () => {
+    // prepopulate the collectible data in the redux state before navigating to the send flow
+    dispatch(saveIsCollectible(true));
+    dispatch(
```
piyalbasu: this is slightly different than how we do this with assets. With assets, we send over query params for the asset and destination. The view decodes them and sets them in Redux

I thought it was a bit cleaner to set all these params at once directly into redux

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -138,102 +170,24 @@ export const CollectibleDetail = ({
...
-                  </div>
-                  <div
-                    className="CollectibleDetail__block__value"
-                    data-testid="CollectibleDetail__description__value"
-                  >
-                    {collectible.metadata?.description}
-                  </div>
-                </div>
-              )}
+              <CollectibleInfo
```
piyalbasu, line 160: All of these components will be reused in the Send flow

extension/src/popup/components/account/CollectibleInfo/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+export const getCollectibleName = (
+  name: string | undefined,
+  tokenId: string,
+) => {
+  // if the name is not defined in the metadata, use the token id to generate a name
+  return name || `Token #${tokenId}`;
+};
+
+export const CollectibleInfoBlock = ({
```
piyalbasu, line 22: most of these components are ported over from the CollectibleDetail view

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -116,9 +118,16 @@ export const SendAmount = ({
...
 
-  const handleContinue = async () => {
+  const handlePaymentContinue = async () => {
     const amount = inputType === "crypto" ? formik.values.amount : priceValue!;
     dispatch(saveAmount(cleanAmount(amount)));
+    await handleContinue();
+  };
+
+  const handleContinue = async () => {
+    if (!transactionFee) {
```
piyalbasu, line 210: this fixes a bug that I've been observing while testing this feature and that Esteban mentioned on Discord. During high network congestion, Soroban tx's were being rejected due to insufficient fee. It was because the base fee wasn't being added to the minimum resource fee from the simulation unless you edited the fee first. This didn't affect classic payments. It also only seemed to cause an issue during high network congestion.

I suspect this is why our send payments e2e tests have been extra flakey lately.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -331,248 +344,273 @@ export const SendAmount = ({
...
-                          {parsedSourceAsset.code}
-                        </div>
-                      </>
-                    )}
-                    {inputType === "fiat" && (
+            {transactionData.isCollectible ? (
+              <div className="SendAmount__collectible-display">
+                <SelectedCollectible goToChooseDest={goToChooseDest} />
+              </div>
+            ) : (
```
piyalbasu, line 518: Either show the Collectible or the Amount selector

extension/src/popup/components/sendCollectible/SelectedCollectible/hooks/useSimulateTxData.ts
```diff
@@ -0,0 +1,183 @@
...
+    new BigNumber(response.simulationResponse.minResourceFee),
+    CLASSIC_ASSET_DECIMALS,
+  );
+  return {
+    payload: response,
+    recommendedFee: baseFee.plus(new BigNumber(minResourceFee)).toString(),
+  };
+};
+
+function useSimulateTxData({
```
piyalbasu, line 81: this hook is similar to the simulateTx hook used by sending a token

extension/src/popup/components/send/SendAmount/index.tsx
aristidesstaffieri: Nice this should improve this issue but imo it can be improved further. What happens now is that we get the base fee at the time that you see the review sheet, then once the user hits "review" we simulate the tx and get the real fee. You can see this happen in the first video at around 0:11s(fee will change after hitting review but before sheet comes in). I'd say we should either simulate before showing any kind of fee value or move the fee to only be shown at the review sheet after simulation(cc @sdfcharles).

extension/src/popup/components/InternalTransaction/CollectiblesList/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <div className="CollectiblesList__collections">
+      {flattenedCollections.map((collection) => {
+        const title =
+          collection.metadata?.name || `${collection.collectionName}`;
+
+        return (
+          <div
+            className="CollectiblesList__collection"
+            key={title}
```
aristidesstaffieri: is this a unique key? I'd suggest `${collection.collectionAddress}-${collection.tokenId}`

extension/src/popup/components/sendCollectible/SelectedCollectible/hooks/useSimulateTxData.ts
```diff
@@ -0,0 +1,183 @@
...
+  const [state, dispatch] = useReducer(
+    reducer<SimulateTxData, string>,
+    initialState,
+  );
+
+  const fetchData = async () => {
+    dispatch({ type: "FETCH_DATA_START" });
+    try {
+      // Read transactionFee from Redux state inside fetchData to get the latest values
+      // Memos should not be allowed for collectible transactions
```
aristidesstaffieri, line 106: memos are not allowed in any Soroban transaction, and it looks like this comment has no enforcement in this function. Should we enforce this constraint in the simulation helper itself?

extension/src/popup/views/Send/index.tsx
```diff
@@ -129,7 +149,11 @@ export const SendPayment = () => {
         emitMetric(METRIC_NAMES.sendPaymentConfirm);
         return (
           <TransactionConfirm
-            xdr={simulationState.data?.transactionXdr!}
+            xdr={
+              isCollectible
+                ? collectibleSimulationState.data?.transactionXdr!
```
aristidesstaffieri: it shouldn't be necessary to assert these as non-null since the data hook gives us a way to narrow the type down to a success or error before the render happens. It would be safer to explicitly handle these states since `TransactionConfirm` expects that data in a non nullable way.

extension/src/popup/components/account/CollectibleDetail/index.tsx
```diff
@@ -76,6 +93,21 @@ export const CollectibleDetail = ({
     setIsPopoverOpen(false);
   };
 
+  const handleSendCollectible = () => {
+    // prepopulate the collectible data in the redux state before navigating to the send flow
```
aristidesstaffieri: in the past we have run into issues when we set up a direct dependency between views, where one view is required to setup state for an unrelated view(If another component navigates to Send with isCollectible=true but forgets to set collectibleData, it will break). The send asset flow uses internal state, optionally populated through URL parameters to solve this flow. This has the benefit of being stateless as far the store is concerned(does not have to be reset, will always be correct and sync'd to the view). Is there any reason to not do that for collectibles also?

extension/src/popup/components/sendCollectible/SelectedCollectible/hooks/useSimulateTxData.ts
piyalbasu: The enforcement in this function comes from not allowing the `memo` param to be passed into the `simulateTx` helper (unlike its token simulation counterpart) on line 124. 

We have a more complete solution for memos in Soroban tx's in this PR: https://github.com/stellar/freighter/pull/2471

aristidesstaffieri: 👍 thanks

extension/src/popup/components/account/CollectibleDetail/index.tsx
piyalbasu: This is a fair argument. My thinking here was to try to avoid either having to set 6 related query params and then decode them on the other end. Using query params in this manner always feels a little convoluted to me as you then need to now validate every passed value and make sure that all passed values align with one collectible the user has. It's not the worst thing in the world, but could start to get a little unwieldy. Query params, IMO, work better when the params are separate configs (like when sending an `asset` plus an optional `destination`. One works without the other)

I think maybe a happy medium here is to just pass the `collectionAddress` and `tokenId` as query params and then grab the rest of the metadata from the user's collections list. (You need to iterate over the collections list to do this, which is also not the most efficient, but perhaps this is the best way to maintain internal state while keeping the code clean

extension/src/popup/components/send/SendAmount/index.tsx
piyalbasu: Yup, definitely agree. I think this is a bit out of scope for this PR as it's fundamental to every type of Soroban tx. We should some spend some time thinking about this. I created a separate ticket here for us to discuss and design options: https://github.com/stellar/freighter/issues/2475

extension/src/popup/components/account/CollectibleDetail/index.tsx
aristidesstaffieri: if it's really worth the convenience then maybe the component consuming the redux state can be more resilient to state of the store? Can there be logic to handle the missing store state if it were not set properly?

piyalbasu: I think I landed on a pretty good solution. Sorry, it's actually split over 3 commits.

The first 2 commits address the pr comments: https://github.com/stellar/freighter/pull/2473/commits/49db53dc751db29576e7e2d18a4d86d82356c567
https://github.com/stellar/freighter/pull/2473/commits/eeb1c11e9a61d2c0c3afce18dcdb024ac80357bd

The 3rd commit just adds tests: https://github.com/stellar/freighter/pull/2473/commits/63b28b2244b021e538f1875ce159fe137ecff3a0

aristidesstaffieri: nice lgtm, good call with the explicit xdr check. 💯

aristidesstaffieri (approved):

extension/src/popup/components/send/SendAmount/index.tsx
piyalbasu: This fix closes #2477

PR #2481, opened 2026-01-02
fix memo required check; rm horizon check for destination

piyalbasu:
Fixing a couple of bugs on the release branch:

1. There was a race condition with fetching "memo required" accounts during the Send flow. I've fixed that while also updating a related incorrect test
2. Removes a superfluous call to Horizon when showing a muxed account in Account History. Horizon returns the base G address when you send to/receive from a muxed account/ We need the XDR to show the user the muxed address. The fix that was merged called Horizon to get the payment's xdr by tx hash, but this is actually an unnecessary API call. We already have the XDR; no need to refetch. I should have noticed this during the PR review process, but only caught this while QA'ing. 
3. There was a duplicate `Review Send` button in the Send flow. We had a few big PR's all landing at the same time and when I rebased, I missed this

### DIFF
--- a/extension/src/popup/helpers/__tests__/useValidateTransactionMemo.test.tsx
+++ b/extension/src/popup/helpers/__tests__/useValidateTransactionMemo.test.tsx
@@ -1,0 +1,110 @@
+import React from "react";
+import { Provider } from "react-redux";
+import * as StellarSdk from "stellar-sdk";
+import { renderHook, act } from "@testing-library/react";
+import { useValidateTransactionMemo } from "../useValidateTransactionMemo";
+import { makeDummyStore } from "popup/__testHelpers__";
+import * as ApiInternal from "@shared/api/internal";
+
+describe("useValidateTransactionMemo", () => {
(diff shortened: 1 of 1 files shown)


extension/src/popup/helpers/useValidateTransactionMemo.ts
```diff
@@ -176,7 +176,12 @@ export const useValidateTransactionMemo = (incomingXdr?: string | null) => {
     setIsMemoMissing(true);
     setIsValidatingMemo(false);
 
-    if (!shouldValidateMemo || !xdr || !networkDetails) {
```
aristidesstaffieri, line 179: wdyt about adding a unit test that asserts this condition?

piyalbasu: Sure I can do that. Fwiw, the e2e tests pick this up (they were failing before), but no harm in adding a unit test 👍

CassioMG (approved):

PR #2503, opened 2026-01-09
[FEATURE] Surface transfers in history

### LINKED ISSUE #2317: [Extension] Improve `transfer` interactions in history, using Horizon
- Right now, certain smart contract interactions are not richly communicated in a user's history page. 
- This makes it difficult for users to determine what interactions were done, which can confuse the user and make them believe funds are lost.

- I'd like to see more verbose smart contract interactions in the History tab
- For example, when a user deposits, or withdraws in Near Intents or Blend, I should be able to see the specific smart contract interaction in the History tab (to the extent that that's possible; may require seeing, parsing, and storing signed message details)
- On transaction details in the History page, I'd want to see what transfers (multiple and net) occurred

- Leaving as is, which will increase confusion for the user
- Not relying on Horizon (which is deprecated), and waiting until other solutions that address this issue are implemented

aristidesstaffieri:
Closes #2317 

Adds the use of Horizon's new `asset_balance_changes` in order to surface transfers that happened as part of another operation(Soroban ops).

Design Reference: [Figma](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=8653-21378&p=f&m=dev)

A few notes about the changes:

- the credit/debit summary has a new visual style
- history rows will show "Multiple" instead of the credit/debit amounts in the case of multiple credits/debits
- The to/from field is only visible in the history item detail in the case of having 1 credit/debit, otherwise you only see a stacked list of the credits/debits fro your account.
- small visual tweaks, like changes to padding in asset diff summary and color change for debits

Screenshot 2026-01-09 at 9 19 46 AM

Screenshot 2026-01-09 at 9 20 04 AM

Screenshot 2026-01-09 at 9 20 33 AM

Screenshot 2026-01-09 at 9 21 01 AM

Screenshot 2026-01-09 at 9 21 25 AM

Screenshot 2026-01-09 at 9 21 53 AM

### DIFF
--- a/extension/src/popup/locales/pt/translation.json
+++ b/extension/src/popup/locales/pt/translation.json
@@ -136,7 +136,7 @@
   "Create Account": "Criar Conta",
   "Create Contract": "Criar Contrato",
   "Create New Address": "Criar Novo Endereço",
-  "Create new wallet": "Create new wallet",
+  "Create new wallet": "Criar nova carteira",
   "Current Network": "Rede Atual",
   "Custom": "Personalizado",
   "Dapps": "Dapps",


extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -75,337 +74,228 @@ export const TransactionDetail = ({
...
+      title = `${activeOperation.action} ${activeOperation.rowText}`;
+    } else if (activeOperation.metadata.isSwap) {
+      title = `${activeOperation.action} ${activeOperation.metadata.srcAssetCode} to ${activeOperation.metadata.destAssetCode}`;
+    } else if (activeOperation.metadata.isPayment) {
+      title = `${activeOperation.action} ${activeOperation.rowText}`;
+    } else if (
+      activeOperation.metadata.type ===
+      Horizon.HorizonApi.OperationResponseType.changeTrust
+    ) {
+      title = `${activeOperation.rowText} for ${activeOperation.metadata.destAssetCode}`;
```
piyalbasu: nit: we should translate the word `for`

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -325,6 +331,113 @@ const getIconUrl = async ({
...
+        networkDetails,
+      });
+    } catch (error) {
+      // If decimals cannot be fetched, skip this asset entirely and report to Sentry
+      Sentry.captureException({
+        message: `Failed to fetch decimals for asset ${assetCode} (${assetIssuer}), skipping`,
+        error,
+        assetCode,
+        assetIssuer,
+      });
```
piyalbasu: Is this actually an error? Or is it possible that a user is just interacting with a token that's not compliant with SEP-41? We've seen some users interacting w non-compliant tokens (specifically ones that didn't have `decimals`) in Sentry and the alerts were a little noisy. Especially if they have (for ex) like 15 history rows that reference a token without decimals - that'll trigger an alert every time they view history

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -424,12 +537,10 @@ export const getRowDataByOpType = async (
     "source_amount" in operation ? operation.source_amount : null;
 
   if (isSwap) {
-    const nonLabelAmount = `${formatAmount(
-      new BigNumber(amount!).toString(),
-    )} ${destAssetCode}`;
-    const formattedAmount = `+${nonLabelAmount}`;
+    const nonLabelAmount = formatAmount(new BigNumber(amount!).toString());
```
piyalbasu, line 555: I know you're just copy/pasting this from above, but we should maybe remove the `!` after amount. We've hit a situation a couple of times where we asserted something as non-null and it ended up being null, which caused a runtime error

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -473,27 +585,24 @@ export const getRowDataByOpType = async (
...
+          </span>
           <span>{destAssetCode}</span>
         </div>
       ),
     };
   }
 
   if (isPayment) {
-    // Extract destination from XDR to get muxed address if present
-    const actualDestination = await extractDestinationFromXDR(
```
piyalbasu, line 485: is removing this part of a bad rebase/merge or is there actually a reason to do this? I don't think we want to drop this logic to pull the muxed address from the xdr

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -473,27 +585,24 @@ export const getRowDataByOpType = async (
...
+
     // default to Sent if a payment to self
-    const isReceiving = actualDestination === publicKey && from !== publicKey;
+    const isReceiving = destination === publicKey && sender !== publicKey;
     const paymentDifference = isReceiving ? "+" : "-";
-    const nonLabelAmount = `${formatAmount(
-      new BigNumber(amount!).toString(),
-    )} ${destAssetCode}`;
-    const formattedAmount = `${paymentDifference}${nonLabelAmount}`;
+    const nonLabelAmount = formatAmount(new BigNumber(amount!).toString());
```
piyalbasu, line 619: similar comment as above - should we be asserting as non-null?

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -542,6 +651,45 @@ export const getRowDataByOpType = async (
       rowText: i18n.t("Contract Function"),
     };
 
+    const assetDiffs = await processAssetBalanceChanges(
```
piyalbasu, line 669: oh cool, it looks like even invokeHostFn ops will have `asset_balance_changes` without having to dig into the `invokedContract.args()`. is that right?

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -75,337 +74,228 @@ export const TransactionDetail = ({
...
+    metadata: PaymentMetadata,
+  ): AssetDiffSummary[] => {
+    const { to, from, nonLabelAmount, destIcon, destAssetCode, isReceiving } =
+      metadata;
+
+    return [
+      {
+        assetCode: destAssetCode,
+        assetIssuer: null,
+        decimals: 7,
```
piyalbasu: nit: we have a var for classic decimals called `CLASSIC_ASSET_DECIMALS` in `popup/helpers/soroban`.

You're using it in `getDecimalsForAsset` :)

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -325,6 +331,113 @@ const getIconUrl = async ({
...
+            code: assetCode,
+            networkDetails,
+            homeDomains,
+            icons,
+          });
+
+    // Fetch decimals based on whether it's a Soroban contract
+    let decimals: number;
+    try {
+      decimals = await getDecimalsForAsset({
```
piyalbasu: I believe if a user has multiple interactions with a sep-41 token in their history, you're going to be making a request for token details for each one. It can lead to a lot of requests to the BE for the same info.

Check out `fetchTokenDetails` on ln 736. This utilizes a cache to prevent repeated calls to the BE

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: it's hard to know for sure, I guess all that we can assume is that a contract emitted an event that changed a balance. I can remove this log.

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: `extractDestinationFromXDR` finds the first payment op and returns the destination from that xdr, so when it is used on any other operation in a transaction it is incorrect. 

I didn't realize the top level destination in the op would not be the muxed address, so it seems like the best thing to do here is fix this helper to extract destinations from the correct operation, wdyt?

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: it is right and really that is the case we are most interested in, surfacing balance changes in soroban ops since classic ops are already accounted for in history(as far as being able to represent the balance change).

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: I don't see how to cleanly remove this in this case, `amount` is nullable in `HistoryItemOperation`. In order to change this we would need to refactor `HistoryItemOperation` to be a union of operation types(which it should be) so we can specifically narrow the type down to a type where that key is either `string` or `null`.

I do think this is wrong but imo this is a large refactor outside of the scope of this change. If you agree I can start an issue to track this work.

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
piyalbasu: Ah yeah, this is something that Leo uncovered in his work with muxed addresses. `to` in a payment is the root address, not the muxed. Though it looks like Horizon provides a `to_muxed`. Maybe that's a better way to do this

Screenshot 2026-01-09 at 1 52 18 PM

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
piyalbasu: Ah got it. Yeah, you're right - it's a more fundamental issue here. Another ticket makes sense 👍

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -473,27 +585,24 @@ export const getRowDataByOpType = async (
       rowText: (
         <div className="HistoryItem__description__swap-label">
           <span>{srcAssetCode}</span>
-          <Icon.ArrowRight className="swap-label-direction" />
+          <span className="HistoryItem__description__swap-label__separator">
+            to
```
CassioMG: should we translate this "to" string?

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
CassioMG: this is awesome

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: I created this issue to describe the improvement https://github.com/stellar/freighter/issues/2504

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
```diff
@@ -542,6 +651,45 @@ export const getRowDataByOpType = async (
...
+      operation,
+      publicKey,
+      networkDetails,
+      homeDomains,
+      icons,
+    );
+
+    if (assetDiffs.length > 0) {
+      // Use first asset diff for the row display amount
+      const primaryDiff = assetDiffs[0];
```
CassioMG, line 680: why are we using only the first asset diff here? what happens when there is more than 1 asset diff?

CassioMG: I see we are passing all asset diffs on the TransactionDetails component here: https://github.com/stellar/freighter/pull/2503/changes#diff-dacdc62063df120813b20d41a256cb1a25bb61b226443a9c6cbd1225bedc4023R224

CassioMG: humm I guess we need to elect a `primaryDiff` here for the cases where we have only 1 diff so we are able to display the amount directly on the history row, whereas when we have more than 1 diff we will display  "Multiple" and omit the `primaryDiff` amount. Is my thought correct?

aristidesstaffieri: that's correct

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
```diff
@@ -75,337 +74,228 @@ export const TransactionDetail = ({
...
-            <div className="TransactionDetailModal__title-details">
+  interface CreateAccountMetadata {
+    to?: string;
+    from?: string;
+    nonLabelAmount: string;
+    isReceiving: boolean;
+  }
+
+  // Normalization helper functions
+  const normalizePaymentToAssetDiffs = (
```
CassioMG, line 103: would it make sense to extract the 3 "normalize_" functions to external helpers to keep the UI component lean?

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: thanks, I refactored to use this in 638bb3cf55912363d69a539f715982931f0268ab

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
aristidesstaffieri: thanks, added in 638bb3cf55912363d69a539f715982931f0268ab

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: removed in 638bb3cf55912363d69a539f715982931f0268ab

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
aristidesstaffieri: yeah so I went back and forth on this, left the helpers in this file because in this repo we don't co-locate helpers so helpers that are very specific to a component tend to be in the component file themselves. The helper that these replace(renderBody) was also in this file. 

I tend to like co-located helper files for this use case but it's not the pattern established already. I don't think it makes sense to pull these kinds of helper that will only be used here into general helper files.

Let me know if you still disagree.

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: yes thanks, added in b5a2732d9cb7311442c230495fb7a31fba5ae900

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
aristidesstaffieri: thanks, added in b5a2732d9cb7311442c230495fb7a31fba5ae900

extension/src/popup/views/AccountHistory/hooks/useGetHistoryData.tsx
aristidesstaffieri: ah very cool that makes this easier, I added this in b311e2069f66fdac4387a90b71d229cc00ac8138

extension/src/popup/components/accountHistory/TransactionDetail/index.tsx
CassioMG: Fine by me, I think it makes sense to keep the codebase pattern for consistency

piyalbasu (approved):
Looks like there's some CI tests failing. These should definitely all pass.

Also, I did a little cleanup on the e2e tests, so they should hopefully be a _little_ more reliable right now

PR #2510, opened 2026-01-15
[FEATURE] Hide collectible

### LINKED ISSUE #2331: [Collectibles] Hide Collectible
Let's por this from mobile.

Users should be able to hide/unhide a collectible.

[Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6323-194459&t=i6DQHvjsIvsFy5wX-1):
Image

aristidesstaffieri:
Closes #2331 

Adds hide/unhide workflows for collectibles, as well as a hidden collectibles view.

Design Reference: [Figma](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=6322-191836&m=dev)


extension/src/popup/components/account/AccountCollectibles/index.tsx
```diff
@@ -1,23 +1,53 @@
...
+  collections: Collection[];
+  showHidden: boolean;
+  hiddenCollectibles: Record<CollectibleKey, string>;
+  onCloseCollectible: () => void;
+}) => {
   const { t } = useTranslation();
   const [selectedCollectible, setSelectedCollectible] =
     useState<SelectedCollectible | null>(null);
+  // Keep track of the last selected collectible so we can render it during close animation
+  const lastSelectedCollectible = useRef<SelectedCollectible | null>(null);
```
aristidesstaffieri: The reason this is needed is because the state that controls wether the detail sheet is open/closed is also the state which defines which collectible is to be rendered. Since we lock these 2 concerns together, we have to use a hack in order to get the sheet to behave normally when dismissed(slide down instead of just disappear). 

Imo this implies that we should refactor how we handle sheets into two states(something like `isDetailOpen`, `detailData`) in order to be able to take tighter control of the dismissal. Because this task impacts all sheets and will raise the scope of this task, I suggest that we tackle this task a standalone refactor and keep this hack in this feature's change set.

extension/src/popup/locales/pt/translation.json
```diff
@@ -239,6 +239,8 @@
   "Got it": "Entendi",
   "Hash": "Hash",
   "Help": "Ajuda",
+  "Hidden collectibles": "Hidden collectibles",
+  "Hidden Collectibles": "Hidden Collectibles",
```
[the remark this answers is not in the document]
aristidesstaffieri: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/stellar/freighter/pull/2510#discussion_r2695472147)

extension/src/popup/components/account/AccountCollectibles/index.tsx
piyalbasu: Ah got it. This design made sense for the prior implementation but makes sense why this doesn't make sense when we have to factor in a hidden collectible.

Having 2 states to control is a good path forward. I'm not quite sure how this change affects all sheets, though. Why can't you just do `<Sheet open={isDetailOpen}`? I'm probably just missing something

extension/src/popup/components/account/AccountCollectibles/index.tsx
```diff
@@ -29,97 +58,161 @@ const CollectionsList = ({ collections }: { collections: Collection[] }) => {
     );
   }
 
-  return collections.map(({ collection, error }) => {
-    // if the collection is missing or has an error, skip rendering
-    if (error || !collection) {
-      return null;
+  // Get the collectible data for the detail view
+  const getCollectibleForDetail = () => {
```
piyalbasu: hm, why do we need to another loop here to find the right collectible? Is there a reason we can't do this (for ex) in the `onClick` on ln 137? It looks like you should have all the necessary collectible data there. This would prevent looping over all the collectibles twice plus I believe you only need to store the collectible detail for a a collection that's showing. This loop checks all collections, regardless of show/hide state

extension/src/popup/components/account/HiddenCollectibles/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  }, [selectedCollectible]);
+
+  // Refresh hidden collectibles when sheet opens
+  useEffect(() => {
+    if (isOpen) {
+      refreshHiddenCollectibles();
+    }
+  }, [isOpen, refreshHiddenCollectibles]);
+
+  const isCollectibleHidden = (collectionAddress: string, tokenId: string) => {
```
piyalbasu: this helper is repeated on ln 41 of AccountCollectibles/index.tsx. Let's re-use the logic

extension/src/background/messageListener/handlers/changeCollectibleVisibility.ts
```diff
@@ -0,0 +1,21 @@
...
+  localStore,
+}: {
+  request: ChangeCollectibleVisibilityMessage;
+  localStore: DataStorageAccess;
+}) => {
+  const { collectibleVisibility } = request;
+
+  const { hiddenCollectibles } = await getHiddenCollectibles({ localStore });
+  hiddenCollectibles[collectibleVisibility.collectible] =
+    collectibleVisibility.visibility;
```
piyalbasu, line 17: this is pretty minor - but wdyt about making this a boolean rather than a string? For now, we only care about where it is hidden or not hidden. It prevents you from having to check `const isHidden = hiddenCollectibles[key] === "hidden"` down below.

aristidesstaffieri: the main reason for using this pattern is that it follows what we do for tokens. I agree that a boolean is better for this use case but is there a difference in the way we hide collectibles vs tokens? I think it would be better to maintain the same pattern for both vs break away from the pattern here.

extension/src/popup/components/account/HiddenCollectibles/index.tsx
aristidesstaffieri: thanks, I combined the helpers and refactored the usages of it in a427cc96ba8db685f7d411f8aa37c44ba139ed0d

extension/src/popup/components/account/AccountCollectibles/index.tsx
aristidesstaffieri: thanks this makes sense, removed the redundant logic in ea9d4d68eb485b92f87846f78c543b8addb385fe

extension/src/background/messageListener/handlers/changeCollectibleVisibility.ts
aristidesstaffieri: yeah good point, even though other sheets follow this pattern we can just fix it for this one now since the other sheets are not dismissed under the same conditions.

I refactored this in 3fc2b20b938f4cb308581055ee132c97b3d384d0

piyalbasu: > the main reason for using this pattern is that it follows what we do for tokens. I agree that a boolean is better for this use case but is there a difference in the way we hide collectibles vs tokens? I think it would be better to maintain the same pattern for both vs break away from the pattern here.

Yeah, if this is how tokens are doing it this way, let's follow the same pattern. This approach allows us to scale more easily if we want to introduce an additional state here

PR #2512, opened 2026-01-15
use ping endpoint to check rpc health

### LINKED ISSUE #2342: Restore rpc-health check

piyalbasu:
Closes #2342 

We were hardcoding the RPC healthy value because our RPC health checks previously were not working properly. Using the v2 [health check](https://github.com/stellar/freighter-backend-v2/pull/64), we can now restore this notification.

Screenshot 2026-01-16 at 11 10 26 AM

# Summary

This pull request enhances the way the extension checks and reports the health of the Soroban RPC service, aligning the backend and frontend logic with the latest API and improving test coverage. The main changes include updating the RPC health check to use a new endpoint, adjusting the backend settings loader to use this updated check, and adding an end-to-end test to verify the notification appears when the RPC is unhealthy.

**RPC Health Check Improvements:**

* Updated `getIsRpcHealthy` in `account.ts` to use the new `INDEXER_V2_URL` and the `/ping?network=*` endpoint, and to extract the `rpc_health` field from the response. The function now takes a `localStore` parameter and retrieves network details internally. [[1]](diffhunk://#diff-337c0386fa91dd3917d65696501c2ab7b726af5b72dfda3615d07a8a49293e65L24-R24) [[2]](diffhunk://#diff-337c0386fa91dd3917d65696501c2ab7b726af5b72dfda3615d07a8a49293e65L300-R319)
* Updated the `loadBackendSettings` handler to call `getIsRpcHealthy` with the correct parameters and return the actual RPC health status instead of a hardcoded value.
* Modified the popup message listener to pass the `localStore` to `loadBackendSettings` for proper context.

**Testing Enhancements:**

* Added a new end-to-end test (`rpcHealthNotification.test.ts`) that stubs the RPC health endpoint as unhealthy and verifies that the appropriate notification is shown in the UI.

### DIFF
--- a/extension/src/background/helpers/account.ts
+++ b/extension/src/background/helpers/account.ts
@@ -313,9 +313,10 @@ export const getIsRpcHealthy = async (localStore: DataStorageAccess) => {
 
       if (!res.ok) {
         captureException(`Failed to load rpc health for Soroban`);
+        rpcHealth = { status: "unhealthy" };
+      } else {
+        rpcHealth = await res.json();
       }
-
-      rpcHealth = await res.json();
     } catch (e) {
       captureException(
         `Failed to load rpc health for Soroban - ${JSON.stringify(e)}`,


extension/src/background/messageListener/handlers/loadBackendSettings.ts
```diff
@@ -1,11 +1,17 @@
...
-export const loadBackendSettings = async () => {
+export const loadBackendSettings = async ({
+  localStore,
+}: {
+  localStore: DataStorageAccess;
+}) => {
   const featureFlags = await getFeatureFlags();
-  const isRpcHealthy = true;
+  const isRpcHealthy = await getIsRpcHealthy(localStore);
   const userNotification = await getUserNotification();
```
piyalbasu, line 15: these settings are loaded async after balances have already been loaded. This ensures that this call doesn't slow down time to largest contentful paint.

aristidesstaffieri: @piyalbasu @sdfcharles what are your thoughts about making these notifications render outside of the layout(like mobile does or general toast patterns)? 

My concern here is that rendering it where it is now will cause a layout shift, which is especially frustrating when it happens async and the user may already be interacting with the layout.

piyalbasu: Yeah, that's a good point. I can implement shadcn's toast here instead

piyalbasu: Toast added here: https://github.com/stellar/freighter/pull/2512/commits/05e5d5da0a9fe7c5ab183f5bf5d930396189a8df


https://github.com/user-attachments/assets/234f37a3-7ba3-47aa-95dc-ea9ec3ed6ee4

extension/src/background/helpers/account.ts
```diff
@@ -297,20 +297,24 @@ export const getIsHideDustEnabled = async ({
...
+  const networkDetails = await getNetworkDetails({ localStore });
+
   if (isCustomNetwork(networkDetails)) {
     // TODO: use server.getHealth method to get accurate result for standalone network
     rpcHealth = { status: "healthy" };
   } else {
     try {
       const res = await fetch(
-        `${INDEXER_URL}/rpc-health?network=${networkDetails.network}`,
+        `${INDEXER_V2_URL}/rpc-health?network=${networkDetails.network}`,
```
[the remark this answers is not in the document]
piyalbasu: Verified

piyalbasu:
This PR got messy with rebasing - re-rolling this in #2577

PR #2564, opened 2026-01-29
fix: sign-transactions balances call should fail silently

### LINKED ISSUE #2437: If Account Balances fetch fails, Sign TX doesn't handle error
Related to the issue here: https://stellarorg.sentry.io/issues/7086506632/?referrer=slack&notification_uuid=99f50784-b087-49b9-a299-6a829e53699f&alert_rule_id=15034055&alert_type=issue

A user wasn't able to load `/account-balances` (possibly due to network connectivity issues?) in Sign Transaction. When this happens, we throw an ugly runtime error

Let's make this UI more resilient to being unable to fetch API's. A user should still be able to sign a tx in this case as having network connectivity isn't required to sign a tx and transmit the signed tx to a dapp

We should just show a warning that we weren't able to load data if anything in useSigntxData fails

aristidesstaffieri:
Closes #2437 

Why
If a user is in the sign-transaction flow and the call to fetch their balances fails, we fail harshly and block them from completing the transaction.

What
This handles the failure silently and allows the user to continue in their flow, the only side effect is that we cannot warn them about a potential insufficient balance.

### DIFF
--- a/extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
+++ b/extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
@@ -28,7 +28,7 @@ import { AssetListResponse } from "@shared/constants/soroban/asset-list";
 export interface ResolvedData {
   type: AppDataType.RESOLVED;
   scanResult: BlockAidScanTxResult | null;
-  balances?: AccountBalances;
+  balances: AccountBalances | null;
   publicKey: string;
   signFlowState: {
     allAccounts: Account[];
@@ -90,7 +90,7 @@ function useGetSignTxData(
 
       // Fetch balances with soft failure handling - if this fails, we continue
       // without balance data (balance-related warnings will be skipped)
-      let balancesResult: AccountBalances | undefined;
+      let balancesResult: AccountBalances | null = null;
       try {
         const fetchResult = await fetchBalances(
           publicKey,
@@ -99,13 +99,11 @@ function useGetSignTxData(
           false,
           true,
         );
-        if (isError<AccountBalances>(fetchResult)) {
-          balancesResult = undefined;
-        } else {
+        if (!isError<AccountBalances>(fetchResult)) {
           balancesResult = fetchResult;
         }
       } catch {
-        balancesResult = undefined;
+        // Balance fetch failed - continue without balance data
       }
 
       // handle auto selecting the right account based on `accountToSign`
--- a/extension/src/popup/views/__tests__/SignTransaction.test.tsx
+++ b/extension/src/popup/views/__tests__/SignTransaction.test.tsx
@@ -1102,7 +1102,7 @@ describe("SignTransactions", () => {
             request_id: "1",
           },
           icons: {},
-          balances: undefined, // Balances unavailable due to fetch failure
+          balances: null, // Balances unavailable due to fetch failure
           publicKey: mockAccounts[1].publicKey,
           signFlowState: {
             allAccounts: mockAccounts,
--- a/extension/src/popup/views/SignTransaction/hooks/__tests__/useGetSignTxData.test.tsx
+++ b/extension/src/popup/views/SignTransaction/hooks/__tests__/useGetSignTxData.test.tsx
@@ -343,11 +343,11 @@ describe("useGetSignTxData", () => {
       await result.current.fetchData();
     });
 
-    // Should still succeed but with undefined balances
+    // Should still succeed but with null balances
     expect(result.current.state.state).toBe<RequestState>(RequestState.SUCCESS);
     expect(
-      (result.current.state.data as { balances?: unknown })?.balances,
-    ).toBeUndefined();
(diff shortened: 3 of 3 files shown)


extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -1089,4 +1089,119 @@ describe("SignTransactions", () => {
...
+        state={{
+          auth: {
+            allAccounts: mockAccounts,
+            publicKey: mockAccounts[0].publicKey,
+          },
+          settings: {
+            isExperimentalModeEnabled: true,
+            networkDetails: {
+              ...defaultSettingsState.networkDetails,
+              networkPassphrase: "Test SDF Future Network ; October 2022",
```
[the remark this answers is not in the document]
aristidesstaffieri: fixed in 0b7b207916297ff945ddaf4cc5367f59c987b1fe

extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -1089,4 +1089,119 @@ describe("SignTransactions", () => {
...
+    });
+    jest
+      .spyOn(SignTxDataHooks, "useGetSignTxData")
+      .mockReturnValue(currentSignTxDataMock);
+
+    jest.spyOn(ApiInternal, "loadSettings").mockImplementation(() =>
+      Promise.resolve({
+        networkDetails: {
+          ...defaultSettingsState.networkDetails,
+          networkPassphrase: "Test SDF Future Network ; October 2022",
```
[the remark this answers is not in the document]
aristidesstaffieri: fixed in 0b7b207916297ff945ddaf4cc5367f59c987b1fe

extension/src/popup/views/__tests__/SignTransaction.test.tsx
```diff
@@ -1089,4 +1089,119 @@ describe("SignTransactions", () => {
...
+          publicKey: mockAccounts[1].publicKey,
+          signFlowState: {
+            allAccounts: mockAccounts,
+            accountNotFound: false,
+            currentAccount: mockAccounts[0],
+          },
+          applicationState: APPLICATION_STATE.MNEMONIC_PHRASE_CONFIRMED,
+          networkDetails: {
+            ...defaultSettingsState.networkDetails,
+            networkPassphrase: "Test SDF Future Network ; October 2022",
```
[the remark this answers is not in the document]
aristidesstaffieri: fixed in 0b7b207916297ff945ddaf4cc5367f59c987b1fe

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -87,13 +87,26 @@ function useGetSignTxData(
...
-        publicKey,
-        isMainnetNetwork,
-        networkDetails,
-        false,
-        true,
-      );
+
+      // Fetch balances with soft failure handling - if this fails, we continue
+      // without balance data (balance-related warnings will be skipped)
+      let balancesResult: AccountBalances | undefined;
```
piyalbasu: Super NIT: maybe it's better to use `AccountBalances | null`, to make it more clear that we attempted to fetch but something went wrong

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
```diff
@@ -87,13 +87,26 @@ function useGetSignTxData(
...
+          true,
+        );
+        if (isError<AccountBalances>(fetchResult)) {
+          balancesResult = undefined;
+        } else {
+          balancesResult = fetchResult;
+        }
+      } catch {
+        balancesResult = undefined;
+      }
```
piyalbasu, line 107: what do you think about moving this balance fetch after `firstRenderPayload` is dispatched? This would move the wait for the balances out of the critical path, allowing for the view to load just a bit faster.

The only downside is that displaying the has-enough-xlm warning would be async, so we'd maybe want to animate it in.

Just a thought!

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
aristidesstaffieri: sure thing, done in 0b266b8e1fe2b3dd09d3666b645e324e8b507287

extension/src/popup/views/SignTransaction/hooks/useGetSignTxData.tsx
aristidesstaffieri: I agree with this approach but this introduces a design change and optimization unrelated to this bug. Wdyt about capturing this separately from this PR so that it can go through the proper design/dev lifecycle. Details like animation/transition aren't just syntax and need design input imo.

piyalbasu: Yeah, good call. New ticket sounds good

PR #2580, opened 2026-02-09
Adjust unfunded account errors

### LINKED ISSUE #2549: Adjust details of failure when sending non-XLM to unfunded account
## Summary
- Update the unfunded-destination details text to: “This is a new account and needs 1 XLM in order to get started. Any transaction to send non-XLM to an unfunded account will fail.”
- Keep the inline toast/label copy: “Transaction is expected to fail.”
- Ensure the “Learn more” text is visually clear as a link (SDS Link component styling).

## Tasks
- Replace the unfunded-destination details copy with the new text.
- Confirm the inline toast/label uses “Transaction is expected to fail.”
- Adjust the “Learn more” styling so it’s obviously clickable.

leofelix077:
Closes #2549 

Screenshot 2026-02-10 at 17 05 18
Screenshot 2026-02-10 at 17 07 44
Screenshot 2026-02-10 at 17 07 55

### DIFF
--- a/extension/src/popup/helpers/blockaid.ts
+++ b/extension/src/popup/helpers/blockaid.ts
@@ -249,8 +249,11 @@ export const useBlockaidOverrideState = () =>
  */
 export const useIsAssetSuspicious = () => {
   const blockaidOverrideState = useBlockaidOverrideState();
-  return (blockaidData?: BlockAidScanAssetResult | null) =>
-    isAssetSuspicious(blockaidData, blockaidOverrideState);
+  return useCallback(
+    (blockaidData?: BlockAidScanAssetResult | null) =>
+      isAssetSuspicious(blockaidData, blockaidOverrideState),
+    [blockaidOverrideState],
+  );
 };
 
 export const isAssetMalicious = (
@@ -272,8 +275,11 @@ export const isAssetMalicious = (
  */
 export const useIsAssetMalicious = () => {
   const blockaidOverrideState = useBlockaidOverrideState();
-  return (blockaidData?: BlockAidScanAssetResult | null) =>
-    isAssetMalicious(blockaidData, blockaidOverrideState);
+  return useCallback(
+    (blockaidData?: BlockAidScanAssetResult | null) =>
+      isAssetMalicious(blockaidData, blockaidOverrideState),
+    [blockaidOverrideState],
+  );
 };
 
 /**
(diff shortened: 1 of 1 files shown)


extension/e2e-tests/helpers/login.ts
```diff
@@ -111,9 +111,10 @@ export const loginToTestAccount = async ({
   if (context && !isIntegrationMode) {
     // Wait for any background activity to complete
     await stubAllExternalApis(page, context);
-    if (stubOverrides) {
-      await stubOverrides();
-    }
+  }
+  // Always apply stubOverrides, even in integration mode, as tests explicitly request these stubs
+  if (stubOverrides) {
```
leofelix077: moved it outside the main if-case to allow overrides for all cases

piyalbasu: doing this defeats the purpose of having an integration mode. It does not allow us to test Freighter BE in any capacity. Why do this?

extension/src/popup/components/send/SendAmount/hooks/useSimulateTxData.tsx
```diff
@@ -73,6 +73,78 @@ export interface SimulateTxData {
...
+  isDestinationFunded?: boolean;
+  assetCanonical: string;
+  amount: string;
+}) => {
+  if (isDestinationFunded !== false) {
+    return null;
+  }
+
+  if (assetCanonical !== "native") {
+    return "Blockaid unfunded destination";
```
piyalbasu: let's make it clear that these are the keys to the translation and will actually map to a fuller description

extension/e2e-tests/helpers/login.ts
leofelix077: I understood this was intended for all cases (e.g. known to be flaky APIs or unimportant calls, to allow it for override in all cases so the main points can be tested. e.g. if I'm testing transaction, I could stub blockaid response, without interference etc ). but can revert it, if that's not the case

piyalbasu: Yes, the design is that all tests, by default, use stubs. But, for certain tests, we want to be able to actually test against live API's only when the integration mode flag is set to true in CI ( we only do this on the branches that will be merged to master - not regular PR's). Therefore, it's actually important to only use the stubs when the integration mode flag is set to false. 

Here's the previously shared design doc that explains this in further detail: https://docs.google.com/document/d/16jNoGSiQRW4MtYEzsrxXhR72D_K0_UkwZM4gOYeoFhw/edit?tab=t.0#heading=h.ubnp7vygs2ah

extension/e2e-tests/blockaidScan.suspicious.test.ts
```diff
@@ -288,18 +298,25 @@ test.describe("BlockAid Scan - Suspicious States", () => {
...
-    await stubScanTxSuspicious(page);
-    await stubMemoRequiredAccounts(
+    await loginToTestAccount({
       page,
-      "GA6SXIZIKLJHCZI2KEOBEUUOFMM4JUPPM2UTWX6STAWT25JWIEUFIMFF",
-    );
-    await loginToTestAccount({ page, extensionId });
+      extensionId,
+      context,
+      stubOverrides: async () => {
```
leofelix077, line 308: adjusted the formats here to use the stubOverrides to make it consistent with the other tests. was failing on CI

PR #2607, opened 2026-02-18
feat(background): add TTL-based cleanup for message queues

aristidesstaffieri:
Transaction, token, blob, and auth entry queues have no TTL or cleanup mechanism. Orphaned entries from closed popups accumulate indefinitely.

Changes:
 - Add createdAt timestamp to queue item types (ResponseQueueItem, TransactionQueueItem, BlobQueueItem, AuthEntryQueueItem, TokenQueueItem)
 - Create queueCleanup utility with cleanupQueue() and startQueueCleanup()
 - Set 5-minute TTL for queue items with 1-minute cleanup interval
 - Initialize periodic cleanup in popupMessageListener on module load
 - Update freighterApiMessageListener to include createdAt on all queue pushes
 - Update addToken handler for new TokenQueueItem structure
 - Add tests for queue cleanup functionality

### DIFF
--- a/extension/e2e-tests/integration-tests/freighterApiIntegration.test.ts
+++ b/extension/e2e-tests/integration-tests/freighterApiIntegration.test.ts
@@ -2,7 +2,14 @@ import { expect, test, expectPageToHaveScreenshot } from "../test-fixtures";
 import { TEST_TOKEN_ADDRESS } from "../helpers/test-token";
 import { loginToTestAccount } from "../helpers/login";
 import { allowDapp } from "../helpers/allowDapp";
-import { stubAccountBalances, stubIsSac } from "../helpers/stubs";
+import {
+  stubAccountBalances,
+  stubAccountHistory,
+  stubIsSac,
+  stubScanDapp,
+  stubTokenDetails,
+  stubTokenPrices,
+} from "../helpers/stubs";
 
 const TX_TO_SIGN =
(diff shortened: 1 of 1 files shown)


aristidesstaffieri:
> A couple of thoughts on this one:
> 
> 1. My understanding is if a queue is cleaned up while a user (for ex) has a sign transaction modal open, it will fail silently because the handler will not have a transaction to sign in the queue anymore. We should surface an error message to the user to let them know that the signing request expired
> 2. I think the 5 min wait to clean up generally makes sense, but we should chat about this with product. I'm wondering if it's a bit too restrictive or if we're missing some cases where a user might want to take longer to consider signing. If I had to redo a flow on a dapp because I took (for ex) 6 min to confirm, I'd be pretty annoyed. IMO, since this is "low" severity - we should err on a longer waiting period

Yeah this is a good point, I see that the messages are in the queue even after being consumed by the modal action and need to be there through the confirmation action. A few possible solutions - 

```
1. Surface the error to the user 

  Modify useSetupSigningFlow to check the response and show an error instead of silently closing:

   const signAndClose = async () => {
     // ...
     const result = await dispatch(signFn({ apiVersion, uuid } as Arg));
     if (result.payload?.error) {
       // Show error to user (e.g., "This signing request has expired")
       return;
     }
     window.close();
   };

  This doesn't prevent the issue but at least tells the user what happened.

  2. Don't clean up items with open popups

  Track which requests have active popups:

   - When popup opens, send a message to background marking that uuid as "active"
   - Cleanup skips active items
   - When popup closes (confirm/reject/window close), mark inactive

  3. Popup refreshes its own TTL

  When the popup loads, it could send a "heartbeat" message to the background that updates createdAt for its uuid. This resets the 5-min clock whenever the user is actively viewing the
  modal.
```

I think option 2 is the cleanest but I could also see the value in keeping it simple and using a much longer ttl.

PR #2625, opened 2026-03-02
Fix: only show "unable to scan" warning on mainnet in ChangeTrustInte…

piyalbasu:
Fix: only show "unable to scan" warning on mainnet in ChangeTrustInternal

Asset scanning is not supported on non-mainnet networks, so scanAsset() returns null, which was incorrectly triggering the "unable to scan" warning. Gate the warning behind an isMainnet() check so it only appears when scanning is actually expected to work.

This pull request updates the asset management flow to ensure that the "unable to scan" security warning is only shown on mainnet, not on testnet. It adds new tests to verify this behavior and refactors the logic in the `ChangeTrustInternal` component to check the network before displaying the warning.

**Asset scan warning logic improvements:**

* The logic for showing the "unable to scan" warning in `ChangeTrustInternal` was updated to only display the warning on mainnet, using the new `isMainnet` helper and the `networkDetails` prop. [[1]](diffhunk://#diff-6762f03c9434fe2fcaa9297d49e97e91e7119d01186dbd0075fd27ce8836dd64R10) [[2]](diffhunk://#diff-6762f03c9434fe2fcaa9297d49e97e91e7119d01186dbd0075fd27ce8836dd64R168-R173)
* The conditional rendering for the warning pane was updated to use the new mainnet-specific check.

**Test coverage enhancements:**

* Two new tests were added to `ManageAssetRows.test.tsx` to verify that the "unable to scan" warning appears on mainnet but not on testnet when the asset scan returns null.
* The test file now imports both `MAINNET_NETWORK_DETAILS` and `TESTNET_NETWORK_DETAILS` to support these new tests.

### DIFF
--- a/extension/e2e-tests/helpers/login.ts
+++ b/extension/e2e-tests/helpers/login.ts
@@ -162,7 +162,8 @@ export const loginToTestAccount = async ({
 export const switchToMainnet = async (page: Page) => {
   await page.getByTestId("network-selector-open").click();
   await page.getByText("Main Net").click();
-  await expect(page.getByTestId("account-view")).toBeVisible({
-    timeout: 30000,
-  });
+  await expect(page.getByTestId("network-selector-open")).toContainText(
+    "Main Net",
+    { timeout: 30000 },
+  );
 };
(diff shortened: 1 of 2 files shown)


piyalbasu:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/stellar/freighter/pull/2625#pullrequestreview-3878992534)

PR #2649, opened 2026-03-12
Improve fees display on send

### LINKED ISSUE #2475: Improve minResourceFee display in Send flow
In the Send flow, a user can send either a classic payment or send a Soroban token/collectible. In the former case, we show the user the suggested base fee (which is generally enough to get the tx onto the network). 

In the latter, we show the suggested fee base, but then we simulate the tx to get the minimum resource fee, and then add that to the user's base fee. This change happens quickly in the UI and we don't explain the changing value to the user. 

What are our options to improve this?

Maybe show no fee for a Soroban tx until simulation?

Related discussion: https://github.com/stellar/freighter/pull/2473#discussion_r2640366849

leofelix077:
closes #2475 

https://github.com/user-attachments/assets/3ada1b6a-aa32-4d52-9a64-4406593646b9

https://github.com/user-attachments/assets/dd5cffa3-218a-4526-9bcb-73290a7a01b4


@shared/api/internal.ts
```diff
@@ -2208,6 +2208,7 @@ export const simulateTransaction = async (args: {
     body: JSON.stringify({
       xdr,
       network_passphrase: networkDetails.networkPassphrase,
+      network_url: networkDetails.sorobanRpcUrl,
```
piyalbasu: why was this change needed? It's actually only used for Soroswap, which is disabled right now

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -785,15 +858,17 @@ export const SendAmount = ({
           />
         </>
       ) : null}
-      {isEditingSettings ? (
+      {isEditingSettings && !isShowingFeesPane ? (
```
piyalbasu, line 969: 1 small UX improvement we can make here. I noticed if you edit the fee, then click the inclusion fee i bubble, it resets the fee field. A user would lose any change they've made in this field

https://github.com/user-attachments/assets/22459fbb-2ec4-45ea-b0c6-7352b78ea471

CassioMG:
## PR Review: Improve fees display on send (Extension)

Cross-reference: This review covers this PR alongside the mobile counterpart at stellar/freighter-mobile#774. A comparison section is included at the bottom.

---

### Individual Review of Extension PR

**Overall**: Clean, well-structured implementation. The `FeesPane` component is simple and reusable, the pane state machine extension is elegant, and the E2E test coverage is excellent.

#### Things I Like
- **Comprehensive E2E tests** (358 lines) covering token send, classic send, collectible send, EditSettings flow, auto-simulation, and re-simulation with destination change. This is significantly stronger test coverage than the mobile PR.
- **`SimulateTxData` consolidated into `types/transactions.ts`** — removes duplicate interface definitions from two separate hooks.
- **`lastInclusionFeeRef` pattern** — persists the inclusion fee across re-simulations so the EditSettings input doesn't flash back to the total fee. Well thought out.
- **Base fee reset before re-simulation** in both `handleContinue` and the auto-simulate `useEffect` — prevents the compounding fee inflation bug where total gets used as baseFee.
- **`simulationDataRef` for change detection** — clean approach to detect destination/asset changes without watching `simulationState.data` (which would cause loops).
- **`FeesPane` takes `simulationState` directly** — it can determine Soroban vs classic from the data itself (`resourceFee` presence), which is simpler than the mobile's dual `isSoroban`/`isSorobanContext` approach.
- **`enableReinitialize` on Formik** — ensures EditSettings reflects updated fee values on re-open.

#### Issues & Suggestions

**1. Auto-simulation `useEffect` depends on `simulationState.state`** ⚠️
```ts
useEffect(() => {
  if (simulationState.state === RequestState.LOADING) return;
  // ...
  if (destChanged || assetChanged) {
    fetchSimulationData();
  }
}, [destination, asset, isToken, isCollectible, simulationState.state]);
```
Including `simulationState.state` in deps means this effect re-runs on every state transition (IDLE → LOADING → SUCCESS). The `simulationDataRef` guard prevents duplicate calls in the happy path, but if a simulation _fails_ (`ERROR` state), the effect re-runs, sees the same destination/asset (already tracked in the ref), and does nothing — the user is stuck without a retry mechanism. Consider adding `ERROR` handling or a manual retry path.

**2. `FeesPane` determines Soroban vs classic based on `simulationState.data?.resourceFee`** 📝
```tsx
{simulationState.data?.resourceFee
  ? t("Fees description soroban")
  : t("Fees description classic")}
```
This works for the fee breakdown description, but has a subtle issue: while loading (`data` is null), the description shows the classic text even for a Soroban transaction. For the mobile PR, an explicit `isSorobanContext` prop solves this. Consider passing a similar hint so the description is correct during loading states too.

**3. `FeesPaneOverlay` style class is defined but unused** 🧹
In `styles.scss`, the `.FeesPaneOverlay` class is defined (13 lines) but never referenced in any component. Looks like a leftover from an earlier approach — can be removed.

**4. Info button always shown on ReviewTransaction fee row** 📝
```tsx
<button
  className="ReviewTx__Details__Row__FeesInfoBtn"
  data-testid="review-tx-fee-info-btn"
  onClick={() => setActivePaneIndex(paneConfig.feesIndex)}
>
  <Icon.InfoCircle />
</button>
```
The info circle button is always rendered (both Soroban and classic). This is fine since the `FeesPane` adapts its content, but it's a design choice worth confirming — the mobile PR only shows the info icon for Soroban transactions on the review screen.

**5. `network_url` added to simulate-tx API call** 📝
```ts
body: JSON.stringify({
  xdr,
  network_passphrase: networkDetails.networkPassphrase,
  network_url: networkDetails.sorobanRpcUrl,
}),
```
Small but important change in `@shared/api/internal.ts`. Worth noting this is a backend contract change — if the backend doesn't expect `network_url`, it should be harmless (ignored), but worth confirming the backend accepts this field.

**6. Fee display shows "Calculating..." only for Soroban** ✅
Good UX — classic transactions don't need this since they don't require simulation. The `isLoading` check in the `FeesPane` also correctly treats `IDLE` as loading.

---

### Cross-Platform Comparison: Extension (#2649) vs Mobile (#774)

Both PRs implement the same feature (Soroban fee breakdown: Inclusion Fee + Resource Fee + Total Fee) across the two Freighter platforms. Here's how they differ:

| Aspect | Extension (#2649) | Mobile (#774) |
|--------|-------------------|---------------|
| **State management** | Redux + `useRequest` pattern (`simulationState` with `RequestState` enum) | Zustand store (`useTransactionBuilderStore`) with `isSoroban`, `sorobanResourceFeeXlm`, `sorobanInclusionFeeXlm` |
| **Fee data location** | `SimulateTxData` interface (shared type) carries `inclusionFee?` and `resourceFee?` | Stored directly in Zustand store; computed in components via `BigNumber` |
| **Soroban detection (FeesPane)** | Inferred from `simulationState.data?.resourceFee` presence | Dual system: `isSoroban` from store + `isSorobanContext` prop from parent context |
| **Fee breakdown component** | `FeesPane` — SCSS-styled, integrated into `MultiPaneSlider` | `FeeBreakdownBottomSheet` — NativeWind/Tailwind, uses `@gorhom/bottom-sheet` |
| **Auto-simulation trigger** | `useEffect` watching `destination` + `asset` with `simulationDataRef` for dedup | `useEffect` with `lastAutoSimulatedKey` ref; uses amount `"0"` for early estimation |
| **Fee estimation approach** | Simulates with the real amount; auto-triggers when destination changes | Simulates with amount `"0"` to get fees before user enters amount; `skipAmountValidation` flag |
| **Base fee inflation prevention** | `lastInclusionFeeRef` + explicit `saveTransactionFee` reset before re-simulation | Inclusion fee derived from `params.transactionFee \|\| MIN_TRANSACTION_FEE` in the store |
| **Total fee computation** | Computed in `simulateTx` → `recommendedFee = baseFee + minResourceFee`; single source of truth | Computed independently in both `FeeBreakdownBottomSheet` and `SendReviewBottomSheet` (duplicated) |
| **EditSettings fee label** | "Inclusion Fee" for Soroban, "Transaction Fee" for classic; info button opens FeesPane | "Inclusion fee" for Soroban, "Transaction fee" for classic; info button opens FeeBreakdownBottomSheet |
| **Info icon on review screen** | Always shown (classic and Soroban) | Only shown for Soroban transactions |
| **Simulation return type** | Unchanged — `useSimulateTxData` hook returns data via existing pattern | **Breaking change**: `simulateContractTransfer` and `simulateCollectibleTransfer` now return `{ preparedTransaction, minResourceFee }` instead of raw XDR string |
| **Error handling** | Existing patterns | New `extractErrorMessage` helper for plain-object API errors |
| **Test coverage** | **358-line E2E test suite** (6 test cases including re-simulation) | Unit tests updated for store + service changes; no E2E tests |
| **i18n** | EN + PT translations | EN + PT translations (matching keys/descriptions) |
| **Loading state** | "Calculating..." text in FeesPane and fee display | `ActivityIndicator` spinner in bottom sheet and fee row |

#### Key Architectural Differences Worth Aligning

1. **Total fee computation**: The extension computes the total fee in one place (`recommendedFee` in the simulation hook). The mobile computes it in two components independently. The extension approach is DRY-er — consider centralizing this in the mobile's Zustand store.

2. **Fee estimation with amount 0**: The mobile PR introduces a `feeEstimationAmount` parameter and `skipAmountValidation` to simulate with amount 0 before the user enters a value. The extension doesn't do this — it only simulates when the destination changes (using the real amount). This is a meaningful UX difference: the mobile user sees estimated fees earlier, but the extension approach avoids the complexity of dummy builds.

3. **Soroban detection strategy**: The extension's `FeesPane` infers Soroban from the data (`resourceFee` presence), which is simpler but shows classic text during loading. The mobile uses an explicit `isSorobanContext` prop, which is more correct during loading but adds complexity. Consider aligning on one approach.

4. **Breaking vs non-breaking simulation API**: The mobile changes `simulateContractTransfer`'s return type (breaking), while the extension adds `inclusionFee`/`resourceFee` to the existing `SimulateTxData` type (non-breaking). The extension approach is safer.

5. **Review screen info icon visibility**: The extension always shows the fee info icon; the mobile only shows it for Soroban. Worth deciding on a consistent approach across platforms.

6. **Unrelated change in mobile PR**: The cancel button logic change in `SendReviewFooter` (`tertiary`/`destructive` conditions) doesn't appear in the extension PR and seems unrelated to fee breakdown.

---
_Generated by [Claude Code](https://claude.ai/code)_

@shared/api/internal.ts
leofelix077: it was throwing an error when trying to submit a collectible, mentioning it's a required field by the backend. on my MP collectibles

extension/src/popup/components/send/SendAmount/index.tsx
leofelix077: nice one! 

I adjusted the fee persistence inside the flow, so if user manually edits it, it's kept inside the context (soroban vs classic) and won't change again after explicit user input, unless user exits the transaction flow 


https://github.com/user-attachments/assets/d8bcde91-7b60-4413-bdf8-3cc92243b36d

leofelix077:
@CassioMG @piyalbasu I adjusted the last points on the code and checked the Claude comments

most of the comments are related to data source and state management, which are inherently different on mobile vs extension, and some of the UI parts that are also different on Mobile (e.g. on mobile we have the info sheet for the transaction fee, on extension we always show the fees breakdown)

Funcionality-wise they should be in parity now

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  fee,
+  simulationState,
+  isSoroban = false,
+  onClose,
+}: FeesPaneProps) => {
+  const { t } = useTranslation();
+
+  const isLoading =
+    simulationState.state === RequestState.IDLE ||
+    simulationState.state === RequestState.LOADING;
```
CassioMG: **Critical — FeesPane renders a misleading Total on `RequestState.ERROR`**

```ts
const isLoading =
  simulationState.state === RequestState.IDLE ||
  simulationState.state === RequestState.LOADING;
```

`ERROR` is neither `IDLE` nor `LOADING`, so `!isLoading` is true — but `simulationState.data` is null. The Inclusion Fee row (line 48) and Resource Fee row (line 62) gate on `simulationState.data?.inclusionFee` / `.resourceFee` and hide, while the Total Fee row at line 76-82 hits the `${fee} XLM` branch.

Result: when Soroban simulation fails, FeesPane silently renders a "Total Fee" that is just the base inclusion fee (≈ 0.00001 XLM), with no breakdown rows and no error indication. The user thinks that's the real cost of a ~0.5 XLM Soroban tx.

This is exactly the failure mode the feature is supposed to eliminate, appearing in the opposite direction. Fix: add an explicit error branch. For example:

```ts
const isError = simulationState.state === RequestState.ERROR;

// in the Total row:
{isLoading ? t("Calculating...") : isError ? "—" : `${fee} XLM`}
```

And render an error message in the card body so the user knows the breakdown is unavailable rather than "0.00001 XLM".

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -429,9 +542,12 @@ export const SendAmount = ({
                   {t("Fee")}:
                 </span>
                 <span data-testid="send-amount-fee-display">
-                  {inputType === "crypto"
-                    ? `${fee} ${t("XLM")}`
-                    : recommendedFeeUsd}
+                  {(isToken || isCollectible) &&
+                  simulationState.state === RequestState.LOADING
```
CassioMG: **Critical — `send-amount-fee-display` shows a misleading total on Soroban simulation failure**

This block handles `LOADING` ("Calculating…") but not `ERROR`. On `RequestState.ERROR`, control falls through to the `inputType === "crypto" ? \`${fee} ${t("XLM")}\` : recommendedFeeUsd` branch, where `fee = transactionFee || recommendedFee`. If the user hasn't edited the inclusion fee, they see `0.00001 XLM` (the base inclusion) as the "Fee" for a Soroban tx whose real cost is much higher.

Compounding: the Review-Send button at line 615 only adds `isLoading` for `LOADING`, so on `ERROR` the button stays enabled — user can proceed to review and sign, still seeing the wrong fee.

Fix: either treat `ERROR` the same as `LOADING` for display (show `Calculating...` or an em-dash) and disable the Continue button, or surface an inline error next to the fee with a retry affordance.

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+import "./styles.scss";
+
+export interface FeesPaneProps {
+  fee: string;
+  simulationState: State<SimulateTxData, string>;
+  isSoroban?: boolean;
+  onClose: () => void;
+}
+
+export const FeesPane = ({
```
CassioMG, line 18: **Important — No component tests for FeesPane**

FeesPane is a new 97-line component with distinct rendering branches for:
- `RequestState.IDLE` / `LOADING` (spinner row + `Calculating...` total)
- `RequestState.SUCCESS` with Soroban data (inclusion + resource + total)
- `RequestState.SUCCESS` classic (only total)
- `RequestState.ERROR` (currently broken — see C1)
- Soroban vs. classic description copy via `isSoroban` prop

The 823-line e2e test (`reviewTxFees.test.ts`) is thorough for the happy path and custom-fee lifecycle, but it doesn't cover the failure states or the branching logic directly — and it exercises through the whole flow, which is expensive and slow compared to a unit test.

Please add Jest/RTL component tests for FeesPane covering at least the five branches above. These would have caught C1 immediately, and would also guard against future regressions when the pane is reused in other flows (Swap already uses it today, Soroban-swap will need the breakdown fields later).

@shared/api/internal.ts
CassioMG: Adding supporting evidence on Piyal's question, since it's still open:

Checked the current `freighter-backend` `master` — `/simulate-tx` in [`src/route/index.ts` L1294-1337](https://github.com/stellar/freighter-backend/blob/main/src/route/index.ts):

```ts
schema: {
  body: {
    type: "object",
    required: ["xdr", "network_passphrase"],   // network_url NOT required
    properties: {
      xdr: { type: "string" },
      network_passphrase: { type: "string" },  // network_url not even listed
    },
  },
},
handler: async (request) => {
  const { xdr, network_passphrase } = request.body;  // network_url not read
  // ... server derived from network_passphrase via getServer(networkName, stellarRpcConfig)
}
```

No `additionalProperties: false` on the schema, so unknown fields are silently accepted and dropped — the handler never sees `network_url`. `/simulate-token-transfer` L1352-1383 also doesn't list it.

@leofelix077 — worth double-checking: the backend error you saw about "required field" might have been from a different field (e.g. a required entry you were missing), from a stale/fork backend, or a different endpoint. On current `freighter-backend/master` this field is dead on the wire. If the error was reproducible, a backend stack trace or the exact error payload would help pin it down before we land the client change.

Either the freighter-backend needs a companion PR to actually use `network_url`, or this line can be removed.

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
CassioMG: this may also be present on mobile's PR

extension/src/popup/components/send/SendAmount/index.tsx
CassioMG: this may also be present on mobile's PR

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
CassioMG: this may also be needed on mobile's PR

@shared/api/internal.ts
leofelix077: I double checked it now. now it's working again without it, but the error message was "network_url is a required field". 

it was removed by this PR here https://github.com/stellar/freighter-backend/commit/988c040592c21a9a4c7e5581738818b3536e405b (2 weeks ago), after I had tested it

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -803,18 +967,48 @@ export const SendAmount = ({
...
+          setIsModalOpen={() => {
+            setIsShowingFeesPane(false);
+            setIsEditingSettings(true);
+          }}
+          isModalOpen={isShowingFeesPane}
+        >
+          <View.Inset>
+            <div className="SendAmount__FeesPane">
+              <FeesPane
+                fee={feesPaneTotal}
```
CassioMG: **Critical: Stale `feesPaneTotal` when simulation completes while pane is open**

`feesPaneTotal` is initialized from `fee` (line 195) and only mutated inside `onShowFeesInfo` (line 953). When the FeesPane is opened from EditSettings while a simulation is in-flight:

- the `Inclusion Fee` row updates live (driven by `simulationState.data?.inclusionFee`)
- the `Resource Fee` row updates live (driven by `simulationState.data?.resourceFee`)
- but the `Total Fee` row stays stale, because `fee={feesPaneTotal}` is a one-time snapshot

This re-introduces the silent-fee-change UX that issue #2475 was meant to fix. Derive total inside `FeesPane` (or via `useMemo` here) from the live inclusion + resource values so the total updates with the breakdown rows. The Soroban e2e tests don't catch this because they all open the pane after simulation has settled.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -109,9 +125,44 @@ export const SendAmount = ({
...
+  // Persist the last-known inclusion fee across re-simulations so the
+  // EditSettings input never jumps back to the total fee while LOADING
+  // (the request reducer sets data: null on FETCH_DATA_START).
+  const lastInclusionFeeRef = useRef<string | null>(null);
+  if (
+    simulationState.state === RequestState.SUCCESS &&
+    simulationState.data?.inclusionFee
+  ) {
+    lastInclusionFeeRef.current = simulationState.data.inclusionFee;
+  }
```
CassioMG: **Critical: Render-time ref mutation**

Mutating `lastInclusionFeeRef.current` during render works but is semantically a side effect. Under React 19's StrictMode/concurrent rendering, components may render-then-bail, leaving this ref in an unintended state across discarded renders. Move into a `useEffect`:

```tsx
useEffect(() => {
  if (
    simulationState.state === RequestState.SUCCESS &&
    simulationState.data?.inclusionFee
  ) {
    lastInclusionFeeRef.current = simulationState.data.inclusionFee;
  }
}, [simulationState.state, simulationState.data?.inclusionFee]);
```

Same pattern issue at line 148: `useRef<string | null>(manualTransactionFee)` — the initializer only runs once. If Redux's `manualTransactionFee` ever changes from another dispatcher, the ref drifts from Redux. Either read straight from Redux every render, or sync via `useEffect`.

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <div className="FeesPane__Card__Row">
+          <span className="FeesPane__Card__Row__Label FeesPane__Card__Row__Label--total">
+            {t("Total Fee")}
+          </span>
+          <span
+            className="FeesPane__Card__Row__Value FeesPane__Card__Row__Value--total"
+            data-testid="review-tx-total-fee"
+          >
+            {isError ? "—" : `${fee} XLM`}
+          </span>
```
CassioMG: **Critical: Typography rule violation**

This component uses raw `<span>` elements throughout (title, row labels, row values) with custom font sizes hardcoded in `styles.scss`:

```scss
&__Title { font-size: pxToRem(18px); }
&__Card__Row__Label { font-size: pxToRem(14px); }
&__Card__Row__Value { font-size: pxToRem(14px); }
&__Description { font-size: pxToRem(14px); }
```

`CLAUDE.md` is explicit:

> Always use `Text` or `Heading` from `@stellar/design-system` — never raw HTML text elements with custom font styles.

Replace with `<Heading as="h2" size="md">` for the title and `<Text as="div" size="md">` / `<Text as="span" size="sm">` for the row labels/values, and drop the explicit `font-size` declarations from SCSS. The title `<div><span>` wrapping at line 43–45 also has no semantic role — `<Heading>` fixes that for screen readers too.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -235,11 +293,30 @@ export const SendAmount = ({
...
-    await fetchSimulationData();
-    setIsReviewingTx(true);
+    const simResult = await fetchSimulationData();
+    // fetchSimulationData returns the SimulateTxData payload on success and
+    // the caught Error on failure. Only open the review modal when the
+    // simulation succeeded — on failure the fee display shows the error
+    // state and the user can retry.
+    if (!(simResult instanceof Error)) {
+      setIsReviewingTx(true);
+    }
```
CassioMG, line 327: **Critical: Loose `Promise<unknown>` typing + brittle `instanceof Error` check**

The prop is declared as `fetchSimulationData: () => Promise<unknown>` (line 107) and the caller branches on `simResult instanceof Error`. The hook (`useSimulateTxData.tsx`) catches `error: unknown` from a try block where `throw new Error(...)` calls coexist with code paths that could throw non-`Error` rejections (axios, RPC layer, anything that calls `Promise.reject(non-Error)`). A non-`Error` rejection would be silently classified as success, opening the review modal with invalid XDR.

Tighten with a discriminated return type so the contract is verified at the type level:

```ts
type SimulateResult =
  | { ok: true; data: SimulateTxData }
  | { ok: false; error: string };

fetchSimulationData: () => Promise<SimulateResult>;
```

Then `if (simResult.ok) setIsReviewingTx(true);` becomes both safer and more readable than the current `instanceof` heuristic.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -235,11 +293,30 @@ export const SendAmount = ({
...
+      dispatch(
+        saveTransactionFee(
+          hasManuallySetFeeRef.current ??
+            lastInclusionFeeRef.current ??
+            recommendedFee,
+        ),
+      );
+    } else if (!transactionFee) {
       dispatch(saveTransactionFee(fee));
     }
```
CassioMG, line 314: **Important: `handleContinue` writes Redux fee twice for Soroban**

For Soroban this always dispatches `saveTransactionFee(...)` with the inclusion-fee chain — that's correct as a "reset before re-simulating" — but `fetchSimulationData` immediately overwrites it again with `inclusion + resource` inside the hook (`useSimulateTxData.tsx`). End result is correct, but it's two writes back-to-back; somebody not following the entire chain will misread the code.

Either (a) gate the reset on `transactionFee !== inclusionFee`, or (b) leave a one-line comment noting that the second write happens inside `fetchSimulationData` and is intentional.

extension/src/popup/components/InternalTransaction/FeesPane/styles.scss
```diff
@@ -0,0 +1,101 @@
...
+.FeesPane {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  padding: 0 0;
+  gap: pxToRem(16px);
+
+  .multi-pane-slider & {
+    padding: 0;
+  }
```
CassioMG: **Important: Redundant SCSS**

```scss
padding: 0 0;     // same as `padding: 0`
gap: pxToRem(16px);
.multi-pane-slider & { padding: 0; }   // same as parent — no-op
```

Either delete both lines, or actually pick a different padding for the `.multi-pane-slider &` variant if that was the original intent. As-is, the override is dead code.

extension/src/popup/components/InternalTransaction/FeesPane/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+        )}
+        <div className="FeesPane__Card__Row">
+          <span className="FeesPane__Card__Row__Label FeesPane__Card__Row__Label--total">
+            {t("Total Fee")}
+          </span>
+          <span
+            className="FeesPane__Card__Row__Value FeesPane__Card__Row__Value--total"
+            data-testid="review-tx-total-fee"
+          >
+            {isError ? "—" : `${fee} XLM`}
```
CassioMG: **Important: Placeholder strings are inconsistent and undocumented**

Three different placeholders are used across the rows:

- Loading inclusion (line 58): `${simulationState.data?.inclusionFee ?? fee} XLM` — the live base fee
- Loading resource (line 75): literal `"-"` (hyphen-minus, U+002D)
- Error all rows (lines 57, 72, 87): literal `"—"` (em dash, U+2014)

Using em dash for error and hyphen for "not yet computed" is intentional and visually distinguishable, but it's not documented anywhere. Lift these into named constants near the top of the file:

```ts
const PLACEHOLDER_PENDING = "-";   // U+002D — "computing"
const PLACEHOLDER_ERROR = "—";     // U+2014 — "unavailable"
```

so future readers don't accidentally normalize them.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -300,6 +377,39 @@ export const SendAmount = ({
...
+              lastInclusionFeeRef.current ??
+              recommendedFee,
+          ),
+        );
+      }
+      simulationDataRef.current = { destination, asset };
+      fetchSimulationData();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [destination, asset, isToken, isCollectible, simulationState.state]);
```
CassioMG, line 420: **Important: Comment claims a retry that doesn't exist**

The header comment at lines 380–384 says:

> simulationState.state is included so that if a change arrives while a simulation is already in-flight, we retry once it finishes.

There's no retry logic. The effect early-returns at line 389 when state is `LOADING`, and when state transitions back to `SUCCESS`/`IDLE` the effect re-fires; if the live `destination`/`asset` still differ from `simulationDataRef.current`, it triggers — which works for the happy case.

But: if the user changes destination twice while a simulation is in flight, only the **second** change ends up being simulated (the first change never had a chance to update the ref). That's fine UX, but the comment should reflect what actually happens:

> simulationState.state is included so that the effect re-runs when an in-flight simulation completes; the destination/asset diff is then re-checked against the latest values.

Or drop the early-return and use `simulationDataRef.current` to debounce instead.

extension/src/types/transactions.ts
```diff
@@ -1,9 +1,18 @@
...
     tags: Array<string>;
   };
 }
 
+export interface SimulateTxData {
+  transactionXdr: string;
+  scanResult?: BlockAidScanTxResult | null;
+  inclusionFee?: string;
+  resourceFee?: string;
+}
```
CassioMG: **Important: `SimulateTxData` migration is incomplete**

This PR cleanly migrates `send` and `SelectedCollectible` over to this shared `SimulateTxData` type — good move. However, `extension/src/popup/components/swap/SwapAmount/hooks/useSimulateSwapData.tsx:79` still defines its own local `SimulateTxData` interface (without `inclusionFee`/`resourceFee`).

If this shared type is positioning itself as the canonical contract for simulation results, swap should be folded in too. Otherwise we'll end up with two types of the same name diverging over time. Out of scope for this PR — please file a follow-up so it doesn't get lost.

CassioMG (approved):

extension/src/popup/components/send/SendAmount/hooks/useSimulateTxData.tsx
CassioMG, line 344: This is an existing bug from before this PR but I think is worth addressing since it's related to fee handling (this is [already fixed on mobile's PR](https://github.com/stellar/freighter-mobile/pull/774/changes#diff-3412a4b2c7afff366c56904c4a4e05ec7059c23cb5a94a80bd74cdb5daf03d60R741)): it looks like the `transactionFee` is being passed here but it's being discarded inside `simulateTokenTransfer` which means it's `not being passed to the backend` so the backend is always using the default fee of 100 stroops. This means the inclusion fee users are setting on transaction settings are not being effectively used.


Below is a more in-depth analysis from AI:

**The user-selected inclusion fee is dropped before reaching the backend on the indexer/mainnet path.**

`transactionFee` is correctly threaded into `simulateTokenTransfer({...transactionFee})` here, but in [`@shared/api/internal.ts`](https://github.com/stellar/freighter/blob/master/extension/%40shared/api/internal.ts) the non-custom-network branch (around line 2186) builds `requestBody` without including any `fee` field:

```ts
const requestBody = {
  address,
  pub_key: publicKey,
  memo: memo || "",
  params,
  network_passphrase: networkDetails.networkPassphrase,
  // fee is NOT included
};
```

The freighter-backend `/simulate-token-transfer` handler [accepts an optional `fee` and falls back to `BASE_FEE` when absent](https://github.com/stellar/freighter-backend/blob/main/src/route/index.ts#L1395-L1402):

```ts
const _fee = fee || Sdk.BASE_FEE;
const builder = new Sdk.TransactionBuilder(sourceAccount, { fee: _fee, ... });
// ... assembleTransaction → returned preparedTransaction.fee = _fee + minResourceFee
```

Since the client submits the returned `preparedTransaction` XDR verbatim, the on-chain inclusion bid is always `BASE_FEE` (100 stroops) regardless of what the user picks in the slider. The new FeesPane shows the user's selected value as the inclusion fee, but the chain sees BASE_FEE.

**Impact:** invisible during normal traffic (BASE_FEE suffices), but during surge pricing — the case the slider exists for — bumping the fee has no effect; the tx loses inclusion auctions and times out.

**Fix:** add one line to the request body in `internal.ts`:

```ts
const requestBody = {
  address,
  pub_key: publicKey,
  memo: memo || "",
  fee: xlmToStroop(transactionFee).toFixed(),
  params,
  network_passphrase: networkDetails.networkPassphrase,
};
```

The mobile companion PR ([stellar/freighter-mobile#774](https://github.com/stellar/freighter-mobile/pull/774)) made this same change (`fee: xlmToStroop(fee).toString()`), and the freighter-backend already supports the field.

CassioMG (changes requested):
Found one [last issue](https://github.com/stellar/freighter/pull/2649#discussion_r3157236967) that would be worth addressing on this PR before merging it, specially since same bug is being addressed on the mobile PR so we keep parity.

piyalbasu:
### Code review

Found 2 issues:

1. The `send-amount-fee-display` "Calculating..." gate only fires for `RequestState.LOADING`, dropping the `RequestState.ERROR` arm an earlier branch commit (`c0973c98`, "sync state with extension and harden validation for simulation errors") had explicitly added. On a Soroban simulation failure the label silently reverts to the prior-simulation `fee` value while the Continue button is disabled with no other affordance, leaving users to wonder why they can't proceed.

https://github.com/stellar/freighter/blob/95e68194f8782179781d81cc0b09073902e08aad/extension/src/popup/components/send/SendAmount/index.tsx#L558-L567

2. `goToChooseAssetAction` clears `manualTransactionFee` and `hasManuallySetFeeRef`, but unlike the sibling `goBackAction` (which dispatches both `saveTransactionFee("")` and `saveManualTransactionFee(null)`) it leaves `transactionFee` in Redux. The block comment says "Clear it here before navigating so post-remount the fee is derived freshly from the new asset's simulation," but `fee = transactionFee || recommendedFee` will still pick up the prior asset's saved total on remount until the auto-simulation `useEffect` runs, contradicting the stated invariant.

https://github.com/stellar/freighter/blob/95e68194f8782179781d81cc0b09073902e08aad/extension/src/popup/components/send/SendAmount/index.tsx#L509-L532

piyalbasu:
Also, I noticed a few UI bugs here:

1). If you update the fee and then click the i bubble, the modal shows the old fee
2) The `default` button next to the fee appears to have stopped working
3) i'm able to save with an empty inclusion fee


https://github.com/user-attachments/assets/c0c59900-4175-4b6e-a967-ad6edc0d6b30

leofelix077:
@piyalbasu @CassioMG  thanks for checking!

I updated it now the UI + verifying that it's carrying forward the fee to the simulation 


https://github.com/user-attachments/assets/b1567da1-496f-4cc6-adab-002f57d358a6



Screenshot 2026-04-29 at 13 01 04

CassioMG (commented):
Re-reviewed at e9e5a7a1. The previous concern about `fee` being dropped before reaching the backend is **resolved** — the new `fee: xlmToStroop(transactionFee).toFixed()` line in `@shared/api/internal.ts` and the test in `@shared/api/__tests__/internal.test.ts` confirm the wiring end-to-end.

Two Important findings, and a Small finding remain — see inline comments.

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -597,6 +597,7 @@ export const SwapAmount = ({
           <div className="SlippageWrapper">
             <EditSettings
               fee={fee}
+              defaultFee={fee}
```
CassioMG: **Important — `defaultFee` is a no-op when the user has a custom fee.**

`defaultFee={fee}` looks safe, but `fee = transactionFee || recommendedFee` (line 96 in this file). If the user has manually saved a fee via EditSettings, `transactionFee` is set and `fee = the manual fee`. Pressing the new "Default" button to revert to the recommended value then does nothing — it sets the input back to the user's current fee.

**Fix:** pass the actual recommendation:

```tsx
defaultFee={recommendedFee}
```

`recommendedFee` is already destructured from `useNetworkFees()` at line 77 — same source the rest of the component uses for the "Recommended" tooltip and idle-state display.

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -785,36 +966,85 @@ export const SendAmount = ({
...
         </>
       ) : null}
-      {isEditingSettings ? (
+      {isEditingSettings && !isShowingFeesPane ? (
         <>
           <div className="EditMemoWrapper">
             <EditSettings
-              fee={fee}
+              fee={draftFeeForDisplay ?? editSettingsFee}
+              defaultFee={editSettingsFee}
```
CassioMG: **Important — same `defaultFee` issue as in `SwapAmount`, with an extra wrinkle for Soroban.**

`defaultFee={editSettingsFee}` where:
```ts
editSettingsFee = isToken || isCollectible
  ? (hasManuallySetFeeRef.current ?? lastInclusionFeeRef.current ?? recommendedFee)
  : fee;
```

For Soroban with a manually-saved fee, `editSettingsFee = hasManuallySetFeeRef.current` (the manual value). Pressing "Default" reverts to itself — no-op. For classic, `editSettingsFee = fee = transactionFee || recommendedFee`, which after a Soroban prior simulation could even be the **total** that was written back to Redux at `useSimulateTxData.tsx:555`.

The unit test in `EditSettings.test.tsx:25` exercises the component with `defaultFee: "0.00001"` (a literal) and passes — so it doesn't catch this integration bug at the call site.

**Fix:** pass `recommendedFee` explicitly:

```tsx
defaultFee={recommendedFee}
```

If the design intent was "revert to last simulated base fee" (preserving any auto-bump from network congestion), then:

```tsx
defaultFee={lastInclusionFeeRef.current ?? recommendedFee}
```

Either way, **must not include `hasManuallySetFeeRef.current`** in the chain — that's what makes the button a no-op.

@shared/api/internal.ts
```diff
@@ -2186,6 +2187,7 @@ export const simulateTokenTransfer = async (args: {
     address,
     pub_key: publicKey,
     memo: memo || "", // Backend requires memo as string, use empty string if undefined
```
CassioMG, line 2189: **Small — first-mount simulation sends `fee: "0"` to the backend.**

Redux's initial `transactionFee` is `""` (`transactionSubmission.ts:495`). On the very first call to `simulateTokenTransfer` after mounting `SendAmount` for a new asset, `transactionFee = ""` flows in. Then:

```ts
xlmToStroop("").toFixed()  // = "0"  (Number("") === 0)
```

The backend's fallback at [freighter-backend `index.ts:1395`](https://github.com/stellar/freighter-backend/blob/main/src/route/index.ts#L1395) is `_fee = fee || Sdk.BASE_FEE` — but `"0"` is a **truthy** string in JS, so the fallback doesn't fire. `_fee = "0"` is used as the inclusion fee. The resulting prepared XDR has `fee = 0 + minResourceFee`. **If submitted, that XDR is rejected for insufficient fee** (Soroban requires `>= BASE_FEE + minResourceFee`).

**Why it's mostly latent today:** `handleContinue` in `SendAmount/index.tsx:305-311` explicitly resets `transactionFee` to `lastInclusionFeeRef.current ?? recommendedFee` before re-simulating, so the XDR the user actually signs and submits is from a re-sim with a proper fee. The first-mount sim's output is discarded.

**Why it's still worth fixing:**
- Wasted backend call on every Send entry.
- Brief UX divergence: the FeesPane's first render shows `inclusionFee = stroopToXlm(BASE_FEE) = 0.00001 XLM` (because `simulateTx` in `useSimulateTxData.tsx:325` falls back to `BASE_FEE` when `recommendedFee` is empty), while the backend was actually given `0` stroops.
- Fragile: any future code path that submits the first-sim XDR (e.g., a refactor that skips the `handleContinue` reset) will silently produce broken transactions.

**Fix options:**

1. **Client-side (in `useSimulateTxData.tsx:439-440`):** default `currentTransactionFee` to a non-empty value:
   ```ts
   const currentTransactionFee =
     currentTransactionData.transactionFee ||
     transactionFee ||
     stroopToXlm(BASE_FEE).toString();
   ```
2. **Server-side (in freighter-backend):** tighten the fallback so it catches `"0"`:
   ```ts
   const _fee = (fee && fee !== "0") ? fee : Sdk.BASE_FEE;
   ```
   This protects future callers too.

The mobile companion PR ([stellar/freighter-mobile#774](https://github.com/stellar/freighter-mobile/pull/774)) doesn't have this issue because its store is initialized to `MIN_TRANSACTION_FEE` (not `""`).

extension/src/popup/components/sendCollectible/SelectedCollectible/hooks/useSimulateTxData.ts
CassioMG, line 1: **Important — this hook misses the first-mount fee fix that was applied to the send hook.**

Lines 109-110 (outside this PR's diff hunks, hence the file-level comment) still use the old pattern:

```ts
const currentTransactionFee =
  currentTransactionData.transactionFee || transactionFee;
```

The new `getCurrentTransactionFee` helper added in `send/SendAmount/hooks/useSimulateTxData.tsx:123-132` (with the `BASE_FEE` fallback) was the fix for the first-mount-fee-zero issue I flagged at https://github.com/stellar/freighter/pull/2649#discussion_r3163110447. The send hook now uses it; this hook does not.

**Why it matters here too:** on first-mount of a collectible send, `currentTransactionData.transactionFee = ""` and `transactionFee = ""`, so `currentTransactionFee = ""`. Then `simulateTx` (line 124) receives `recommendedFee: ""` → `simulateSendCollectible` is called with `transactionFee: ""` → `xlmToStroop("").toFixed() === "0"` is set as the `TransactionBuilder` `fee` ([@shared/api/internal.ts:2080](https://github.com/stellar/freighter/blob/6e26d05a/@shared/api/internal.ts#L2080)). stellar-sdk's `TransactionBuilder` rejects fees below `BASE_FEE` at submit time.

**Why it's mostly latent today:** like the send case, `handleContinue` paths typically reset `transactionFee` before re-simulation, so the user-submitted XDR is from a re-sim with a proper fee. But the first sim still wastes a build cycle and produces an invalid XDR, and any future code path that submits the first-sim output silently breaks.

**Suggested fix:** export `getCurrentTransactionFee` from a shared location (e.g. `popup/helpers/fees.ts` or `popup/helpers/soroban.ts`) and use it here too:

```ts
import { getCurrentTransactionFee } from "popup/helpers/fees";

const currentTransactionFee = getCurrentTransactionFee({
  currentTransactionFee: currentTransactionData.transactionFee,
  fallbackTransactionFee: transactionFee,
});
```

Keeping the helper inside the `send/SendAmount` hook directory while the collectible hook needs it would force this hook to reach across component boundaries, which is a smell — the helper lives in the wrong place even before the parity fix.

extension/src/popup/components/sendCollectible/SelectedCollectible/hooks/useSimulateTxData.ts
```diff
@@ -8,6 +8,7 @@ import { initialState, reducer } from "helpers/request";
 import { NetworkDetails } from "@shared/constants/stellar";
 import { stroopToXlm } from "helpers/stellar";
 import { getBaseAccount } from "popup/helpers/account";
+import { getCurrentTransactionFee } from "popup/components/send/SendAmount/hooks/useSimulateTxData";
```
CassioMG: should we extract it to a helper as suggested [here](https://github.com/stellar/freighter/pull/2649#discussion_r3170018066)?

CassioMG (approved):
Lgtm

Issue #2538, opened 2026-01-23
Implement Amplitude SDK


Keeeeeeeks:
We should remove the Amplitude API requests and use the Amplitude SDK instead like we are doing on mobile.

While implementing the SDK, let's try uploading same user context properties we currently [upload on mobile](https://github.com/stellar/freighter-mobile/blob/main/src/services/analytics/core.ts#L201-L223) for consistency:
```
  const context: Record<string, unknown> = {
    publicKey: account?.publicKey ?? "N/A",
    platform: Platform.OS,
    platformVersion: Platform.Version,
    network: network.toUpperCase(), // Stellar network (TESTNET, PUBLIC, FUTURENET)
    connectionType, // Internet connectivity (wifi, cellular, etc.)
    appVersion: getVersion(),
    buildVersion: getBuildNumber(),
    bundleId: getBundleId(),
  };
```

PR #2659, opened 2026-03-23
add amplitude SDK to extension

leofelix077:
https://github.com/stellar/freighter/issues/2538

https://github.com/user-attachments/assets/b826706a-d527-46d7-ac2a-7fec426b7282

https://github.com/user-attachments/assets/2a9095cf-7aad-4812-808f-f82dc64fc329

### DIFF
--- a/.github/actions/validate-branch-from/action.yml
+++ b/.github/actions/validate-branch-from/action.yml
@@ -1,0 +1,95 @@
+name: "Validate Branch From"
+description:
+  "Validates that branch_from is 'master' or a semver tag and provides available
+  version tags"
+
+inputs:
+  branch_from:
+    description:
+      "The source ref to branch from: either 'master' or a version tag of the
+      form X.Y.Z"
+    required: true
+
+outputs:
+  is_valid:
+    description:
+      "Whether the branch_from input is a valid 'master' ref or semver tag"
+    value: ${{ steps.validate.outputs.is_valid }}
+  branch_or_tag:
+    description: "The validated source ref name ('master' or an X.Y.Z tag)"
+    value: ${{ steps.validate.outputs.branch_or_tag }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Fetch tags
+      id: fetch_tags
+      shell: bash
+      run: |
+        if ! git fetch --tags 2>/dev/null; then
+          echo "Error: Failed to fetch git tags. Please check repository access and network connectivity." >&2
+          exit 1
+        fi
+
+        TAGS=$(git tag --list '[0-9]*' --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | head -n 10)
+
+        TAGS_CSV=$(echo "$TAGS" | tr '\n' ',' | sed 's/,$//')
+        echo "tags=$TAGS_CSV" >> $GITHUB_OUTPUT
+
+        DELIM="TAGS_$(openssl rand -hex 16)"
+        {
+          echo "tags_list<<$DELIM"
+          echo "$TAGS"
+          echo "$DELIM"
+        } >> $GITHUB_OUTPUT
+
+        echo "Available tags:"
+        echo "$TAGS" | while read tag; do
+          echo "  - $tag"
+        done
+
+    - name: Validate branch_from
+      id: validate
+      shell: bash
+      env:
+        RAW_BRANCH_FROM: ${{ inputs.branch_from }}
+        FETCHED_TAGS: ${{ steps.fetch_tags.outputs.tags_list }}
+      run: |
+        BRANCH_FROM="$RAW_BRANCH_FROM"
+        TAGS="$FETCHED_TAGS"
+
+        if [ "$BRANCH_FROM" = "master" ]; then
+          echo "Using master branch as base"
+          echo "is_valid=true" >> $GITHUB_OUTPUT
+          echo "branch_or_tag=master" >> $GITHUB_OUTPUT
+          exit 0
+        fi
+
+        # Strip "v" prefix if present
+        BRANCH_FROM=$(echo "$BRANCH_FROM" | sed 's/^v//')
+
+        if [[ ! "$BRANCH_FROM" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          echo "Error: Invalid format. Must be 'master' or a version tag like '5.38.0'"
+          echo ""
+          echo "Available tags (last 10):"
+          echo "$TAGS" | while read tag; do
+            echo "  - $tag"
+          done
+          echo "is_valid=false" >> $GITHUB_OUTPUT
+          exit 1
+        fi
+
+        if git tag --list "$BRANCH_FROM" | grep -qx "$BRANCH_FROM"; then
+          echo "Using tag $BRANCH_FROM as base"
+          echo "is_valid=true" >> $GITHUB_OUTPUT
+          echo "branch_or_tag=$BRANCH_FROM" >> $GITHUB_OUTPUT
+        else
+          echo "Error: Tag $BRANCH_FROM does not exist"
+          echo ""
+          echo "Available tags (last 10):"
+          echo "$TAGS" | while read tag; do
+            echo "  - $tag"
+          done
+          echo "is_valid=false" >> $GITHUB_OUTPUT
+          exit 1
+        fi
--- a/.github/workflow-drafts/prPreview.yml
+++ b/.github/workflow-drafts/prPreview.yml
@@ -1,0 +1,92 @@
+# name: PR Preview
+# env:
+#   INDEXER_URL: ${{ secrets.INDEXER_URL }}
+#   INDEXER_V2_URL: ${{ secrets.INDEXER_V2_BETA_URL }}
+# on:
+#   pull_request:
+#     types: [opened, synchronize, reopened, closed]
+# concurrency:
+#   group: pr-preview-${{ github.event.pull_request.number }}
+#   cancel-in-progress: true
+# permissions:
+#   contents: write
+# jobs:
+#   build-and-release:
+#     if: github.event.action != 'closed' && github.event.pull_request.head.repo.full_name == github.repository
+#     name: Build and Release PR Preview
+#     runs-on: ubuntu-latest
+#     steps:
+#       - if: ${{ env.INDEXER_URL == '' }}
+#         run: |
+#           echo "Missing INDEXER_URL"
+#           gh run cancel ${{ github.run_id }}
+#           gh run watch ${{ github.run_id }}
+#       - if: ${{ env.INDEXER_V2_URL == '' }}
+#         run: |
+#           echo "Missing INDEXER_V2_URL"
+#           gh run cancel ${{ github.run_id }}
+#           gh run watch ${{ github.run_id }}
+#       - name: Checkout code
+#         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+#       - name: Update manifest-v3.json name
+#         uses: jossef/action-set-json-field@2a0f7d953b580b828717daf4de7fafc7e4135e97 #v2
+#         with:
+#           file: ./extension/public/static/manifest/v3.json
+#           field: name
+#           value: Freighter PR Preview ${{ github.event.pull_request.number }}
+#       - name: Update manifest-v3.json version_name
+#         uses: jossef/action-set-json-field@2a0f7d953b580b828717daf4de7fafc7e4135e97 #v2
+#         with:
+#           file: ./extension/public/static/manifest/v3.json
+#           field: version_name
+#           value: pr-preview-${{ github.event.pull_request.number }}
+#       - name: Enable Corepack
+#         run: corepack enable
+#       - name: Build extension
+#         uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+#         with:
+#           node-version: 22
+#       - run: yarn && yarn build:freighter-api && yarn build:extension:production
+#       - name: Use BETA icons
+#         run: |
+#           rm -rf ./extension/build/images
+#           mv ./extension/build/beta_images ./extension/build/images
+#       - name: Remove scripts tag
+#         uses: restackio/update-json-file-action@f8ef1561cb15ba86a6367b547216375bc60e7f91 #v2.1
+#         with:
+#           file: ./extension/build/manifest.json
+#           fields: '{"background": {"service_worker": "background.min.js"}}'
+#       - name: Install zip
+#         uses: montudor/action-zip@0852c26906e00f8a315c704958823928d8018b28 #v1.0.0
+#       - name: Zip extension build
+#         run: zip -qq -r ./build.zip *
+#         working-directory: ./extension/build
+#       - name: Delete existing release
+#         env:
+#           GH_TOKEN: ${{ github.token }}
+#           PR_NUMBER: ${{ github.event.pull_request.number }}
+#         run: |
+#           gh release delete "pr-preview-${PR_NUMBER}" --yes --cleanup-tag || true
+#       - name: Create GitHub Release
+#         env:
+#           GH_TOKEN: ${{ github.token }}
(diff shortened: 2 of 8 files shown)


extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,480 @@ export interface MetricsData {
...
-  cache = [];
-  if (!AMPLITUDE_KEY) {
-    console.log("Not uploading metrics", toUpload);
-    return;
+/**
+ * Truncates a Stellar public key for analytics/storage use.
+ * Returns first 8 + last 4 chars (e.g. "GABCDEFG…WXYZ"), which is enough
+ * for session correlation without storing the full address.
+ */
+const truncatePublicKey = (publicKey: string): string =>
```
piyalbasu: we actually already have a helper for this one in `freighter/extension/src/helpers/stellar.ts` called `truncatedPublicKey`

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,480 @@ export interface MetricsData {
...
+    const cutoff = Date.now() - DEBUG_EVENT_TTL_MS;
+    return events.filter((e) => e.timestamp >= cutoff);
+  } catch {
+    return [];
   }
+};
+
+// Initialize the snapshot from localStorage so the debug panel shows
+// persisted events immediately after a page refresh (dev only).
+let recentEventsSnapshot: DebugEvent[] = isDev ? readPersistedEvents() : [];
```
piyalbasu: I see that we're adding a debug screen for dev environments that records the last 50 analytics calls. Can I ask what problem we're solving by implementing this? I don't necessarily think it's wrong! It just seems like we're adding a lot of code for a feature that we haven't discussed or prioritized as a team

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,480 @@ export interface MetricsData {
...
-    if (storedData) {
-      metricsData = JSON.parse(storedData);
-    } else {
-      throw new Error("No metrics data found in localStorage");
+    amplitude.init(AMPLITUDE_KEY, undefined, {
+      // Disable Amplitude's built-in cookie/localStorage identity persistence.
+      // We manage the user ID ourselves via `getUserId()` + localStorage so that
+      // we control the identifier format and lifecycle. Amplitude's default
+      // identity storage would create a second, SDK-managed identifier that could
+      // conflict with ours and is unnecessary given we call setUserId() manually.
```
piyalbasu: hm, what is the benefit to us handling this ourselves rather than letting the SDK do it for us?

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,480 @@ export interface MetricsData {
...
+const getAppVersion = (): string => {
+  const g = globalThis as any;
+
+  // Chrome / Chromium-based extensions
+  const chromeVersion = g?.chrome?.runtime?.getManifest?.()?.version;
+  if (chromeVersion) return chromeVersion;
+
+  // Firefox extensions (browser.* API)
+  const browserVersion = g?.browser?.runtime?.getManifest?.()?.version;
+  if (browserVersion) return browserVersion;
```
piyalbasu: the `browser` lib should work for all browsers (including Chrome). You shouldn't need to distinguish between Chrome and Firefox

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,480 @@ export interface MetricsData {
...
+
+  // Dev server or unknown environment — use build-time constant
+  return APP_VERSION;
```
piyalbasu: Why not try this first? There should be an APP_VERSION in almost all scenarios

extension/src/popup/views/Swap/index.tsx
```diff
@@ -31,6 +31,27 @@ export const Swap = () => {
...
+    if (activeStep === lastEmittedStep.current) return;
+    lastEmittedStep.current = activeStep;
+
+    const metricByStep: Partial<Record<STEPS, string>> = {
+      [STEPS.SWAP_CONFIRM]: METRIC_NAMES.swapConfirm,
+      [STEPS.SET_DST_ASSET]: METRIC_NAMES.swapTo,
+      [STEPS.AMOUNT]: METRIC_NAMES.swapAmount,
+      [STEPS.CONFIRM_AMOUNT]: METRIC_NAMES.swapAmountReview,
+      [STEPS.SET_FROM_ASSET]: METRIC_NAMES.swapFrom,
+    };
```
piyalbasu: nice fix!

.github/workflows/submitBeta.yml
```diff
@@ -82,7 +82,10 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: 22
-      - run: yarn && yarn build:freighter-api && yarn build:extension:production
+      - run: >
+          yarn && yarn build:freighter-api && yarn build:extension:production
+          --env AMPLITUDE_KEY="${{ secrets.AMPLITUDE_KEY }}" SENTRY_KEY="${{
+          secrets.SENTRY_KEY }}"
```
piyalbasu: If we were going to add analytics/Sentry to Beta, I don't think we should use the Prod keys here.

IMO, I don't think we need Sentry tracking at all in Beta.

I can see an argument for being able to test Amplitude in Beta, but we should use a key specific to the beta environment for this so we don't pollute Production data

extension/src/constants/env.ts
```diff
@@ -1,8 +1,27 @@
...
-export { _AMPLITUDE_KEY as AMPLITUDE_KEY, _SENTRY_KEY as SENTRY_KEY };
+
+/** Version from package.json, injected at build time. */
+declare const APP_VERSION: string;
+const _APP_VERSION = APP_VERSION;
+
+/**
+ * Build type derived from webpack flags at compile time.
+ * - "development" — `yarn start` (dev server)
+ * - "beta" — `yarn build` (non-production extension build)
```
piyalbasu: this actually isn't accurate. You'll see that `submitBeta` uses the yarn:build:production command. The reason for this is that we want the build to be as close as possible to the production build when we're QA'ing

extension/src/helpers/metrics.ts
leofelix077: I added this more to keep in line with the mobile implementation, that we log the events locally as well for debugging, but honestly not sure how often we need to use it. for initial implementation it helped a lot, but it's a fair point / open if we need or want to keep it in the long run, as it likely wont change much after the implementation is done

@CassioMG  wdyt? in regards to mobile as well. maybe instead we go the opposite way and remove it on both sides if it's not used too often (or at all)

.github/workflows/submitBeta.yml
leofelix077: the amplitude key itself can be the same, but we have different build types for prod/beta/dev, so it can be filtered by the different environments. so it shouldn't be a problem to segment and filter it properly without having polluted data

extension/src/helpers/metrics.ts
leofelix077: oh yes. it should always be available. we don't need this logic

extension/src/helpers/metrics.ts
leofelix077: good catch. it came from taking the mobile implementation as a base (there it's generated manually)

getUserId from Math.random -> Claude suggested a different approach -> I changed to this with crypto. but we can just leave it for the SDK directly

extension/src/helpers/metrics.ts
CassioMG: We've added it on mobile because by default we avoid sending any events to Amplitude when developing locally (__DEV__) to prevent flooding Amplitude with trash events, which means we don't use the Amplitude API KEY when running things locally on mobile. So this debug screen was added to help seeing the events locally without having to manually connect to Amplitude locally (which basically means changing 1-2 lines of code locally).

Tbh, I'm not strongly in favor of keeping this analytics debug screen as I also never use it. But one thing to notice is that on mobile the Analytics logs on the debug screen is just one of the sections of the debug screen. This same debug screen is used for setting other convenient debug values locally like overriding Blockaid responses or Transaction failure responses so it's easy to simulate those scenarios.

extension/src/helpers/metrics.ts
CassioMG: @piyalbasu @leofelix077 I'm also not sure there is a lot of value in ourselves creating the id here but for the moment I'd keep it the same as mobile for consistency. We can remove it once we implement Freighter's user model so we'll have proper unique user ids which we could also use on Amplitude. How does that sound?

extension/src/popup/views/Send/index.tsx
```diff
@@ -107,6 +107,24 @@ export const Send = () => {
     });
 
   const [activeStep, setActiveStep] = React.useState(STEPS.AMOUNT);
+  const lastEmittedStep = React.useRef<STEPS | null>(null);
+
+  // Emit a screen-view metric only once per step transition.
+  React.useEffect(() => {
```
CassioMG:
```suggestion
  useEffect(() => {
```

extension/src/popup/views/Send/index.tsx
```diff
@@ -107,6 +107,24 @@ export const Send = () => {
     });
 
   const [activeStep, setActiveStep] = React.useState(STEPS.AMOUNT);
+  const lastEmittedStep = React.useRef<STEPS | null>(null);
```
CassioMG:
```suggestion
  const [activeStep, setActiveStep] = useState(STEPS.AMOUNT);
  const lastEmittedStep = useRef<STEPS | null>(null);
```

extension/src/popup/views/Send/index.tsx
```diff
@@ -107,6 +107,24 @@ export const Send = () => {
...
+  // Emit a screen-view metric only once per step transition.
+  React.useEffect(() => {
+    if (activeStep === lastEmittedStep.current) return;
+    lastEmittedStep.current = activeStep;
+
+    const metricByStep: Partial<Record<STEPS, string>> = {
+      [STEPS.AMOUNT]: METRIC_NAMES.sendPaymentAmount,
+      [STEPS.PAYMENT_CONFIRM]: METRIC_NAMES.sendPaymentConfirm,
+      [STEPS.DESTINATION]: METRIC_NAMES.sendPaymentRecentAddress,
+    };
```
CassioMG: this could probably live outside the useEffect since it's a static object

extension/src/popup/views/Swap/index.tsx
```diff
@@ -31,6 +31,27 @@ export const Swap = () => {
   const navigate = useNavigate();
   const location = useLocation();
   const [activeStep, setActiveStep] = React.useState(STEPS.AMOUNT);
```
CassioMG:
```suggestion
  const [activeStep, setActiveStep] = useState(STEPS.AMOUNT);
```

extension/.env.example
```diff
@@ -0,0 +1,11 @@
+INDEXER_URL=
+INDEXER_V2_URL=
+
+# For local development, analytics events are logged to console but not uploaded (no key needed).
+# To test Amplitude locally, add the line below with your Amplitude key, then run: yarn start
+#
+# AMPLITUDE_KEY=
+#
```
CassioMG, line 8: should we also add AMPLITUDE_EXPERIMENT_DEPLOYMENT_KEY here?

CassioMG: I see it's being handled on this other PR: https://github.com/stellar/freighter/pull/2666/changes#diff-5e8b4d2df7c32bf287ba46a22ace2c8dff10bd859a615b97173d2da3c8e4d758R8

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,440 @@ export interface MetricsData {
...
+
+  // Navigator.connection is not available in all browsers
+  const nav = navigator as Navigator & {
+    connection?: { type?: string; effectiveType?: string };
+  };
+
+  const context: Record<string, unknown> = {
+    publicKey: activePublicKey ? truncatePublicKey(activePublicKey) : "N/A",
+    platform: METRICS_PLATFORM,
+    platformVersion: navigator.userAgent,
```
CassioMG: I'm not sure we should log the whole `userAgent` here, maybe we could try coarsening to `browser name + major version`

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,440 @@ export interface MetricsData {
...
+    network: (networkDetails?.networkName ?? "").toUpperCase(),
```
CassioMG: on mobile we are sending the network identifier here, e.g. "PUBLIC" | "TESTNET"
I think we should do the same on extension for consistency, instead of using the network name which will log it as "MAIN NET" | "TEST NET"

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,440 @@ export interface MetricsData {
...
+  const activePublicKey = publicKeySelector(state);
+  const networkDetails = settingsNetworkDetailsSelector(state);
+
+  // Navigator.connection is not available in all browsers
+  const nav = navigator as Navigator & {
+    connection?: { type?: string; effectiveType?: string };
+  };
+
+  const context: Record<string, unknown> = {
+    publicKey: activePublicKey ? truncatePublicKey(activePublicKey) : "N/A",
```
CassioMG: on mobile we are sending the whole public key without any truncation

extension/src/helpers/metrics.ts
```diff
@@ -83,116 +79,440 @@ export interface MetricsData {
...
+    platform: METRICS_PLATFORM,
+    platformVersion: navigator.userAgent,
+    network: (networkDetails?.networkName ?? "").toUpperCase(),
+    connectionType: nav.connection?.type ?? "unknown",
+    appVersion: getAppVersion(),
+    bundleId: BUILD_TYPE,
```
CassioMG: I think using the build type as the `bundleId` is too generic considering both extension and mobile send events to the same project.

I'd suggest making it something like **bundleId: `extension.${BUILD_TYPE}`** , so we will have `"org.stellar.freighterdev" / "org.stellar.freighterwallet"` for mobile and `"extension.dev" / "extension.beta" / "extension.production"` for extension which removes any ambiguity.

Wdyt?

leofelix077: oh yes. makes sense. also makes easier for readability and finding it at a glance

extension/src/helpers/metrics.ts
leofelix077: this came from a comment from opencode to obfuscate it, not to have a direct identifier to any wallet. wdyt? it's not like we can easily track a wallet owner, but there are some extra info available that could track down an approximate wallet + location combination. wdyt?

extension/src/helpers/metrics.ts
piyalbasu: Yeah, in dev mode in the extension, we just console.log metrics rather than sending them off to Amplitude. If you navigate around the app in dev, you can see the history of metrics that would have been emitted in your console. This conse is probably a little more accessible in the browser than in mobile. I'd suggest not adding this so we don't have to worry about maintaining a feature that doesn't seem to have much of a value add

CassioMG: sounds good to me, let's remove it

CassioMG: I think we can also remove it on mobile to be honest

extension/src/helpers/metrics.ts
CassioMG: @leofelix077 humm yeah in this case let's also truncate on mobile so we keep both in sync. Would you mind doing that?

extension/src/helpers/metrics.ts
leofelix077: sounds good to me too. will clean it up 👌

leofelix077: https://github.com/stellar/freighter-mobile/issues/795 also created a ticket for mobile. if we have time during this sprint, I will pull it to my board during the week

extension/src/helpers/metrics.ts
leofelix077: for sure. I added to the same other ticket https://github.com/stellar/freighter-mobile/issues/795 (as more of a catch all, not to have 2 tiny tickets) 1. to remove the debug local logging and 2. truncate the addresses

PR #2666, opened 2026-03-25
Add maintenance mode through Amplitude

### LINKED ISSUE #2531: Maintenance mode through Amplitude
Currently, if we want to show users a maintenance mode banner, we can turn this on and configure the message in freighter backend v1. 

But what would we do if BE v1 goes down? We wouldn't be able to push a custom message to users. If we configure this using a feature flag in Amplitude, this removes our dependence on a functioning BE to be able to communicate with users.

1) Set up feature flags in Amplitude
2) Implement checking these flags in Freighter 
3) Configure the notification in Freighter based on flag: https://github.com/stellar/freighter/blob/master/extension/src/popup/views/Account/index.tsx#L242

*Bonus* Can we configure the message in Amplitude?

leofelix077:
Closes #2531 

banner-error
banner-modal-error
banner-modal-primary
banner-modal-warning
banner-primary
banner-secondary
banner-tertiary
banner-warning
screen-full-overlay
screen-minimal
Screenshot 2026-03-25 at 16 17 51
Screenshot 2026-03-25 at 16 17 55
Screenshot 2026-03-25 at 16 35 31

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,10 @@
 enableScripts: false
 
+npmMinimalAgeGate: 7d
+
+# Uncomment to bypass the age gate when we need a stellar-sdk release immediately
+# (e.g. critical bug fix or new Stellar protocol upgrade that Freighter must support on day one)
+# npmPreapprovedPackages:
+#   - "@stellar/stellar-sdk"
+
 nodeLinker: node-modules
--- a/package.json
+++ b/package.json
@@ -144,5 +144,5 @@
     "debug@4.4.2": "4.4.1",
     "ansi-styles@6.2.2": "6.2.3"
   },
-  "packageManager": "yarn@4.9.4"
+  "packageManager": "yarn@4.10.0"
 }


extension/src/popup/ducks/remoteConfig.ts
```diff
@@ -0,0 +1,200 @@
...
+ * Boolean flags — variant value is checked against ON_VARIANT_VALUES.
+ * Empty for now; add flag names here as new boolean flags are introduced.
+ */
+const BOOLEAN_FLAGS = [] as const;
+
+/**
+ * String flags — variant value is used directly.
+ * Empty for now; add flag names here as new string flags are introduced.
+ */
+const STRING_FLAGS = [] as const;
```
CassioMG: on mobile we don't have STRING_FLAGS, we actually have VERSION_FLAGS which receive a special treatment where we parse 1_2_3 => 1.2.3

CassioMG: I'd suggest updating it to VERSION_FLAGSn + special parsing so we don't need to worry about it whenever we need to use a version flag with extension so we keep same behavior on both platforms for consistency

extension/src/popup/ducks/remoteConfig.ts
```diff
@@ -0,0 +1,200 @@
...
+export const fetchFeatureFlags = createAsyncThunk(
+  "remoteConfig/fetchFeatureFlags",
+  async (): Promise<RemoteConfigState> => {
+    const client = getExperimentClient();
+
+    if (!client) {
+      return { ...initialState, isInitialized: true };
+    }
+
+    await client.fetch();
```
CassioMG: I think we need to fetch the flags passing `user_properties + bundleId` to rest assure it'll fetch the flags when users have opt out of analytics, the same we do on mobile:
```
      await experimentClient.fetch({
        user_properties: {
          // We need to explicitly pass the "Bundle Id" here to make sure the
          // correct flag values will be assigned regardless if (iOS) users
          // have ever enabled tracking permission or not.
          // Other common properties like "Platform" and "Version" appear to be
          // always forwarded regardless of tracking permission.
          [ANALYTICS_CONFIG.BUNDLE_ID_KEY]: getBundleId(),
        },
      });
```

extension/.env.example
```diff
@@ -5,7 +5,9 @@ INDEXER_V2_URL=
 # To test Amplitude locally, add the line below with your Amplitude key, then run: yarn start
 #
 # AMPLITUDE_KEY=
+# AMPLITUDE_EXPERIMENT_DEPLOYMENT_KEY=
```
CassioMG, line 8: I think we need to polish the comments on lines 4-5 a little bit to account for this new variable

extension/src/popup/ducks/remoteConfig.ts
CassioMG: could you please apply this change and test what happens when a user opt-out of analytics? the analytics metrics should stop being sent to Amplitude but we should still be able to fetch the feature flags with no issues

extension/src/popup/helpers/maintenance/parseMaintenanceContent.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Picks the localized string array for the current browser language.
+ * Falls back to English, then to an empty array.
+ *
+ * @param map - Map of language code → translated string array
+ * @returns The best matching localized string array
+ */
+function resolveLocalizedArray(map: LocalizedStringArray): string[] {
+  const lang = getLang();
+  return map[lang] ?? map.en ?? [];
+}
```
CassioMG, line 46: on mobile we do some filtering here to take only the string values from the array. E.g.:  if the Amplitude payload contains [\"text\", 42, null], mobile filters to ["text"] but extension would pass all values through. 

I think extension should add the same .filter() guard for consistency.

This is there I saw it on mobile:
```
const getLocalizedStringArray = (
  obj: LocalizedStringArray | undefined,
  fallback: string[],
): string[] => {
  if (!obj) return fallback;
  const lang = getDeviceLanguage();
  const value = obj[lang] ?? obj.en;
  if (!Array.isArray(value)) return fallback;
  return value.filter(
    (v): v is string => typeof v === "string" && v.trim().length > 0,
  ); // ========================> filter out non-string values
};
```

leofelix077:
@CassioMG  finished adjusting the UI + open comments from review


Screenshot 2026-03-30 at 13 49 24
Screenshot 2026-03-30 at 13 49 28
Screenshot 2026-03-30 at 13 57 39
Screenshot 2026-03-30 at 13 57 39

extension/src/popup/ducks/remoteConfig.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    if (!client) {
+      return { ...initialState, isInitialized: true };
+    }
+
+    // Pass user_properties explicitly so flags are fetched correctly even when
+    // the user has opted out of analytics (mirrors mobile implementation).
+    await client.fetch({
+      user_properties: {
+        "Bundle Id": `extension.${BUILD_TYPE}`,
```
CassioMG: could we extract "Bundle Id" to a const and `extension.${BUILD_TYPE}` to a getBundleId helper? I think we use it in multiple places now

leofelix077: oh yes. done 👌

Issue #2640, opened 2026-03-05
Sign Auth Entry should show network information


piyalbasu:
In Sign Auth Entry, we show the address that is signing and the authorizations, but we don't show the network. We should show the network information here, as well

<img width="712" height="1196" alt="Image" src="https://github.com/user-attachments/assets/64812ad0-c258-43cc-892f-a58a329ee147" />

PR #2667, opened 2026-03-26
Show network on dApp requests

leofelix077:
closes https://github.com/stellar/freighter/issues/2640

Also hardens network validation across Sign Auth Entry and Sign Message flows to prevent bypassing network checks by embedding a mismatched network passphrase in the request body. Previously, it was possible to send a request claiming one network (e.g. Testnet) while the auth entry itself was configured for a different network (e.g. Mainnet). This fix brings the extension in line with mobile validation behavior.

---

🔒 GitHub Advanced Security automatically protects Copilot coding agent pull requests. You can protect all pull requests by enabling Advanced Security for your repositories. [Learn more about Advanced Security.](https://gh.io/cca-advanced-security)

### DIFF
--- a/extension/e2e-tests/integration-tests/freighterApiIntegration.test.ts
+++ b/extension/e2e-tests/integration-tests/freighterApiIntegration.test.ts
@@ -929,9 +929,7 @@ test("should show network mismatch warning when signing auth entry for wrong net
   const popup = await popupPromise;
 
   await expect(
-    popup.getByText(
-      "The authorization entry is for a different network than the one you are connected to.",
-    ),
+    popup.getByText(/The authorization entry is for Test Net/),
   ).toBeVisible();
   await expect(
     popup.getByText(
@@ -1005,35 +1003,6 @@ test("should show invalid entry warning when auth entry is not a Soroban authori
     popup.getByText(
       "The authorization entry is malformed or contains invalid data.",
     ),
-  ).toBeVisible();
-});
-
-test("should show network warning when signing message without network passphrase", async ({
-  page,
-  extensionId,
-  context,
-}) => {
-  await loginToTestAccount({ page, extensionId, context, isIntegrationMode });
-  await allowDapp({ page });
-
-  const pageTwo = await page.context().newPage();
-  await pageTwo.waitForLoadState();
-
-  const popupPromise = page.context().waitForEvent("page");
-  await pageTwo.goto("https://docs.freighter.app/docs/playground/signMessage");
-  await pageTwo.getByRole("textbox").first().fill(MSG_TO_SIGN);
-  // leave networkPassphrase field (nth(1)) empty — falsy value bypasses the old guard
-  await pageTwo.getByText("Sign message").click();
-
-  const popup = await popupPromise;
-
-  await expect(
-    popup.getByText(
-      "The requester did not specify a network for this message.",
-    ),
-  ).toBeVisible();
-  await expect(
-    popup.getByText("Signing this message is not possible at the moment."),
   ).toBeVisible();
 });
 


extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -200,6 +200,15 @@ export const SignAuthEntry = () => {
                   <KeyIdenticon publicKey={publicKey} />
                 </div>
               </div>
+              <div className="SignAuthEntry__Metadata__Row">
```
leofelix077, line 248: Original ticket only mentions authEntry, but wanted to keep it inline with mobile, where it's displayed in all dApp requests, so added for the other functions as well

extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -124,19 +120,62 @@ export const SignAuthEntry = () => {
...
+    );
+  }
+
+  // Cryptographically validate the networkId embedded in the XDR against the
+  // wallet's active network. This is stronger than trusting the dApp-supplied
+  // networkPassphrase string (which may be absent or spoofed).
+  let sorobanAuth: ReturnType<typeof preimage.sorobanAuthorization>;
+  try {
+    sorobanAuth = preimage.sorobanAuthorization();
+    const embeddedNetworkId = sorobanAuth.networkId();
```
leofelix077, line 148: while doing the tests I noticed that there were some differences with mobile in regards to validation

didn't think it was worth creating a separate ticket just for that. maybe broaden the original ticket to cover more? wdyt?

before it was possible to bypass the network check, because it could send e.g. as test net as part of the body ,but inside the auth entry, have it configured as main net, so it would bypass the checks

CassioMG: I'm fine keeping it on same PR, thanks for fixing this

@shared/constants/stellar.ts
```diff
@@ -72,3 +72,9 @@ export const DEFAULT_NETWORKS: Array<NetworkDetails> = [
 
 export const BASE_RESERVE = 0.5 as const;
 export const BASE_RESERVE_MIN_COUNT = 2 as const;
+
+export const PASSPHRASE_TO_NETWORK_NAME: Record<string, string> = {
+  [Networks.PUBLIC]: NETWORK_NAMES.PUBNET,
+  [Networks.TESTNET]: NETWORK_NAMES.TESTNET,
+  [FUTURENET_NETWORK_DETAILS.networkPassphrase]: NETWORK_NAMES.FUTURENET,
```
CassioMG, line 79: this is looking a bit confusing, the PASSPHRASE_TO_NETWORK_NAME const name gives the impression that it maps passphrase to network name but this is only happening for futurenet. Is this correct? Why only futurenet?

CassioMG: aaah, nevermind I realize now that it's using `Networks` from the Stellar SDK which is actually the passphrase 👍

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -450,6 +450,15 @@ export const SignTransaction = () => {
...
                   </div>
+                  <div className="SignTransaction__Metadata__Row">
+                    <div className="SignTransaction__Metadata__Label">
+                      <Icon.Globe02 />
+                      <span>{t("Network")}</span>
+                    </div>
+                    <div className="SignTransaction__Metadata__Value">
+                      <span>{networkName}</span>
+                    </div>
+                  </div>
```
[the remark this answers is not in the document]
CassioMG: @copilot it appears this is fixed already, right? since we are not using `signTxState.data?.networkDetails!` anymore on this file

extension/src/popup/views/SignMessage/index.tsx
```diff
@@ -165,8 +166,9 @@ export const SignMessage = () => {
         header={`${t("Freighter is set to")} ${networkName}`}
       >
         <p>
-          {`${t("The requester expects you to sign this message on")} `}
-          {blobNetworkPassphrase}.
+          {blobNetworkPassphrase
+            ? `${t("The requester expects you to sign this message on")} ${PASSPHRASE_TO_NETWORK_NAME[blobNetworkPassphrase] ?? blobNetworkPassphrase}.`
+            : t("The requester did not specify a network for this message.")}
```
piyalbasu: This line actually isn't reachable. If `blobNetworkPassphrase`  is empty, we won't render this component at all. While this is technically harmless, it is dead code and it does send the message to future maintainers that networkPassphrase should not be optional, when it is

extension/src/popup/views/SignTransaction/index.tsx
```diff
@@ -261,7 +262,21 @@ export const SignTransaction = () => {
...
+  // Handle potential missing networkDetails
+  if (!signTxState.data?.networkDetails) {
+    return (
+      <WarningMessage
+        variant={WarningMessageVariant.warning}
+        handleCloseClick={() => window.close()}
+        isActive
+        header={t("Unable to load network details")}
+      >
+        <p>{t("Please try again or contact support if the issue persists.")}</p>
```
piyalbasu: Very NIT: I think "contact support" is kind of a confusing directive because we don't really provide a user a way to do that in this message (like providing a phone number or email). I would maybe just put `Please try again.`. Ideally Sentry should report this to us automatically

extension/src/popup/views/SignAuthEntry/index.tsx
piyalbasu: Yeah, I'm also fine with doing it here. let's just make a note in the GH issue to track that we're addressing this

extension/src/popup/views/SignAuthEntry/index.tsx
```diff
@@ -124,19 +120,62 @@ export const SignAuthEntry = () => {
...
+      return (
+        <WarningMessage
+          variant={WarningMessageVariant.warning}
+          handleCloseClick={() => rejectAndClose()}
+          isActive
+          header={`${t("Freighter is set to")} ${networkName}`}
+        >
+          <p>
+            {t(
+              "The authorization entry is for a different network than the one you are connected to.",
```
piyalbasu: Let's tell the user what network the authorization is configured for. We do that on the other "mismatch" messages

PR #2669, opened 2026-03-27
Auto Focus password fields

### LINKED ISSUE #2565: Auto focus password field when wallet popup opens
When opening the wallet extension popup, the cursor is not automatically placed in the password input. Would it be possible to auto focus the password field once the popup finishes loading?

Image

leofelix077:
Closes #2565 

https://github.com/user-attachments/assets/fd58f074-1239-4585-93a9-8678e795b4b4

https://github.com/user-attachments/assets/ce69632f-90c8-4db4-90cd-b3a369cf266f

### DIFF
--- a/.github/actions/validate-branch-from/action.yml
+++ b/.github/actions/validate-branch-from/action.yml
@@ -1,0 +1,95 @@
+name: "Validate Branch From"
+description:
+  "Validates that branch_from is 'master' or a semver tag and provides available
+  version tags"
+
+inputs:
+  branch_from:
+    description:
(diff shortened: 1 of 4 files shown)


extension/src/popup/components/EnterPassword/index.tsx
```diff
@@ -101,6 +101,7 @@ export const EnterPassword = ({
                         isPassword
                         fieldSize="md"
                         autoComplete="off"
+                        autoFocus
```
leofelix077, line 104: added for both pw fields for consistency

on lock screen and when entering the show recovery phrase

PR #2676, opened 2026-04-03
Update default build type for local development

leofelix077:
Screenshot 2026-04-03 at 11 25 01
Screenshot 2026-04-03 at 11 27 19


piyalbasu (approved):
I agree with the 2 copilot comments. But otherwise lgtm!

PR #2687, opened 2026-04-09
add v1 of freighter extension best practices skill

leofelix077:
# Skill Eval: `freighter-best-practices` — Benchmark Report

**Date:** 2026-04-09
**Iterations:** 1
**Repo:** [stellar/freighter](https://github.com/stellar/freighter)
**Model:** Claude Opus 4.6

## Summary

# on further passes both get close to 100%, as Claude picks up automatically on the rules and the entry points

| Metric | With Skill | Without Skill | Delta |
|---|---|---|---|
| Pass rate | **81.5%** | **74.1%** | +7.4pp |

### anti-patterns-refactor

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Adds error handling with captureException (no empty catch block) | 1/1 (100%) | 0/1 (0%) |
| 2 | Changes from default export to named export | 1/1 (100%) | 1/1 (100%) |
| 3 | Changes from function declaration to arrow function expression | 1/1 (100%) | 1/1 (100%) |
| 4 | Extracts inline onClick to useCallback | 1/1 (100%) | 0/1 (0%) |
| 5 | Moves enrichedTokens computation to createSelector or useMemo | 1/1 (100%) | 1/1 (100%) |
| 6 | Replaces Function with proper typed callback | 1/1 (100%) | 1/1 (100%) |
| 7 | Replaces any[] with proper TypeScript types | 1/1 (100%) | 1/1 (100%) |
| 8 | Replaces array index key with stable unique identifier | 1/1 (100%) | 1/1 (100%) |
| 9 | Replaces chrome.runtime.sendMessage with sendMessageToBackground | 1/1 (100%) | 1/1 (100%) |
| 10 | Uses double quotes in string literals | 1/1 (100%) | 1/1 (100%) |
| 11 | Wraps list item in React.memo() | 1/1 (100%) | 0/1 (0%) |
| 12 | Wraps user-facing strings in t() | 1/1 (100%) | 1/1 (100%) |

### architecture-redux-duck

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Background store hydration saves to chrome.storage.session on changes | 0/1 (0%) | 0/1 (0%) |
| 2 | Does NOT add a new React context for data sharing | 1/1 (100%) | 1/1 (100%) |
| 3 | Does NOT put view-local filter state in Redux | 1/1 (100%) | 1/1 (100%) |
| 4 | No module-level mutable state in background code | 1/1 (100%) | 0/1 (0%) |
| 5 | Uses ActionStatus enum for async operation tracking | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses Redux (createSlice/createAsyncThunk) for the cross-cutting dApp connections... | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses useReducer with RequestState pattern for view-local filtering/sorting | 1/1 (100%) | 1/1 (100%) |

### architecture-state-decision

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | All user-facing error messages wrapped in t() | 1/1 (100%) | 1/1 (100%) |
| 2 | Named export (not default export) | 1/1 (100%) | 1/1 (100%) |
| 3 | No magic strings - error codes referenced via enum values | 1/1 (100%) | 1/1 (100%) |
| 4 | Provides both English (en) and Portuguese (pt) translations for all error messag... | 0/1 (0%) | 0/1 (0%) |
| 5 | Uses enum for Horizon error codes (not union type or plain strings) | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses t() with interpolation for dynamic values, not template literals | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses unknown (not any) for untyped error inputs | 1/1 (100%) | 1/1 (100%) |

### code-style-scss

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | File extension is .scss (not .css or .module.css) | 1/1 (100%) | 1/1 (100%) |
| 2 | No hardcoded color values (hex, rgb, etc.) | 1/1 (100%) | 1/1 (100%) |
| 3 | No raw pixel values for spacing/sizing | 0/1 (0%) | 0/1 (0%) |
| 4 | Uses BEM naming convention with kebab-case block names | 0/1 (0%) | 0/1 (0%) |
| 5 | Uses pxToRem() for size values instead of raw pixels | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses var(--sds-clr-*) design system color tokens | 1/1 (100%) | 1/1 (100%) |

### code-style-todo-magic

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | No module-level mutable state for caching (if background, uses chrome.storage) | 0/1 (0%) | 0/1 (0%) |
| 2 | Numeric literals like cache duration and multiplier are SCREAMING_SNAKE_CASE con... | 1/1 (100%) | 1/1 (100%) |
| 3 | TODO comment references a GitHub issue number: // TODO(#1234): description | 1/1 (100%) | 0/1 (0%) |
| 4 | Uses ?? for nullish coalescing instead of || | 0/1 (0%) | 0/1 (0%) |
| 5 | Uses named export | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses unknown or proper types instead of any | 1/1 (100%) | 1/1 (100%) |

### err-handling-bg-handler

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Adds new entry to SERVICE_TYPES enum | 0/1 (0%) | 0/1 (0%) |
| 2 | Calls captureException for unexpected errors | 1/1 (100%) | 0/1 (0%) |
| 3 | Does NOT call captureException for expected errors like contract-not-found | 1/1 (100%) | 1/1 (100%) |
| 4 | Handler function never throws - always returns structured response object | 1/1 (100%) | 1/1 (100%) |
| 5 | Registers handler with case statement in the message listener | 0/1 (0%) | 0/1 (0%) |
| 6 | Returns { error: string } for error cases, not thrown exceptions | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses sendMessageToBackground from popup side, NOT direct chrome.runtime.sendMess... | 1/1 (100%) | 0/1 (0%) |

### err-handling-thunk

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Calls captureException from Sentry for unexpected errors | 1/1 (100%) | 1/1 (100%) |
| 2 | Does NOT use console.error for error reporting | 1/1 (100%) | 0/1 (0%) |
| 3 | Error handling uses rejectWithValue with both instanceof Error and fallback bran... | 1/1 (100%) | 1/1 (100%) |
| 4 | Uses createAsyncThunk from Redux Toolkit | 1/1 (100%) | 1/1 (100%) |
| 5 | Uses createSelector for any derived state | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses createSlice with ActionStatus enum for loading/success/error states | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses interface for state shape, not type alias | 1/1 (100%) | 1/1 (100%) |
| 8 | Uses named exports, not default exports | 1/1 (100%) | 1/1 (100%) |

### i18n-settings

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | All user-facing strings wrapped in t() with interpolation syntax, not template l... | 1/1 (100%) | 1/1 (100%) |
| 2 | Boolean prop uses is/has/should prefix | 1/1 (100%) | 0/1 (0%) |
| 3 | Does NOT use React.FC<Props> pattern | 1/1 (100%) | 1/1 (100%) |
| 4 | Event handler props use on prefix, internal handlers use handle prefix | 1/1 (100%) | 1/1 (100%) |
| 5 | Provides both English (en) and Portuguese (pt) translation entries | 1/1 (100%) | 1/1 (100%) |
| 6 | SCSS uses BEM naming and var(--sds-clr-*) tokens | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses enum (not union type) for auto-lock timeout options | 1/1 (100%) | 1/1 (100%) |
| 8 | Uses enum (not union type) for currency options | 1/1 (100%) | 1/1 (100%) |
| 9 | Uses interface for component props shape | 1/1 (100%) | 1/1 (100%) |

### messaging-service-type

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Adds value to SERVICE_TYPES enum (not creating a new enum) | 1/1 (100%) | 1/1 (100%) |
| 2 | Creates handler function in handlers/ directory | 1/1 (100%) | 1/1 (100%) |
| 3 | Does NOT call browser.runtime.sendMessage or chrome.runtime.sendMessage directly... | 1/1 (100%) | 1/1 (100%) |
| 4 | Handler returns structured response object, never throws | 1/1 (100%) | 1/1 (100%) |
| 5 | Popup sends via sendMessageToBackground() helper | 1/1 (100%) | 1/1 (100%) |
| 6 | Registers handler in listener with case statement | 1/1 (100%) | 1/1 (100%) |
| 7 | Response uses domain-specific fields for success, { error: string } for failure | 0/1 (0%) | 1/1 (100%) |

### performance-list

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | All user-facing text wrapped in t() | 1/1 (100%) | 0/1 (0%) |
| 2 | CSS uses var(--sds-clr-*) design tokens, not hardcoded colors | 1/1 (100%) | 1/1 (100%) |
| 3 | Callbacks passed as props wrapped in useCallback | 1/1 (100%) | 1/1 (100%) |
| 4 | List item component is wrapped in React.memo() | 1/1 (100%) | 1/1 (100%) |
| 5 | No inline arrow functions inside .map() or list render | 0/1 (0%) | 0/1 (0%) |
| 6 | SCSS uses BEM naming convention with kebab-case | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses createSelector for derived/computed values from Redux state | 1/1 (100%) | 0/1 (0%) |
| 8 | Uses pxToRem() instead of raw pixel values | 0/1 (0%) | 1/1 (100%) |
| 9 | Uses stable unique identifiers as keys, not array index | 1/1 (100%) | 1/1 (100%) |

### performance-selectors

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Does NOT compute derived state inline in the component | 0/1 (0%) | 0/1 (0%) |
| 2 | Does not add a new React context for cross-slice data - uses Redux selectors | 1/1 (100%) | 1/1 (100%) |
| 3 | Objects/arrays created for props are wrapped in useMemo | 0/1 (0%) | 1/1 (100%) |
| 4 | Uses ?? (nullish coalescing) instead of || for fallback values | 0/1 (0%) | 0/1 (0%) |
| 5 | Uses createSelector to derive data across multiple Redux slices | 1/1 (100%) | 1/1 (100%) |
| 6 | Uses interface for props shape | 0/1 (0%) | 1/1 (100%) |
| 7 | Uses template literals (not string concatenation) for any string building | 1/1 (100%) | 1/1 (100%) |

### security-ext-msg

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Adds new entry to EXTERNAL_SERVICE_TYPES enum (not SERVICE_TYPES) | 1/1 (100%) | 1/1 (100%) |
| 2 | All user-facing strings wrapped in t() from useTranslation | 1/1 (100%) | 1/1 (100%) |
| 3 | Background handler validates message type against enum | 1/1 (100%) | 0/1 (0%) |
| 4 | Background handler validates sender origin | 1/1 (100%) | 1/1 (100%) |
| 5 | Content script only forwards messages with valid EXTERNAL_SERVICE_TYPES | 0/1 (0%) | 1/1 (100%) |
| 6 | Handler returns structured response, never throws | 1/1 (100%) | 1/1 (100%) |
| 7 | Popup uses navigateTo with ROUTES enum for navigation | 0/1 (0%) | 0/1 (0%) |
| 8 | Uses double quotes in all string literals | 1/1 (100%) | 1/1 (100%) |

### security-key-export

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Calls captureException for unexpected errors during encryption/decryption | 1/1 (100%) | 0/1 (0%) |
| 2 | Does NOT use non-null assertion (!) on key material - handles undefined explicit... | 1/1 (100%) | 1/1 (100%) |
| 3 | Does not store any decrypted secrets in chrome.storage.local | 1/1 (100%) | 1/1 (100%) |
| 4 | Never logs private keys, mnemonics, or decrypted key material | 1/1 (100%) | 1/1 (100%) |
| 5 | Reads decrypted key material from chrome.storage.session (not chrome.storage.loc... | 0/1 (0%) | 1/1 (100%) |
| 6 | Returns structured response object, never throws | 1/1 (100%) | 1/1 (100%) |
| 7 | Uses unknown or proper types instead of any for error handling | 1/1 (100%) | 1/1 (100%) |

### testing-thunk

| # | Assertion | With Skill | Without Skill |
|---|---|---|---|
| 1 | Creates test store with configureStore() | 1/1 (100%) | 1/1 (100%) |
| 2 | Dispatches thunks and asserts via store.getState() | 1/1 (100%) | 1/1 (100%) |
| 3 | Test file located in __tests__/ directory | 1/1 (100%) | 1/1 (100%) |
| 4 | Test file uses .test.ts extension, not .spec.ts | 1/1 (100%) | 1/1 (100%) |
| 5 | Tests error path verifying captureException is NOT called for expected errors | 0/1 (0%) | 0/1 (0%) |
| 6 | Tests success path with proper state assertions | 1/1 (100%) | 1/1 (100%) |
| 7 | Tests unexpected error path verifying captureException IS called | 0/1 (0%) | 0/1 (0%) |
| 8 | Uses jest.requireMock for typed mock access | 1/1 (100%) | 0/1 (0%) |


docs/skills/freighter-best-practices/references/anti-patterns.md
```diff
@@ -0,0 +1,164 @@
...
+time. Any variable declared outside a function will be reset.
+
+**Use `chrome.storage.session`** for ephemeral data or
+**`chrome.storage.local`** for persistent data.
+
+## 5. Direct chrome.runtime.sendMessage from Popup
+
+```typescript
+// WRONG: bypasses the shared API layer
+chrome.runtime.sendMessage({ type: SERVICE_TYPES.GET_DATA });
```
piyalbasu, line 70: We should `browser.runtime` for cross browser functionality throughout this doc

Maybe this should also be it's own entry: always prefer the `browser` lib unless you're doing something specific for Chrome or Firefox

piyalbasu (commented):
I love the `with skill` vs `without skill` breakdown! It seems like a lot of these were already solved without the skill (100% without skill). Is it fair to say we don't need to include the skill attribute if it Claude was already handling this issues before?

leofelix077:
@piyalbasu so I'm not quite sure if we can safely remove it or keep it as enforcement, due to the non-deterministic nature of the LLM. 

I did run the tests on my local repo, and from what I know is that Claude keeps some project knowledge/context (not long-term memory, but kind of a cache), so there's the possibility of not getting it quite right on a completely clean environment + just ignoring the rules from time to time depending on the files you are changing. 

just thought about it now, but in this eval it runs through the entire codebase more or less, with Opus that has a large context window, so I'd say this is the best case scenario, which might not translate or more punctual changes


I will try to figure out running it on a weaker model / clean environment / asking just for a subset of the changes. at least in one of these PRs to have a more definite answer on how it compares if you would run on Sonnet or Haiku, right after cloning the repo. If it still gets the same ones right the entire time

leofelix077:
### Code review

Found 2 issues:

1. Broken symlink — the target has one too many `../` levels. The symlink lives at `.claude/skills/`, so going up 3 levels exits the repository. Should be `../../docs/skills/freighter-best-practices`, not `../../../docs/skills/freighter-best-practices`.

https://github.com/stellar/freighter/blob/d43294e420c8fdbb076d207eb0eccb652dadea13/.claude/skills/freighter-best-practices

2. `navigateTo` "CORRECT" example is missing its required second argument. The actual signature is `navigateTo(path: ROUTES, navigate: NavigateFunction, queryParams?: string)` with no default for `navigate`, so `navigateTo(ROUTES.sendPayment)` is a compile error. Should be `navigateTo(ROUTES.sendPayment, navigate)`.

https://github.com/stellar/freighter/blob/d43294e420c8fdbb076d207eb0eccb652dadea13/docs/skills/freighter-best-practices/references/anti-patterns.md#L91-L96

piyalbasu:
> @piyalbasu so I'm not quite sure if we can safely remove it or keep it as enforcement, due to the non-deterministic nature of the LLM.
> 
> I did run the tests on my local repo, and from what I know is that Claude keeps some project knowledge/context (not long-term memory, but kind of a cache), so there's the possibility of not getting it quite right on a completely clean environment + just ignoring the rules from time to time depending on the files you are changing.
> 
> just thought about it now, but in this eval it runs through the entire codebase more or less, with Opus that has a large context window, so I'd say this is the best case scenario, which might not translate or more punctual changes
> 
> I will try to figure out running it on a weaker model / clean environment / asking just for a subset of the changes. at least in one of these PRs to have a more definite answer on how it compares if you would run on Sonnet or Haiku, right after cloning the repo. If it still gets the same ones right the entire time

I did some preliminary investigation into this, and Claude doesn't seem to think the skills as presented here will bring much value to agents ☹️ Here's what my investigation came up with. It might be worth exploring this a bit on your end:

```
Took a look at the eval and the skill structure to figure out whether the skill is actually pulling its weight on top of `AGENTS.md`. I think there's a structural redundancy worth resolving before we land
  more skills like this.

  ## The skill's `SKILL.md` duplicates `AGENTS.md`'s routing

  `SKILL.md` is 41 lines: frontmatter + a one-paragraph intro + an 11-row routing table mapping concerns → `references/*.md` files.

  `AGENTS.md` already has a "Best Practices Entry Points" section (lines 150–166) with the same 11 rows pointing at the same files with the same "When to Read" hints. Destinations are literally identical (just
  relative vs absolute paths).

  `AGENTS.md` is auto-loaded into context at session start, so its routing is present in every session in this repo without invoking the skill. This means there's no marginal routing the skill provides that
  `AGENTS.md` doesn't.

  ## Implication for the +7.4pp benchmark result

  I don't think the eval is measuring what we'd want it to. The PR adds `AGENTS.md` + reference files + skill together. Unless the "without skill" arm also strips `AGENTS.md` and the references — which it
  almost certainly doesn't, since Claude Code auto-loads `AGENTS.md` — the delta is the marginal effect of "skill mechanism on top of an already-loaded `AGENTS.md` that points at the same files."

  That marginal effect should be near zero by inspection, and a noisy +7.4pp at N=1 is consistent with near-zero (95% CI on a single run of 95 binary trials is roughly ±10pp).

  The point Leo flagged — that this is a best-case scenario and may not generalize — is also real, but I think it's downstream of this. The first thing to nail down is whether the skill mechanism adds anything
  over `AGENTS.md` at all.

  ## The reference content is real and worth keeping

  To be clear: the value-add of this PR is the reference files in `docs/skills/freighter-best-practices/references/`. Those didn't exist in the repo before, and `AGENTS.md` (which is also new in this PR) routes
   to them. That's where the actual contribution is, and it's substantive.

  The thing I think is dead weight is the skill mechanism wrapping it.

  ## Proposed path forward

  1. **Keep** all the `references/*.md` files — they're the real contribution.
  2. **Delete** `SKILL.md` and the `.claude/skills/freighter-best-practices` symlink (which is broken anyway per the earlier review comment). The skill is a duplicate router.
  3. **Rename** `docs/skills/freighter-best-practices/references/` → `docs/best-practices/` since it's no longer a skill. Update the `AGENTS.md` routing table.
  4. **Skip the larger eval** — the static check answers the load-bearing question, so the run-it-on-Sonnet/Haiku follow-up isn't needed for this PR.

  If we want a decision rule for future skills: a skill should only land if its content (or routing) isn't already in `AGENTS.md`, and we can show one before/after diff where its presence visibly changes
  Claude's output. If a contributor can't produce that diff in 30 minutes, the skill probably isn't doing anything detectable, and the maintenance cost isn't worth it.
  ```

**Comment by leofelix077** [comment]
## Clean-Slate Benchmark: 4-Config Comparison

> This benchmark was run from a **completely cold-start environment** — no existing project knowledge, fresh repo clone, no prior context. The point was to test how the skill performs for a developer cloning the repo for the first time and asking Claude to do common tasks, compared to other guidance strategies.

### Methodology

4 configurations tested against the same 4 prompts and 20 assertions (5 per prompt):

| Config | What it gets |
|--------|-------------|
| **Baseline** | No skill, no references — just the task prompt |
| **Minimal refs** | 132 lines of targeted doc sections covering only the 7 patterns the baseline missed |
| **Full refs** | All 11 reference files injected (1,855 lines) |
| **With skill** | `freighter-best-practices` SKILL.md only |

Each config ran as an independent subagent with no memory of the other runs. Same 4 eval prompts, same 20 assertions graded identically.

---

### Overall Results

| Config | Passed | Score |
|--------|--------|-------|
| 🔴 Baseline (no skill, no refs) | 13/20 | **65%** |
| 🟡 Minimal refs (targeted 132 lines) | 14/20 | **70%** |
| 🟠 Full refs (all 11 docs, 1,855 lines) | 15/20 | **75%** |
| 🟢 With skill | 20/20 | **100%** |

---

### Per-Eval Breakdown

#### Eval 1 — Anti-patterns: refactor `TokenList` with perf best practices

| Assertion | Baseline | Minimal refs | Full refs | With skill |
|-----------|:--------:|:------------:|:---------:|:----------:|
| `React.memo()` wrapping | ✅ | ✅ | ✅ | ✅ |
| Custom `isEqual` comparator (2nd arg to memo) | ❌ | ❌ | ✅ | ✅ |
| Named `interface TokenListItemProps` | ✅ | ✅ | ✅ | ✅ |
| `TokenListItem` is a named export | ❌ | ❌ | ❌ | ✅ |
| No inline arrow fn in JSX `onClick` | ✅ | ✅ | ✅ | ✅ |
| **Score** | **3/5** | **3/5** | **4/5** | **5/5** |

#### Eval 2 — Error handling: background service handler + popup caller

| Assertion | Baseline | Minimal refs | Full refs | With skill |
|-----------|:--------:|:------------:|:---------:|:----------:|
| `SERVICE_TYPES` SCREAMING_SNAKE_CASE enum entry | ✅ | ✅ | ✅ | ✅ |
| `captureException()` from `@sentry/browser` | ❌ | ✅ | ❌ | ✅ |
| Clean discriminated union `{ domain } \| { error }` | ❌ | ❌ | ❌ | ✅ |
| Uses `sendMessageToBackground()` | ✅ | ✅ | ✅ | ✅ |
| Safe default before `try` block | ✅ | ✅ | ✅ | ✅ |
| **Score** | **3/5** | **4/5** | **3/5** | **5/5** |

#### Eval 3 — Architecture: Redux duck for recent dApps connections

| Assertion | Baseline | Minimal refs | Full refs | With skill |
|-----------|:--------:|:------------:|:---------:|:----------:|
| `ActionStatus` from `@shared/api/types` in state | ❌ | ❌ | ❌ | ✅ |
| `createSelector` for all exported selectors | ✅ | ✅ | ✅ | ✅ |
| Named `reducer` export (`export const { reducer }`) | ✅ | ✅ | ✅ | ✅ |
| Named interfaces for state + model | ✅ | ✅ | ✅ | ✅ |
| Named constant cap (`MAX_RECENT_DAPPS`) | ✅ | ✅ | ✅ | ✅ |
| **Score** | **4/5** | **4/5** | **4/5** | **5/5** |

#### Eval 4 — Code style: Privacy settings component + SCSS

| Assertion | Baseline | Minimal refs | Full refs | With skill |
|-----------|:--------:|:------------:|:---------:|:----------:|
| Enum values are PascalCase (not SCREAMING_SNAKE_CASE) | ❌ | ❌ | ❌ | ✅ |
| BEM `__` double-underscore for elements (not `--`) | ✅ | ❌ | ✅ | ✅ |
| All px values wrapped in `pxToRem()` | ✅ | ✅ | ✅ | ✅ |
| All strings wrapped with `t()` | ✅ | ✅ | ✅ | ✅ |
| Named export, no `React.FC` annotation | ✅ | ✅ | ✅ | ✅ |
| **Score** | **4/5** | **3/5** | **4/5** | **5/5** |

---

### Key Findings

**3 patterns the skill uniquely encodes — 0% without it regardless of references:**

1. **`ActionStatus` from `@shared/api/types`** — The Redux duck pattern requires including a status field using the project's shared `ActionStatus` enum. None of the three non-skill configs did this, even with all 11 reference files injected. The concept only appears buried in existing duck files (not documented explicitly), so the skill encodes it as an explicit rule.

2. **PascalCase enum values for domain types** — Models like `AnalyticsConsent.Enabled` (not `ENABLED`) follow a Freighter-specific convention that contradicts what TypeScript developers expect by default (SCREAMING_SNAKE_CASE). All 3 non-skill configs defaulted to SCREAMING_SNAKE_CASE. The skill names this explicitly.

3. **Clean discriminated union return types** — Freighter bg handlers should return `{ nftCollections } | { error }`, not `{ nftCollections, error: '' }`. Without the skill, agents consistently initialize a mixed result object and return it from both success and error paths — never achieving a true discriminated union.

**The `isEqual` comparator and BEM `__` separator** were the only assertions where injecting the full reference docs helped (vs. baseline), suggesting those patterns are present in the docs but buried in context. Still not 100% reliable across configs.

**Minimal refs underperformed full refs** — injecting only 132 targeted lines actually degraded BEM behavior (eval 4 regressed from ✅ to ❌ vs baseline) while helping `captureException` (eval 2). This is a signal that partial context can mislead; the skill wins by being explicit and declarative rather than relying on example comprehension.

---

### Summary

The skill provides a **35-point lift over baseline** (100% vs 65%) and a **25-point lift over injecting all reference docs** (100% vs 75%), with zero token overhead vs. loading 1,855 lines of raw reference files. The patterns the skill uniquely solves are exactly the ones that diverge from TypeScript/React defaults — conventions a new contributor would have no reason to know without being explicitly told.

**Review by piyalbasu — APPROVED** [review]
Thanks for providing that detailed analysis! This looks to be really helpful for our codebase. Aside from the Copilot comments that should be addressed, this is LGTM!

File: docs/skills/freighter-best-practices/references/performance.md [hunk 2]
```diff
@@ -0,0 +1,241 @@
+# Performance -- Freighter Extension
+
+Based on analysis of 224 component files. Current performance score: **5.9/10**.
```
CassioMG: **[Comment 1 — Suggestion] Hard-coded statistics throughout the skill will go stale immediately**

This line states `Based on analysis of 224 component files. Current performance score: **5.9/10**.` — both numbers are point-in-time snapshots that will become incorrect as the codebase evolves.

This is the most prominent example of a pattern repeated across multiple skill files:

**`code-style.md`:** "~200+ type definitions", "85% arrow functions", "75% inline destructuring", "12% renaming", "659 vs 405 vs 41 occurrences", "95-97% named exports", "60% / 40%", "169 files", "1.3:1 ratio, 162 vs 124", "554 occurrences", "99%+", "17 occurrences", "50+ files", "30 / 40 / 30 / 15 files", "167 occurrences in 71 files", "330 occurrences", "41 occurrences", "13 occurrences", "110 SCSS files", "922 test cases across 95 files", "88 of 95 (93%)".

**`anti-patterns.md`:** "Found ~50 times".

**`performance.md`:** "5.9/10", "1 memo() usage", "13 occurrences across 7 files", "15 occurrences across 9 files", "130 inline arrow functions across 54 files", "15 inline style objects across 8 files", "createSelector in 4 files", "161 useSelector calls across 81 files", "48 react-hooks/exhaustive-deps suppressions", "3 contexts only", "4 index-as-key anti-patterns".

When the AI model consumes this skill, stale counts mislead it about the codebase's current state. The rules are the valuable part — consider replacing exact counts with qualitative descriptions ("Well-adopted", "Common anti-pattern", "Critical gap") or adding a `last audited: YYYY-MM-DD` note. This was [addressed in PR #810 by removing the counts](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4124772078).

docs/skills/freighter-best-practices/references/performance.md
```diff
@@ -0,0 +1,241 @@
...
+      <h1>{title}</h1>
+      <span>{formattedBalance}</span>
+    </div>
+  );
+};
+```
+
+**Never produce only selectors when a component is requested.**
+
+## React.memo -- CRITICAL GAP
```
CassioMG: **[Comment 2 — Suggestion] "CRITICAL GAP" framing makes aspirational rules collide with codebase reality**

This section says `## React.memo -- CRITICAL GAP` because there's currently only 1 `memo()` usage. But the rule below says `Components rendered in lists ... MUST use React.memo()`. The same pattern shows up below for selectors:

> Line 154: `RULE: All derived/computed state MUST use createSelector. Never compute inline in components.`

But only 4 files actually use `createSelector` today. An agent applying these rules defensively could try to wrap every list-rendered component in `React.memo()` or refactor every `useSelector` to `createSelector` — even on screens that aren't hot paths and don't need it.

This was the same issue I flagged in PR #810 ([Comment 1](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4151971144) and [Comment 16](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4174042204)), eventually softened in the mobile skill to "prefer for new code, don't refactor existing".

Suggested reframing:

> "**RULE: For new components in lists or hot paths, prefer `React.memo()`.** The codebase currently has only 1 `memo()` usage — this is a known gap, not the established pattern. Don't refactor existing components to add `memo()` unless you're already touching them or there's a measured perf problem."

docs/skills/freighter-best-practices/references/performance.md
```diff
@@ -0,0 +1,241 @@
...
+// CRITICAL ANTI-PATTERN — 130 inline arrows found in codebase
+// WRONG:
+<AssetList onClick={() => handleClick(asset)} />
+
+// CORRECT:
+const onClick = useCallback(() => handleClick(asset), [asset]);
+<AssetList onClick={onClick} />
+```
+
+**High-priority files needing useCallback refactoring:**
```
CassioMG: **[Comment 3 — Suggestion] "Files needing refactoring" lists are action-plan content, not skill rules**

This section calls out specific files (`AccountAssets/index.tsx — 10+ inline handlers`, `ManageAssetRows/ChangeTrustInternal/index.tsx — 10 inline handlers`, `AccountHeader/index.tsx — 10+ inline handlers`). The same pattern repeats:

- Line 213-219: `4 files use array index as key. Fix these:` followed by a list of 4 files
- Line 117-122: "High-priority files needing useCallback refactoring"

This is **work-to-do tracking**, not stable best-practices content. Once those files are fixed, the lists become outdated; once new violations appear, the lists drift the other way. PR #810 had the same Priority Actions issue that was eventually [moved out of the skill](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4124772078).

Suggest moving these to a GitHub issue ("Performance backlog: useCallback refactoring") and keeping the skill focused on the rules.

docs/skills/freighter-best-practices/references/performance.md
```diff
@@ -0,0 +1,241 @@
...
+state:
+
+- Use `useReducer` for local component state (loading spinners, form validation,
+  step tracking)
+- Use Redux dispatch for global state changes (account data, network state,
+  transaction results)
+
+## Performance Priority Actions
+
+| Priority | Action                                            | Impact                       |
```
CassioMG: **[Comment 4 — Suggestion] Move "Performance Priority Actions" table to a GitHub issue**

This P0/P1/P2 table is an action plan for fixing performance debt — not a reference rule for agents writing new code. As items are addressed, the table becomes stale; the impact estimates ("30-40% fewer re-renders") will likely never be verified.

Same pattern as PR #810 — that table was [removed from the skill](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4124772078) and the perf concerns were tracked in issues instead.

Suggest:
1. Open a "Performance backlog" GitHub issue and move this table there
2. Keep `performance.md` focused on timeless rules

AGENTS.md
```diff
@@ -0,0 +1,167 @@
...
+| Error Handling       | `docs/skills/freighter-best-practices/references/error-handling.md` | Adding error states, catch blocks, or user-facing errors |
+| Internationalization | `docs/skills/freighter-best-practices/references/i18n.md`           | Adding or modifying user-facing strings                  |
+| Messaging            | `docs/skills/freighter-best-practices/references/messaging.md`      | Adding popup/background/content script communication     |
+| Git & PR Workflow    | `docs/skills/freighter-best-practices/references/git-workflow.md`   | Branching, committing, opening PRs, CI                   |
+| Dependencies         | `docs/skills/freighter-best-practices/references/dependencies.md`   | Adding, updating, or auditing packages                   |
+| Anti-Patterns        | `docs/skills/freighter-best-practices/references/anti-patterns.md`  | Code review, avoiding common mistakes                    |
+| Troubleshooting      | `docs/docs/troubleshooting-guide.md`                                | Build failures, setup issues, known bugs and gotchas     |
```
CassioMG, line 167: **[Comment 5 — Suggestion] AGENTS.md vs SKILL.md table drift**

This row adds "Troubleshooting" to the AGENTS.md "Best Practices Entry Points" table (12 rows total), but the matching table in [`docs/skills/freighter-best-practices/SKILL.md`](docs/skills/freighter-best-practices/SKILL.md) (called "Reference Guide") has only 11 rows — no Troubleshooting entry.

The two tables duplicate each other and have already drifted on this row. Once new reference files are added, drift will continue.

Same pattern as PR #810 ([Comment 5](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4124888671)). The eventual fix on the mobile side was to **delete the duplicate table from AGENTS.md and replace it with a single link to SKILL.md**:

```markdown
## Best Practices

See [`docs/skills/freighter-best-practices/SKILL.md`](./docs/skills/freighter-best-practices/SKILL.md) for the full best-practices skill, including per-concern reference files.

For build failures, setup issues, and known bugs, see [`docs/docs/troubleshooting-guide.md`](./docs/docs/troubleshooting-guide.md).
```

This avoids the drift problem entirely. SKILL.md becomes the single source of truth for the per-concern table.

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,417 @@
+---
+id: troubleshooting-guide
+title: Troubleshooting Guide
+slug: /troubleshooting-guide
+---
+
+_Last updated: 2026-04-08_
```
CassioMG: **[Comment 6 — Nit] Update the "Last updated" date before merging**

This says `2026-04-08` but the PR has been actively iterated on since then. Worth bumping to today's date so the staleness signal is accurate. Same nit as the PR #810 troubleshooting guide ([Comment 23](https://github.com/stellar/freighter-mobile/pull/810#pullrequestreview-4199875389)).

CassioMG (commented):
Initial review. This PR mirrors the V1 state of [PR #810 (mobile)](https://github.com/stellar/freighter-mobile/pull/810), which went through ~18 rounds of review to address systemic issues. Many of the same patterns are here. The good news: most of the lessons from the mobile PR transfer directly, so this can hopefully iterate faster.

**Verified accurate** (no need to fix):
- `sendMessageToBackground` in `@shared/api/helpers/extensionMessaging.ts` ✓
- `SERVICE_TYPES` / `EXTERNAL_SERVICE_TYPES` enums in `@shared/constants/services.ts` ✓
- `npmMinimalAgeGate: 7d` in `.yarnrc.yml` ✓
- Pre-commit hook content matches doc ✓
- Playwright 1.57.0 matches the troubleshooting guide claim ✓
- `ROUTES` enum and `navigateTo()` helper exist as described ✓

The 6 inline comments below cover the systemic issues.

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,417 @@
+---
+id: troubleshooting-guide
+title: Troubleshooting Guide
+slug: /troubleshooting-guide
```
CassioMG: The Docusaurus frontmatter (`id`, `title`, `slug: /troubleshooting-guide`) means this page will be published at `https://docs.freighter.app/docs/troubleshooting-guide` and will appear in the sidebar of the public docs site.

The content is largely developer-internal — Husky/nvm setup, ESLint IDE plugin troubleshooting, Webpack memory tuning, the Playwright 1.57 memory bug, MV3 service-worker pitfalls. Worth confirming this is intended for end users browsing the wallet docs.

If not, consider moving this to a `CONTRIBUTING.md`-style doc that lives outside the Docusaurus content tree, or stripping the slug so it doesn't surface in the public sidebar.

docs/skills/freighter-best-practices/references/performance.md
```diff
@@ -0,0 +1,241 @@
...
+))}
+```
+
+## Selector Memoization -- createSelector in 4 files (GOOD)
+
+`createSelector` from reselect is used in `remoteConfig.ts`, `settings.ts`,
+`cache.ts`, `accountServices.ts`. 161 `useSelector` calls across 81 files.
+
+**RULE: All derived/computed state MUST use `createSelector`. Never compute
+inline in components.**
```
CassioMG: Suggest softening this. *"All derived/computed state MUST use `createSelector`. Never compute inline in components"* is too absolute — trivial computations like `const isMainnet = network === NETWORKS.PUBLIC` (which the doc itself uses elsewhere as a "NOT NEEDED — simple primitive" example for `useMemo`) shouldn't be wrapped in reselect.

The bolded MUST will be cited out of context by reviewers / Claude. Maybe rephrase as *"For non-trivial derived state, prefer `createSelector` over inline computation"* and let the existing `useMemo` guidance further down clarify when it's worth it.

AGENTS.md
```diff
@@ -0,0 +1,167 @@
...
+yarn test:e2e                      # Playwright e2e
+```
+
+## Environment Setup
+
+Create `extension/.env`:
+
+```env
+INDEXER_URL=https://freighter-backend-prd.stellar.org/api/v1
+INDEXER_V2_URL=https://freighter-backend-v2.stellar.org/api/v1
```
CassioMG: The production indexer URL shouldn't be committed verbatim here — the [README]( ../README.md) already uses a placeholder (`INDEXER_V2_URL=your_backend_v2_prod_url_here`) for exactly this reason.

Suggest matching that pattern:

```env
INDEXER_URL=your_backend_url_here
INDEXER_V2_URL=your_backend_v2_url_here
```

AGENTS.md
```diff
@@ -0,0 +1,167 @@
+# Freighter
+
+> Non-custodial Stellar wallet browser extension for Chrome, Firefox, and
+> Safari. Monorepo: extension, client SDK (`@stellar/freighter-api`), shared
```
CassioMG: Safari is mentioned here as a supported target, and again in [SKILL.md](docs/skills/freighter-best-practices/SKILL.md) and [git-workflow.md](docs/skills/freighter-best-practices/references/git-workflow.md).

`submitSafari.yml` does exist in `.github/workflows/`, but it isn't actively used — there's no live Safari distribution today. Suggest dropping the Safari references in the agent-facing docs so contributors / Claude don't treat it as a target that needs cross-browser verification on every change. The workflow can stay parked without being advertised.

CassioMG (changes requested):
Posted a few comments above based on the many iterations we did on the mobile PR to polish the skills. I think we are in good state on the mobile PR now so sounds like a good moment to port those common findings over - thanks!

docs/skills/freighter-best-practices/references/code-style.md
```diff
@@ -0,0 +1,488 @@
...
+### Action Type Strings
+
+The reducer action types (`"FETCH_DATA_START"`, `"FETCH_DATA_SUCCESS"`,
+`"FETCH_DATA_ERROR"`) are string literals in the shared reducer. This is
+acceptable for the shared `useReducer` pattern, but Redux slice action types
+should always use `createSlice` (which auto-generates them).
+
+## Function Style
+
+Arrow functions dominate (~85%). Function declarations are used rarely:
```
CassioMG: **[Comment 7 — Suggestion] Hard-coded statistics still present in `code-style.md` and `anti-patterns.md`**

Follow-up to [Comment 1 from my initial review](https://github.com/stellar/freighter/pull/2687#pullrequestreview-4207841285). `performance.md` got the great cleanup treatment in `13a83fbc` ("enhance best practices"), but the same pattern persists in this file and in `anti-patterns.md` — the counts will go stale immediately as the codebase evolves.

**Still in `code-style.md`** (~25 instances):
- Line 101: `all 37+ enums in the codebase`
- Line 125: `~4 remaining union types that should be enums`
- Line 130: `observed in ~200+ type definitions`
- Line 189: `Arrow functions dominate (~85%)` (anchor)
- Line 192/196: `(85% of codebase)` / `(15% of codebase)`
- Line 285: `dominant pattern (75% of components)`
- Line 288/291: `(75%)` / `(25%)`
- Line 307: `(~12% of destructuring operations)`
- Line 315: `&& (659 occurrences) more than ?. (405 occurrences)`
- Line 320–321: `(405 occurrences)` / `(659 occurrences)`
- Line 324: `?? underutilized (41 occurrences)`
- Line 337: `Named exports dominate (95-97%) ... rare (~17 occurrences)`
- Line 340/344: `(97%)` / `(3%)`
- Line 348: `~15-20 instances`
- Line 353: `Separate Props interface preferred (60%) over inline (40%)`
- Line 356/364: `(60%)` / `(40%)`
- Line 368: `only 17 occurrences remain`
- Line 392: `(~15-20 files)`
- Line 402: `1.3:1 ratio, 162 vs 124 occurrences`
- Line 404: `554 occurrences vs <2%`
- Plus several more ratios deeper in the file (`(>97%)`, `(<3%)`, `(60%)/(40%)`, `(50+ files)/(15 files)`, `30 files / 40 files / 30 files / 15 files / 167 occurrences in 71 files`, `~330 occurrences`, `41 occurrences`, `13 occurrences`, `110 SCSS files`, `922 test cases across 95 files`, `88 of 95 (93%)`)

**Still in `anti-patterns.md`:**
- Line 11: `Found ~50 times in the codebase`

Suggest applying the same approach used for `performance.md`:
- Replace exact counts with qualitative descriptions ("common", "dominant", "rare", "well-adopted", "few remaining")
- Drop counts that don't change the rule's meaning (e.g., percentages between two acceptable styles)
- Keep counts only where they're load-bearing and add a `last audited: YYYY-MM-DD` note

The rules themselves are the valuable part — the surrounding statistics will be wrong within a few sprints regardless of how careful the author is.

CassioMG (commented):
Follow-up on the hard-coded statistics finding from my [initial review](https://github.com/stellar/freighter/pull/2687#pullrequestreview-4207841285).

CassioMG:
@leofelix077 I think we should probably be good to go with this PR once the remaining 2 open comments are addressed ([here](https://github.com/stellar/freighter/pull/2687#discussion_r3170243367) and [here](https://github.com/stellar/freighter/pull/2687#discussion_r3185131968))

PR #2688, opened 2026-04-09
troubleshooting guide

leofelix077:


docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+`localhost:9000`.
+
+**Solution:** This is by design. Production builds have security guardrails that
+block dev server connections. Use `yarn build:extension` for development.
+
+## Testing Issues
+
+### Playwright tests fail on stale build
+
+**Symptom:** E2e tests fail with unexpected behavior, missing elements, or
```
piyalbasu:
```suggestion
**Symptom:** e2e tests fail with unexpected behavior, missing elements, or
```

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+yarn test:e2e --ui
+```
+
+1. Let **all** tests run to completion first
+2. Filter for failed tests only
+3. Re-run the failed tests 2-3 times
+4. Tests that fail consistently across multiple runs are real failures; tests
+   that pass on retry are flaky
+
+This avoids wasting time debugging timing-related flakes. If a test is
```
piyalbasu, line 202: I think we should move away from this advice and move towards using Playwright's playwright-test MCP server to solve flakey tests. The MCP will often put in a more scalable fix than just increasing timeouts, which is often a band aid

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+
+### Pre-commit hook adds unexpected locale changes
+
+**Symptom:** `git diff` shows changes to locale files you didn't touch.
+
+**Solution:** This is expected. The pre-commit hook auto-runs
+`yarn build:extension:translations` and stages locale files. These are generated
+— commit them as-is.
+
+## Playwright & CI Compatibility Issues
```
piyalbasu: I think let's group all the Playwright related stuff together

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+Changes to content script code require a full `yarn build:extension` and
+extension reload. The dev server only hot-reloads popup code.
+
+### Message passing type mismatches
+
+When adding new message types between popup/background/content script, update
+the typed message enums in all three contexts. A mismatch causes silent message
+drops.
+
+### Forgetting to test both Manifest V3 contexts
```
piyalbasu: I'm a little confused by what this title means. What are the 2 v3 contexts that this referring to? Is this the DOM and the service worker?

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+
+Background scripts run as service workers (Manifest V3), which means:
+
+- No persistent state between events (use `chrome.storage`)
```
piyalbasu: actually, one common pitfall we can document: we should always use `browser.storage` rather than th API `chrome` to maintain cross-browser compatibility

docs/docs/troubleshooting-guide.md
```diff
@@ -0,0 +1,470 @@
...
+behavior.
+
+**Root cause:** React 19 treats functions returned from ref callbacks as cleanup
+functions (similar to `useEffect`). Components that accidentally return values
+from ref callbacks will break.
+
+**Solution:** If a ref callback doesn't need cleanup, ensure it returns
+`undefined` (or nothing).
+
+## Known Issues & Recent Fixes
```
piyalbasu: I think this is useful today, but we should think about how we can keep this section updated

docs/docs/troubleshooting-guide.md
leofelix077: oh yes

docs/docs/troubleshooting-guide.md
leofelix077: it was regarding the chrome killing long running background tasks, but not really relevant day-to-day. I removed it

docs/docs/troubleshooting-guide.md
leofelix077: I removed this section as well to keep something more generic / only major issues and not require active maintenance . otherwise it's gonna end up being somewhat of a changelog

piyalbasu (approved):

PR #2720, opened 2026-04-23
Skip unfunded-destination warning for pure Soroban custom tokens

JakeUrban:
## Summary
- Pure Soroban custom tokens (asset ids with a C-address issuer) no longer surface the "Blockaid unfunded destination" warning when sending to an unfunded address. Their transfer is a contract invocation and doesn't require a funded classic destination account, so the "expected to fail" label was misleading.
- `getExpectedToFailReason` now checks whether the asset's issuer is a Soroban (C-address) issuer via the existing `isSorobanIssuer` helper and returns `null` in that case. Classic sends (native XLM, credit_alphanum) and SAC-wrapped classic assets keep the warning — SACs normalize to their underlying G-issuer and their `transfer` still fails to an unfunded G-account.
- Collectibles go through a separate simulation hook that never called this predicate, so they're naturally unaffected.
- Exported `getExpectedToFailReason` and added unit tests covering classic credit_alphanum4/12, SAC-wrapped classic, pure Soroban, and native-XLM reserve edge cases.

## Related
Parallel fix in Freighter mobile: stellar/freighter-mobile (branch `fix-send-warnings`).


piyalbasu:
### Code review

extension/src/popup/components/send/SendAmount/hooks/__tests__/useSimulateTxData.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        }),
+      ).toBe("Blockaid unfunded destination");
+    });
+  });
+
+  describe("destination is unfunded, SAC-wrapped classic assets", () => {
+    // SACs normalize to their underlying classic G-issuer, so the canonical
+    // is indistinguishable from a classic credit_alphanum send — which is
+    // correct: a SAC's `transfer` still fails to an unfunded G-account.
+    it("returns the unfunded warning for a SAC-wrapped USDC (classic G-issuer canonical)", () => {
```
piyalbasu: NIT: this is the same test as what's on line 35 and doesn't really provide much value. To actually test this flow of a selecting a SAC and then receiving the unfunded destination, you would likely need to either 

1) add a unit test for the sac normalization (with the subsequent error message being captured by the above test)
or
2) use an e2e test to capture both the SAC normalization and the subsequent error message in 1 test

piyalbasu (approved):
This bug fix itself is looking good. Just left 1 comment about improving the testing here

extension/src/popup/components/send/SendAmount/hooks/__tests__/useSimulateTxData.test.ts
JakeUrban: Actually there is no real scenario where the user selects a SAC -- Freighter references stellar assets using the G address format, so I think we can actually just remove this test.

PR #2744, opened 2026-04-30
federated addr. memo type inference

### LINKED ISSUE #2717: Memo type inference for federated address
Coming from analysis on https://github.com/stellar/freighter-mobile/pull/821, the memo and memo_type returned from federated server are discarded

specified on https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md

>   The extension has a latent bug: if an exchange federation address returns memo_type: "id" with a numeric memo, the
>    extension sends it as Memo.text() instead of Memo.id() — which many exchanges reject.

Note on Freighter extension
The browser extension appears to have the same bug. All three federation-resolution call sites only read fedResp.account_id and discard memo / memo_type:

[extension/src/popup/components/send/SendTo/hooks/useSendToData.tsx](https://github.com/stellar/freighter/blob/main/extension/src/popup/components/send/SendTo/hooks/useSendToData.tsx) (send flow)
extension/src/popup/helpers/account.ts
extension/src/popup/views/SignTransaction/index.tsx
Worth either fixing there in parallel or filing a GH issue on [stellar/freighter](https://github.com/stellar/freighter/issues) so both clients converge on correct SEP-0002 behavior.

leofelix077:
closes #2717 

This PR adds safer support for federation-provided memos in Send
It now recognizes the main SEP-0002 memo formats (text, id, hash) instead of treating everything the same way.
It improves validation/error handling when federation servers return invalid data (including invalid destination addresses).
It also fixes send-flow behavior so memo data is kept consistent when users change recipient input.
 \+ it adds test coverage for the new federation memo scenarios, including prefill, invalid responses, and recipient switching.

[federated-switch-address-clears-memo.webm](https://github.com/user-attachments/assets/b5554254-e056-4e3b-8675-c939ba3f3039)

[federated-hash-memo.webm](https://github.com/user-attachments/assets/ca7afd20-5dcd-4f60-b54b-8539e440549f)

[federated-id-memo.webm](https://github.com/user-attachments/assets/56d80c29-582d-48c9-9893-e55cb778c150)

[federated-text-memo.webm](https://github.com/user-attachments/assets/3b39940d-b529-4c60-a36e-94d8c295efad)

[federated-invalid-account-id-memo.webm](https://github.com/user-attachments/assets/da0c5f6b-a95e-4083-8101-4ece8741940d)


extension/src/popup/helpers/useValidateMemo.ts
```diff
@@ -1,33 +1,47 @@
 import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { Memo } from "stellar-sdk";
+import { FederationMemoType } from "popup/helpers/federationMemo";
 
 const MAX_MEMO_BYTES = 28;
+const MAX_UINT64 = BigInt("18446744073709551615");
+const HEX_32_BYTES_RE = /^[0-9a-fA-F]{64}$/;
```
piyalbasu: these var were set in `federationMemo.ts`. Let's reuse

piyalbasu:
### Code review

Found 2 issues:

1. State leak: when the user navigates back from `SendAmount` to `SendTo` and replaces a federation address (e.g. `name*exchange.com` resolving to `memo=12345`, `memo_type=id`) with a regular `G...` recipient, the else branch resets `memoType` to `""` but never clears `memo`. The previous federation memo silently rides on the new transaction with an empty type. Fix: also `dispatch(saveMemo(""))` in the else branch.

https://github.com/stellar/freighter/blob/647a0f819d8f9b902037c3b0e7a0257b2927b34d/extension/src/popup/components/send/SendTo/index.tsx#L125-L136

2. UX trap: `EditMemo`'s `onSubmit` dispatches `saveMemo` but never `saveMemoType`. Once a federation address has set `memoType` to `hash` or `id`, the user cannot clear or change the memo type from the edit dialog — they're locked into typing 64 hex chars or a uint64, with no way to revert to a freeform text memo without going back and changing the recipient. If federation-supplied types are intentionally enforced (per SEP-0002), consider rendering the field read-only or hiding the editor when a federation memo type is present; otherwise dispatch `saveMemoType("")` on submit so the user can escape the constraint.

https://github.com/stellar/freighter/blob/647a0f819d8f9b902037c3b0e7a0257b2927b34d/extension/src/popup/components/send/SendAmount/index.tsx#L759-L772

piyalbasu:
The Claude review had 2 good finds here:

For 1, it wasn't clear to me if `saveMemo` is being called somewhere else and we're trying to avoid overwriting it here. I think it's a bit of a footgun in that memo and memo type aren't being saved together

Similar concern on issue 2

leofelix077:
@piyalbasu yep. it's a good catch. I updated it to save memo and type together, and clearing memo when switching from federated address + added the e2e tests for it

piyalbasu (approved):

extension/src/popup/helpers/federationMemo.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    case FederationMemoType.Hash: {
+      if (!HEX_32_BYTES_RE.test(memo)) {
+        throw new Error(
+          "Federation memo hash must be a 64-character hex string (32 bytes)",
+        );
+      }
+      break;
+    }
+
```
CassioMG: SEP-0002 specifies hash memos as **base64-encoded**, but this validates them as hex (`/^[0-9a-fA-F]{64}$/`). A 32-byte hash in base64 is 44 chars, not 64 hex chars — so any spec-compliant federation server returning `memo_type: "hash"` will fail validation here and the user sees "Failed to resolve federated address".

Spec quote: *"for `hash` this should be base64-encoded"* — [SEP-0002](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md).

Mobile PR stellar/freighter-mobile#821 handles this with `Buffer.from(memo, "base64")` and asserts `length === 32` before calling `Memo.hash(buffer)`.

Two changes needed:
1. Validate as base64 → 32 bytes here (replace `HEX_32_BYTES_RE`).
2. Line 94 below — `Memo.hash(memo)` also needs the decoded buffer: `Memo.hash(Buffer.from(memo, "base64"))`. `Memo.hash` accepts a 64-char hex *string* or a 32-byte Buffer, so passing the raw base64 string silently produces an incorrect hash.

The same hex-only assumption is mirrored in `useValidateMemo.ts:23-30` (the EditMemo path), which would similarly reject a real hash memo if the user opens the editor on a federated address.

After fixing, the e2e test in `sendPayment.test.ts` that stubs `hashMemo = "93a4c1c4...64chars"` will need to use a base64 string instead.

CassioMG (approved):

PR #2746, opened 2026-05-01
fix: prevent SVG NFTs from collapsing to 0x0

AdamBuzzard:
## Description
Fixes an issue where valid SVG NFTs (e.g. SEP-50 standard tokens) with `width='100%'` and `height='100%'` collapse to a 0x0 height inside the `CollectibleInfo__image` container. By adding an explicit `aspect-ratio: 1 / 1;`, it ensures that the container maintains correct square proportions and standard SVGs render visibly.

## Changes
- Added `aspect-ratio: 1 / 1` and `width: 100%` to `.CollectibleInfo__image` in `styles.scss`.

### DIFF
--- a/extension/src/popup/components/account/CollectibleInfo/styles.scss
+++ b/extension/src/popup/components/account/CollectibleInfo/styles.scss
@@ -5,6 +5,8 @@
     border-radius: pxToRem(24px);
     margin-bottom: pxToRem(12px);
     overflow: hidden;
+    width: 100%;
+    aspect-ratio: 1 / 1;
 
     img {
       width: 100%;
--- a/extension/src/popup/components/account/CollectibleInfo/styles.scss
+++ b/extension/src/popup/components/account/CollectibleInfo/styles.scss
@@ -11,7 +11,7 @@
     img {
       width: 100%;
       height: 100%;
-      object-fit: cover;
+      object-fit: contain;
     }
 
     &--small {


piyalbasu:
### Code review

Found 1 issue:

1. The `object-fit: cover` → `object-fit: contain` change isn't needed to fix the SVG collapse and diverges from how freighter-mobile renders collectibles.

https://github.com/stellar/freighter/blob/045e086c8dd3153321210455334d9367cd888f10/extension/src/popup/components/account/CollectibleInfo/styles.scss#L11-L15

The SVG visibility fix is entirely the container additions (`width: 100%` + `aspect-ratio: 1 / 1`) — they give the percentage-sized SVG a concrete box to resolve `height: 100%` against. The `object-fit` switch is a separate change: it applies to the shared `.CollectibleInfo__image img` selector used by the grid/thumbnail views too, so non-square raster NFTs go from crop-fill to letterboxed (empty bars).

freighter-mobile solves the same problem the same way — definite container box + crop-fill — but keeps `cover`:
- Container gets an explicit square box: [`CollectibleDetailsScreen.tsx#L223-L226`](https://github.com/stellar/freighter-mobile/blob/main/src/components/screens/CollectibleDetailsScreen/CollectibleDetailsScreen.tsx#L223-L226) (`style={{ height: windowWidth - 2 * padding }}`)
- Fit mode stays `cover`, never overridden: [`CollectibleImage.tsx#L43`](https://github.com/stellar/freighter-mobile/blob/main/src/components/CollectibleImage.tsx#L43) (`resizeMode = "cover"`)

Suggest reverting line 14 back to `object-fit: cover` and keeping the `width`/`aspect-ratio` additions. That preserves the SVG fix while keeping raster NFTs crop-filled, consistent with mobile.

AdamBuzzard:
😅🙏🏼

PR #2764, opened 2026-05-11
Linearize Send Flow

### LINKED ISSUE #2686: Linearize Send flow from Home screen
Currently, the Send flow from the Home screen drop the user on the amount screen where they can choose token, amount, and recipient in any order. This can be confusing (e.g., entering an amount before realizing the wrong token is selected).

Proposed flow from Home screen: Send → token selection → amount → recipient → review. Each step gives the user one thing to do at a time.

When starting from the Token Details screen, keep the current behavior (navigate straight to Amount with the token pre-selected).

This should be coordinated with the [equivalent mobile change](https://github.com/stellar/freighter-mobile/issues/809) to maintain a consistent cross-platform experience.

Context: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1775602561479619

leofelix077:
Closes #2686 

This PR linearizes the extension Send flow so users move more consistently from recipient selection to amount entry to review, with cleaner state handoff between steps.
It improves recipient UX by keeping Recents and My Accounts visible while searching, making Suggestions pressable like other destination options, and debouncing validation/search feedback to avoid premature errors.
It also hardens amount handling by preserving the exact source value when switching between fiat and token views, improving numeric sanitization, and fixing edge cases like formatted thousands leading to invalid amount states.
Additionally, it includes reliability and polish updates across query-param handling, submit safeguards, collectible grouping/styling, locale cleanup, and Send flow test coverage.

https://github.com/user-attachments/assets/0885cd4b-2c9a-4836-b794-7ef1ca3eb54b

https://github.com/user-attachments/assets/ddf29001-91e5-4126-83b9-42e61e949c24

https://github.com/user-attachments/assets/ce6727f1-cc03-4365-a7f6-bde929a04666

https://github.com/user-attachments/assets/20d2a329-5b61-42e1-8141-e2927f976587

https://github.com/user-attachments/assets/68cf6d9b-5004-4bcd-abb8-e3e911bb7a93

https://github.com/user-attachments/assets/1cf62c7d-40c4-43a5-aff6-d4a610a01435

Adding back the memo and fee row that was missing from previous videos:

Screenshot 2026-05-14 at 15 28 48

Screenshot 2026-05-14 at 15 29 00


extension/src/popup/views/Send/index.tsx
```diff
@@ -112,9 +115,87 @@ export const Send = () => {
...
+    navigateTo(ROUTES.account, navigate);
+  };
+
+  const initialStepRef = useRef<STEPS>(
+    (() => {
+      const params = new URLSearchParams(location.search);
+      return params.has("asset") || params.has("collection_address")
+        ? STEPS.DESTINATION
+        : STEPS.SELECT_SOURCE_ASSET;
+    })(),
```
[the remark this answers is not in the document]
leofelix077: intended. following figma

piyalbasu:
### Code review

Found 1 issue:

1. The Soroban "Account doesn't exist" warning gate from master commit [`d8ee94c8`](https://github.com/stellar/freighter/commit/d8ee94c8) (`#2778`, landed 2026-05-15) is not applied in the refactored `SendTo`. Master rendered `<AccountDoesntExistWarning />` behind `shouldShowAccountDoesntExistWarning({ assetCanonical: asset, destination, isCollectible, isFunded })` (from `popup/helpers/sendWarnings.ts`) so that pure-Soroban custom tokens sent to a contract or unfunded destination wouldn't display the misleading "Send at least 1 XLM to create account" notice. This PR reverts to an unconditional `!isFunded` check, which after a rebase will re-introduce the bug `#2778` fixed.

   https://github.com/stellar/freighter/blob/d11ef2f0368d7aa67cc1b5bcd1b551a5b71cbd7a/extension/src/popup/components/send/SendTo/index.tsx#L307-L312

   Suggest pulling `shouldShowAccountDoesntExistWarning` into the new flow (and re-selecting `asset`/`isCollectible` from `transactionDataSelector`) to preserve the fix.

extension/e2e-tests/loadAccount.test.ts
```diff
@@ -642,6 +642,7 @@ test("Loads collectibles data with successful metadata", async ({
   extensionId,
   context,
 }) => {
+  test.slow();
```
piyalbasu: I notice we're using a lot of `test.slow()`. Why is this necessary? This isn't necessarily wrong, but it does triple the time a test has to run (15s vs 45s). If a test fails, you'll have to wait 45s for it terminate.

It's best to avoid this unless it's absolutely necessary to make sure our tests run and fail quickly

extension/e2e-tests/reviewTxFees.test.ts
```diff
@@ -550,22 +543,36 @@ test("Custom fee resets to default when re-entering send flow from home screen",
...
 
-  // resetSubmission cleared destination → no auto-simulation fires on mount.
-  // Fee display shows the base fee, NOT the previous override total (0.0093738).
+  // resetSubmission cleared fee state. Auto-simulation fires on destination set,
+  // using the reset base fee (0.00001) — NOT the previous override (0.00005).
+  // Total = 0.00001 + 0.0093238 = 0.0093338 (not the previous 0.0093738).
   await expect(page.getByTestId("send-amount-fee-display")).toHaveText(
-    "0.00001 XLM",
+    "0.0093338 XLM",
+    { timeout: 10000 },
```
piyalbasu: 10s is a pretty long timeout. Shouldn't this happen pretty quickly if the API calls are stubbed?

extension/e2e-tests/sendPayment.test.ts
```diff
@@ -18,14 +19,66 @@ const UNFUNDED_DESTINATION =
...
   page,
   extensionId,
 }) => {
+  test.slow();
   await login({ page, extensionId });
 
   await page.getByTestId("nav-link-swap").click();
-  await expect(page.getByTestId("AppHeaderPageTitle")).toContainText("Swap");
+  await expect(page.getByTestId("swap-src-asset-tile")).toBeVisible({
+    timeout: 15000,
```
piyalbasu: similar comment about the timeouts here

extension/src/popup/components/send/SendAmount/index.tsx
```diff
@@ -75,6 +71,52 @@ import { settingsNetworkDetailsSelector } from "popup/ducks/settings";
 import "../styles.scss";
 
 const DEFAULT_INPUT_WIDTH = 25;
+const PERCENTAGE_OPTIONS = [
+  ["25%", 25],
+  ["50%", 50],
+  ["75%", 75],
+] as const;
+
+const normalizeNumericString = (value: string) => {
```
piyalbasu: NIT: it might be nice to move this helper and the 2 below this to the `formatters` helpers file so all the formatting utils are in 1 place

piyalbasu:
Occasionally, I'm seeing a loader briefly under the destination address bar that pushes everything down. It's not clear what it's loading:


https://github.com/user-attachments/assets/06ef8a49-21b7-499d-922d-a473822cf537

piyalbasu:
The number input in the Swap amount input isn't quite displaying correctly

https://github.com/user-attachments/assets/6be46684-fd22-4470-9003-eeb9edefb152

piyalbasu:
If you click Swap from the Asset Detail, pressing back goes back to the Account screen

But, if you click Send from the Asset Detail, pressing back goes back to the Asset Detail screen

https://github.com/user-attachments/assets/af9dc79c-f2c8-44e2-bbda-919070e04bc0

CassioMG:
Found a few issues:

1 - Weird double AQUA balances when entering fiat amount
Screenshot 2026-05-15 at 14 51 32

2 - When you start the Send flow from the Token Details screen it'll pre-select the token and it won't let you change the token using the picker. The picker opens, you tap to select a different token but nothing happens.

extension/e2e-tests/loadAccount.test.ts
leofelix077: not needed. I re-ran all the tests without it and it works fine

extension/e2e-tests/reviewTxFees.test.ts
leofelix077: same for the others. re-ran the tests adjusting the timeouts and the .slow()

leofelix077:
@piyalbasu -> The swap flow is untouched (except for the issue with the input. there were some overlapping classes on Swap and Send. both were using the SendAmount classes. Adjusted now 

@CassioMG adjusted the token change when coming from details

and also the other minor parts added to the comments

also merged the last changes from main on both mobile and extension


https://github.com/user-attachments/assets/0cc2ae67-0367-4346-be6f-0b33cf5d1b7f


https://github.com/user-attachments/assets/988c44e3-6d4f-4b10-a87d-66b1c9ca2902

CassioMG:
@leofelix077 I'm still seeing the [double token balances issue](https://github.com/stellar/freighter/pull/2764#issuecomment-4464048818) when entering **fiat** amount, see video below.

I'll cross-post here my comment from [the mobile PR](https://github.com/stellar/freighter-mobile/pull/856#issuecomment-4537415396) which has the same issue so you know how to repro it:

What appears to happen here is that the "token available balance" value which usually appears aligned on the right side (right below the dropdown) is conflicting with the "token amount to send" value which usually appears aligned on the left side. This probably only happens for big string amounts. I'd suggest applying some UI constraints here to prevent this from happening, or keep things flexible, but ensure that:
- the "token available balance" always appear below the dropdown in all cases instead of jumping to the other side
- the "refresh-cw-03" icon always appear right next to the "amount to send" value instead of breaking the line which results in the icon being placed on the left side of the "token available balance"

https://github.com/user-attachments/assets/621e310b-ad29-4eb7-8ef6-d82fa61eba2c

extension/src/popup/constants/send-payment.ts
```diff
@@ -1,4 +1,5 @@
 export enum STEPS {
+  SELECT_SOURCE_ASSET = "select-source-asset",
```
CassioMG: nit: should we name it as `SET_SOURCE_ASSET` for consistency with `SET_DESTINATION_ASSET` below or do they represent distinct things?

leofelix077:
@CassioMG not sure I'm catching what you mean by double token balances

one of them is the conversion from USD to the token equivalent, and the second line is the total of the tokens

UI-wise I had already discussed with @sdfcharles for breaking it into two lines. one for the current and one for total spendable, just missing the refresh icon staying on the first line, instead of dropping down to the second line. He also gave the green light after I showed the mocks, but I think he didn't see this small tweak that it should stay on the first line as it converts back and forth to the typed amount equivalent

extension/src/popup/constants/send-payment.ts
leofelix077: it can be SET. it leads to the same action

CassioMG:
> @CassioMG not sure I'm catching what you mean by double token balances
> 
> one of them is the conversion from USD to the token equivalent, and the second line is the total of the tokens

Sorry I think the expression I used was indeed ambiguous, by "double token balances" I meant the 2 amounts stacked on top of each other as it appear on the video, which is basically just the result of the line break that happens when the 2 amounts get too big and don't fit in a single line. I think now that the [refresh icon position is fixed](https://github.com/stellar/freighter/pull/2764/commits/c84e2160569756b60dd05feb7976f5fea3d8b8da) on the last commit it is looking better. My initial thought was that the available token balance should always stay right below the token dropdown for consistency as jumping to the opposite side of the screen could be confusing for users. But I'm not super strong on it, I'll defer the UI treatment to @minkyeongshin. What do you think?

https://github.com/user-attachments/assets/ca234c88-1fef-4bb3-8f2d-0e7520d04f03

leofelix077:
@CassioMG Ahh I see. Now I get what you meant, hadn't fully understood it before. but not sure as well how we could handle it. will leave it like this for now until we get feedback

if we keep it fixed below the token dropdown, then the first line needs to drop below it. might also get somewhat confusing

Screenshot 2026-05-27 at 12 57 38
Screenshot 2026-05-27 at 12 58 25

Another option could be to display above
Screenshot 2026-05-27 at 13 01 12
Screenshot 2026-05-27 at 13 03 27



For displaying below, long texts we probably would need some different UI I think, breaking into two lines on opposite sides doesn't look good imo. and not sure if we should ellipsize the amount values to make it fit into one line 

or if we keep as is, as it's kind of an edge-case even for small token to USD conversion it needs a very long value

minkyeongshin:
Leo's suggestion could work. we can move the available balance above the dropdown, which gives more room to display the balance. (screen 1)

For cases where the typed amount gets long, we can reduce the text size up to a certain length to prevent it from being cut off. (screen 2) However, I think it would be better to set a maximum input length(screen 3). 

What do you think? @CassioMG @leofelix077 

Screenshot 2026-05-27 at 2 40 28 PM


[design file here](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=8214-16406&t=5QfvxsqMJim8SQpx-1)

leofelix077:
Posting for both here so it's easier to keep track, but imo it looks good / doesn't look out of place. @minkyeongshin @CassioMG 

Implemented with dynamic font size and in case it's still too big it ellipsizes, to avoid breaking the UI into two lines

what do you think?



Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 13 22 28

Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 13 22 48

Screenshot 2026-05-28 at 13 36 40

Screenshot 2026-05-28 at 13 37 24

Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 13 41 36


Also adjusted the logic to hide the fiat input when not supported (following extension logic)

Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 14 00 46

minkyeongshin:
Looks good to me! 👍👍 @leofelix077

CassioMG:
> Posting for both here so it's easier to keep track, but imo it looks good / doesn't look out of place. @minkyeongshin @CassioMG
> 
> Implemented with dynamic font size and in case it's still too big it ellipsizes, to avoid breaking the UI into two lines
> 
> what do you think?
> 

@leofelix077 looks good to me 👍 thanks for making those changes. I think we could do the same on mobile for consistency 📱

leofelix077:
@CassioMG I applied for both. I meant to keep the conversation only in one thread, but screenshots 1,2,5,6 are for mobile ✅ . will crosspost the thread there

extension/src/popup/views/__tests__/Send.test.tsx
```diff
@@ -10,6 +10,9 @@ import {
 import * as ApiInternal from "@shared/api/internal";
```
CassioMG, line 10: just checking, is this test file in the expected folder?

extension/src/popup/locales/en/translation.json
```diff
@@ -76,6 +76,7 @@
   "At the end of this process, Freighter will only display accounts related to the new backup phrase.": "At the end of this process, Freighter will only display accounts related to the new backup phrase.",
   "Authorizations": "Authorizations",
   "Authorize": "Authorize",
+  "available": "available",
```
CassioMG, line 79: it seems this string is not being used

extension/src/popup/helpers/formatters.ts
```diff
@@ -1,9 +1,51 @@
...
+
+  return numericValue.isNaN() ? null : numericValue;
+};
+
+export const isValidPositiveAmount = (value: string) => {
+  const numericValue = getValidBigNumber(value);
+
+  return Boolean(numericValue && numericValue.gt(0));
+};
+
```
CassioMG, line 48: do we have tests for those 3 new helpers?

CassioMG (approved):

PR #2778, opened 2026-05-12
Skip unfunded-destination warning for Soroban on Search address screen

### LINKED ISSUE #2777: Skip unfunded-destination warning for Soroban on Search address screen
This PR [Skip unfunded-destination warning for pure Soroban custom tokens](https://github.com/stellar/freighter/pull/2720) partially fixed the issue.

✅  It removed the warning only from the transaction Review step below:
Image

But, we also need to remove it from the Search address step which is still showing it:
Image

Freighter-mobile already has it fixed on both places, so we just need to make this final adjustment on extension.

Slack thread with more info: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1778550106064029?thread_ts=1778548460.141749&cid=C03347FNAHK

JakeUrban:
## Problem

PR #2720 removed the misleading "Blockaid unfunded destination" warning for
pure Soroban custom tokens on the **transaction Review** step. The same
warning still appears one step earlier on the **Search address** screen
(`extension/src/popup/components/send/SendTo`), which is what issue #2777
tracks. Freighter Mobile already addresses both screens.

## Root cause

Two surfaces independently decide whether to show the unfunded-destination
warning:

- **Search address** (`SendTo/index.tsx:282-285`) renders
  `<AccountDoesntExistWarning />` based purely on `!destinationBalances.isFunded`,
  with no asset-side check at all.
- **Review** (`SendAmount/hooks/useSimulateTxData.tsx:94-123` —
  `getExpectedToFailReason`) checks `assetCanonical !== "native"` and
  `isContractId(issuer)`, but doesn't consider `isCollectible` or
  contract destinations.

The exported helper `shouldAccountDoesntExistWarning(isFunded, assetID, amount)`
in `SendTo/index.tsx:51-58` is dead code (verified via repo-wide grep) and
encodes a third, incomplete version of the rule.

### 1. New shared helper

Create `extension/src/popup/helpers/sendWarnings.ts`:

```ts
import { isContractId } from "popup/helpers/soroban";

/**
 * Returns true when the "destination is unfunded" warning rule applies to
 * this send — i.e. the asset transfers via a classic Stellar payment op AND
 * the destination would be a classic G-account.
 *
 *  - Native XLM, credit_alphanum4/12, and SAC-wrapped classic assets all
 *    use a classic payment op that fails to an unfunded G-account (with
 *    the special case of native ≥ 1 XLM, which succeeds via create-account
 *    — that quantitative check is left to the caller; this helper only
 *    answers the qualitative "should the rule even fire?" question).
 *  - Pure Soroban custom tokens (issuer is a C-address) and collectibles
 *    transfer via contract invocation; the classic destination ledger is
 *    never touched.
 *  - Contract (C...) destinations have no classic account at all — their
 *    "balance" lives in the token contract's storage.
 */
export const shouldCheckUnfundedDestinationWarning = ({
  assetCanonical,
  destination,
  isCollectible,
}: {
  assetCanonical: string;
  destination: string;
  isCollectible: boolean;
}): boolean => {
  if (isCollectible) return false;
  if (destination && isContractId(destination)) return false;

  if (assetCanonical && assetCanonical !== "native") {
    const [, issuer] = assetCanonical.split(":");
    if (issuer && isContractId(issuer)) return false;
  }

  return true;
};
```

`isContractId` is the same predicate already used in
`getExpectedToFailReason`, so the two surfaces will agree by construction.

The qualitative/quantitative split (helper answers "does the rule apply?";
caller decides "is the destination actually unfunded enough to fail?") is
deliberate. The native ≥ 1 XLM create-account threshold lives where the
amount is finalized — `getExpectedToFailReason`'s native branch — and the
SendTo screen never needs it (asset is fixed, amount is not yet known).
This addresses the naming concern by making the helper's semantics
unambiguous: it is a precondition gate, not a "this send will fail"
predicate.

### 2. Use the helper in `SendTo/index.tsx`

- Pull `asset` and `isCollectible` from the existing `transactionDataSelector`
  (already imported and used for `destination`/`federationAddress`).
- Replace the inline `!isFunded` check with a single compound
  helper call (see Tests, below, for why the `=== false` and the
  qualitative gate are co-located):

```tsx
{shouldShowAccountDoesntExistWarning({
  assetCanonical: asset,
  destination: sendDataState.data.validatedAddress,
  isCollectible,
  isFunded: sendDataState.data.destinationBalances?.isFunded,
}) && <AccountDoesntExistWarning />}
```

`validatedAddress` is the right thing to pass: federation domains are
resolved to G-addresses there, and `useSendToData` only fetches
`destinationBalances` when the resolved address is non-contract anyway —
but passing it through the helper keeps the rule self-consistent at the
read site.

Note the change from `!destinationBalances.isFunded` to `destinationBalances?.isFunded === false`: `isFunded` is `boolean | null` (`@shared/api/types/backend-api.ts:10-13`), and Review-step logic only warns on `=== false` (`useSimulateTxData.tsx:105-123`). Aligning Search with that strict check means an unknown/null funding state (e.g. mid-fetch or fetch failure) no longer triggers a misleading warning. (Critic round 2.)

- **Delete** the dead `shouldAccountDoesntExistWarning` export and its base
  reserve constant. They are unused (grep confirms). Leaving a stale,
  near-correct helper next to the new one is a footgun for the next
  person tracing this logic. The min-1-XLM check it encoded for native
  sends already lives in `getExpectedToFailReason`'s native branch.

### 3. Refactor `getExpectedToFailReason` — in-place, no API surface change

Per critic feedback (round 1, item 1): do **not** thread new args through
`popup/views/Send/index.tsx`. The hook already has what it needs:

- `useSimulateTxData` already accepts `destination` as a hook arg
  (`useSimulateTxData.tsx:394-405`).
- `useSimulateTxData` already destructures from `transactionDataSelector`
  (`useSimulateTxData.tsx:410-412`); add `isCollectible` to that destructure.
- The single `getExpectedToFailReason(...)` call inside the hook
  (`useSimulateTxData.tsx:472-477`) gains two locally-scoped parameters
  (`destination`, `isCollectible`).
- `Send/index.tsx` is untouched; the hook signature is untouched.

Inside `getExpectedToFailReason`, replace the bespoke
`assetCanonical !== "native"` + issuer parsing with:

```ts
if (
  isDestinationFunded !== false ||
  !shouldCheckUnfundedDestinationWarning({
    assetCanonical,
    destination,
    isCollectible,
  })
) {
  return null;
}
// native min-1-XLM branch unchanged
```

### 4. Tests

**Unit tests** for `shouldCheckUnfundedDestinationWarning` covering:
- Native XLM → true
- credit_alphanum4 with G-issuer → true
- SAC-wrapped classic (G-issuer) → true
- Pure Soroban custom token (C-issuer) → false
- `isCollectible: true` (regardless of asset) → false
- Contract (C...) destination (regardless of asset) → false
- Empty `assetCanonical` (initial-state edge case before
  `useSendQueryParams` resolves) → defaults to "treat as classic" →
  document why this is safe (the warning would only render once
  destinationBalances loads, which gates on a real address, and the
  default redux value for `asset` is `"native"` per
  `transactionSubmission.ts:488`, so empty-string is not a real
  runtime path; covered as a defensive case).

**`SendTo` warning gate as a unit-testable helper.** The
`extension/src/popup/views/__tests__/SendPayment.test.tsx` integration
suite is currently `describe.skip("Send", ...)` (line 128, verified
round 3) — anything added there would not run. Rather than unskip an
unrelated suite or stand up an ad-hoc integration harness for one
boolean, extract the wrapping `=== false` + helper composition into a
second pure helper alongside `shouldCheckUnfundedDestinationWarning`
in `sendWarnings.ts`:

```ts
export const shouldShowAccountDoesntExistWarning = ({
  assetCanonical,
  destination,
  isCollectible,
  isFunded,
}: {
  assetCanonical: string;
  destination: string;
  isCollectible: boolean;
  isFunded: boolean | null | undefined;
}): boolean =>
  isFunded === false &&
  shouldCheckUnfundedDestinationWarning({
    assetCanonical,
    destination,
    isCollectible,
  });
```

The `SendTo` JSX then collapses to:

```tsx
{shouldShowAccountDoesntExistWarning({
  assetCanonical: asset,
  destination: sendDataState.data.validatedAddress,
  isCollectible,
  isFunded: sendDataState.data.destinationBalances?.isFunded,
}) && <AccountDoesntExistWarning />}
```

This puts the strict-`=== false` rule in the same module as the
qualitative gate (one place to update) and makes the integration cases
pure unit tests in `extension/src/popup/helpers/__tests__/sendWarnings.test.ts`:

- Default state (`asset: "native"`, `isCollectible: false`,
  `isFunded: false`) + non-contract G-destination → true.
- Pure Soroban asset (`asset: "TOKEN:CABC...XYZ"`,
  `isCollectible: false`, `isFunded: false`) + G-destination → false.
- Collectible state (`isCollectible: true`, `asset: "native"` —
  reproduces the early-return query-param hydration in
  `useSendQueryParams.ts:64-85` where `saveIsCollectible(true)` runs
  but the asset destructure path is short-circuited) +
  `isFunded: false` → false.
- Contract destination (`destination: "C..."`) regardless of asset +
  `isFunded: false` → false. (Also documents that
  `useSendToData.tsx:92-93` already skips balance fetch for contract
  destinations, so this is belt-and-braces.)
- Unknown funding state (`isFunded: null`, e.g. mid-fetch or backend
  fallback) with classic asset + non-contract destination → false
  (regression guard for the strict-`=== false` change, per critic
  round 2).
- `isFunded: undefined` (no balances loaded yet) with classic asset →
  false (defensive, matches the optional-chaining at the call site).
- `isFunded: true` with classic asset → false (sanity).

**Update** `useSimulateTxData.test.ts`'s existing
`getExpectedToFailReason` cases to pass the new `destination` /
`isCollectible` params, and add a collectible case (collectible flag
true with classic asset / unfunded G-destination → returns `null`).

### 5. Out of scope (file as follow-ups if confirmed)

- Equivalent screens for Swap/Soroswap, account-merge, and the QR-scan
  recipient flow — none called out in #2777, each touches separate UI.
- Aligning extension's helper with mobile's `buildUnfundedContext` shape
  (which feeds Blockaid result merging on mobile). Extension's Review
  step uses a simpler "expected to fail reason" string; cross-platform
  convergence is a separate refactor.
- Promoting the helper into `@shared/helpers/` for cross-submodule
  reuse. Mobile's helper is shaped differently (returns
  `UnfundedDestinationContext`); a shared module would have to bridge
  both shapes, which is out of scope for a one-screen bug fix.

## Verification

Per `freighter/AGENTS.md` (also stored in agent memory): from repo root
run `yarn test:ci` and `yarn build:extension`. Manually exercise the
SendTo screen in the extension dev build with (a) a classic asset →
unfunded G-address (warning shown) and (b) a pure Soroban custom token →
unfunded G-address (warning hidden).

---

Closes #2777

### DIFF
--- a/extension/src/popup/components/send/SendAmount/hooks/__tests__/useSimulateTxData.test.ts
+++ b/extension/src/popup/components/send/SendAmount/hooks/__tests__/useSimulateTxData.test.ts
@@ -1,5 +1,6 @@
 import { getExpectedToFailReason } from "../useSimulateTxData";
 
+const G_DEST = "GA4UFF2WJM7KHHG4R5D5D2MZQ6FWMDOSVITVF7C5OLD5NFP6RBBW2FGV";
 const t = (key: string) => key;
 
 describe("getExpectedToFailReason", () => {
@@ -11,6 +12,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
@@ -23,6 +26,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
@@ -37,6 +42,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBe("Blockaid unfunded destination");
@@ -49,6 +56,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "LONGCODE12:GDMTVHLWJTHSUDMZVVMXXH6VJHA2ZV3HNG5LYNAZ6RTWB7GISM6PGTUV",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBe("Blockaid unfunded destination");
@@ -66,6 +75,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "PBT:CAZXRTOKNUQ2JQQF3NCRU7GYMDJNZ2NMQN6IGN4FCT5DWPODMPVEXSND",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
@@ -78,6 +89,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "PBT:CAZXRTOKNUQ2JQQF3NCRU7GYMDJNZ2NMQN6IGN4FCT5DWPODMPVEXSND",
           amount: "0",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
@@ -93,6 +106,8 @@ describe("getExpectedToFailReason", () => {
           assetCanonical:
             "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef:lp",
           amount: "10",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBe("Blockaid unfunded destination");
@@ -106,6 +121,8 @@ describe("getExpectedToFailReason", () => {
           isDestinationFunded: false,
           assetCanonical: "native",
           amount: "0.5",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBe("Blockaid unfunded destination native");
@@ -117,6 +134,8 @@ describe("getExpectedToFailReason", () => {
           isDestinationFunded: false,
           assetCanonical: "native",
           amount: "1",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
@@ -128,6 +147,8 @@ describe("getExpectedToFailReason", () => {
           isDestinationFunded: false,
           assetCanonical: "native",
           amount: "5",
+          destination: G_DEST,
+          isCollectible: false,
           t,
         }),
       ).toBeNull();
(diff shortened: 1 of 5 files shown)


leofelix077 (approved):
Verified behavior, unit tests and other E2E tests for regression. All look fine

There's one E2E test that is failing, but also failing on master, due to wrong mocks (Send token payment to C address), but works fine when doing manually. Checking it separately

https://github.com/user-attachments/assets/5cc6fce5-d849-4e05-b5d4-baff4ce937e1

https://github.com/user-attachments/assets/89c6f7b8-b2d3-4970-a895-38aac8125e9e

https://github.com/user-attachments/assets/29493b84-eadc-49cd-902b-2f45fe0f800d

PR #2810, opened 2026-05-25
Order transactions by date

leofelix077:
Quick fix PR to get all fetched transactions and order them by their created at date, instead of depending on the returned groupings of the backend, which can group failed together, and successful together, and order them by date internally, instead of ordering the entire result/outcome by date

Screenshot 2026-05-25 at 11 39 32
Screenshot 2026-05-25 at 11 46 58


piyalbasu (commented):
Three observations before merge — none blocking, the fix is fine to ship.

**1. The sort is load-bearing for the section reducer in `useGetHistoryData.tsx`.** That reducer creates a new `monthYear` bucket each time the month changes between adjacent items, so it requires sorted input — without this sort, history would render with duplicate non-contiguous month headers (e.g., two separate 'March 2025' sections). Worth a one-line comment on the sort here noting that downstream grouping depends on it, otherwise a future refactor that drops the sort produces a silent visual bug with no type/test failure.

**2. Worth a follow-up at the backend.** With Mercury disabled in prod, `freighter-backend/src/helper/horizon-rpc.ts` `fetchAccountHistory` is the only path serving real users — a thin passthrough of one Horizon call (`order("desc").includeFailed(true)`). If Horizon really is returning grouped data, a one-line sort there gives every consumer (extension, mobile, any future v2 reimplementation) the invariant without each client re-implementing it. The client sort here is still useful as defense-in-depth (and the section reducer needs it locally), but the canonical fix probably belongs upstream.

**3. Inline notes** on the rationale comment and the sort itself below.

extension/src/helpers/hooks/useGetHistory.tsx
```diff
@@ -27,10 +27,16 @@ function useGetHistory() {
     try {
       const cachedHistoryData =
         cachedHistory[networkDetails.network]?.[publicKey];
-      const data =
+      const fetched =
         useCache && cachedHistoryData
           ? cachedHistoryData
           : await getAccountHistory(publicKey, networkDetails);
+      // Backends can return failed and successful operations in separate
+      // groupings, so sort by created_at desc to guarantee chronological order.
```
piyalbasu: This comment plants a server-side cause that hasn't been reproduced against the actual prod path. With Mercury disabled in prod, `/account-history` is a thin Horizon passthrough (`order("desc").includeFailed(true)`) — Horizon orders operations by id desc, which is chronological including failed. So either Horizon itself is returning grouped data (a quirk of `includeFailed`+`join("transactions")` worth verifying with a real curl), or the reshuffle is happening somewhere we haven't found yet.

Two options:

- Soften to a defensive rationale that doesn't make an unverified claim: `// Defensive: guarantee desc order regardless of upstream behavior. The section reducer in useGetHistoryData.tsx requires sorted input.`
- Or do a 10-minute repro against a reproducing account and document what you actually saw — that's load-bearing context for whoever picks this up next.

As written, a future maintainer reads this, assumes the backend is at fault, and never investigates further — while freighter-mobile (or any future client) still sees the bug.

extension/src/helpers/hooks/useGetHistory.tsx
```diff
@@ -27,10 +27,16 @@ function useGetHistory() {
...
+      const data = fetched
+        .map((op) => ({ op, ts: Date.parse(op.created_at) }))
+        .sort((a, b) => b.ts - a.ts)
+        .map(({ op }) => op);
```
piyalbasu: Two small things on the sort itself:

**Schwartzian is overkill here.** `N` is capped at `TRANSACTIONS_LIMIT=100` and `Date.parse` on ISO strings is ~microseconds — the decorate-sort-undecorate idiom isn't paying for itself, and readers have to mentally unwind it to confirm what it does. The plain form is shorter and equivalent in practical cost:

```ts
const data = [...fetched].sort(
  (a, b) => Date.parse(b.created_at) - Date.parse(a.created_at),
);
```

**`Date.parse(undefined)` returns `NaN`**, and a NaN-returning comparator yields implementation-defined non-stable order. Types say `created_at` is required `string`, so this is defensive — but a single malformed/missing record silently corrupts the entire sort. Cheapest guard inline: `Date.parse(x.created_at) || 0` (or `|| -Infinity` to bucket bad records at the bottom).

extension/src/helpers/hooks/useGetHistory.tsx
piyalbasu: @leofelix077 Wanted to follow up on the above Claude-generated comment:

Looks like you have a test account that you were able to replicate this. Would you mind quickly using Claude to check if the reason this is coming back in the wrong order is because of something `freighter-backend` is doing or because stellar-sdk/Horizon is returning these in the wrong order. Ideally, the BE would return things in the right order. Maybe there's some learnings that we can apply to the freighter-backend-v2 version of this.

Either way, I think it's fine to apply this client-side fix for now

piyalbasu (approved):

extension/src/helpers/hooks/useGetHistory.tsx
leofelix077: This is actually coming from the e2e test, not my own account to reproduce it

Checking against the backend it confirms your suspicions regarding Mercury, so I updated the comment not to mention backend but just force the sorting, as suggsted above

PR #2812, opened 2026-05-26
Truncate long memos dynamically

### LINKED ISSUE #2615: Fix memo display for long memos
In the send flow, when a user has a long memo, it should be truncated once it reaches a certain limit to prevent it from crashing into the label.

Example of the current UI:

Image

leofelix077:
Closes #2615 

Quick fix to truncate long memos as needed, depending on the window size

No tests added due to e2e infrastructure being fullscreen mode, not window mode, so it will not cover the important scenarios in which the memos might need to be truncated, as it only happens on the small viewport

https://github.com/user-attachments/assets/da4d9604-523f-411d-82b3-b57519cc6f8a

https://github.com/user-attachments/assets/24ce639f-5883-4cb3-8653-ad7d5d804b56

https://github.com/user-attachments/assets/9d8f79e8-46a8-494e-ab66-9f70476ee53a


extension/src/popup/views/SignTransaction/Preview/Summary/styles.scss
```diff
@@ -25,7 +25,14 @@
     &__value {
       color: var(
         --sds-clr-gray-12
-      ) !important; //sds rule does not allwo changing colors on p tags
+      ) !important; //sds rule does not allow changing colors on p tags
```
piyalbasu: 🤖 _From Claude:_ Stale comment. The PR converts the memo `__value` from `<p>` to `<span>` (`Summary/index.tsx:58`), so this comment is now half-true — the `!important` is still needed for the sibling sequence-number `<p className="TxInfoBlock__value">` at line 52, but the comment misleads readers about what `__value` actually is. Since this very line is being retouched for the `allwo`→`allow` typo, easy to clarify in the same pass (e.g. `// sds rule does not allow changing colors on the sibling <p> tags`).

extension/src/popup/components/TruncatedMemo/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  fallback?: string;
+  inline?: boolean;
+  /**
+   * Optional character limit. When set, memos longer than `maxChars` are
+   * pre-truncated with an ellipsis before being handed to CSS. Useful in
+   * tight popup layouts where the value cell shares its row with a fixed
+   * suffix (e.g. a memo-type tag) and CSS-only ellipsis would still let
+   * the value push past its share.
+   */
+  maxChars?: number;
```
piyalbasu: 🤖 _From Claude:_ `maxChars` is dead API — none of the four production callsites (ReviewTransaction, TransactionDetail, Summary, SignTransaction) pass it; only the unit tests exercise it. The JSDoc justifies it for cases like the Summary memo-type tag, but Summary actually solves that with `flex` + `max-width: 60%` on the parent, not via `maxChars`. Suggest deleting `maxChars`, the `ellipsize` helper, the `as number` cast, and the three dedicated tests until a real caller needs it.

extension/src/popup/components/TruncatedMemo/index.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+
+import "./styles.scss";
+
+interface TruncatedMemoProps {
+  memo?: string | null;
+  fallback?: string;
```
piyalbasu: 🤖 _From Claude:_ `fallback` is also unused at every production callsite — they all rely on the default `t("None")`. Drop the prop and its default-resolution branch; if a future caller ever needs a different fallback, they can render it themselves and pass an empty memo. Shrinks the API from 7 props to 5.

extension/src/popup/components/accountHistory/TransactionDetail/styles.scss
```diff
@@ -286,7 +286,35 @@
...
+        text-align: right;
+      }
+
       &__value {
+        &--memo {
+          min-width: 0;
+          flex: 1 1 0;
+          max-width: 60%;
+          overflow: hidden;
+          text-align: right;
```
piyalbasu, line 310: 🤖 _From Claude:_ Redundant `text-align: right` and `overflow: hidden`. The parent `__value--memo` here (310–315) sets both, the child `__memo` (301–307) sets `text-align: right` again, and `.TruncatedMemo` itself already owns `overflow: hidden`. Same duplication in `SignTransaction/styles.scss` (`__Value--memo` 284–289 + `__Memo` 275–281). Three locations encode the same rules — a future tweak has to touch each to stay in sync. Suggest letting `.TruncatedMemo` own the truncation properties and the per-row class own only layout (`width: 100%`, alignment).

piyalbasu (commented):
_Code-review pass by Claude (run by @piyalbasu). 4 cleanup nits — no correctness bugs found. Feel free to bundle or punt to follow-up._

extension/src/popup/components/TruncatedMemo/index.tsx
piyalbasu: I tend to agree that until this is needed, it's not worth adding

extension/src/popup/components/TruncatedMemo/index.tsx
piyalbasu: Same here - no need to over-optimize for scenarios that we're not dealing with yet

piyalbasu (approved):

PR #2830, opened 2026-06-05
Sort token pickers by USD value via TokenList

### LINKED ISSUE #2820: Apply value sort to Send flow
Follow-up feedback on #2227 (sort balances by USD value) from Emir Ayral:

~1. XLM is currently sorted by USD amount like any other token. For a Stellar wallet, XLM is the native asset and should always be the first token in the list regardless of value.~
2. The Send flow's token picker is still sorted alphabetically — it didn't get the value-sort change applied in #2227.

~- Token / balance list: sort by USD value, **but** always pin XLM to the top.~
- Send flow token picker: same ordering as the balance list (sort by USD value, XLM pinned to top).

- ~Keep XLM in value-sorted position (current behavior) — rejected by feedback.~
- Sort Send picker independently — inconsistent UX.

### Context
Slack thread: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1779982000509779

JakeUrban:
### Problem

The Account view sorts token balances by descending USD value (via `sortBalancesByValue` in `popup/helpers/balance.ts`). The Send flow's token picker (`SendDestinationAsset`) and the Swap flow's token picker (`SwapAsset`) both render tokens in `sortBalances` order — XLM first, LP shares last, everything else in API insertion order. This is not value-sorted and is inconsistent with the Account view.

### Solution

Centralize the value-sort inside `TokenList` itself. `TokenList` is exclusively a picker component — it is only used by `SendDestinationAsset` (`popup/components/send/SendDestinationAsset/index.tsx`) and `SwapAsset` (`popup/components/swap/SwapAsset/index.tsx`), never by the Account view (which has its own rendering path via `AccountAssets` + `useStableSortedBalances`). Since `TokenList` already receives `tokenPrices` as a prop and is the single shared picker renderer, it is the natural place to own the picker ordering policy.

**Why TokenList and not a shared hook?** Putting the sort in a hook or helper that callers must remember to invoke is the same pattern that produced this bug — the sort existed (`sortBalancesByValue`), but nobody called it in the picker paths. Placing it inside `TokenList` makes correct ordering structural: every picker gets it automatically, and no future caller can accidentally skip it. The Account view is unaffected because it uses a completely separate rendering path (`AccountAssets` component with `useStableSortedBalances`).

### Changes

**File: `extension/src/popup/components/InternalTransaction/TokenList/index.tsx`**

1. Import `sortBalancesByValue` from `popup/helpers/balance`.
2. At the top of the component body, sort the incoming `tokens` before rendering:
   ```tsx
   const sortedTokens = sortBalancesByValue(tokens, tokenPrices);
   ```
3. Replace `tokens` with `sortedTokens` in the render body (the `.filter().filter().map()` chain).

This is a single change point — both callers automatically get value-sorted rendering with zero modifications.

### Testing

**New tests for `TokenList`:**
- Tokens render in descending USD value order when `tokenPrices` has entries.
- Original order is preserved when `tokenPrices` is empty `{}` (no-op behavior).
- Hidden assets and LP shares remain excluded regardless of sort order.

Note: `sortBalancesByValue` already has direct unit tests in `popup/helpers/__tests__/balance.test.js` covering its sorting logic, edge cases, and tie-breaking. No additional helper-level tests are needed.

### Edge Cases

- **Swap search filtering:** In `SwapAsset`, filtering only applies for search terms > 2 characters (`useSwapFromData.tsx:107-127`). When no search is active, the full balance list reaches `TokenList`; when search is active, a filtered subset reaches it. In both cases, `sortBalancesByValue` applied inside `TokenList` produces the correct descending-value order for whatever subset it receives.
- **Empty prices:** `sortBalancesByValue` is a no-op when prices is empty or null — tokens keep their existing `sortBalances` order (XLM first, LP last).
- **Zero price:** Assets with price `"0"` are treated as priced with value 0 (sorting them below assets with positive value but above unpriced assets). This matches the existing `sortBalancesByValue` semantics.

### What this does NOT change

- XLM pinning is not applied (per updated issue description).
- The stable-sort behavior on the Account view is unchanged (different rendering path entirely).
- The baseline `sortBalances` in `useGetBalances` remains (provides tie-breaking insertion order for assets without price data).

---

Closes #2820

### DIFF
--- a/extension/src/popup/components/InternalTransaction/TokenList/__tests__/TokenList.test.tsx
+++ b/extension/src/popup/components/InternalTransaction/TokenList/__tests__/TokenList.test.tsx
@@ -1,0 +1,153 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import BigNumber from "bignumber.js";
+
+import { defaultBlockaidScanAssetResult } from "@shared/helpers/stellar";
+
+import { TokenList } from "../index";
+
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+jest.mock("popup/components/account/AccountAssets", () => ({
+  AssetIcon: () => <div data-testid="asset-icon" />,
+}));
+
+const USDC_ISSUER = "GCK3D3V2XNLLKRFGFFFDEJXA4O2J4X36HET2FE446AV3M4U7DPHO3PEM";
+const AQUA_ISSUER = "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5";
+
+const USDC_CANONICAL = `USDC:${USDC_ISSUER}`;
+const AQUA_CANONICAL = `AQUA:${AQUA_ISSUER}`;
+
+const XLM_BALANCE = {
+  token: { code: "XLM", type: "native" as const },
+  total: new BigNumber("500"),
+  available: new BigNumber("500"),
+  blockaidData: defaultBlockaidScanAssetResult,
+};
+
+const USDC_BALANCE = {
+  token: {
+    code: "USDC",
+    issuer: { key: USDC_ISSUER },
+  },
+  total: new BigNumber("200"),
+  available: new BigNumber("200"),
+  blockaidData: defaultBlockaidScanAssetResult,
+};
+
+const AQUA_BALANCE = {
+  token: {
+    code: "AQUA",
+    issuer: { key: AQUA_ISSUER },
+  },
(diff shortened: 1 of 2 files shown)


CassioMG (approved):

CassioMG:
The fix work, but there is room for improvement like memoization


Memoization: sortBalancesByValue runs on every render, and the Swap picker re-renders on each (debounced) keystroke. List sizes are small so this is negligible, but a useMemo(() => sortBalancesByValue(tokens, tokenPrices), [tokens, tokenPrices]) would make intent clear and avoid resorting on unrelated re-renders.

PR #2872, opened 2026-06-25
Swap to New Token

CassioMG:
## Summary

Enables swapping from a held token to **any Stellar classic asset** in a single flow — token discovery (held + popular + free-text search), Blockaid coverage on non-held destinations, and an atomic `changeTrust + pathPaymentStrictSend` when a trustline is needed.

This is the browser-extension counterpart to the mobile feature. See the reference PR and its design doc:

- 📱 **Mobile PR:** [stellar/freighter-mobile#879](https://github.com/stellar/freighter-mobile/pull/879)
- 📄 **Design doc:** [`extension/specs/swap-to-new-token-design.md`](../blob/feature/swap-to-new-token/extension/specs/swap-to-new-token-design.md) — kept in sync with what shipped; mirrors the mobile design doc.

**Scope:** 189 files changed, ~16.3k LOC across the branch:

- `src/` (+ `@shared`) — ~7.7k lines (47%)
- tests (unit + Playwright E2E) — ~7.7k lines (47%)
- design doc — ~730 lines (4%)
- i18n (en + pt) — ~130 lines (1%)

Most of the large diffs are net-new files (new picker sections, new hooks, new sheets) plus tests written alongside them, and shared UI components that Send and the account-home balances list now reuse.

---

## What's new

- **`useSwapTokenLookup`** — owns destination-token discovery: held balances, popular tokens (volume7d ∩ verified lists), and free-text search via stellar.expert. Closes a security gap: every non-held destination is Blockaid-scanned before it can be selected.
- **`SwapAsset` picker, parameterized by `selectionType`** — one picker drives both the Sell and Receive sides, replacing the old single-list picker.
- **`SwapPickerSections`** — the labeled sections (Your tokens / Popular / Verified / Unverified), the `(i)` info sheets, and the empty / Soroban / fallback states.
- **`SwapAmount` rewrite** — built on the new shared `AmountCard` + `PercentageButtons`, with a Sell/Receive layout, a swap-direction chevron, crypto⇄fiat toggle, and a **debounced live "You receive" quote** that updates as you type.
- **Atomic transaction** — `buildChangeTrustOperation` is extracted and prepended so a new-token swap is a single `changeTrust + pathPaymentStrictSend` XDR (fee split total-across-ops); Blockaid scans the combined XDR for defence in depth.
- **`XlmReserveSheet`** — a pre-flight when the account lacks XLM headroom for the new 0.5 XLM trustline reserve; offers a "Swap for 0.5 XLM" affordance, a copy-address fallback, and a "why do I need XLM?" link.
- **Trustline banner + info sheet + rate row** on the review screen — explains the trustline reserve up-front before the user confirms.
- **Transaction details sheet (Swap + Send)** — the shared review screen gains an expandable, operation-by-operation breakdown (amounts, network fee, memo, per-op fields), now surfaced in both the Swap and Send review flows.
- **Unified Blockaid coverage** — a single reusable `BlockaidBanner` across Add-a-token, Sign Transaction, Sign Message, ChangeTrust, Grant Access, and Swap; verdict-colored token badges (amber = suspicious, red = malicious); a scan-on-select recovery path so a token picked before its verdict lands still gets assessed.
- **Quote-expiry recovery** — Horizon `op_under_dest_min` / `op_too_few_offers` at submit are classified and surfaced as a toast, and the flow recovers to the amount screen with a fresh quote instead of dead-ending.
- **Shared UI components** — `AmountCard`, `PercentageButtons`, `AssetListRow`, and `BalanceRow` are now shared by Send, Swap, and the account-home balances list, removing duplicated markup.

---

## Search results strategy

When the user types a search term, the picker renders **three labeled, mutually-exclusive sections** — chosen over a flat "Results" list so users can visually distinguish levels of trust:

1. **Your tokens** — held tokens matching the term.
2. **Verified** — verified-list tokens. The section header has an `(i)` icon opening a sheet explaining what the verified-token lists are.
3. **Unverified** — remaining stellar.expert matches. Its header has its own `(i)` icon for the unverified caveat.

`G…` issuer addresses surface the assets that account issues via stellar.expert's fuzzy `?search=`. `C…` contract pastes fall under a Soroban-specific empty state explaining the flow is classic-only for now — Soroban contract tokens are filtered out at every stage.

---

## Caching & pre-warm

Popular ("top tokens") lists are cached in **two layers**, keyed per network:

1. **Persistent disk cache** — `chrome.storage.local` (background-owned; the popup messages the background, matching every other cache in the codebase), with a **30-minute staleness window**. Because it survives popup close, reopening the extension paints Popular straight from disk with no spinner.
2. **In-memory Redux `cache` slice** — the session working copy, read first on each open.

On open the picker reads Redux → disk → network in freshness order, using whichever layer is still fresh. Search/discover results repaint instantly from the last cached result and revalidate silently in the background (SWR-style), so re-typing a term never flashes a spinner.

**Pre-warm.** `useSwapTopTokensPrewarm` mounts on the account/home screen and — deferred ~1s past first paint, mainnet-only, aborting on unmount — populates both cache layers when the persisted entry is stale or absent. So by the time the user opens Swap the Popular list is usually already warm (the analog of mobile's tab-mount pre-warm). It costs at most one trending request per staleness window.

**Blockaid asset scans are deliberately _not_ persisted.** Unlike Popular tokens, per-token Blockaid verdicts stay in an in-memory (module-scoped) cache and are re-scanned each session. Persisting verdicts to disk would carry a cache-poisoning risk — a tampered entry could mark a malicious token "safe" — so we keep them in memory only.

**Defence in depth at review.** The scan caches are a first-paint UX optimization. The review step still issues a fresh `scanTransaction` against the built XDR every time, so transaction-level signals are always evaluated live per confirmation.

---

## Graceful fallback (stellar.expert unreachable)

Held-to-held swaps keep working even when stellar.expert is down:

- **Popular tokens section** — omitted.
- **Search** — degrades to held-only in-memory matches.
- **Soft inline notice** — "Token discovery is temporarily unavailable" is rendered at the top of the picker; non-blocking.

Path-finding is unaffected, since it uses Horizon strict-send paths, not stellar.expert.

---

## Notes for reviewers

- **State** lives in the Redux `transactionSubmission` slice (`destinationTokenDetails`, best-path, cache slots), not a separate store.
- **Overlays** use the existing `SlideupModal` / `LoadingBackground` pattern rather than native bottom sheets; toasts use `sonner`.
- **`SwapAmount` business logic** is split into focused hooks + helpers (`useSwapLiveQuote`, `useSwapQuoteExpiry`, `useSwapDestinationScan`, `getSwapDerivedData`, …), mirroring the mobile `SwapScreen` decomposition, so the component stays presentation-focused.

---

## Telemetry

Eight new events instrument the flow (screen-view events like `swap: amount` already existed on master):

| Event | Fires when | Key properties |
|---|---|---|
| `swap: picker opened` | The Sell/Receive picker is opened | `side` (source/destination), `source` (cta/dropdown) |
| `swap: source selected` | A sell token is picked | `tokenCode`, `tokenIssuer`, `source` |
| `swap: destination selected` | A receive token is picked | `tokenCode`, `tokenIssuer`, `requiresTrustline`, `source` |
| `swap: direction toggled` | The user flips sell ⇄ receive | — |
| `swap: xlm reserve shown` | The XLM-reserve pre-flight sheet is shown | — |
| `swap: quote expired` | A quote expires at submit (Horizon `op_under_dest_min` / `op_too_few_offers`) | `sourceToken`, `destToken`, `sourceAmount`, `destAmount`, `allowedSlippage`, `resultCode` |
| `swap: trustline added` | A new-token swap settles on-chain | `tokenCode`, `tokenIssuer` |
| `swap: success` | The swap settles on-chain | `sourceToken`, `destToken`, `sourceAmount`, `destAmount`, `allowedSlippage` |

`swap: trustline added` and `swap: success` fire **post-confirmation** — once the transaction actually settles, not at review time — mirroring mobile.

---

## Shared Blockaid banner

Now we have a single reusable `BlockaidBanner` component that renders in every flow that surfaces a security verdict — SAFE renders nothing, **suspicious = amber**, **malicious = red** — replacing the old per-flow warning components. The screenshot below covers all of them: **Grant access / connect, Add a token, Sign message, Sign transaction, Send review, and Swap review**.

Screenshot 2026-07-01 at 14 20 41

## Transaction details sheet

The shared transaction-details sheet (operation-by-operation breakdown), is now available in both the **Swap** and **Send** review flows and aligned with mobile UI.

Screenshot 2026-07-01 at 14 29 23

---

## Testing

Manually exercised in Chrome on the popup and sidebar. Full unit suite passing locally; the Playwright E2E swap spec was updated for the new flow (12/12), and the broader Blockaid / add-token / send E2E specs were updated to the new shared components and pass.

#### Swapping to a new token

https://github.com/user-attachments/assets/9b47bad6-06c0-416f-9ce4-8c170b6c9bf8

#### Swapping to a held token

https://github.com/user-attachments/assets/56f5f7e2-3b2e-462e-b807-7b8ab4586201

#### Swapping with not enough XLM for reserves

https://github.com/user-attachments/assets/a4b9b671-e127-4c22-b38a-269279af7bbf

#### Using the search input

https://github.com/user-attachments/assets/a770a9c5-e202-4543-be78-56fdbc4d42d8

#### Quote expiration warning + retry

https://github.com/user-attachments/assets/59110c41-1b61-4440-831d-cae68abc1d49

#### Swapping to/from a token with no price

https://github.com/user-attachments/assets/4455f9d2-8f88-4653-9aa7-4d95221e7aac

#### Blockaid warning

https://github.com/user-attachments/assets/99320634-d93d-4f44-a254-d96051dce477

#### Graceful degradation when stellar.expert is down

https://github.com/user-attachments/assets/b8afd58d-2a34-41d7-94c8-b32bbe7b6664

#### Amount and path errors

https://github.com/user-attachments/assets/865ef59a-1d81-4635-a7dd-aed5d88fb0ed


JakeUrban (commented):
Some things I found during testing:

1. When switching to/from token units vs. dollars, the amounts change.

https://github.com/user-attachments/assets/0727249e-c046-4baf-9b0c-3614232820a4

2. The button that switches the tokens being sold & bought is a down arrow. Should it be a cycle? Or include both up and down arrows?

3. When I opened the token picker for the first time, for both the buy and sell tokens, there was a decently long loading state before rendering. Can we start loading this data ahead of time, so users don't have to wait to see the tokens they can pick?

4. The buy amount unexpectedly updates after I select "Review Swap".

5. When I flip the buy and sell tokens, the buy token and both amounts are cleared

https://github.com/user-attachments/assets/78221321-cdca-43c8-a738-7da2ec6681e3

CassioMG:
@JakeUrban thanks for the quick review!

> 1. When switching to/from token units vs. dollars, the amounts change.

Nice catch, I'll take a look.

> 2. The button that switches the tokens being sold & bought is a down arrow. Should it be a cycle? Or include both up and down arrows?

I'm fine either way, I'll defer it to @minkyeongshin but just adding more context on the current behavior:
- If both sides are held tokens then both sides flip
- If just the sell side is a held token and the receive side is a new (non-held) token, then the token on the sell side moves to the receive side and the sell side becomes blank since you can't have a non-held token on the sell side

> 3. When I opened the token picker for the first time, for both the buy and sell tokens, there was a decently long loading state before rendering. Can we start loading this data ahead of time, so users don't have to wait to see the tokens they can pick?

We are already loading data ahead of time from the moment users land on Home screen (more precisely 1 second after users land on Home screen so we give some breath for other home-related stuff to load). I did notice though that the app randomly takes longer to release the spinner in random places (e.g. when starting any of the main flows we have: Send, Swap, History). Would you mind deleting the extension and trying again to see if it's consistent for you? Below is how it's usually behaving for me on first install:

https://github.com/user-attachments/assets/dbe691fb-0015-4136-8f9a-bd0d3fbc8d0c

> 4. The buy amount unexpectedly updates after I select "Review Swap".

I haven't seen it, I'll try to repro here. I'm assuming that by "buy amount" you mean the receive side, let me know if that's not it.

> 5. When I flip the buy and sell tokens, the buy token and both amounts are cleared

This is implemented like that, every time you change the source (sell) token the amount gets cleared to avoid having "Amount too high errors" when you change from a token with high balance to a token with small balance. I think it also helped with other small async issues (from when I was working with mobile) which I don't quite remember now but an alternative to avoid having the "Amount too high errors" would be capping the amount to the max available balance for the next token (which may or may-not result in a small "UI glitch" while the amount jumps from the old value to the capped value). Tagging @minkyeongshin in case she has some inputs here as well

minkyeongshin:
> 2. The button that switches the tokens being sold & bought is a down arrow. Should it be a cycle? Or include both up and down arrows?

I think we can keep the current down arrow icon. Users are already familiar with tapping that button to swap the two tokens, and the primary action here is swapping from token to another. Keeping the icon simple also helps reduce visual noise. 

> 5. When I flip the buy and sell tokens, the buy token and both amounts are cleared

I also think clearing the amounts when flipping tokens is the better. From a UX perspective, asking users to explicitly enter the amount again is more predictable and avoids unintended behavior.
@CassioMG @JakeUrban

@shared/api/types/message-request.ts
```diff
@@ -325,6 +325,18 @@ export interface CacheAssetIconMessage extends BaseMessage {
...
+export interface GetCachedSwapTopTokensMessage extends BaseMessage {
+  type: SERVICE_TYPES.GET_CACHED_SWAP_TOP_TOKENS;
+  network: string;
+}
+
+export interface CacheSwapTopTokensMessage extends BaseMessage {
+  type: SERVICE_TYPES.CACHE_SWAP_TOP_TOKENS;
+  network: string;
+  // Opaque to the background — the popup owns the trending-asset schema.
+  tokens: unknown[];
```
piyalbasu: is this really unknown? Is there any way to narrow this type down?

extension/e2e-tests/blockaidScan.suspicious.test.ts
```diff
@@ -253,33 +253,23 @@ test.describe("BlockAid Scan - Suspicious States", () => {
...
-    await page.getByTestId("swap-dst-asset-tile").click({ force: true });
-    await expect(page.getByText("Swap to")).toBeVisible();
-    await expect(page.getByText(/USDC/)).toBeVisible();
-    await page.getByTestId("USDC-balance").click();
-
-    await expect(page.getByTestId("AppHeaderPageTitle")).toContainText("Swap");
-    await expect(page.getByTestId("send-amount-amount-input")).toBeVisible({
-      timeout: 10000,
+    await expect(page.getByTestId("swap-sell-card")).toBeVisible({
+      timeout: 15000,
```
piyalbasu: Just curious - did you need a really long timeout here for these tests to pass? Just wondering if this is a code smell

extension/webpack/amplitude-autocapture-stub.js
```diff
@@ -46,6 +46,9 @@ export const autocapturePlugin = noopPlugin(
 export const frustrationPlugin = noopPlugin(
   "@amplitude/plugin-frustration-browser",
 );
+export const performancePlugin = noopPlugin(
+  "@amplitude/plugin-autocapture-browser/performance",
```
piyalbasu, line 50: why is this needed?

extension/src/background/messageListener/handlers/cacheSwapTopTokens.ts
```diff
@@ -0,0 +1,18 @@
...
+  request: CacheSwapTopTokensMessage;
+  localStore: DataStorageAccess;
+}) => {
+  const cache = (await localStore.getItem(CACHED_SWAP_TOP_TOKENS_ID)) || {};
+  cache[request.network] = {
+    tokens: request.tokens,
+    updatedAt: Date.now(),
+  };
+  await localStore.setItem(CACHED_SWAP_TOP_TOKENS_ID, cache);
+};
```
piyalbasu, line 21: ### `cacheSwapTopTokens` handler returns no response object

**Low severity — not a merge blocker** (currently masked; see below), but a genuine defect worth a one-line fix before merge.

**TL;DR:** The background handler that persists the swap "top tokens" cache doesn't return a response, so the code that calls it throws an error every time it runs. It's invisible today only because the cache write finishes first and the caller silently swallows the error — but the handler's error-handling path is effectively dead and this is inconsistent with every other handler of its kind.

**Steps to reproduce:**
1. Open the swap flow so the popular/top-tokens list is fetched and cached.
2. Observe that persisting the cache rejects with a `TypeError` (caught and swallowed by the caller). The disk write itself still succeeds.

---

<details>

**Root cause:** the handler performs its `setItem` and then returns `undefined` (no `return {}`):

https://github.com/stellar/freighter/blob/70709a390dc66f99db8ede9a6e03ad70539ee778/extension/src/background/messageListener/handlers/cacheSwapTopTokens.ts#L5-L18

The dispatcher returns the handler's value verbatim...

https://github.com/stellar/freighter/blob/70709a390dc66f99db8ede9a6e03ad70539ee778/extension/src/background/messageListener/popupMessageListener.ts#L620-L622

...so the popup-side caller destructures `error` from `undefined`, which throws `TypeError: Cannot destructure property 'error' of 'undefined'`:

https://github.com/stellar/freighter/blob/70709a390dc66f99db8ede9a6e03ad70539ee778/@shared/api/internal.ts#L2564-L2578

**Why it's currently masked (and why it's still worth fixing):**
- `localStore.setItem(...)` is `await`ed *before* the return, so the disk write completes successfully — the persisted cache actually works.
- The only caller, `setPersistedPopularTokens`, wraps the call in a swallowing `try/catch`, so the `TypeError` never surfaces.

Net effect: the handler's entire error-propagation contract (`if (error) throw new Error(error)`) is dead code, a `TypeError` is thrown-and-swallowed on every cache write, and the code is one reorder away from a live break. Every sibling void handler already returns `{}` (e.g. `MARK_QUEUE_ACTIVE`, `OPEN_SIDEBAR`).

**Suggested fix:**
Return an (optionally-typed) response object from the handler:
```ts
export const cacheSwapTopTokens = async ({
  request,
  localStore,
}: {
  request: CacheSwapTopTokensMessage;
  localStore: DataStorageAccess;
}) => {
  const cache = (await localStore.getItem(CACHED_SWAP_TOP_TOKENS_ID)) || {};
  cache[request.network] = { tokens: request.tokens, updatedAt: Date.now() };
  await localStore.setItem(CACHED_SWAP_TOP_TOKENS_ID, cache);
  return {}; // match every other void handler; keeps the caller's `{ error }` destructure safe
};
```

</details>

piyalbasu:
Code-wise, this is looking pretty good! I just left a couple of small comments plus a comment from Claude. I'm still doing some additional manual testing

Something I'm noticing - it seems like there's a few different refactors that are being folded into this PR:

1. Blockaid UX
2. SEP-42 validation
3. Webpack updates
4. Updating `inputWidthContext`
plus some others

I appreciate that this work is getting done, but I'm wondering if some of this could have been separate PR's to make this diff a little smaller and the review process a little easier

extension/webpack/amplitude-autocapture-stub.js
CassioMG: This is actually related to an issue we currently have on `master`, after you build you get this:
```
[build:extension    ] WARNING in ../node_modules/@amplitude/analytics-browser/lib/esm/browser-client.js 305:54-71
[build:extension    ] export 'performancePlugin' (imported as 'performancePlugin') was not found in '@amplitude/plugin-autocapture-browser' (possible exports: DataExtractor, autocapturePlugin, default, enableVisualTagging, frustrationPlugin, plugin)
[build:extension    ]  @ ../node_modules/@amplitude/analytics-browser/lib/esm/index.js 5:0-52 5:0-52
[build:extension    ]  @ ./src/helpers/metrics.ts 1:0-58
[build:extension    ]  @ ./src/popup/App.tsx 5:0-67 33:76-93 35:0-13
[build:extension    ]  @ ./src/popup/index.tsx 3:0-28 8:32-35
[build:extension    ] 
[build:extension    ] webpack compiled in 19358 ms (affe348a794346a21ef1)
```

And then on `http://localhost:9000/` you get this:
Screenshot 2026-07-02 at 11 26 18

I should have pushed it in a separate PR but ended up forgetting about it 😅

@shared/api/types/message-request.ts
CassioMG: good catch, updated here: https://github.com/stellar/freighter/pull/2872/changes/625f02686f6e0f030f4eefe8e879173644d0314e

extension/src/background/messageListener/handlers/cacheSwapTopTokens.ts
CassioMG: nice hardening, applied here: https://github.com/stellar/freighter/pull/2872/changes/07b7cff1b99023f8e1bf70d14ecfab74a5c38a97

extension/src/popup/components/amount/AmountCard/index.tsx
```diff
@@ -0,0 +1,321 @@
...
+            {amount || "0"}
+          </span>
+          <span
+            ref={fiatSpanRef}
+            className={fontClass}
+            style={{
+              position: "absolute",
+              visibility: "hidden",
+              whiteSpace: "pre",
+            }}
```
piyalbasu: we should apply these styles with a class

extension/src/popup/components/InternalTransaction/ReviewTransaction/components/SwapRateRow.tsx
```diff
@@ -0,0 +1,53 @@
...
+  sendAmount: string;
+  destinationAmount: string;
+}
+
+// The rate value can be long ("1 yXLMUSD ≈ 0.00000012 yBTCETH") while the
+// "Rate" label is short, so the row gives the label only the width it needs
+// and lets the value fill the rest, stepping the value's font-size down by
+// length so it isn't cropped. Mirrors the AVAILABLE_BALANCE_FONT_SIZES scale
+// used on the swap amount screen.
+const RATE_VALUE_FONT_SIZES = [
```
piyalbasu, line 19: ### Consolidate the duplicated "shrink font as the string grows" scales

**Cleanup — not a merge blocker.** No behavior change; the values work as-is.

**TL;DR:** This is the third hand-rolled "step the font size down as the text gets longer" lookup table in the codebase, and the base version of it is copy-pasted identically between the Send and Swap amount screens. Since this PR is already consolidating the amount UI into shared components, these little scales could ride along into one shared helper instead of being duplicated a third time.

---

<details>

**What's duplicated:** there are now three copies of the same `{ maxLen, sizePx }[]` + `.find(len <= maxLen).sizePx` idiom:

`RATE_VALUE_FONT_SIZES` (this row) — a 4-step variant, used only here:

https://github.com/stellar/freighter/blob/6e764125d79f0efe0f88b9976e0a0efe0f02f79a/extension/src/popup/components/InternalTransaction/ReviewTransaction/components/SwapRateRow.tsx#L19-L27

`AVAILABLE_BALANCE_FONT_SIZES` in the Send flow:

https://github.com/stellar/freighter/blob/6e764125d79f0efe0f88b9976e0a0efe0f02f79a/extension/src/popup/components/send/SendAmount/index.tsx#L78-L82

`AVAILABLE_BALANCE_FONT_SIZES` in the Swap flow — **byte-for-byte identical** to the Send one (same thresholds, same sizes):

https://github.com/stellar/freighter/blob/6e764125d79f0efe0f88b9976e0a0efe0f02f79a/extension/src/popup/components/swap/SwapAmount/helpers/swapAmountDisplay.ts#L3-L7

So: `RATE_VALUE_FONT_SIZES` itself is **not reused** anywhere (only the one `.find()` in `SwapRateRow`), and the sibling `AVAILABLE_BALANCE_FONT_SIZES` it "mirrors" is itself a straight copy across Send and Swap.

**Suggested fix (increasing depth):**
1. *Minimal:* hoist the single shared `AVAILABLE_BALANCE_FONT_SIZES` constant to one module (e.g. next to the extracted `AmountCard`) and import it in both Send and Swap, removing the duplicate.
2. *Root cause:* add one tiny helper — `fitFontSizePx(value: string, scale: { maxLen: number; sizePx: number }[]): number` — and express all three cases as data (the two balance scales collapse to one; `RATE_VALUE_FONT_SIZES` stays its own scale but shares the lookup logic). Removes the repeated `.find(...)!.sizePx` idiom.

Low priority; fine as a follow-up if you'd rather not grow this PR.

</details>

extension/src/popup/components/swap/SwapAmount/helpers/getSwapDerivedData.ts
```diff
@@ -0,0 +1,245 @@
...
+  const supportsUsd = isMainnet(data.networkDetails) && assetPrice;
+  const dstPriceValueUsd = dstAssetPrice
+    ? formatAmount(
+        roundUsdValue(
+          new BigNumber(dstAssetPrice)
+            .multipliedBy(new BigNumber(cleanAmount(destinationAmount || "0")))
+            .toString(),
+        ),
+      )
+    : null;
```
piyalbasu, line 113: NIT: calculating `priceValueUsd` and `dstPriceValueUsd` follows a pretty similar pattern. This could be a helper

extension/e2e-tests/blockaidScan.suspicious.test.ts
CassioMG: it's actually not needed, updated here: https://github.com/stellar/freighter/pull/2872/changes/6e764125d79f0efe0f88b9976e0a0efe0f02f79a

CassioMG:
> Code-wise, this is looking pretty good! I just left a couple of small comments plus a comment from Claude. I'm still doing some additional manual testing
> 
> Something I'm noticing - it seems like there's a few different refactors that are being folded into this PR:
> 
> 1. Blockaid UX
> 2. SEP-42 validation
> 3. Webpack updates
> 4. Updating `inputWidthContext`
>    plus some others
> 
> I appreciate that this work is getting done, but I'm wondering if some of this could have been separate PR's to make this diff a little smaller and the review process a little easier

@piyalbasu thanks for reviewing! Yeah, you are right that those could be separate PRs, I will keep that in mind for a next big feature like this so we ship it in smaller pieces 🚢

extension/src/popup/components/swap/SwapAsset/hooks/useSwapTokenLookup.ts
```diff
@@ -0,0 +1,831 @@
...
+import { AppDispatch, store } from "popup/App";
+import {
+  fetchTrendingAssets,
+  TrendingAsset,
+} from "popup/helpers/trendingAssets";
+
+// Re-export RequestState for consumers
+export { RequestState };
+
+const MAX_ASSETS_TO_SCAN = 10;
```
piyalbasu: ### `MAX_ASSETS_TO_SCAN = 10` is duplicated across the swap and add-token flows

**Cleanup — not a merge blocker.** No behavior change; both copies hold the same value.

**TL;DR:** The batch size for Blockaid asset scans is defined separately in the swap token lookup and the add-token lookup, with the same value. They could share one constant so the two flows can't silently drift apart. (For clarity to future readers: this is a scan *batch size*, not a cap on how many tokens get searched or scanned — both flows scan the full result set in chunks.)

---

<details>

Two independent definitions of the same constant, both `= 10`:

This PR's swap lookup (module scope):

https://github.com/stellar/freighter/blob/6e764125d79f0efe0f88b9976e0a0efe0f02f79a/extension/src/popup/components/swap/SwapAsset/hooks/useSwapTokenLookup.ts#L51

Pre-existing add-token lookup (function scope):

https://github.com/stellar/freighter/blob/6e764125d79f0efe0f88b9976e0a0efe0f02f79a/extension/src/popup/components/manageAssets/SearchAsset/hooks/useAssetLookup.ts#L75

Both feed the `/scan-asset-bulk` chunking loop and neither caps the number of assets scanned — swap chunks `scanCandidates` (all classic results across popular/verified/unverified) into groups of 10 run in parallel; add-token chunks all `assetRows` into groups of 10 run sequentially. Same intent, same batch size, two sources of truth.

**Suggested fix:** hoist a single `MAX_ASSETS_TO_SCAN` (next to `scanAssetBulk` / the Blockaid helpers) and import it in both hooks, so a future tuning change to the batch size applies everywhere. Purely a consistency/DRY improvement — fine as a follow-up.

</details>

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -256,64 +340,138 @@ export const SwapAmount = ({
...
+    // The receive side becomes XLM — a held token, so no trustline is needed.
+    dispatch(saveDestinationAsset("native"));
+    dispatch(saveDestinationTokenDetails(null));
+
+    if (!sourceIsNonXlmClassic) {
+      // Switching the sell side to a different token: reset the amount, since
+      // any prior amount was denominated in the now-replaced source.
+      dispatch(saveAsset(sellCanonical));
+      dispatch(saveAmount("0"));
+      dispatch(saveAmountUsd("0.00"));
```
piyalbasu: the string's `0` and `0.00` are repeated a few times in this file. Could be saved as const's

piyalbasu:
1 small thing I noted, the asset list schema is being loaded 3 times when I click an asset to `Swap To` (assuming this is loading once for every asset list). Not a huge deal because the browser is smart enough to load it from disk cache rather than doing a roundtrip, but something worth noting:
Screenshot 2026-07-02 at 12 40 05 PM

extension/src/popup/components/amount/AmountCard/index.tsx
CassioMG: https://github.com/stellar/freighter/pull/2872/changes/bbc81698d75f205406483bd1a9f70190613c09ee

extension/src/popup/components/InternalTransaction/ReviewTransaction/components/SwapRateRow.tsx
CassioMG: https://github.com/stellar/freighter/pull/2872/changes/19421361d44e1a5be55d6479b23682e1b636d595

extension/src/popup/components/swap/SwapAmount/helpers/getSwapDerivedData.ts
CassioMG: https://github.com/stellar/freighter/pull/2872/changes/d1e21ab3e6e77a846f36724314d6b3a86be6a435

extension/src/popup/components/swap/SwapAsset/hooks/useSwapTokenLookup.ts
CassioMG: https://github.com/stellar/freighter/pull/2872/changes/acb199d05f247ba48084e22756f268d80c1fc409

extension/src/popup/components/swap/SwapAmount/index.tsx
CassioMG: https://github.com/stellar/freighter/pull/2872/changes/8911a5fed47a19729473d401edf517fa1746a4ad

CassioMG:
> 1 small thing I noted, the asset list schema is being loaded 3 times when I click an asset to `Swap To` (assuming this is loading once for every asset list). Not a huge deal because the browser is smart enough to load it from disk cache rather than doing a roundtrip, but something worth noting: Screenshot 2026-07-02 at 12 40 05 PM

@piyalbasu good catch, and exactly right — `schemaValidatedAssetList` runs once per enabled asset list, and each call re-fetched the same SEP-0042 schema (3 lists → 3 requests, disk-cached by the browser but still issued every time).

Fixed in https://github.com/stellar/freighter/pull/2872/changes/c40fd69afe3ce629a6ec4ca81af703cba979631b: the schema is now fetched once and memoized, so all list validations in a session reuse it — a single request instead of one per list (and it's shared across the Swap-to / search / Add-a-token flows too, not just deduped per open). A failed fetch isn't cached, so a transient error still retries on the next attempt.

CassioMG:
> 1. When switching to/from token units vs. dollars, the amounts change.
> 4. The buy amount unexpectedly updates after I select "Review Swap".

> 3. When I opened the token picker for the first time, for both the buy and sell tokens, there was a decently long loading state before rendering. Can we start loading this data ahead of time, so users don't have to wait to see the tokens they can pick?

@JakeUrban the points 1 & 4 above have been fixed on both Swap and Send flows, you should see stable crypto and fiat amounts now.

About the point 3, as I mentioned earlier we are already loading data ahead of time which makes it very hard to repro the long loading time on first Swap picker open. I had to delete and reinstall the extension several times to be able to repro it once, and I also had to immediately open the picker right after opening the pop-up to be able to repro it. Then on the rare times I was able to repro it it happened just on the very first rendering, after that the result is cached and you should always see the popular tokens rendering immediately on following renders.

So from my testing it appears that it is very unlikely that a regular user will see it unless they open the Swap picker blazing fast right after updating the app version for the first time, and, in case they see it, they will only see it once because the next time there should be a cache in place already.

I'm thinking we could release it like this and revisit it in case we start seeing long loading times happening more often. How does that sound?

Thanks!

PR #2873, opened 2026-06-25
Fix wrong asset on "Sent!" screen after switching token mid-send

### LINKED ISSUE #2871: Wrong asset shown on "sent success" screen

Prod Version 5.42.1

### What did you do?

Sent AQUA and saw USDC on the "sent success" screen.

Repro steps:
1 - Open USDC token details
2 - Tap Send
3 - Switch to any other token and send it
4 - "Sent success" screen will display the original USDC token instead of the token you just sent

This bug happens on 100% of times.

"Sent success" screen showing the actual token I just sent.

### What did you see instead?

"Sent success" screen showing a token I didn't send.

https://github.com/user-attachments/assets/6f440e4a-98f8-4e1e-923d-61d36bb0e209

CassioMG:
Fixes #2871.

Opening Send from a token's detail page keeps `?asset=<token>` in the URL for the whole (linearized) flow. [`useSendQueryParams`](https://github.com/stellar/freighter/blob/master/extension/src/popup/views/Send/hooks/useSendQueryParams.ts) re-applied that param on every effect run, and its deps include the collections cache — which refreshes on a successful submit. So switching to another token mid-flow was reverted to the URL token right at the "Sending…" → "Sent!" transition, and the success screen (which reads `transactionData.asset`) showed the wrong token. 100% reproducible.

The `currentAssetRef` guard added in #2764 only covered the no-param / invalid-param branches — not the valid-param branch that does the reverting. This gates the destination/asset URL pre-population on an actual `location.search` change, so re-runs from other deps no longer clobber an in-flow selection. The collectible block still re-runs (it needs asynchronously-loaded collections).

Regression test added.

https://github.com/user-attachments/assets/1c199e1e-a74f-4b5a-8777-ab8da2dc7cdb

### DIFF
--- a/extension/src/popup/views/Send/hooks/__tests__/useSendQueryParams.test.tsx
+++ b/extension/src/popup/views/Send/hooks/__tests__/useSendQueryParams.test.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Provider } from "react-redux";
 import { useLocation } from "react-router-dom";
-import { renderHook } from "@testing-library/react";
+import { renderHook, act } from "@testing-library/react";
 import { StrKey } from "stellar-sdk";
 import {
   makeDummyStore,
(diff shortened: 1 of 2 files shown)


piyalbasu:
### Code review

piyalbasu (approved):
This LGTM!

Issue #2769, opened 2026-05-11
[Extension] Derive auth keypair from seed for backend authentication


piyalbasu:
Derive the Ed25519 auth keypair on the extension via `HMAC-SHA256(seedBytes, "freighter-auth-v1")`. The hex-encoded pubkey is the user's anonymous backend user ID — separate from any Stellar G keypair, never used for wallet signing.

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — Auth Flow + Key properties.

Blocks: [Extension] Generate per-request backend JWT in `@shared/api`.

**Acceptance:**
- Same seed produces identical auth keypair (and therefore identical user ID) on extension and mobile.
- Auth pubkey is not a valid Stellar G address (cryptographically independent from the wallet keypair).
- No wallet-signing prompt is triggered by derivation or use.

PR #2876, opened 2026-06-26
feat(auth): derive backend auth keypair from seed (#2769)

piyalbasu:
## TL;DR

Adds the extension-side primitive that turns a wallet's recovery phrase into the user's anonymous backend identity — the first piece of Cross-Platform Contact Sync ([#2769](https://github.com/stellar/freighter/issues/2769)). It derives a dedicated auth key from the seed that is cryptographically independent from the user's Stellar wallet key, so the backend can recognize a returning user without ever learning their wallet address and without triggering any signing prompt.

This PR is **derivation only** — it deliberately does not generate request tokens, call any backend, or touch the contacts UI. Those are follow-up tickets. Nothing in the product changes yet; this is a self-contained, fully-tested building block.

The same recovery phrase must produce the exact same identity on the extension and on mobile, so this ships committed cross-platform test vectors that the mobile app will be held to as well.

> **Design doc** lives in `wallet-eng-monorepo` (`design-docs/contact-lists/Freighter Auth Keypair Derivation Design Doc.md`), alongside the Contact Sync design doc — not in this repo.
>
> **Draft:** opening for early review of the crypto contract and test vectors before the dependent token-generation work builds on it.

<details>

**What changed (all new, under `@shared/api/helpers/`):**

- [`deriveAuthKeypair.ts`](https://github.com/stellar/freighter/blob/6245cddfdebf93b31dbb8d67835ccf23c3f19e3b/@shared/api/helpers/deriveAuthKeypair.ts) — the primitive:
  - `deriveAuthSeed(mnemonic)`: `HMAC-SHA256` (via `crypto.subtle`) with **key = the 64-byte BIP39 seed** (`bip39.mnemonicToSeedSync(mnemonic)`) and **message = `utf8("freighter-auth-v1")`**, returning 32 bytes. Marked `@internal` (returns private-key material; exported only for test assertions).
  - [`deriveAuthKeypair(mnemonic)`](https://github.com/stellar/freighter/blob/6245cddfdebf93b31dbb8d67835ccf23c3f19e3b/@shared/api/helpers/deriveAuthKeypair.ts#L59-L66): feeds that 32-byte seed to `Keypair.fromRawEd25519Seed`; `userId = keypair.rawPublicKey().toString("hex")` (lowercase hex — matches the backend's canonical `sub`). Pure: no logging, no keyManager, no messaging, no persistence.
- [`authKeypairVectors.ts`](https://github.com/stellar/freighter/blob/6245cddfdebf93b31dbb8d67835ccf23c3f19e3b/@shared/api/helpers/authKeypairVectors.ts) — committed cross-platform vectors (`mnemonic → authSeedHex → userId`). Kept **outside `__tests__/`** on purpose: Jest's default `testMatch` collects every file under `__tests__/` as a suite and fails a fixture with "must contain at least one test." The intermediate `authSeedHex` is included so a failing mobile test localizes the divergence (HMAC step vs Ed25519 step). **freighter-mobile must mirror these.**
- [`__tests__/deriveAuthKeypair.test.ts`](https://github.com/stellar/freighter/blob/6245cddfdebf93b31dbb8d67835ccf23c3f19e3b/@shared/api/helpers/__tests__/deriveAuthKeypair.test.ts) — 10 tests mapped to acceptance criteria: vector parity (#1), determinism, lowercase-64-hex format, independence from the wallet key (#2), no messaging side-effects (#3), invalid-mnemonic rejection.
- `@shared/api/package.json` / `yarn.lock` — declares `bip39@3.1.0` (exact pin; promoted from a transitive dep). No new library is introduced to the project.

**Notable decision — `bip39` directly instead of `stellar-hd-wallet`:** `stellar-hd-wallet` cannot run under jest/jsdom (its compiled bip39 import throws `Cannot read properties of undefined (reading 'wordlists')`; verified that adding it to the transform allowlist does not fix it). `bip39.mnemonicToSeedSync` is byte-identical to what `stellar-hd-wallet` wraps, so cross-platform parity is unaffected.

**Acceptance #2 wording correction:** the ticket says "auth pubkey is not a valid Stellar G address," which is technically false (any 32 bytes StrKey-encode to a format-valid `G…`). The true, tested property is **cryptographic independence from the wallet keypair**. Ticket text should be updated.

**Verification:** `yarn jest @shared/api` (full collection) → 8 suites / 63 tests pass under `jest-fixed-jsdom` (real WebCrypto); `tsc -p @shared/api/tsconfig.json` clean. Test vectors were independently regenerated from the algorithm and matched.

**Known minor follow-ups (non-blocking, not yet applied):**
- [ ] Clarify the `authKeypairVectors.ts` header comment to state HMAC arg order explicitly (key = seedBytes, message = salt) for mobile implementers.
- [ ] Strengthen the independence test to also assert the expected `userId` (today it only asserts inequality with the wallet key; correctness is already covered by the vector test).
- [ ] Drop the redundant `Buffer.from(...)` wraps and hoist `AUTH_SALT` bytes to a module constant (micro-cleanups).
- [ ] Sibling ticket: `freighter-mobile` mirrors the algorithm + vectors.
- [ ] Update ticket #2769 acceptance #2 wording.

**The two `translation.json` keys** in the diff ("Auto-lock timer" en/pt) are auto-generated by the repo's husky `i18next-scanner` pre-commit hook filling a pre-existing gap on `master`; unrelated to this feature.

</details>


@shared/api/helpers/authKeypairVectors.ts
```diff
@@ -0,0 +1,28 @@
+// Canonical cross-platform auth-keypair derivation vectors.
```
JakeUrban: Can we indicate in this file that its used only for tests?

@shared/api/helpers/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,66 @@
...
+export const AUTH_SALT = "freighter-auth-v1";
+
+/**
+ * Computes the 32-byte auth seed: HMAC-SHA256 keyed on the wallet's 64-byte
+ * BIP39 seed, over the salt. Throws "Invalid mnemonic (see bip39)" on a
+ * malformed mnemonic.
+ *
+ * Implementation note: stellar-hd-wallet@1.0.2's fromMnemonic() is an ESM
+ * module whose internal bip39 default-import hits a Jest CJS interop edge case
+ * (bip39 sets __esModule:true without a .default export). We call bip39 named
```
JakeUrban, line 19: Can we file an issue on bip39 for this?

@shared/api/helpers/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @returns keypair stellar-sdk Keypair; the JWT ticket signs with keypair.sign().
+ */
+export const deriveAuthKeypair = async (
+  mnemonic: string,
+): Promise<{ userId: string; keypair: Keypair }> => {
+  const authSeed = await deriveAuthSeed(mnemonic);
+  const keypair = Keypair.fromRawEd25519Seed(Buffer.from(authSeed));
+  const userId = keypair.rawPublicKey().toString("hex");
+  return { userId, keypair };
+};
```
JakeUrban, line 70: ~nit: if `keypair` is only returned for testing purposes, you could mark this function as `@internal`, create another function `deriveUserId(mnemonic: string): string` that calls this function, and move the userId derivation line into the new function. That way, real users of this functionality don't need to concern themselves with the underlying keypair.~

JakeUrban (approved):
Looks good but left some suggestions you can take or leave

@shared/api/helpers/deriveAuthKeypair.ts
JakeUrban: Never mind, I realized that keypair is necessary to sign JWTs, proving possession of the userId's associated private key.

@shared/api/helpers/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,66 @@
...
+ * module whose internal bip39 default-import hits a Jest CJS interop edge case
+ * (bip39 sets __esModule:true without a .default export). We call bip39 named
+ * exports directly — bip39 is already a transitive dep of stellar-hd-wallet,
+ * and the derivation is byte-for-byte identical.
+ * @internal Exported only for tests (intermediate-value assertions). The
+ * returned bytes are private-key seed material — do not use as a standalone key.
+ */
+export const deriveAuthSeed = async (mnemonic: string): Promise<Uint8Array> => {
+  // Validate mnemonic first, matching stellar-hd-wallet's error contract.
+  if (!validateMnemonic(mnemonic)) {
```
aristidesstaffieri: **Consider pinning the english wordlist on `validateMnemonic`**

`validateMnemonic(mnemonic)` relies on bip39's implicit default wordlist. That default is set by require-order in bip39's `_wordlists.js` — the last list that successfully `require`s wins (english today), and the file's own comment warns *"Bundles may remove wordlists they don't need."* Everywhere else in the wallet we pin it explicitly: `StellarHDWallet.validateMnemonic(mnemonic, "english")`.

This isn't broken today — webpack bundles all the statically-`require`d wordlist JSON, so english stays the default and CI is green — but the default is implicit and diverges from our other mnemonic paths. Suggest pinning it to be safe and consistent:

```ts
import { mnemonicToSeedSync, validateMnemonic, wordlists } from "bip39";
// ...
if (!validateMnemonic(mnemonic, wordlists.english)) {
  throw new Error("Invalid mnemonic (see bip39)");
}
```

(`mnemonicToSeedSync` takes no wordlist — it just PBKDF2s the normalized string — so it needs no change.)

@shared/api/helpers/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,66 @@
...
+    throw new Error("Invalid mnemonic (see bip39)");
+  }
+  // BIP39 seed, 64 bytes, empty passphrase — identical to what StellarHDWallet
+  // .fromMnemonic(mnemonic).seedHex produces.
+  const seedBytes = Buffer.from(mnemonicToSeedSync(mnemonic));
```
aristidesstaffieri: it looks like `mnemonicToSeedSync` already returns a Buffer so the outer buffer is redundant.

@shared/api/helpers/deriveAuthKeypair.ts
piyalbasu: Fixed in 6304c206. Now `validateMnemonic(mnemonic, wordlists.english)` — pins the wordlist explicitly to match the rest of the wallet's mnemonic paths.

@shared/api/helpers/deriveAuthKeypair.ts
piyalbasu: Fixed in 6304c206. Dropped the redundant `Buffer.from()` — `mnemonicToSeedSync` already returns a Buffer.

Issue #2879, opened 2026-06-28
[Extension] Route all freighter-backend-v2 calls through the authed fetch wrapper


piyalbasu:
Centralize the scattered `INDEXER_V2_URL` fetches through a single backend-v2 client built on `authedFetch` (#2770), so every request to `freighter-backend-v2` carries a per-request JWT when the user is authenticated.

Today each call site builds its own `fetch`/`URL` against `INDEXER_V2_URL`:
- `@shared/api/internal.ts` (e.g. `/protocols`)
- `@shared/api/helpers/fetchCollectibles.ts` (`/collectibles`)
- `@shared/api/helpers/getLedgerKeyAccounts.ts` (`/ledger-key/accounts`)
- `extension/src/background/helpers/account.ts` (`/rpc-health`)
- (plus any others reaching `INDEXER_V2_URL`)

Route them all through one wrapper that:
- attaches the per-request JWT (`buildAuthJwt` / `authedFetch`, #2770) when the background has an unlocked auth keypair (#2769);
- sends anonymously when no seed is unlocked (the backend stays permissive — see freighter-backend-v2 companion ticket);
- retries once on 401 (already in `authedFetch`).

Part of the Unified User Model auth rollout: "whenever Freighter clients call freighter-backend-v2, they go through a wrapped fetch that derives the auth key and sends the JWT."

**Depends on:** #2769, #2770.

**Acceptance:**
- All backend-v2 requests flow through the shared wrapper; no call site constructs its own `fetch` to `INDEXER_V2_URL`.
- Authenticated requests (unlocked wallet) carry a valid JWT; locked / onboarding / watch-only requests still succeed anonymously.
- No regression in existing endpoint behavior.

PR #2880, opened 2026-06-29
feat(auth): route all backend-v2 calls through the authed fetch chokepoint (#2879)

piyalbasu:
## TL;DR

Routes **every** Freighter-extension call to `freighter-backend-v2` through a single background "chokepoint" that attaches the user's per-request auth token when the wallet is unlocked, and sends anonymously otherwise ([#2879](https://github.com/stellar/freighter/issues/2879)). This is the client groundwork so the backend can attribute requests to a user and, eventually, require authentication.

It's **non-breaking**: the backend stays permissive, so old clients (and locked/onboarding states) keep working — updated clients just start including the token, which shows up in the backend's auth metrics as adoption signal. No user-facing change.

**Invariant:** a backend-v2 request carries the per-request JWT only when the wallet is unlocked and goes out anonymously otherwise. The mnemonic / derived auth keypair never leaves the background (only `{ status, body }` returns to the popup) and never appears in a request body or the JWT payload beyond the pubkey `sub` claim.

> **Stacked on #2770** ([PR #2877](https://github.com/stellar/freighter/pull/2877)) — its base is `feat/2770-per-request-backend-jwt`, so this diff shows only the #2879 work. Merge order: #2876 → #2877 → this.
>
> **Design doc** lives in `wallet-eng-monorepo` (`design-docs/contact-lists/Freighter Backend-V2 Authed Fetch Chokepoint Design Doc.md`).
>
> **Draft** for review of the chokepoint architecture + the message-routing of the migrated call sites.

<details>

**The chokepoint** — [`extension/src/background/helpers/callBackendV2.ts`](https://github.com/stellar/freighter/blob/00c2a9ee00a71ed6aae45834a28a89a661316c48/extension/src/background/helpers/callBackendV2.ts): derives the auth keypair on-demand from the unlocked session mnemonic (else `null` → anonymous), and either delegates to `authedFetch` (#2770, JWT + retry-once-on-401) or does a plain `fetch`. Returns `{ status, body }`.
- **Signed path correctness:** `INDEXER_V2_URL` already includes `/api/v1`, so the JWT's `methodAndPath` is derived from the resolved URL's `pathname + search` (full server request-target incl. query, e.g. `POST /api/v1/ledger-key/accounts?network=PUBLIC`) — matching the server's `r.URL.RequestURI()`. Unit-tested by decoding the JWT claim.

**Wiring (option A — route through the background):** popup callers reach the chokepoint via a new `FETCH_BACKEND_V2` message (`SERVICE_TYPES` + `message-request.ts` union + a `popupMessageListener` case **gated by `isFromExtensionPage`** so content scripts can't trigger authed calls). The seed never leaves the background; only `{ status, body }` returns to the popup; the base URL is server-fixed (not an open proxy).

**Migrated call sites:**
- `getDiscoverData` → `GET /protocols` (popup → message)
- `getLedgerKeyAccounts` → `POST /ledger-key/accounts?network=` (popup → message)
- `fetchCollectibles` → `POST /collectibles?network=` (popup → message)
- `getTokenPrices` (v2 path only) → `POST /token-prices?network=` (popup → message). The `useV2=false` path stays a direct fetch to the legacy v1 indexer. This site was added to `master` by #2870 *after* this branch diverged, so it wasn't in the original migration; routing it fixed a `TS2552` (stray `INDEXER_V2_URL`) and closed the "no call site builds its own fetch to `INDEXER_V2_URL`" gap.
- rpc-health (`account.ts`) → `GET /rpc-health?network=` (background → direct `callBackendV2` with `skipAuth: true` — this endpoint is never auth-gated, so it sends anonymously and skips the keypair/PBKDF2 derivation a JWT would trigger on every health check)

**Notable:** routing rpc-health introduced a circular import (`ducks/session → account.ts → callBackendV2 → helpers/session → store → ducks/session`); fixed by lazy-importing `callBackendV2` inside `getIsRpcHealthy`. `npx madge --circular` confirms **zero cycles remain**.

**Verification:** `yarn jest @shared/api` 76/76; `yarn jest extension/src/background` 33 suites / 0 failed. New unit tests cover the chokepoint (unlocked/locked, POST body+Content-Type, full signed query path, non-2xx) and each migrated wrapper's message shape + mapping. Reviewed per-task + a whole-branch review (verdict: ready to merge).

**Scope / follow-ups (out of scope here):**
- Backend permissive→**required** flip (`freighter-backend-v2#114`), gated on extension+mobile adoption metrics + a pre-flip audit that none of these endpoints is hit by a locked/watch-only account.
- Mobile parity (`freighter-mobile#865`).
- **Perf:** `getTokenPrices` is the first *high-frequency* chokepoint caller (the Account view polls it every 30s, uncached), so it now runs `deriveAuthKeypair` → synchronous PBKDF2 (~18ms block) per poll. Correct but worth a follow-up (short-lived in-background keypair memo, or let the poll honor cache). Tracked separately.
- The two `translation.json` keys in the diff are the husky `i18next-scanner` hook backfilling a pre-existing `master` gap — unrelated.

</details>


extension/src/background/messageListener/popupMessageListener.ts
```diff
@@ -647,6 +647,17 @@ export const popupMessageListener = (
       return userActivity({ sessionTimer, sessionStore, localStore });
     }
 
+    case SERVICE_TYPES.FETCH_BACKEND_V2: {
+      if (!isFromExtensionPage) return { error: "Unauthorized" };
```
aristidesstaffieri: **Bug: this gate breaks all backend-v2 calls in the DEV_SERVER workflow.**

When the popup runs via the webpack dev server (`DEV_SERVER=true`), `sendMessageToBackground` falls back to relaying through the content script (`extensionMessaging.ts` → `redirectMessagesToBackground.ts`), so the message arrives here with `sender.tab` set and `sender.url` on the dev-server origin — `isFromExtensionPage` is `false` and every `FETCH_BACKEND_V2` call returns `{ error: "Unauthorized" }`.

Downstream effects in dev: `getTokenPrices`/`getDiscoverData` destructure `status === undefined` and throw, while `fetchCollectibles`/`getLedgerKeyAccounts` silently return empty. Discover, token prices, collectibles, and ledger-account import all worked as direct popup fetches before this PR and are now broken under the dev server.

The gate needs a DEV_SERVER carve-out (like other dev-relayed message types) or the relay path needs to preserve an extension-page sender.

extension/src/background/helpers/callBackendV2.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+  } else {
+    res = await doFetch(fullUrl.href, {
+      method,
+      headers: body ? { "Content-Type": "application/json" } : {},
+      body,
+    });
+  }
+
+  return { status: res.status, body: res.ok ? await res.json() : null };
```
aristidesstaffieri: **Bug: server error bodies are discarded on non-2xx, so Sentry reports lose all diagnostic detail.**

`body: res.ok ? await res.json() : null` hands the popup `{ status: 500, body: null }` when backend-v2 returns an error with a JSON payload. The direct-fetch code this replaces parsed the error body and propagated it (plus `statusText`) into the thrown Error and Sentry capture.

Result: `getTokenPrices`/`getDiscoverData` now throw `Error("null")`, and every Sentry report for a backend-v2 failure reads `Failed to fetch token prices - 500: null` — the server's error detail never reaches telemetry.

Suggest parsing the body on non-2xx too (falling back to `null` only if it isn't JSON), e.g.:

```ts
const body = await res.json().catch(() => null);
return { status: res.status, body };
```

extension/src/background/helpers/callBackendV2.ts
```diff
@@ -0,0 +1,98 @@
...
+  try {
+    const mnemonic = await getEncryptedTemporaryData({
+      sessionStore,
+      localStore,
+      keyName: TEMPORARY_STORE_EXTRA_ID,
+    });
+    if (!mnemonic) return null;
+    const { keypair } = await deriveAuthKeypair(mnemonic);
+    return keypair;
+  } catch (_e) {
```
aristidesstaffieri: **Bug: this catch only ever swallows *unexpected* errors, and does so silently.**

The comment says the catch handles "Locked (no hashKey) or no mnemonic", but neither case throws: `getEncryptedTemporaryData` returns `""` gracefully both when the entry is missing and when the store is locked, and that's already handled by the `if (!mnemonic) return null;` above. So the only errors that reach this catch are unexpected ones — a corrupted `temporaryStoreExtra` entry, a WebCrypto/PBKDF2 failure — and they're dropped with no logging or `captureException`.

For an unlocked user hitting one of those, every backend-v2 request silently downgrades to anonymous. The moment the backend enforces auth on these endpoints, they get empty collectibles, failed price loads, and Discover errors with zero telemetry signal to diagnose from. Suggest a `captureException(_e)` (or at least `console.error`) before returning null.

extension/src/background/helpers/callBackendV2.ts
```diff
@@ -0,0 +1,98 @@
...
+  sessionStore,
+  localStore,
+  fetchImpl,
+  skipAuth,
+}: CallBackendV2Params): Promise<CallBackendV2Result> => {
+  const doFetch = fetchImpl ?? fetch;
+  const fullUrl = new URL(`${INDEXER_V2_URL}${path}`);
+  const keypair = skipAuth
+    ? null
+    : await tryGetAuthKeypair(sessionStore, localStore);
```
aristidesstaffieri, line 90: **Perf: every authed request pays a synchronous PBKDF2 in the MV3 service worker.**

Each non-`skipAuth` call re-runs `deriveAuthKeypair`, whose `mnemonicToSeedSync` is 2048 rounds of synchronous PBKDF2-HMAC-SHA512 on the single-threaded background worker. Token prices are polled repeatedly and collectibles/protocols fire on account load, so concurrent popup→background messages (unlock, signing, balance loads) queue behind repeated key derivation — noticeable on low-end machines. `skipAuth` currently exempts only rpc-health.

I understand never caching the keypair at rest is deliberate (#2769), but worth a deliberate decision here rather than an accident: e.g. the async `mnemonicToSeed`, or a short-lived in-memory memo scoped to the unlocked session (cleared on lock), which doesn't change the at-rest posture.

@shared/api/internal.ts
```diff
@@ -631,18 +637,41 @@ export const getTokenPrices = async (
     if (!filteredTokens.length) {
       return {};
     }
-    url = new URL(`${INDEXER_V2_URL}/token-prices`);
-    url.searchParams.append("network", priceNetwork);
-  } else {
-    url = new URL(`${INDEXER_URL}/token-prices`);
+
+    const { status, body } = await sendMessageToBackground<{
```
aristidesstaffieri: **Cleanup: this ~10-line `sendMessageToBackground<{ status, body }>` block is copy-pasted four times.**

The same inline-typed `FETCH_BACKEND_V2` message (response type, `activePublicKey: null`, method/path/body) appears here, at `getDiscoverData` (internal.ts:691), `fetchCollectibles.ts:94`, and `getLedgerKeyAccounts.ts:18`. Each new v2 endpoint will copy it again, and the four inline response types can drift independently — notably, none of them models the handler's `{ error: "Unauthorized" }` reply shape today, so each call site already degrades differently (throws on `undefined`, returns `{}`, etc.).

Suggest extracting a single `fetchBackendV2({ method, path, body })` helper in `@shared/api/helpers` (next to `extensionMessaging.ts`) that owns the message shape and the typed result, including the error reply.

extension/src/background/messageListener/popupMessageListener.ts
piyalbasu: Good catch — this broke every v2 call under the dev server. Added a `DEV_SERVER` carve-out so the content-script relay works locally, while the gate stays intact in production (`DEV_SERVER=false`).

Fixed in `3dd05b59`: https://github.com/stellar/freighter/blob/3dd05b593336b1e482b1faef192e60c56dbfa0b5/extension/src/background/messageListener/popupMessageListener.ts#L650-L657

extension/src/background/helpers/callBackendV2.ts
piyalbasu: Fixed — `callBackendV2` now parses the body on non-2xx too, so the server's error payload reaches Sentry instead of `null`; falls back to `null` only when the body isn't JSON. Added a test asserting a JSON error body is preserved.

`3dd05b59`: https://github.com/stellar/freighter/blob/3dd05b593336b1e482b1faef192e60c56dbfa0b5/extension/src/background/helpers/callBackendV2.ts#L106-L110

extension/src/background/helpers/callBackendV2.ts
piyalbasu: Fixed — the catch now `captureException`s the error, and I corrected the comment: locked/missing-entry is handled by the `!mnemonic` check above, so only unexpected failures (corrupted store / WebCrypto) reach here. New test drives a derive-throw and asserts both the capture and the anonymous fallback.

`3dd05b59`: https://github.com/stellar/freighter/blob/3dd05b593336b1e482b1faef192e60c56dbfa0b5/extension/src/background/helpers/callBackendV2.ts#L52-L66

extension/src/background/helpers/callBackendV2.ts
piyalbasu: Agreed this warrants a deliberate call. Deferring to #2897 rather than folding into this routing PR: derivation is deterministic, so the fix (a session-scoped in-memory memo cleared on lock, or async `mnemonicToSeed`) is equally secure and doesn't change the at-rest posture (#2769). Keeping it out here to avoid a rushed keypair-lifecycle change; both options are captured in #2897.

Resolving — happy to reopen if you'd rather it land in this PR.

@shared/api/internal.ts
piyalbasu: Done — extracted `fetchBackendV2({ method, path, body })` in `@shared/api/helpers`, now used by all four call sites. It owns the message shape + typed result and normalizes the `{ error }` reply to a 401 (previously unmodeled — the reason each site degraded differently). Test covers passthrough + error normalization.

`3dd05b59`: https://github.com/stellar/freighter/blob/3dd05b593336b1e482b1faef192e60c56dbfa0b5/@shared/api/helpers/fetchBackendV2.ts#L26-L50

@shared/api/helpers/fetchBackendV2.ts
```diff
@@ -0,0 +1,50 @@
...
+    method,
+    path,
+    body,
+  });
+
+  if (response && "error" in response) {
+    return { status: 401, body: { error: response.error } };
+  }
+  return response;
+};
```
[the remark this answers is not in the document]
piyalbasu: Fixed — `fetchBackendV2` now maps only the sender-rejection error (`"Unauthorized"`) to 401; any other `{ error }` reply (e.g. `"Message type not supported"`) maps to 500 so callers don't misread an unrelated failure as an auth problem. It also guards a falsy/undefined message-channel response into a defined 500 result instead of returning it and breaking `{ status, body }` destructuring. Added tests for both the non-auth-error and missing-response cases.

`8f63926d`: https://github.com/stellar/freighter/blob/8f63926d5a6859a8b0eda5e7599ecf7c521febab/@shared/api/helpers/fetchBackendV2.ts#L46-L57

extension/src/background/helpers/callBackendV2.ts
```diff
@@ -0,0 +1,111 @@
...
+  } catch (e) {
+    // getEncryptedTemporaryData returns "" for both a locked store and a
+    // missing entry (handled by the `!mnemonic` check above), so anything
+    // reaching here is unexpected — a corrupted temporaryStoreExtra entry or a
+    // WebCrypto/PBKDF2 failure. Capture it: otherwise an unlocked user silently
+    // downgrades to anonymous with zero telemetry the moment the backend
+    // enforces auth on these endpoints.
+    captureException(
+      `callBackendV2: unexpected error deriving auth keypair - ${JSON.stringify(e)}`,
+    );
```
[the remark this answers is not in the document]
piyalbasu: Fixed — `tryGetAuthKeypair` now passes the original exception to `captureException` (preserving the stack) with the context string in `extra`, instead of `JSON.stringify(e)` — which flattens an `Error` to `{}` and can itself throw on non-serializable values. The test now asserts the original `Error` object plus the `extra` context reach `captureException`.

`8f63926d`: https://github.com/stellar/freighter/blob/8f63926d5a6859a8b0eda5e7599ecf7c521febab/extension/src/background/helpers/callBackendV2.ts#L60-L67

aristidesstaffieri (approved):

PR #2885, opened 2026-07-02
Fix SEP-41/SAC add-token regressions (feeds into #2869)

leofelix077:
## Summary
I lost write access to `stellar/freighter` (now read-only), so I can no longer push directly to my PR branch `feat/addtoken-sac-changetrust-review` (used by #2869). This PR targets that branch directly (not `master`) so a maintainer can merge it in without needing to grant me access back.

Contains one commit addressing Piyal's latest review round on #2869:
- Fix a residual race where a successful SAC trustline could still report "user rejected" to the dApp if Done was clicked before the add-token round-trip finished
- Remove dead/unreachable `ADD_TOKEN_FAILURE` plumbing
- Fix a SEP-41 storage-write failure incorrectly reporting success to the dApp
- Make SEP-41 a true one-step flow (submit + auto-close, matching pre-SAC-review behavior) instead of a second "Done" click
- Scope the "already has a trustline" message to SAC only (SEP-41 has no trustline)
- Allow re-adding SEP-41 tokens multiple times (local-only, no cost) instead of blocking on a prior add
- New Playwright e2e regression test + updated unit tests

## Test plan
- [x] `tsc --noEmit` clean
- [x] Unit tests pass (`yarn jest`)
- [x] e2e tests pass (`yarn test:e2e`), including a new deterministic regression test for the Done/race issue


piyalbasu (approved):
These changes look good. I'm merging into the main feature branch and will be doing a more thorough review

PR #2929, opened 2026-08-04
chore(wallets): extract WalletRow, RenameWallet and AddWallet into components

CassioMG:
## Why this PR exists

This is the **first step of the Home refresh** project. That work redesigns the Home screen plus
six related screens, and it's being split into small, independently reviewable PRs rather than one
large one.

Three of those upcoming PRs each restyle a different piece of the Wallets screen:

| Upcoming PR | Restyles |
| --- | --- |
| Wallets list | the account rows + a new 4-button header |
| Rename modal | the rename dialog |
| Add wallet | the "add a wallet" sheet |

The problem: **all three of those components currently live in the same 494-line file**,
`views/Wallets/index.tsx`. Three PRs rewriting one file would conflict on every merge, and each
review would show the other two PRs' churn as noise.

This PR splits that file so those three PRs can each touch a different file and be reviewed on
their own.

## What this PR does

**Pure extraction — no behavior changes.** Nothing about the app changes: same rendered DOM, same
CSS, same props, same analytics. Only *which file* the code lives in.

Three components move out of `views/Wallets/index.tsx` into their own directories, following the
existing convention of the 28 other components already under `popup/components/account/`:

## Demo sample (no behavior changes)

https://github.com/user-attachments/assets/d31b8f75-b411-45d0-bab0-210e569c9619

## Deliberately not done

- **No new unit tests.** These components are moved, not written. The existing
  `loadAccount.test.ts` flow is the regression guard, and any test written against the ellipsis
  dropdown would be deleted by the Wallets-list PR, which removes that dropdown.
- **Four dead/undefined CSS classes are carried unchanged** — `.Wallets__fail`,
  `.detail-indentifier` (typo'd, no consumer), `.WalletRow__options-actions`, and
  `.AddWallet__row-link`. Removing them would have changed the compiled CSS and forfeited the
  equivalence check above. Each is assigned to the follow-up PR that rewrites its area.
- **Two pre-existing `'await' has no effect` warnings** in `views/Wallets/index.tsx` are left
  alone. They come from `await dispatch(...)` on plain `createSlice` actions and are unchanged
  from `master` — only their line numbers shifted. The Wallets-list PR rewrites that exact block
  and will clean them up.

## Merge note for the follow-ups

The three follow-up PRs each own a different component file, but all three still touch small
regions of `views/Wallets/styles.scss` (the two overlay wrapper classes plus the dropdown block
being deleted). Merging **Wallets list → Rename modal → Add wallet** in that order keeps the
rebases trivial.


CassioMG:
### Code review

CassioMG:
Merged this right away since it's pure component extraction verbatim with no behavior change to unblock other work as described on the PR

PR #2932, opened 2026-08-04
feat(add-funds): refresh the add funds screen

CassioMG:
Part of the **[Home refresh](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=9567-16765&t=LfFmnsAINU8vVAWe-1)** project.

## What changed

The Add funds screen, per [Figma frame](https://www.figma.com/design/C3G0a4Gd6RQyplRBppGDsL/Freighter-Extension?node-id=9593-26775&t=LfFmnsAINU8vVAWe-1) `9593-26775`:

- **Each option card puts its icon on its own line above the title**, instead of beside the text block.
- **Header:** `X` close icon, and the "Choose your method" subtitle is removed.
- **Coinbase card copy** shortens to "Transfer from Coinbase & other options".
- **The `* payment methods may vary based on your location` footnote is removed** — and so is the `*` in the Coinbase description that referenced it. Leaving an orphaned asterisk with no footnote would have been worse than either state.
- **Transfer card:** purple QR icon, description "Receive funds from another wallet".

Both `?isAddXlm=true` and default modes keep their distinct copy.

Screenshot 2026-08-04 at 13 42 58


CassioMG:
### Code review

piyalbasu (approved):
Approved! Claude found 1 small CSS bug worth fixing

piyalbasu:
### `.AddFunds__error` uses a font-weight variable that nothing defines

**Nit — not a merge blocker.** Cosmetic, confined to the on-ramp error message.

**TL;DR:** The new error text is meant to render at medium weight, but the variable it asks for doesn't exist anywhere in the project or in the design system, so the browser throws the declaration away and the text renders at normal weight instead. It's a one-token fix. Worth catching here because a repo-wide cleanup of exactly this mistake is already in flight, and this PR adds a new instance that the cleanup won't cover.

**Steps to reproduce:**
1. Open Add Funds and trigger the on-ramp error state (the error message under the buttons).
2. Inspect the error text — computed `font-weight` is `400`, not the intended `500`.

---

<details>

**Root cause:** `--font-weight-medium` is not defined anywhere — not in `extension/src`, and not in `@stellar/design-system@3.2.8`, which exposes the weight scale as `--sds-fw-*` (`--sds-fw-medium: 500`). Because `font-weight` is an inherited property, a `var()` reference to an undefined custom property is invalid at computed-value time and resolves to `unset`, falling back to the parent's weight (400 here). The declaration is silently dropped.

https://github.com/stellar/freighter/blob/12c61e33b605a82fbe2e337822ed73f9ded6fec7/extension/src/popup/views/AddFunds/styles.scss#L59-L65

This is the same defect documented in the merged #2937 (`d1f328d`), whose commit message states it directly: the row weights "referenced `var(--font-weight-medium)` / `var(--font-weight-regular)`, which nothing defines … the weights never applied." That PR migrated its own files to `--sds-fw-medium`.

**Merge-order hazard:** the systematic sweep of this mistake — `61a7ea38`, "fix(styles): replace undefined `--font-weight-*` vars with SDS tokens", which converts 44 medium-weight sites — is **not** on `master` yet; it's unmerged on `origin/fix/undefined-font-weight-vars`. `master` currently has 44 `var(--font-weight-medium)` against 28 `var(--sds-fw-medium)`. If this PR lands after that sweep, it reintroduces a 45th broken site that the sweep has already passed over, and the cleanup will read as complete when it isn't.

Note also that the restyle removed the old `&__button` rule carrying `font-weight: var(--sds-fw-semi-bold)`, so `AddFunds/styles.scss` now has exactly one `font-weight` declaration and it's the non-functional one.

**Suggested fix:**

```diff
   &__error {
     color: var(--sds-clr-red-09);
     font-size: pxToRem(12px);
-    font-weight: var(--font-weight-medium);
+    font-weight: var(--sds-fw-medium);
     line-height: pxToRem(18px);
     margin-top: pxToRem(8px);
   }
```

</details>

PR #2940, opened 2026-08-05
feat(swap): default the "You receive" token to USDC

### LINKED ISSUE #2914: [Swap] Default the "You receive" token to USDC
## Summary

In the Swap flow, the **You receive** field currently defaults to an empty state with a `Select` button (and a `+` icon on the left). We should default the receive token to **USDC** instead, so users have one less tap to complete a common swap.

## Current behavior

- The **You receive** field shows a `Select` button (empty state).
- A `+` icon is shown on the left of the field.

## Desired behavior

- Default the **You receive** token to **USDC**, except when starting the swap flow from the USDC Asset Details screen in which case the **You receive** token should default to the native **XLM**.
- Apply this default **regardless of whether the user already holds a USDC balance**.
- Users can still open the selector to choose a different receive token.
- If we keep an empty/select state anywhere, remove the `+` icon on the left (see [thread note](https://stellarfoundation.slack.com/archives/C03DK78B43C/p1783960898557799))

## Source

Based on Slack discussion: https://stellarfoundation.slack.com/archives/C03DK78B43C/p1783960898557799

JiahuiWho:
## What

  Closes [#2914](https://github.com/stellar/freighter/issues/2914). 
  1. The swap flow's **You receive** field now defaults to the network's USDC instead of the empty `Select` state.
  2. When the flow starts from USDC itself, the destination defaults to native XLM instead.
  3. Also removes the `+` icon from the empty `Select` pill.
  
  ## Main Changes
**Trustline Flag** 
  - `destRequiresTrustline` is derived from balances and drives everything that depends on "this swap adds a trustline": 
    -   the changeTrust op in `getBuiltTx`, 
    -   the 2-op fee,
    -   the XLM-reserve preflight and Max deduction,
    -   the review screen's trustline banner/info sheet 
  - `useGetBalances` now exposes the pre-filter list, so a hidden held asset never gets a redundant changeTrust, doubled fee, false reserve block, or a trustline limit reset to the SDK default
  - `swapTrustlineAdded` telemetry gates on the changeTrust op in the submitted XDR itself, so it can't disagree with what settled on-chain
  
  **Destination icon**
  - `getAssetIcons` gains `additionalAssetIds` (mirroring token prices') so the unheld destination's logo resolves through the held-token pipeline.
  
   ## Testing

  - `Swap.destinationDefault.test.tsx`: network defaults, USDC-source → XLM, param precedence, unconfigured networks
  - `getBuiltTx` tests: changeTrust prepended for unheld destinations, asset taken from `destAsset`
  - `ReviewTx.trustlineBanner` tests: banner renders from the derived flag with no snapshot; derived flag overrides a stale snapshot
  - Full suite green, tsc clean; verified live on testnet (default + collision fallback, trustline banner and 2-op fee for an unheld destination, no banner/single-op fee for a hidden held destination)

Updated 08/14
https://github.com/user-attachments/assets/a0872e72-62ae-4167-a257-dabb3ac3bbb6


aristidesstaffieri:
`a4ba4f044` moved the trustline decision to a balances-derived flag (`destRequiresTrustline`, `SwapAmount/index.tsx:151`) and wired it into the builder, but the disclosure and telemetry consumers still read the pick-time snapshot. There are now two sources of truth and they disagree on the path this PR adds.

`ReviewTransaction/index.tsx:314`:

```ts
const requiresTrustline = !!destinationTokenDetails?.requiresTrustline;
```

`destinationTokenDetails` is populated only by the picker's `onClickAsset` (`views/Swap/index.tsx:179`); the default path at `:148-150` dispatches `saveDestinationAsset` alone, and `useSwapDestinationScan` bails on `!details` so it cannot bootstrap the value. For a user who does not hold USDC — the target case for this default — `destRequiresTrustline` is `true` and the built transaction carries `changeTrust`, while `requiresTrustline` is `false` and neither `TrustlineBanner` nor `TrustlineInfoSheet` renders.

Net effect: the user signs a two-op transaction that locks 0.5 XLM of reserve, with no disclosure anywhere in the flow. The fee is correctly doubled but unexplained. `shouldShowXlmReservePreflight` was migrated to the new flag, so the `XlmReserveSheet` only covers users who cannot afford the reserve — sufficiently funded accounts get no signal at all.

Same stale gate at `SubmitTransaction/hooks/useSubmitTxData.tsx:114`, so `swapTrustlineAdded` will not fire for any default-destination swap.

Suggest threading `destRequiresTrustline` down to `ReviewTx` (`SwapAmount/index.tsx:876`) and to the submit metric instead. The banner and sheet only need a token code, which is available from `getAssetFromCanonical(destinationAsset).code` without the snapshot.

extension/src/popup/components/swap/SwapAmount/index.tsx
```diff
@@ -147,6 +137,31 @@ export const SwapAmount = ({
...
+  // destination. Derived from holdings (like the picker derives held-ness)
+  // rather than the pick-time snapshot, so a defaulted or deep-linked
+  // destination — which never went through the picker — is covered too.
+  // Empty until balances resolve; every consumer (fee count, reserve
+  // preflight, simulation) only acts post-resolve, behind the spinner.
+  const heldBalances =
+    swapAmountData.data?.type === AppDataType.RESOLVED
+      ? swapAmountData.data.userBalances.balances
+      : [];
+  const destRequiresTrustline =
```
aristidesstaffieri, line 152: `destRequiresTrustline` resolves against the hidden-filtered balance list, so a destination the account already holds but has hidden reads as unheld and the transaction prepends a redundant `changeTrust`.

`SwapAmount/index.tsx:131` requests `{ showHidden: false }`, which threads through `useGetAssetDomainsWithBalances` → `useGetBalances` → `formatBalances` (`useGetBalances.tsx:42-47`) into `filterHiddenBalances`. Assets toggled off in Manage Assets → Asset visibility are absent from `userBalances.balances`, so the `getBalanceCanonicalKey` comparison at `:151` misses and `getBuiltTx` builds the op for an existing trustline.

The option itself predates this PR (#2137) and was correct while it only drove rendering. `a4ba4f044` promoted that display list into a transaction-construction input, so a UI preference now determines which operations get signed.

Consequences, in rough severity order:

- `shouldShowXlmReservePreflight` returns `true` whenever spendable XLM is under 0.5, so a user holding hidden USDC with a thin XLM balance is hard-blocked by the `XlmReserveSheet` from swapping into an asset they already hold.
- `deductNewTrustlineReserve` withholds 0.5 XLM from Max against a reserve that will never be consumed.
- `opCount` is 2, so the fee doubles for a no-op operation.
- `buildChangeTrustOperation` passes no `limit` (`getManageAssetXDR.ts:20`), so the SDK default `MAX_INT64` overwrites any custom trustline limit set elsewhere.

Reachable pre-PR by picking a hidden asset out of the picker — `currencyToRecord` derives `requiresTrustline` from the same filtered set (`useSwapTokenLookup.ts:185`) — but it required deliberate selection, and the review banner still fired on that path. As the default it needs no interaction and, per the previous comment, shows nothing.

Suggest sourcing the held-check from an unfiltered list; visibility is a display preference and should not reach transaction construction. `useGetSwapAmountData` already holds a `useGetBalances({ showHidden: true })` instance at `:46-49`.

aristidesstaffieri (approved):

PR #2943, opened 2026-08-06
🔧 Add fork-safe indexer fallbacks

JFWooten4:
fixes https://github.com/stellar/freighter/pull/2784#issuecomment-5207842011

`freighter-backend-prd.stellar.org/api/v1` and `freighter-backend-v2.stellar.org/api/v1` themselves probably don’t need to be secret—they are compiled into the [browser extension](https://github.com/stellar/freighter/releases/tag/5.44.0) and can be discovered by users in the JavaScript. It could be stored as a public GitHub variable if the backend is intended to be public or simply be observed in the extension’s network traffic.


piyalbasu:
### Right fix — but it needs #2784's stubs to land first, and two tests still reach the network after that

**Not a merge blocker on the approach** — the change itself is correct and I'd like to see it merged. The issue is sequencing: merged on its own today, fork CI stays red, just on different tests.

**TL;DR:** This correctly removes the reason fork PRs can't build. But four tests in the suite quietly depend on a real backend rather than on the stubs, so taking the backend away moves the failure rather than removing it. Three of those four are already fixed in #2784. One more isn't fixed anywhere yet, and there's a screenshot that was captured against a real backend and won't match without one.

Concretely, the proposal is: land #2784 first (it carries the stubs), then this. At that point this becomes a genuine two-line change whose green run actually proves something.

I verified all of this locally rather than reasoning about it — full 236-scenario runs, varying only the build's indexer URL. Numbers below.

---

<details>

**Method**

`origin/master` @ `9d42400`, overriding `INDEXER_URL` / `INDEXER_V2_URL` via the shell (`dotenv-webpack` runs with `systemvars: true`, so shell values win over `extension/.env`). Confirmed the real value was absent from the emitted bundle on every placeholder run. Placeholder used was `https://indexer.e2e.invalid`; behaviour is identical with `http://127.0.0.1`, since what matters is only that the string parses as an absolute URL.

**Why an empty value breaks the build today**

https://github.com/stellar/freighter/blob/9d4240074699d811bfda5df79fb2673b51ba692c/%40shared/api/internal.ts#L950-L966

Empty `INDEXER_URL` makes that `new URL()` argument a relative reference, so the constructor throws before `fetch` is reached. Playwright never sees a request, which is why it looks like a stubbing failure but cannot be fixed by adding stubs. This PR's diagnosis is exactly right.

**Full suite, placeholder build, `--max-failures=0`** (236 scenarios; #2784's `freighterApiIntegration.test.ts` applied):

```
220 passed, 4 failed, 2 flaky, 10 skipped
```

Failures, after discarding one that was a local untracked file on my checkout:

| Test | Fixed by #2784? |
| --- | --- |
| `freighterApiIntegration.test.ts:243` — should not sign transaction when not allowed | partially — see below |
| `sendPayment.test.ts:72` — Swap doesn't throw error when account is unfunded | **yes** (`stubAccountBalances(page, "0")`) |
| `swap.test.ts:886` — recovers to the amount screen when the swap quote expires at submit | **no — not touched by any PR** |

Each was confirmed to pass on the same commit with a real `INDEXER_URL`, so these are backend dependencies, not pre-existing breakage. `onboarding.test.ts:41` and `reviewTxFees.test.ts:758` came back flaky (passed on retry) in both configurations — unrelated to this change.

**`freighterApiIntegration.test.ts:243` is the interesting one**

There is no single assertion that passes in both environments:

| Build | `toBeDisabled()` | `toHaveCount(0)` |
| --- | --- | --- |
| Real URL | passes | fails (`Expected: 0, Received: 1`) |
| Placeholder | fails (absent) | passes |

`sign-transaction-sign` only renders in the non-Blockaid branch of the actions row, and which branch renders depends on scan data. That divergence *is* the bug — the test isn't hermetic, and no assertion choice can paper over it. It needs the same stub treatment the other three got, so the popup renders identically either way. (I've asked on #2784 for the `toHaveCount(0)` edit to be reverted, since it fails against a real backend and would break `master` on merge — but note that reverting alone leaves this test failing for forks until it's properly stubbed.)

Related: `domain-not-allowed-sign-transaction.png` differs by ~5% of pixels without a backend. Any screenshot whose rendering depends on backend data will pass for maintainers and fail for forks, and you can't commit two baselines for one test. Stubbing is the only fix — regenerating the baseline just inverts who fails.

**Suggested plan**

1. Merge #2784 with the `toHaveCount(0)` line reverted. That covers `sendPayment.test.ts:72` and three of the four integration tests.
2. Stub `freighterApiIntegration.test.ts:243` and `swap.test.ts:886` so they don't reach the network. These two are the remainder; both are small.
3. Merge this PR. Its CI run then demonstrates a hermetic suite rather than asserting one.

Steps 2 and 3 could reasonably be the same PR, since step 2 is what makes step 3 meaningful.

**Two smaller notes**

*Placeholder value.* `.invalid` is reserved by [RFC 6761 §6.4](https://datatracker.ietf.org/doc/html/rfc6761#section-6.4): it can never resolve and can never be registered. `127.0.0.1` is a real routable address, so if anything happens to be listening on port 80 of a runner, an unstubbed call gets a real response instead of failing. Low probability on a GitHub macOS runner; `.invalid` just removes the question. Either works — not worth holding the PR over.

*The deeper fix.* You're right in the description that these URLs don't need to be secret — they ship in every published build. Making them repository **variables** would fix this class of problem outright and make the fallback unnecessary, since forks can read variables. That needs someone with repo settings access to create them first (changing the references alone would break CI), so it's a follow-up rather than something to fold in here.

Once the suite is hermetic, a check that no request escapes to a real host would keep it that way and make this whole category of failure impossible to reintroduce.

</details>

piyalbasu (approved):
Approved! Thank you for your work on this @JFWooten4 . I know the scope expanded a little outside of what you originally set out to do, but this will make future contributions much easier for everyone 🎉

PR #2948, opened 2026-08-10
fix(blockaid): stop dropping the site scan verdict on sign transaction

aristidesstaffieri:
## Summary

On the sign-transaction approval screen, a Blockaid `is_malicious` verdict for the requesting dApp could be discarded before it reached the UI: no malicious-site banner rendered, and Confirm was not marked destructive. `useGetSignTxData` starts the site scan unawaited and then dispatches a final payload assembled from a stale local object, while the reducer full-replaces `data` — so whichever write landed last owned the entire payload and dropped the other writer's field. In the opposite resolution order the same defect silently blanked the fetched changeTrust asset icons.

Both contested fields now live in refs shared by the two writers, so neither ordering loses data.

## Root cause

`firstRenderPayload.siteScanData` is `undefined` at construction and is never mutated, because the scan callback returns a new object rather than writing back into it. The final dispatch read that field, so it wrote `undefined` unconditionally; when the scan resolved first, its real verdict was overwritten. `getSiteSecurityStates` treats `undefined` as "scan in flight" and returns all-clear, which suppresses both the banner and the destructive Confirm styling rather than degrading to "unable to scan".

The reverse ordering lost the icons. The guard intended to prevent that inspected `firstRenderPayload.icons`, which is hardcoded `{}` and never mutated, so its condition never held.

Impact is bounded to dApps the user has already allow-listed. `renderBanner` is priority-ordered and shows the domain-not-allowed banner first, so for every other origin the Blockaid site banner was already not the visible signal.

## What's in this PR

- `useGetSignTxData.tsx` — `siteScanDataRef` and `iconsRef` as the shared source of truth between the unawaited scan callback and the final `fetchData` dispatch, reset per `fetchData`.
- `useGetSignTxData.tsx` — removes the previous icon-preservation guard, which could never fire.
- `__tests__/useGetSignTxData.test.tsx` — two tests pinning both resolution orderings, driven by hand-resolved deferreds. `useAsyncSiteScan`, `useScanSite`, the reducer and `getSiteSecurityStates` stay real; only the two network leaves are stubbed.

## Test plan

- [x] Both new tests fail on the unmodified hook and pass with the fix; their mid-flow assertions confirm each value reached state before being overwritten, so neither is tautological
- [x] `yarn jest --ci --selectProjects jsdom --runTestsByPath extension/src/popup/views/SignTransaction/hooks/__tests__/useGetSignTxData.test.tsx` — 6/6
- [x] `yarn jest --ci --selectProjects jsdom --runTestsByPath extension/src/popup/views/__tests__/SignTransaction.test.tsx` — 12/12
- [x] `yarn jest --ci --selectProjects jsdom -t "Blockaid"` — 77 passed, no test failures
- [x] `yarn tsc --noEmit` clean; eslint clean on the changed source
- [ ] CI green
- [ ] Manual: with the dev Blockaid override forcing a malicious site verdict, sign a `changeTrust` transaction for an issuer absent from the icon cache from an allow-listed dApp. The banner must stay visible for the life of the popup and the asset icon must render — the two failure modes are mutually exclusive on `master`, so both holding at once confirms the fix.

## Followups

- The approval UI is interactive before the verdict lands. `isSubmitDisabled` in `SignTransaction/index.tsx` does not consider scan state, so a fast click can approve before the banner renders. This PR guarantees the verdict is not lost, not that it arrives before the user can act.
- The reducer full-replaces `data`, so any future async enrichment added to this hook must extend the same ref bookkeeping or it will reintroduce the clobber. Awaiting the scan promise before the final dispatch — concurrent with the icon fetches, so no added latency in the slow path — would eliminate the class of bug. Deliberately out of scope here.


piyalbasu (approved):
Tested and reviewed with Claude. Lgtm

Issue #1585, opened 2024-10-28
Integrate with sorobandomains


JakeUrban:
### What problem does your feature solve?

Freighter displays account and contract addresses when necessary, such as when reviewing a transaction to sign. However, these addresses are not friendly, and are usually cut off the screen.

 ![Image](https://github.com/user-attachments/assets/373f58cd-d008-4d81-8d06-49bfe7987d27)

### What would you like to see?

Replacing stellar accounts and contract addresses with their registered [soroban domain](https://sorobandomains.org/).

We would need to decide when to do this replacement. For example, should we replace addresses with domains for addresses passed as parameters to contract functions? We'll need input from @sdfcharles on this.

### What alternatives are there?

Continue doing what we're doing.

JakeUrban:
Duplicate of an internal spike issue. cc @Keeeeeeeks

PR #2949, opened 2026-08-10
feat: resolve Soroban Domains (*.xlm) to addresses

salazarsebas:
## Summary

Freighter currently only resolves Federation addresses to a friendly name;
Soroban Domains (`*.xlm`) fall back to the raw `G...`/`C...` address, which is
easy to mistype and gets truncated in the UI. This adds Soroban Domain
resolution end-to-end, following the same pattern already established for
Federation addresses.

- Detects `*.xlm` input and resolves it via the official
  `@creit-tech/sorobandomains-sdk`, using Freighter's own configured Soroban
  RPC only (never the SDK's default third-party RPC)
- Resolved domains persist alongside the address through `SendTo`,
  `AddressTile`, review, submit, and send-amount screens, matching existing
  Federation-address handling
- Contract/simulation errors are never surfaced to the user — only a generic
  translated message, with the real cause reported to Sentry

Relates to #1585 (closed in 2025 as a duplicate of an internal spike) —
opening this as a ready-to-review implementation in case that effort didn't
move forward. Happy to adjust scope against any existing internal design.

## Demo

https://youtube.com/shorts/pSmJFXXbRH8

## Test plan

- `yarn test:ci` passing (new coverage: `sorobanDomains` helper,
  `resolveSorobanDomain` module, `AddressTile`, `SendTo`, `useSendToData`)
- `yarn build:extension` type-checks and bundles clean
- `npx eslint` clean on all touched files
- Manual test: send flow resolves a live `*.xlm` domain end-to-end (see demo
  above)


aristidesstaffieri:
@salazarsebas thanks for putting this together, this looks like a solid start for this feature but we have concerns about the level of adoption that Soroban Domains has seen and frankly wether it is the right underlying mechanism for this kind of feature in Freighter. At this time we're going to choose to not merge this change set. This doesn't mean that we won't support this feature in the future but for now I'd like to turn this into an issue where we can engage @earrietadev and other team members to better understand the health of the project and the long term involvement from the maintainer, as well as consider other options for human friendly identifiers(like an internal contact book).

aristidesstaffieri:
> @earrietadev

I've started the issue - https://github.com/stellar/freighter/issues/2950
@salazarsebas it would be great to get input from you here as well.

xlmdomains:
there is also xlm.domains. we think .xlm should be taken to a top level domain. we are passionate about .xlm and think there is much work to be done before a wallet adopts .xlm. we are willing to work on .xlm as maintainers and would love to collaborate with others who are passionate about the true potential of .xlm

aristidesstaffieri:
> there is also xlm.domains. we think .xlm should be taken to a top level domain. we are passionate about .xlm and think there is much work to be done before a wallet adopts .xlm. we are willing to work on .xlm as maintainers and would love to collaborate with others who are passionate about the true potential of .xlm

@xlmdomains we would certainly consider all protocols available when planning this feature. If you'd like to join the conversation in #2950 we can continue to see where this feature fits into Freighter's future.

salazarsebas:
Makes sense, thanks for the honest take @aristidesstaffieri. I built this on the same pattern Freighter already uses for Federation addresses specifically so it wouldn't be a one-off, but I get that the real question here is bigger than implementation, it's about which naming mechanism the wallet should actually commit to long term. I'd rather this land as the right feature than a fast one, so I'll follow up with more thoughts in #2950. Happy to help with whichever direction the team lands on.

xlmdomains:
The frustrating part is that we had Soroban Domains for 2 years and did not get the process going for .xlm to become a Top Level Domain with SDF/SCF support, despite repeated requests from the community.

What is even more frustrating is that the 2026 ICANN window closed August 12 2026. And there won't be another application window until 2028 to 2030, which means even if the SDF committed to taking .xlm to a TLD today that it would not be able to be live with .xlm until 2030-2032.

NEAR and Unstoppable Domains managed to get in the 2026 cohort and wont even get live until 2028.

Another missed opportunity for Stellar Community. 

Imagine what we could have done with .xlm as a TLD using x402 in 2028...

PR #2952, opened 2026-08-11
Add SEP-53 message signing support for hardware wallets

SmolinPavel:
## What

Enables SEP-53 message signing for hardware wallets. Previously `SignMessage`
returned an "Unsupported signing method" screen whenever the active account was
a hardware wallet.

That guard was added in 5.23.0 (Aug 2024). It predates Ledger Stellar app
**v6.0.0**, which added a dedicated SEP-53 `SIGN_MESSAGE` instruction (APDU
`0x0C`, documented in app-stellar as "Sign the SEP-0053 message"). The pinned
`@ledgerhq/hw-app-str@7.7.4` already exposes `signMessage`, so no dependency
change is needed.

## How

The message path mirrors the existing Soroban auth-entry path — a new
`hardwareSignMessage` map in `hardwareConnect.ts`, an `isSignMessage` branch in
the `signWithHardwareWallet` thunk, and an `isSignMessage` prop on
`HardwareSign`.

The device applies the SEP-53 prefix and hashing itself and displays the message
for review, so it receives the raw UTF-8 bytes. The resulting signature is
identical to what `encodeSep53Message` + a local key produce, so the dApp-facing
response shape is unchanged. This is **not** blind signing — the message renders
as text on the device.

### Two supporting fixes the path needs

- **base64 transport.** The signature crosses `runtime.sendMessage` as base64 and
  is normalized back to a Buffer by a new `toSignatureBuffer` helper. A raw
  `Buffer` is JSON-serialized in transit, so `signedBlob.toString("base64")` in
  `submitBlob` would have produced `"[object Object]"`.
- **`signerAddress`.** `handleSignedHwPayload` now forwards the signing address,
  which SEP-53 consumers need in order to verify the signature they get back.
  The message branch verifies the returned signature against that key before
  reporting it — `connectToLedgerTransport` closes and reopens the connection on
  every call, so the key read by `getWalletPublicKey` is not otherwise provably
  the key that signed.

### Error handling

A Stellar app older than v6.0.0 now reports that it needs updating, rather than
surfacing an opaque transport error. The check is a `getAppConfiguration()`
preflight in `hardwareSignMessage`; an unparseable version defers to the device
rather than refusing. Deliberately *not* matched on the device's generic
`INS_NOT_SUPPORTED (0x6d00)` status word — `parseWalletError` is shared by every
Ledger flow, and that code also fires for e.g. a Soroban auth entry on an older
app, where both the reason and the version number would be wrong.

Second commit also fixes an adjacent problem this path exposed: `HardwareSign`
runs its first signing attempt automatically from a mount effect, but the error
block only rendered after the "Detect device" button had been clicked. A device
that failed immediately left the overlay reading "Connect device to computer"
with no explanation until the user clicked through and failed again. Rendering on
the error itself drops the flag and covers the automatic attempt, so the existing
transaction and auth-entry overlays report failures on first try too.

## Testing

Verified end to end on a **Ledger Flex running Stellar app 6.0.3**: the message
text renders on the device, approval returns a signature, and the dApp verifies
it against the signer address.

- 27 new unit tests covering the signing helper, the version and size gates, the
  error mapping, the thunk branch, the base64 normalization, the `signerAddress`
  passthrough, and a new `HardwareSign` suite for the account guard and the
  first-attempt error state. The thunk and component tests sign with real
  keypairs, so signature verification is exercised rather than stubbed.
- Full Jest suite: 1540 passing. The one failing suite, `testNodeCompat.js`, is
  pre-existing on `master` and needs `@stellar/freighter-api` built first.
- `tsc --noEmit -p extension/tsconfig.json` clean; ESLint clean on changed files.

## Notes for reviewers

- `INTEGRATING_HARDWARE_WALLET.MD` documented only `hardwareSign`; it now also
  covers `hardwareSignAuth` (already required) and `hardwareSignMessage`.
- Two now-unused locale keys were removed alongside the deleted warning screen.
- `getAppConfiguration()` also reports the device's `maxDataSize`, so an
  oversized message is rejected with an actionable error rather than the
  library's `RequestDataTooLarge` (0xB004) remap. Measured in UTF-8 bytes.

### Two pre-existing gaps this work surfaced, deliberately left alone

Both affect the hardware transaction and auth-entry paths equally, so fixing
either only for messages would make the behaviour inconsistent. Flagging rather
than guessing, since both touch areas you own:

1. **`signerAddress` is still `undefined` for hardware transactions and auth
   entries.** `docs/api-integration.md:145` tells integrators to "always verify
   `result.signerAddress === expectedAddress`", which means a dApp following that
   guidance rejects every hardware-wallet transaction. Closing it properly means
   verifying those signatures too (against `tx.hash()` and `hash(authPreimage)`
   respectively) before reporting a key — cheap, but it changes the wallet's
   busiest signing path, and I could only test message signing on a device.
2. **`signWithHardwareWallet` dispatches no metrics.** The signing events in
   `popup/metrics/access.ts` are registered on `signTransaction` / `signBlob` /
   `signEntry`, so hardware signing has never appeared in them. This wants one
   handler covering all three and an event shape you'd want to decide.


extension/src/popup/components/hardwareConnect/HardwareSign/index.tsx
```diff
@@ -174,18 +183,15 @@ export const HardwareSign = ({
           </div>
         </div>
         <div className="HardwareSign__bottom">
-          {isDetectBtnDirty && <WalletErrorBlock error={connectError} />}
+          {connectError && <WalletErrorBlock error={connectError} />}
```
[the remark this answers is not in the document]
SmolinPavel: Added in 04ca8f81. HardwareSign suite covering the first-attempt error state, the account-mismatch refusal, and the message-specific copy. Extended in f95b49c9 with the requested-vs-active case and an empty-message payload.

extension/src/popup/components/hardwareConnect/HardwareSign/index.tsx
```diff
@@ -106,7 +111,11 @@ export const HardwareSign = ({
         } else if (uuid) {
           // right now there are only two cases after signing,
           // submitting to network or handling in background script
-          await handleSignedHwPayload({ signedPayload: res.payload, uuid });
+          await handleSignedHwPayload({
+            signedPayload: res.payload,
+            signerAddress: publicKey,
```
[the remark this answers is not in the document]
SmolinPavel: Fixed in 04ca8f81, then corrected in f95b49c9. Two places this diverged from the suggestion:
- Scoped to the message flow. A transaction or auth entry signed by the wrong device is self-invalidating — the signature won't satisfy the source account — so a hard block there would change shared behavior without adding safety.
- Comparing against the active account wasn't enough. signFlowAccountSelector leaves the previous account active when the requested account isn't in the wallet, and accountNotFound only flags a missing active key, so an active-only check passed in exactly the case that matters. It now compares against accountToSign when the dApp names an account.

extension/src/popup/components/hardwareConnect/HardwareSign/index.tsx
```diff
@@ -32,13 +32,15 @@ import "./styles.scss";
 export const HardwareSign = ({
   walletType,
   isSignSorobanAuthorization,
+  isSignMessage,
```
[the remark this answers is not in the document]
SmolinPavel: Fixed in 04ca8f81. The instruction now follows the payload — "Review message on device" / "Review authorization on device" / "Review transaction on device". Fixing the ternary also corrects the Soroban authorization case, which showed the wrong copy before this PR.

SmolinPavel:
@JakeUrban could you please review this PR?

extension/src/popup/ducks/transactionSubmission.ts
```diff
@@ -232,9 +238,48 @@ export const signWithHardwareWallet = createAsyncThunk<
...
+        // This also pins the device's SEP-53 digest to encodeSep53Message: if
+        // they ever disagreed, signing fails here rather than returning a
+        // signature no verifier can check.
+        const isFromExpectedSigner = Keypair.fromPublicKey(publicKey).verify(
+          encodeSep53Message(transactionXDR),
+          signature,
+        );
+        if (!isFromExpectedSigner) {
+          return thunkApi.rejectWithValue({
+            errorMessage: MISMATCHED_HARDWARE_ACCOUNT_ERROR,
```
JakeUrban: `verify` returning false covers more than a wrong device, but every failure is reported as `MISMATCHED_HARDWARE_ACCOUNT_ERROR` — which the user sees as "The connected device does not match the selected account. Connect the device this account was added from and try again."

If the device's SEP-53 digest ever disagrees with `encodeSep53Message`, or the device returns a malformed signature, the correct device produces that message. Retrying with a different device is a dead end, and the real cause is hidden.

The comment above already names the two cases separately; worth splitting them at the error boundary too — keep the sentinel for the mismatch case, and return a distinct message for a signature that fails to verify from the expected key.

---
_Generated by [Claude Code](https://claude.ai/code)_

SmolinPavel: Good catch — fixed in 785b5326.

Split at the error boundary as you suggested. `MISMATCHED_HARDWARE_ACCOUNT_ERROR` now belongs solely to the pre-sign check in `HardwareSign`, which has actually read and compared two keys and can honestly tell the user to connect a different device. The post-sign verification returns a new `UNVERIFIED_SIGN_MESSAGE_ERROR`:

> The signature returned by your Ledger could not be verified. Disconnect any other devices and try again. If this keeps happening, update the Stellar app on your Ledger.

The two causes are genuinely indistinguishable from inside that `if`, so the copy asserts neither: retrying covers a device swapped between connections, updating covers an app whose SEP-53 digest disagrees with ours, and neither sends the user off to swap hardware that was never the problem.

Regression coverage: the "signature over a different message" case in `signWithHardwareWallet.test.ts` now asserts the new sentinel and explicitly asserts it is *not* the mismatch one, and a `parseWalletError` test pins the two to different strings so they can't silently converge later.

extension/src/popup/components/hardwareConnect/HardwareSign/index.tsx
```diff
@@ -81,6 +95,26 @@ export const HardwareSign = ({
...
+      //
+      // Compare against the requested account, not merely the active one:
+      // signFlowAccountSelector silently leaves the previous account active
+      // when the dApp names an account the wallet does not hold, so checking
+      // the active account alone would pass in exactly that case.
+      const expectedPublicKey = requestedPublicKey || activePublicKey;
+      if (
+        isSignMessage &&
+        expectedPublicKey &&
+        publicKey !== expectedPublicKey
```
aristidesstaffieri: The new guard compares the device key against `requestedPublicKey` with raw string equality, but `accountToSign` is passed through unnormalized — a dApp can legitimately supply a muxed (M...) address whose base account is exactly the account that's active and attached.

In that case `signFlowAccountSelector` finds no match (it also compares exact G-strings), the guard compares a G-address against the M-string, and signing dead-ends with `MISMATCHED_HARDWARE_ACCOUNT_ERROR` — telling the user to connect a different device while the correct one is already attached. Retrying via "Detect device" fails identically forever. The same request works for `signTransaction`, which normalizes muxed and federation forms via `decodeAccountToSign` before comparing.

`isSameAccount` in `extension/src/helpers/stellar.ts` already does the muxed→base normalization synchronously:

```ts
if (isSignMessage && expectedPublicKey && !isSameAccount(publicKey, expectedPublicKey)) {
  throw new Error(MISMATCHED_HARDWARE_ACCOUNT_ERROR);
}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

JakeUrban: We can update this so we're consistent with signTransaction, but a dapp passing a muxed or federated address is invalid isn't it @aristidesstaffieri ?

SmolinPavel: Traced it — the dead-end is real. `signFlowAccountSelector` compares exact strings (`popup/helpers/account.ts:379`), and `currentAccount` is pre-seeded from the active account so `accountNotFound` never fires. The user reaches the signing screen, hits the device error with the correct Ledger attached, and Detect device re-runs the same comparison forever.

One correction: `signTransaction` doesn't work either. `decodeAccountToSign()` runs at `SignTransaction/index.tsx:232`, but `useGetSignTxData(..., accountToSign, ...)` already ran at `:134`, and it mutates a render-local `let` nothing reads afterward. It gets the raw M-address too — SignMessage is just the only flow with a device guard, so it's the only one that fails loudly.

@JakeUrban muxed isn't supported end to end today: even for software accounts `signMessage` returns the base `G`, which fails the `signerAddress === expectedAddress` check we tell integrators to write. `isSameAccount` wouldn't change that — the guard would pass but the dApp still rejects. It also uses the sync `getBaseAccount`, so federation dead-ends identically.

So it's where it fails, not whether. Either `isSameAccount` (downgrades to a clean dApp-side rejection) or reject M/federation up front with copy naming the real problem. Happy to do either — preference?

aristidesstaffieri: hmm yeah I see now that passing anything but a G address is invalid, just not by signature. Making the hardware path consistent with the software path seems correct to me here.

aristidesstaffieri:
Screenshot 2026-08-13 at 9 07 02 AM
 @SmolinPavel it looks like this is pre-existing but since we are improving this API maybe we can fix this bug while we're here as well, it looks like a one-liner interpolation bug.

SmolinPavel:
Good catch — fixed in 1de84abf. It was seven call sites, not one: four in `HardwareSign`, two in `DeviceConnect`, one in `PluginWallet` (`Connect with {walletType}`). Re-keyed to `{{walletType}}` across the call sites and both bundles, carrying the Portuguese translations over.

Added a locale test rejecting any single-brace placeholder so the next one fails CI instead of shipping. It immediately found one more: `SSLWarningMessage` passes `values={{ url }}` against a `{url}` key, so the domain never renders in that warning either. That one's unrelated to hardware wallets so I allowlisted it rather than widen this PR — happy to pick it up separately if you want it.

gmelika:
Just wanted to add a +1 to this PR.  This is a much needed feature for us to allow users to sign in to their accounts with their hardware wallet to voltacircuit.com

extension/src/popup/components/hardwareConnect/HardwareSign/index.tsx
JakeUrban: @SmolinPavel let us know if its unclear how to proceed on this

SmolinPavel: Clear enough — done in e54c0dc2, taking @aristidesstaffieri's read that the hardware path should match the software path.

The guard now compares the device key against the active account only, and `requestedPublicKey` is gone (no remaining caller). That keeps what it was actually for — the attached device must be the account the popup says it is signing as, so a second Ledger can't sign while the UI names someone else — and drops the part that duplicated the `signerAddress` check the dApp already performs.

Whether the requested account was honoured is now decided exactly as it is for software keys: `signFlowAccountSelector` activates it when the wallet holds it, otherwise the active account stands and the true signer goes back for the dApp to verify. The muxed dead-end disappears with it, since an M-address no longer participates in any comparison.

Branch is merged up to latest master. 1548 tests passing.

aristidesstaffieri:
thanks @SmolinPavel this lgtm, I will go ahead and approve/merge this and we will include it in the upcoming Freighter release.

aristidesstaffieri (approved):

PR #2960, opened 2026-08-13
feat(balances): align the $0.00 Testnet balance UI with mobile + tweaks

CassioMG:
Aligns the `$0.00` Testnet balance UI with the mobile app ([freighter-mobile#943](https://github.com/stellar/freighter-mobile/pull/943)), plus a few other small polishes found along the way listed here and it also prunes and review stale translations from the codebase.

## Balance display

*UPDATED, see [PR comment at the bottom](https://github.com/stellar/freighter/pull/2960#issuecomment-5299485841). We will need to port some changes back to mobile as well to keep both in parity.

- `New Testnet behavior`: Wallets rows show a spinner while an account's total is still pending instead of the literal `...`, and settle on `$0.00` rather than an empty cell since we have no prices (or "real-world" value) on Testnet. The Home hero also defaults to `$0.00` for the same reason.

- `New Mainnet behavior`: Wallets rows show a spinner while an account's total is still pending instead of the literal `...`, and settle on its expected total USD balance or `--` if there was an error while fetching prices. The Home hero is also never hidden — it defaults to `$0.00` for **unfunded** accounts and defaults to `--` for **funded** accounts if there was an error while fetching prices.

Home **token** rows are deliberately unchanged: unpriced tokens keep showing the `--` dashes.

### Regular experience with quick load + "$0.00" labels on Testnet

https://github.com/user-attachments/assets/cf0b0ec0-76dd-49ee-a420-1f80d110c7fa

### Experience with long load times showing the row spinners

https://github.com/user-attachments/assets/5d4e0b83-b47c-4161-9792-095aa2185026

## "Imported" label (long standing bug on master)

Recovery pre-loads derivation indices 1..n and was storing them as `imported`, so every account but the first came back labelled "Imported". They come from the recovery phrase, exactly like the ones `addAccount` creates. The label now means what it says: brought in from a secret key.

Fixes new recoveries only — the flag is rehydrated from the keystore, so wallets recovered before this keep the old label until recovered again.

### Before fix (video from master branch)
 
https://github.com/user-attachments/assets/29a3befc-5e49-43a5-9d79-9b5bd61519e7

### After fix (video from this PR's branch)

https://github.com/user-attachments/assets/c1407ae7-aba5-4641-bd88-d5a148f99280

## Unfunded empty state

Test networks stacked two competing funding buttons. Friendbot now replaces "Add XLM" rather than sitting under it, in the same style, one size larger.

### Before fix

Screenshot 2026-08-12 at 17 39 12

### After fix

Screenshot 2026-08-12 at 17 36 37

## Home error state

The hero shows `$0.00` label when the fetch fails, for consistency with the wallets rows and with mobile.
The error notification now gets a horizontal gutter.
The account identicon also went blank here: the header was handed an empty public key once the fetch returned no data, while the name kept coming from Redux. It now falls back to Redux too, which also repairs the header's copy button and account-details link in this state.

### Before fix

Screenshot 2026-08-12 at 14 14 01

### After fix

Screenshot 2026-08-12 at 17 34 01

## Wallets error state (long standing bug on master)

The error branch returned the notification on its own, leaving no close button, no account actions and no Add wallet — an error you could not navigate out of. It is now scoped to the list region.

### Before fix

Screenshot 2026-08-12 at 14 19 55

### After fix

Screenshot 2026-08-12 at 17 34 07

## Fullscreen layout (long standing bug on master)

The layout only sized to its content, so the vertical borders stopped partway down the window. Also gives the Home notifications their horizontal gutter, and repairs an invalid `margin-top` shorthand that left the fullscreen notice glued to the edges.

### Before fix

Screenshot 2026-08-12 at 17 39 54

Screenshot 2026-08-12 at 17 42 05

### After fix

Screenshot 2026-08-12 at 18 04 15

Screenshot 2026-08-12 at 17 34 38

---

`yarn test:ci` 201 suites / 1543 tests, `yarn build:extension` clean. New coverage for `WalletRow`, `NotFundedMessage`, `formatFiatAmount` and the recovery flag.


extension/src/popup/components/account/WalletRow/index.tsx
```diff
@@ -34,9 +41,14 @@ export const WalletRow = ({
...
-  // Balance is its own cell now. While prices are still loading we show an
-  // ellipsis rather than an empty gap, matching the previous subtitle behavior.
-  const balanceLabel = accountValue || (isFetchingTokenPrices ? "..." : "");
+  // Balance is its own cell. Mirrors the mobile app: a spinner only while
+  // this account's total is genuinely still pending — totals arrive in
+  // batches, so early rows show a real value while later ones are unresolved.
+  // Once loading settles the total always shows, defaulting to $0.00 rather
+  // than an empty cell. A falsy check (not `== null`) is deliberate: the data
+  // hook writes "" for an account whose fetch threw, which is also "no total".
+  const isTotalLoading = !accountValue && isFetchingTokenPrices;
```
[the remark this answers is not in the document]
CassioMG: Declining — the mobile behavior described here is the opposite of what mobile does, so the change would introduce the divergence it's meant to fix.

[`AccountItemRow.tsx:62`](https://github.com/stellar/freighter-mobile/blob/main/src/components/screens/HomeScreen/AccountItemRow.tsx#L62):

```ts
// Covers both never-fetched (undefined) and failed (null) totals: a
// failed row is retried by the next cycle, and showing the spinner
// during the retry beats asserting a confident $0.00.
const isTotalLoading = fiatTotal == null && isLoadingFiatTotal;
```

`== null` is loose equality, so it is true for `null` as well as `undefined` — a failed entry spins too, and the comment above it says that is deliberate. The prop's own doc is explicit as well: "`null` means unavailable (fetch failed or fiat-less network). **Both** render a spinner while `isLoadingFiatTotal`".

`!accountValue && isFetchingTokenPrices` is the faithful equivalent here: falsy covers `""` and `undefined` exactly as `== null` covers `null` and `undefined`. A failed row spins while the cycle runs and then settles on `$0.00`, which is what mobile does.

So the test asserting the failed row keeps spinning is pinning the intended behavior rather than a bug.

On the second half of the comment: line 91 is the render site consuming this same predicate, not a separate occurrence.

CassioMG:
### Code review

CassioMG:

piyalbasu:
In the Wallets view, it looks like a particular wallet fails to fetch its USD, balance, we show `$0.00`. I think that can be a little misleading (and potentially a little scary) for users. It's not actually 0, it's just that we can't show the total value. I'd suggest we try to show `--` or some other indicator that there was a failure. I think it's fine to show `$0.00` on testnet

piyalbasu:
### New test file isn't colocated with its component

**Nit / suggestion** — **not a merge blocker.** Cosmetic only; no behavior impact, and it's a one-line file move.

**TL;DR:** One of the two test files this PR adds sits in a different place than the other. The wallet-row test lives next to the component it covers, but the not-funded-message test is placed a level above its component, in a new folder created just for it. Everything else in this part of the codebase keeps tests next to the code they test, so it'd be nice to have the two new files agree with each other.

---

<details>

**The inconsistency:** this PR adds two test files under `popup/components/account/` and places them differently.

Colocated with its component (matches convention):

https://github.com/stellar/freighter/blob/b2c0dbedc32e27fd0175241e7e4109495b985c99/extension/src/popup/components/account/WalletRow/__tests__/WalletRow.test.tsx#L1-L8

One level above its component, in a newly created `account/__tests__/` directory:

https://github.com/stellar/freighter/blob/b2c0dbedc32e27fd0175241e7e4109495b985c99/extension/src/popup/components/account/__tests__/NotFundedMessage.test.tsx#L1-L8

The component under test is at `popup/components/account/NotFundedMessage/index.tsx`, so the colocated path would be `account/NotFundedMessage/__tests__/NotFundedMessage.test.tsx`.

**Why this one is worth a nit rather than a shrug:**

1. It's the stated rule — `testing.md` line 5: *"Test files live in `__tests__/` directories alongside the source they test."*

   https://github.com/stellar/freighter/blob/b2c0dbedc32e27fd0175241e7e4109495b985c99/docs/skills/freighter-best-practices/references/testing.md#L1-L8

2. `account/__tests__/` **did not exist before this PR** and now contains exactly one file. Prior to this change, the only test anywhere under `popup/components/account/` was colocated:

   https://github.com/stellar/freighter/blob/b2c0dbedc32e27fd0175241e7e4109495b985c99/extension/src/popup/components/account/AccountHeader/__tests__/getNetworkDisplayName.test.ts#L1-L5

   So this isn't following an existing local pattern — it establishes a new non-colocated directory in a subtree that was previously 100% colocated.

3. The PR disagrees with itself: same parent directory, same PR, two different conventions.

**Pre-empting the obvious counter-argument:** `popup/components/__tests__/` holds 15 test files for components that live in their own subdirectories (`AssetDetail.test.tsx`, `CollectibleDetail.test.tsx`, etc.), so non-colocated tests clearly exist in the repo. But that's a different directory one level up, and it's legacy — it doesn't establish a precedent inside `account/`, where the count of non-colocated tests before this PR was zero.

**Suggested fix:**

```bash
mkdir -p extension/src/popup/components/account/NotFundedMessage/__tests__
git mv extension/src/popup/components/account/__tests__/NotFundedMessage.test.tsx \
       extension/src/popup/components/account/NotFundedMessage/__tests__/NotFundedMessage.test.tsx
```

That's the whole change — no other edits needed:

- **No import changes.** The file imports via absolute module paths (`popup/components/account/NotFundedMessage`, `popup/__testHelpers__`), resolved by `moduleDirectories: ["node_modules", "<rootDir>/extension/src", "<rootDir>/."]` in `jest.config.js`, so nothing is relative to the file's own location.
- **No jest config change.** `roots` already includes `./extension` and the default `testMatch` picks up `__tests__/` at any depth.

Entirely your call if you'd rather keep it as-is; flagging it because the two new files landing in the same PR with different conventions is the kind of thing that quietly becomes the next precedent.

</details>

CassioMG:
> In the Wallets view, it looks like a particular wallet fails to fetch its USD, balance, we show `$0.00`. I think that can be a little misleading (and potentially a little scary) for users. It's not actually 0, it's just that we can't show the total value. I'd suggest we try to show `--` or some other indicator that there was a failure. I think it's fine to show `$0.00` on testnet

@piyalbasu Yeah, thinking again about this I think it's best to display `--` when there is an error fetching the USD balance so the app is clear about it instead of displaying a misleading (and scary) `$0.00` to users on Mainnet, which is different than defaulting to `$0.00` on Testnet's Home screen where it's expected to have no price or real-world value. I'll adjust it to display `--` in case of error. FYI @aristidesstaffieri @minkyeongshin

CassioMG:
> ### New test file isn't colocated with its component
> **Nit / suggestion** — **not a merge blocker.** Cosmetic only; no behavior impact, and it's a one-line file move.
> 
> **TL;DR:** One of the two test files this PR adds sits in a different place than the other. The wallet-row test lives next to the component it covers, but the not-funded-message test is placed a level above its component, in a new folder created just for it. Everything else in this part of the codebase keeps tests next to the code they test, so it'd be nice to have the two new files agree with each other.
> 

Nice call, fixed in b50168fe

CassioMG:
@piyalbasu thanks for reviewing. I've updated how we handle "$0.00" vs "--" on Mainnet and Testnet depending on if there was an error fetching prices or not:

- `MAINNET behavior`:
   - displays `--` as an account total USD balance if the account is **funded** and there was an error while fetching prices
   - always displays `$0.00` for **unfunded** accounts

- `TESTNET behavior`: 
   - always displays `$0.00` for an account total USD balance regardless of prices erroring since we have no prices (or "real-world" value) on Testnet

I've centralized this total USD balance display logic on this [getTotalUsdLabel](https://github.com/stellar/freighter/pull/2960/changes#diff-ea1bcf3d2361cf8b10d94288078518493a84611c26f6daf8bcd71018d9337204R288) helper so it's easier to maintain and rationale about. I'll port the same on mobile.

Also while testing the price error scenarios I've found a bug that results in an infinite spinner on the "Swap to" screen (which also load prices) that I've fixed on this [other commit](https://github.com/stellar/freighter/pull/2960/changes/fbea7d1d47a3ded4ba28398acb78d887bb3ec0b9).

Below are screenshots so we visualize the different scenarios. FYI @aristidesstaffieri @minkyeongshin 

### Mainnet happy path (all priced)

Screenshot 2026-08-14 at 16 38 54

### Mainnet with prices erroring

Screenshot 2026-08-14 at 16 40 37

### Mainnet for unfunded account

Screenshot 2026-08-14 at 16 41 02

### Testnet in all cases
Always the same UI regardless of prices erroring since we use no prices there.

Screenshot 2026-08-14 at 16 40 04

extension/src/popup/helpers/__tests__/formatters.test.ts
```diff
@@ -151,3 +152,25 @@ describe("formatAmountPreserveCursor", () => {
...
 });
+
+describe("formatFiatAmount", () => {
+  it("formats a USD amount, flooring to two decimals", () => {
+    expect(formatFiatAmount("1149.239")).toBe("$1,149.23");
+    expect(formatFiatAmount("0.5")).toBe("$0.50");
+  });
+
+  // Totals for accounts with no price data at all (non-Mainnet networks, or a
+  // failed price fetch) render this rather than being hidden.
```
piyalbasu: ### Nit: stale comment on the `$0.00` fallback test

**Nit** — comment text only; no assertion or behavior is wrong. **Not a merge blocker.**

**TL;DR:** This comment still says a failed price fetch shows `$0.00`. That stopped being true in this PR — a total that can't be read now renders the `--` placeholder, and `$0.00` is reserved for totals that are genuinely zero. The equivalent note on the formatter itself was updated; this one was missed.

```suggestion
  // A total that is genuinely zero — a network that prices no tokens, or an
  // unfunded account — renders this rather than being hidden. A total that
  // exists but could not be read uses NO_FIAT_VALUE; see getTotalUsdLabel.
```

---

<details>

**Root cause:** the "failed price fetch" half of the parenthetical describes the pre-PR behavior. `getTotalUsdLabel` now routes that case to `NO_FIAT_VALUE`, and sends only the two genuine-zero branches (`!hasPriceFeed || !isFunded`) to `formatFiatAmount()`:

https://github.com/stellar/freighter/blob/7202893e0aa758b357c8f9d34ad54be312d9f990/extension/src/popup/helpers/balance.ts#L303-L320

The `formatFiatAmount` JSDoc already carries the corrected version of the same distinction, which is what makes this read as an oversight rather than a second opinion:

https://github.com/stellar/freighter/blob/7202893e0aa758b357c8f9d34ad54be312d9f990/extension/src/popup/helpers/formatters.ts#L172-L188

**Scope:** the assertions are correct as written — `formatFiatAmount("0")` and `formatFiatAmount()` both do return `$0.00`. `formatFiatAmount` itself has no notion of a failed fetch; the caller decides. So this is purely the comment reaching into `getTotalUsdLabel`'s territory and describing it wrongly.

</details>

piyalbasu (approved):

extension/src/popup/helpers/__tests__/formatters.test.ts
CassioMG: https://github.com/stellar/freighter/pull/2960/changes/497de83790671b84184733bcc2b6e4d6e2f83308

PR #2962, opened 2026-08-17
Remove the "Introducing Freighter Mobile" promo banner

CassioMG:
Removing the purple "Introducing Freighter Mobile" banner from the home screen — mobile launched a year ago, so the promo has run its course and no longer earns the space.

This is a full cleanup, not just hiding the UI:

- `MobileAppBanner` component + styles
- The `logo-freighter-shadow.png` asset (only used by the banner)
- `getMobileAppBannerDismissed` / `dismissMobileAppBanner` background handlers, their `SERVICE_TYPES` entries, `@shared/api/internal` wrappers, and message types
- The `mobileAppBannerDismissed` local-storage key
- The banner's `en` / `pt` locale strings (regenerated via `yarn build:extension:translations`)

No behavior change beyond the banner disappearing.

This PR is removing the purple banner below:
Screenshot 2026-08-17 at 11 32 27 AM (1)


piyalbasu (approved):
This LGTM. Just want to call out that `MOBILE_APP_BANNER_DISMISSED` is now an orphaned value in the user's storage. Not a merge blocker as we have plenty of space for now, but at some point we may need to run a migration to remove orphaned values

PR #2968, opened 2026-08-18
feat(send): verify assembled collectible transfer matches intent before signing

JiahuiWho:
### What

The collectible send flow builds a Soroban transfer with no authorization entries and delegates simulation and assembly to the backend. The authorization entries on the returned transaction are therefore derived from simulating the target contract, not built by the wallet.

The wallet is the only party that knows the call it intended, so it should confirm the assembled transaction still matches that intent before signing, rather than adopting the returned bytes verbatim.

###  Why

Simulation is used to discover the fee and authorization a transaction needs, but the wallet should not treat the simulated result as the definition of what it signs. Re-deriving and comparing against the locally built call keeps the wallet the authority on its own transactions, and holds even if the assembling
service returns something unexpected.

### Known Limitation 

To fix this we can't just copy paste auth-entry guard onto `/simulate-tx` in the freighter-backend, due to Soroswap flow, which also calls this endpoint and legitimately produces sub-invocations.

### Testing
- `verifyPreparedTransaction.test.ts` covers: honest flat transfer accepted, empty auth accepted, sub-invocation under the root rejected, tampered host  function rejected.
- `tsc --noEmit` clean; full jest suite green.

### DIFF
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -29,6 +29,7 @@ import {
   getSymbol,
   transfer,
 } from "@shared/helpers/soroban/token";
+import { verifyFlatTransferPreparedTransaction } from "@shared/helpers/soroban/verifyPreparedTransaction";
 import {
   getAssetFromCanonical,
   getCanonicalFromAsset,
@@ -2211,6 +2212,17 @@ export const simulateSendCollectible = async (args: {
       simulationResponse: SorobanRpc.Api.SimulateTransactionSuccessResponse;
     };
   };
+
+  // The backend derives this transaction's authorization entries from simulating
+  // the target contract, not from what the wallet built. Confirm the assembled
+  // transaction still matches the flat transfer we intended before it can be
+  // signed, so no additional authorization can ride along on the user's
+  // source-account signature.
+  verifyFlatTransferPreparedTransaction({
+    builtTransaction: transaction,
+    preparedTransactionXdr: simulationResponse.response.preparedTransaction,
+    networkPassphrase: networkDetails.networkPassphrase,
+  });
 
   return simulationResponse;
 };
--- a/@shared/api/internal.ts
+++ b/@shared/api/internal.ts
@@ -2213,16 +2213,20 @@ export const simulateSendCollectible = async (args: {
     };
   };
 
-  // The backend derives this transaction's authorization entries from simulating
-  // the target contract, not from what the wallet built. Confirm the assembled
-  // transaction still matches the flat transfer we intended before it can be
-  // signed, so no additional authorization can ride along on the user's
-  // source-account signature.
-  verifyFlatTransferPreparedTransaction({
-    builtTransaction: transaction,
-    preparedTransactionXdr: simulationResponse.response.preparedTransaction,
-    networkPassphrase: networkDetails.networkPassphrase,
-  });
+  // The backend derives this transaction's authorization entries, fee, and
+  // resource data from simulating the target contract, not from what the wallet
+  // built. Confirm the assembled transaction still matches the flat transfer we
+  // intended before it can be signed, so no additional authorization or fee can
+  // ride along on the user's source-account signature.
+  if (simulationResponse.ok) {
+    verifyFlatTransferPreparedTransaction({
+      builtTransaction: transaction,
+      preparedTransactionXdr: simulationResponse.response.preparedTransaction,
+      networkPassphrase: networkDetails.networkPassphrase,
+      maxResourceFee:
+        simulationResponse.response.simulationResponse.minResourceFee,
+    });
+  }
 
   return simulationResponse;
 };
--- a/@shared/helpers/soroban/__tests__/verifyPreparedTransaction.test.ts
+++ b/@shared/helpers/soroban/__tests__/verifyPreparedTransaction.test.ts
@@ -65,11 +65,14 @@ const authorizedInvocation = (
   });
 
 // Rebuild the transaction the way an assembling backend does: reuse the invoked
-// host function verbatim and attach the given auth entries.
+// host function verbatim (or tamper with it), set a fee, and attach auth entries.
 const buildPreparedXdr = (
   built: Transaction,
   auth: xdr.SorobanAuthorizationEntry[],
-  { tamperFunc = false }: { tamperFunc?: boolean } = {},
+  {
+    tamperFunc = false,
+    fee = BASE_FEE,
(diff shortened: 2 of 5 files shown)


CassioMG:
### Missing simulated resource fee collapses the fee ceiling to the built fee

**Possible functional regression** — not a definite merge blocker, but worth confirming against real backend responses before merge, since the failure mode is a hard-failed send rather than a degraded one. The same issue exists in the mobile counterpart of this change, and the shared root cause sits inside the helper, so one fix can cover both.

**TL;DR:** The helper quietly treats an absent simulated resource fee as a resource fee of zero, which sets the fee ceiling at exactly the fee the wallet originally built. Every genuinely assembled Soroban transaction has a fee above that by definition — raising it to cover resource cost is the whole point of simulation — so a missing resource fee turns an ordinary collectible send into a verification failure instead of a send that merely cannot show a fee breakdown. The call site here looks immune because the field is typed as required, but that type comes from an unchecked cast rather than a runtime check, so nothing actually guarantees the field is present.

**Steps to reproduce:**
1. Make the simulation endpoint return a valid `preparedTransaction` but omit the resource fee from the simulation response.
2. Start a collectible send and let the flow reach simulation.
3. The send fails with "fee exceeds the built fee plus the simulated resource fee" rather than proceeding without a fee breakdown.

---

<details>

**Root cause:** the helper coerces a missing value to zero rather than distinguishing "no resource fee" from "resource fee unknown":

https://github.com/stellar/freighter/blob/d0caec552ca338a490d89440de20c7b61a5d2c92/@shared/helpers/soroban/verifyPreparedTransaction.ts#L131-L145

With `maxResourceFee` absent, `resourceFee` is `0`, so `maxFee === builtTransaction.fee`. Since the SDK's `assembleTransaction` sets `fee = builtFee + resourceFee`, any invocation with non-zero resource cost satisfies `prepared.fee > maxFee` and trips the throw.

**Why the type does not protect this call site:** the value is passed with no fallback, which reads as safe because `SorobanRpc.Api.SimulateTransactionSuccessResponse.minResourceFee` is a required `string`:

https://github.com/stellar/freighter/blob/d0caec552ca338a490d89440de20c7b61a5d2c92/@shared/api/internal.ts#L2220-L2229

But that type is imposed by an unchecked `as` cast on the `simulateTransaction` result a few lines above, not validated at runtime:

https://github.com/stellar/freighter/blob/d0caec552ca338a490d89440de20c7b61a5d2c92/@shared/api/internal.ts#L2204-L2214

So `undefined` can reach the helper, where `maxResourceFee || "0"` absorbs it silently. The typechecker will not flag it and there is no `??` at the call site to notice during review.

**Cross-platform note:** the mobile port of this change ([freighter-mobile#984](https://github.com/stellar/freighter-mobile/pull/984)) hits the identical failure mode from the opposite direction — there the field is genuinely declared optional (`minResourceFee?: string`) and the call site writes `?? "0"` explicitly. Mobile's repo history is also relevant: commit `372f63bb` there deliberately changed handling of a missing resource fee to mean "don't display a fee" rather than "the fee is zero", which is the convention this ceiling inverts. Because the coercion lives in the shared helper on both sides, fixing it in the helper fixes both platforms.

1. *Cheap guard:* when the resource fee is absent, skip the fee-ceiling check and document why — the other four checks (host function, envelope fields, auth credentials, sub-invocations) are the ones carrying the security value and remain intact.
2. *Explicit failure:* keep failing closed but with a distinct message such as "resource fee unknown; cannot verify fee ceiling", so the cause is diagnosable instead of presenting as an inflated-fee attack.
3. *Root cause:* type the parameter `maxResourceFee?: string`, drop the internal `|| "0"`, and make each caller decide explicitly. That removes the ambiguity between a legitimately zero resource fee and an absent one, and forces the question at both call sites rather than hiding it behind a cast.

</details>

CassioMG:
### The built transaction's operation count is never validated

**Suggestion** — not a merge blocker. Latent today rather than exploitable, since the current build path produces exactly one operation.

**TL;DR:** The check confirms the assembled transaction has exactly one operation, but never confirms the same of the transaction the wallet built — it just reads the first operation and compares that. So the equality being enforced is "your first operation matches my first operation", not "your operation set matches mine". If a built transaction ever carried more than one operation, a backend could return a transaction that kept the first and dropped the rest, and this would accept it. The mobile port of this change validates both sides symmetrically, so the two implementations differ in strictness here.

---

<details>

**Root cause:** the operation-count guard is applied only to `prepared`:

https://github.com/stellar/freighter/blob/d0caec552ca338a490d89440de20c7b61a5d2c92/@shared/helpers/soroban/verifyPreparedTransaction.ts#L68-L85

`builtTransaction.operations[0]` is read directly with no `builtTransaction.operations.length !== 1` check. The subsequent `builtOp.type !== "invokeHostFunction"` test guards the operation *type* but not the *count*, so extra built operations are invisible to the verification rather than rejected by it.

Note that the missing fee-bump check on `builtTransaction` is genuinely unnecessary and should not be "fixed" — `builtTransaction: Transaction` is compile-time guaranteed because it is the wallet's own in-memory object rather than parsed XDR. Only the count is the gap.

**Why it is currently latent:** `transfer()` in `@shared/helpers/soroban/token` builds a single `invokeHostFunction` operation, so `operations.length` is always 1 on the path that reaches this helper today. The concern is that the invariant is enforced nowhere near the code that depends on it, so a future caller passing a multi-operation transaction would silently weaken the guarantee instead of tripping an assertion.

**Divergence from the mobile port:** [freighter-mobile#984](https://github.com/stellar/freighter-mobile/pull/984) factors this into a `requireInvokeHostFunctionOp` helper and applies it to built and prepared alike, so a multi-operation built transaction is rejected there. Every other check in the two implementations is identical in both logic and error strings, which makes this the one real semantic difference between them — worth closing so the two do not drift.

**Suggested fix:** apply the same count-and-type guard to both transactions, mirroring the mobile structure:

```ts
const requireSingleInvokeHostFunctionOp = (tx: Transaction) => {
  if (tx.operations.length !== 1) {
    throw new PreparedTransactionMismatchError(
      `expected exactly 1 operation, got ${tx.operations.length}`,
    );
  }
  const op = tx.operations[0];
  if (op.type !== "invokeHostFunction") {
    throw new PreparedTransactionMismatchError(
      "operation is not invokeHostFunction",
    );
  }
  return op;
};
```

then call it for `builtTransaction` and `prepared` and drop the separate type check. This also narrows `builtOp` to the `invokeHostFunction` variant in one place, which the later `builtOp.func.invokeContract()` access depends on.

</details>

PR #2974, opened 2026-08-20
feat(src): Earn deposit flow

aristidesstaffieri:
## Summary

Adds an Earn deposit flow to the extension. From a new Earn tile on the Home action row, the user picks a supported token, sees the Blend Fixed pool's supply APY, enters an amount, reviews projected earnings, and submits a `submit` invocation against the pool contract. Gated to networks with an allowlisted pool — pubnet and testnet only.

Deposits use Blend `request_type` 2 (`SupplyCollateral`), not 0, so the position lands in `collateral_tokens` and the user's borrowing power stays open. Review reads before/after from `total_tokens` for that reason.

**Scope: deposit only.** No Positions tab and no withdraw — the Home tile is the only entry point and Home's tabs stay Tokens | Collectibles. If the account doesn't hold the selected token, the flow offers buy / swap / transfer remedies inline, including a swap-within-earn branch that reuses the Swap views inside Earn's sheet stack. Depends on freighter-backend-v2's `/protocols/blend/*` endpoints.

## What's in this PR

- `@shared/helpers/soroban/blend.ts` — request/`submit` op encoding. UDT `ScMap`s are hand-built because `nativeToScVal` sorts keys with `localeCompare`, which is not byte order.
- `@shared/api/{helpers,types}/blend.ts` — client for earn-options, pools, and supplied-tokens. `null` USD/APY means unavailable and is never coalesced to `0`.
- `@shared/constants/blend.ts` — pool allowlist per network plus `isEarnSupportedNetwork`, which gates the tile and the route.
- `popup/views/Earn/` + `popup/constants/earn.ts` — step machine modelled on Send (all visited steps stay mounted). Sheets — pool details, insufficient-balance, network fee, review, swap — are owned by their host step, not steps themselves.
- `popup/components/earn/` — the five screens and their sheets, CTA-state and projected-earnings helpers, pool-stat formatting, asset-icon resolution.
- `popup/helpers/blendDeposit.ts` — build + simulate. Auth comes back as `sorobanCredentialsSourceAccount`, so no `authorizeEntry` round-trip is needed, including on hardware wallets.
- `popup/components/{amount,swap}/` — extracted `percentageAmount` helpers and pinned the Swap footer/receive side so the reused views fit a fixed-height sheet. No behavior change for Send or standalone Swap.
- `background/messageListener/handlers/{get,dismiss}EarnIntro.ts` — persists the one-time intro interstitial.
- `popup/metrics/earn.ts` + `constants/metricsNames.ts` — funnel instrumentation. No amounts or fiat on any earn event; asset codes, `pool_id`, and `apy` only.
- `locales/{en,pt}/translation.json` + `translationParity.test.ts` — new strings and a test that fails on missing or empty parity.

## Notes for the reviewer

- XLM is not held back from a deposit. A `submit` resource fee is ~546k stroops (~5,000× the inclusion fee), so a naive Max on XLM simulates into insufficient balance; the amount screen surfaces that as a fee shortfall rather than silently reserving.
- Pool description prose is keyed by pool ID, not generic — the Fixed pool's admin key is verifiably burned, which is what the copy asserts.

## Test plan

- [x] Unit tests pass — `yarn jest blend earn Earn percentageAmount SwapAmount.layout translationParity metricsNames` (14 suites, 139 tests)
- [x] Manual deposit against mainnet Fixed pool `CAJJZSGM…BXBD`
- [x] CI green
- [x] Manual pass on a hardware wallet

## Followups
- Positions tab / withdraw.

## Example Flow

https://github.com/user-attachments/assets/8c286617-d9f1-4836-b0e8-6475318232d6

https://github.com/user-attachments/assets/80b2a3bd-5111-401a-bc55-940e919c44f3


extension/src/popup/components/earn/EarnAmount/index.tsx
```diff
@@ -0,0 +1,477 @@
...
+      // deposit that leaves nothing for it is caught here rather than by holding
+      // a guessed buffer back from the balance. The submission would otherwise
+      // fail with txINSUFFICIENT_BALANCE after the user had already signed.
+      const shortfall = isXlm
+        ? getXlmFeeShortfall({
+            spendableXlm,
+            amount: enteredAmount.toFixed(),
+            resourceFee: simulation.resourceFee || "0",
+          })
+        : "0";
```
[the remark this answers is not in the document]
aristidesstaffieri: path for network fee warning sheet was added for these cases in 0e00205b68d536612ae801aa8b197968ea4c5bd9

extension/src/popup/components/earn/EarnAmount/index.tsx
```diff
@@ -0,0 +1,477 @@
...
+
+      const simulation = await simulate({
+        publicKey: data.publicKey,
+        assetId: selectedAssetId,
+        amount,
+        decimals,
+        networkDetails: data.networkDetails,
+        transactionFee: recommendedFee,
+        transactionTimeout,
+      });
```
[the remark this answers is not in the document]
aristidesstaffieri: they are both now awaited in 58bc03e768e14a6d9957100e455057a94b737204

extension/src/popup/components/earn/EarnSubmit/hooks/useSubmitEarnTxData.tsx
```diff
@@ -0,0 +1,146 @@
...
+        }),
+      );
+
+      if (submitFreighterSorobanTransaction.fulfilled.match(submitResp)) {
+        trackEarnDepositCompleted({
+          assetCode,
+          poolId,
+          apy,
+          viaSwap,
+        });
```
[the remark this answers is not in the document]
aristidesstaffieri: Closing the tab doesn't actual abort the transaction but the event was mis-representing what happens here so the metrics are updated to better represent this case in c8415a4a74f417457aaf0f617cbe0240480f004a

extension/src/popup/Router.tsx
```diff
@@ -275,6 +276,8 @@ export const Router = () => (
         ></Route>
         <Route path={ROUTES.swap} element={<Swap />}></Route>
         <Route path={`${ROUTES.swap}/*`} element={<Swap />}></Route>
+        <Route path={ROUTES.earn} element={<Earn />}></Route>
+        <Route path={`${ROUTES.earn}/*`} element={<Earn />}></Route>
```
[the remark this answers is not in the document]
aristidesstaffieri: This is acceptable, a user that manually bypasses the navigation gates can fail later downstream.

extension/src/popup/components/earn/PoolDetailsSheet/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+            testId="earn-pool-net-apy"
+          />
+        </div>
+
+        <div className="PoolDetailsSheet__group">
+          <StatRow
+            label={t("Accepted tokens")}
+            value={
+              <div className="PoolDetailsSheet__tokens">
+                {pool.reserves.map((reserve) => {
```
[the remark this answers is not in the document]
aristidesstaffieri: there are now filtered in c2bbe6178647553a50ae9fc1ed8dcc01a541e9ff

extension/src/popup/components/earn/EarnAmount/index.tsx
```diff
@@ -0,0 +1,477 @@
...
+      setIsReviewOpen(true);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      // A simulation failure is the deposit's own pre-flight rejection — the
+      // same analytical unit as `payment.simulation_failed`, and never reaching
+      // the network means it has no result codes to key on, only this message.
+      trackEarnSimulationFailed({
+        assetCode: selected?.code || "",
+        reasonCode: scrubStrKeys(message) || "unknown",
+      });
```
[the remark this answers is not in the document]
aristidesstaffieri: Without the address, this level of attribution is acceptable.

extension/src/popup/components/earn/EarnIntro/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * whole viewport, which a self-measuring SlideupModal cannot give it.
+ *
+ * The artwork slot is a placeholder pending the final asset from design.
+ */
+export const EarnIntro = ({ onStart, onClose }: EarnIntroProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className="EarnIntro" data-testid="earn-intro">
+      <div className="EarnIntro__art">
```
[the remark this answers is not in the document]
aristidesstaffieri: Accepted as pending

extension/src/popup/helpers/blendDeposit.ts
```diff
@@ -0,0 +1,108 @@
...
+    networkDetails.networkPassphrase,
+  );
+  const account = await server.loadAccount(publicKey);
+
+  const request = buildBlendRequestScVal({
+    assetId,
+    // Blend takes the amount in the asset's smallest unit. toFixed(0) because
+    // an i128 cannot carry a fraction, and exponential notation would not parse.
+    amount: parseTokenAmount(amount, decimals).toFixed(0),
+    requestType: BlendRequestType.SupplyCollateral,
```
CassioMG, line 73: I think it'd be worth adding a comment here explaining why we're using `BlendRequestType.SupplyCollateral` instead of `BlendRequestType.Supply` for clarity (like the one we have on PR description)

extension/src/popup/components/earn/EarnAmount/hooks/useSimulateEarnDeposit.tsx
```diff
@@ -0,0 +1,93 @@
...
+        CLASSIC_ASSET_DECIMALS,
+      );
+
+      // Scanned on the PREPARED transaction — the thing the user actually
+      // signs — not the pre-assembly build.
+      const scanResult = await scanTx(
+        preparedTransaction,
+        // Internal flows have no originating dApp; the pool contract is the
+        // subject, so attribute the scan to it.
+        params.assetId,
```
CassioMG: just doing a sanity check here as we are passing a contract id as the `url` param of `scanTx`, is this safe to do so?

CassioMG:
### `earn.max_amount_selected` fires for every percentage shortcut, not just Max

**Suggestion (analytics correctness)** — wrong-by-name event data from the moment the flow ships; cheap to fix now, awkward to fix once dashboards are built on it. **Not a merge blocker.**

**TL;DR:** All four percentage shortcuts on the deposit amount screen — 25%, 50%, 75% and Max — emit the same event named "max amount selected". The event does carry a `percent` property, so the underlying data is recoverable, but anything keyed on the event name alone (dashboards, funnels, alerting) will count a 25% tap as Max usage. The Send flow shipped this exact behaviour, got called out for it in review, and was subsequently fixed to emit only on the 100% tap — this is the same pattern going in the other direction.

---

<details>

**Root cause:** the handler emits unconditionally, and the tracker hard-codes the "max" metric name regardless of the `percent` it receives.

The call site fires for any `pct`:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnAmount/index.tsx#L408-L416

The tracker always emits `METRIC_NAMES.earnMaxAmountSelected`, with the percentage relegated to a property:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/metrics/earn.ts#L127-L139

And the constant it resolves to is explicitly named for Max:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/constants/metricsNames.ts#L57-L59

**Prior art in this repo:** `SendAmount` had the identical issue. Review feedback on #2764 said the metric "will make analytics indistinguishable from actual 'set max' usage. Consider only emitting this metric for the 100% case, or introduce a separate metric (and include the percentage value) for the partial buttons." It now gates on master:

https://github.com/stellar/freighter/blob/2577532ce50d812befe09323561441d0505dd9f4/extension/src/popup/components/send/SendAmount/index.tsx#L569-L574

The Earn implementation took the second half of that suggestion (the `percent` property) without the first (a name that matches what happened), so the ambiguity the reviewer flagged is reintroduced at the name level.

1. *Match Send:* gate the emit on `percent === 100` inside `trackEarnPercentAmountSelected`, leaving partial taps untracked. Smallest diff, consistent with the sibling flow, loses partial-shortcut usage data.
2. *Rename to match behaviour:* rename the constant to `earn.percent_amount_selected` and keep emitting for all four. The `percent` property then fully disambiguates and you retain the richer funnel data. Requires coordinating the rename with whoever consumes the metrics, since the string is the wire format.
3. *Two metrics:* emit `earn.percent_amount_selected` for 25/50/75 and `earn.max_amount_selected` for 100, matching Send's name for the Max case while still capturing partials.

Option 2 or 3 preserves the most signal; option 1 is the safest if Earn's funnel is meant to mirror Send's exactly. Note the existing `reports Max as percent 100` test only exercises `percent: 100`, so it would keep passing under option 1 and would need updating only if the constant is renamed.

</details>

CassioMG:
### Earn submit hook reports `status: "success"` when the submission fails

**Bug (latent)** — the hook's own state is wrong on every failed deposit, but a second check in the consuming component currently keeps the wrong value off the screen. **Not a merge blocker.**

**TL;DR:** When the on-chain submission is rejected, the deposit hook records the attempt as a success anyway — it tracks the failure metric, then falls through and reports success regardless. Nothing is visibly broken today, because the screen decides what to render by also consulting the Redux submission status, which is correctly set to error. The risk is that the hook's own signal is wrong: anyone who later trusts it directly — a refactor, a new consumer, a test exercising the hook in isolation — silently gets "the deposit succeeded" for a deposit that failed.

---

<details>

**Root cause:** the submit-failure branch is missing the early return its sibling branch has. It emits `earn.deposit_failed` and then falls through to the shared success tail.

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnSubmit/hooks/useSubmitEarnTxData.tsx#L132-L140

Control flow rejoins here and dispatches unconditionally, so `RequestState.SUCCESS` is reached on both paths:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnSubmit/hooks/useSubmitEarnTxData.tsx#L164-L168

Compare the sign-failure branch a few lines above, which is the shape the submit branch should have — track, `FETCH_DATA_ERROR`, `return`:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnSubmit/hooks/useSubmitEarnTxData.tsx#L104-L120

**Why nothing is visibly broken today:** `EarnSubmit` does not trust the hook alone. It ands the hook's state together with the Redux `submitStatus`, which the `submitFreighterSorobanTransaction` thunk sets to `ERROR` independently, so the success view stays hidden:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnSubmit/index.tsx#L74-L79

That also means the `earn_success` screen-view metric, which is gated on the same `isSuccess`, does not fire spuriously. So this is a correctness defect in the hook, not a user-facing one — worth fixing because the guard that saves it lives in a different file and is not obviously load-bearing.

**Deterministic repro:** drive the hook directly rather than trying to fail a real submission — make `submitFreighterSorobanTransaction` reject (mock the thunk, or point at an RPC that rejects the envelope), call `fetchData`, and assert on the hook's returned payload and reducer state. Current behaviour: returns `{ status: "success" }` and `state === RequestState.SUCCESS`. Expected: an error state, mirroring the sign-failure path. The existing `EarnSubmit.test.tsx` covers the failure path only through the emitted metric, which is why this passes CI.

**Suggested fix:** mirror the sibling branch exactly — inside the `if (!submitFreighterSorobanTransaction.fulfilled.match(submitResp))` block, after `trackDepositFailed(...)`, dispatch `FETCH_DATA_ERROR` with the rejection payload and return it. The `else` wrapper around the success path then becomes redundant and can be flattened, which also removes the possibility of this class of fall-through recurring.

</details>

CassioMG:
> PR Preview build is ready: https://github.com/stellar/freighter/releases/tag/untagged-6a7858a6668267bb229c Backend: V1 prod + V2 beta (no sandbox configured for @aristidesstaffieri). SDF collaborators only — install instructions in the release description.

@aristidesstaffieri @piyalbasu it seems I can't test the Earn Deposit feature from the PR preview yet because the sandbox is not configured (and I believe the sandbox-per-dev-config is still WIP), is that right?

So in case I'd like to test the full Earn Deposit flow I'd need to checkout the backend-v2 [main-blend](https://github.com/stellar/freighter-backend-v2/tree/main-blend) branch and run/deploy it locally, is that the right approach?

Thanks!

Screenshot 2026-08-24 at 18 22 02

CassioMG:
### `scanTx` result is computed but never read

**Nit / question** — no security impact, **not a merge blocker**.

**TL;DR:** The Earn deposit flow runs a Blockaid scan on the prepared transaction
and stores the verdict on the simulation payload, but nothing downstream ever
reads it — the review sheet doesn't surface it and the Confirm button doesn't
consider it. Flagging it in case wiring up the warning UI was intended and got
left behind; if it was deliberate (the scan and simulate endpoints share a host,
so the verdict adds little here, and the deposit target is a pinned constant
rather than anything caller-supplied), then the round-trip itself is probably
worth dropping.

---

<details>

**What I found:** `useSimulateEarnDeposit` awaits `scanTx(preparedTransaction, …)`
and writes the result onto the `SimulateTxData` payload:

https://github.com/stellar/freighter/blob/1a2587d216f02938b77ec03abc3d8d6e36a6cf30/extension/src/popup/components/earn/EarnAmount/hooks/useSimulateEarnDeposit.tsx#L56-L77

A grep for `scanResult` across `components/earn/` and `views/Earn/` returns only
that hook plus two test fixtures. `EarnReview` destructures `simulationState.data`
for `transactionXdr` only, and `FeesPane` reads just `inclusionFee` / `resourceFee`
/ `state`, so the verdict is dead on arrival.

**For contrast**, `ReviewTransaction` reads the same field and turns a
`Malicious` / `Suspicious` / `UNABLE_TO_SCAN` verdict into a banner plus a
demoted "Confirm anyway" affordance via `getTransactionSecurityLevel` →
`mergeSecurityLevels` → `shouldShowTxWarning` → `ActionButtons`. Earn is the
first signing surface that computes the verdict and drops it.

**Two ways to close it:**
1. *Wire it up:* consume `simulationState.data?.scanResult` in `EarnReview` the
   way `ReviewTransaction` does — render `BlockAidScanLabel` /
   `BlockAidScanExpanded` and require the explicit acknowledgement before
   `onConfirmTx` fires. Keeps Earn consistent with every other signing path.
2. *Drop it:* if the scan was never meant to gate this flow, remove the call so
   the flow isn't paying for a round-trip whose answer is discarded.

</details>

extension/src/popup/components/account/AccountHeader/index.tsx
```diff
@@ -415,6 +416,18 @@ export const AccountHeader = ({
                     </Text>
                   </div>
                 </NavLink>
+                {isEarnSupportedNetwork(networkDetails) ? (
```
CassioMG: Should we also put this button behind a feature flag?

extension/src/popup/views/Swap/index.tsx
```diff
@@ -77,30 +75,9 @@ export const Swap = () => {
...
-        .code,
-      result_code: getQuoteExpiredOperationCodes(submission.error).join(", "),
-    });
-    // Clear only the ERROR status (keep the transaction data + the
-    // isSwapQuoteExpired flag, which drives the amount-screen notification).
-    dispatch(resetSubmitStatus());
-    setActiveStep(STEPS.AMOUNT);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isQuoteExpiredAtSubmit]);
+  const { isQuoteExpiredAtSubmit } = useSwapSubmitQuoteExpiry({
```
CassioMG, line 78: nice, thanks for extracting this into a hook

aristidesstaffieri:
> > PR Preview build is ready: https://github.com/stellar/freighter/releases/tag/untagged-6a7858a6668267bb229c Backend: V1 prod + V2 beta (no sandbox configured for @aristidesstaffieri). SDF collaborators only — install instructions in the release description.
> 
> @aristidesstaffieri @piyalbasu it seems I can't test the Earn Deposit feature from the PR preview yet because the sandbox is not configured (and I believe the sandbox-per-dev-config is still WIP), is that right?
> 
> So in case I'd like to test the full Earn Deposit flow I'd need to checkout the backend-v2 [main-blend](https://github.com/stellar/freighter-backend-v2/tree/main-blend) branch and run/deploy it locally, is that the right approach?
> 
> Thanks!
> 
> Screenshot 2026-08-24 at 18 22 02

@CassioMG correct, the earn APIs are still in our dev cluster so what I have been doing is just running a port forward to the backend v2 instance in dev from my machine.

extension/src/popup/components/account/AccountHeader/index.tsx
aristidesstaffieri: I would lean towards no, my thoughts are that once this feature is live we wont want to remove functionality that users have already had access to. I think in the event of a protocol problem we would want to use our notification system to warn users. wdyt?

aristidesstaffieri:
> ### `scanTx` result is computed but never read
> **Nit / question** — no security impact, **not a merge blocker**.
> 
> **TL;DR:** The Earn deposit flow runs a Blockaid scan on the prepared transaction and stores the verdict on the simulation payload, but nothing downstream ever reads it — the review sheet doesn't surface it and the Confirm button doesn't consider it. Flagging it in case wiring up the warning UI was intended and got left behind; if it was deliberate (the scan and simulate endpoints share a host, so the verdict adds little here, and the deposit target is a pinned constant rather than anything caller-supplied), then the round-trip itself is probably worth dropping.
> 

great catch, it seems like this actually needs design input so I will look into that and get back to this thread.

extension/src/popup/components/account/AccountHeader/index.tsx
CassioMG: hum yeah I'm all in for the notification but I'm afraid some users would still ignore it, I think having a mechanical way to prevent users from doing it would be a little bit safer. The notification would also make it clear for users why the deposit button is missing or disabled with a "This feature is temporarily unavailable" message or something like that. An alternative approach would be disabling our endpoints in an emergency case, but the feature flag would degrade more gracefully I think (and also easier than having to change an endpoint).

Users would still be able to check their positions or withdraw from it, they would just not be able to make new deposits. Wdyt?

I don't think it's a blocker though, we could work on it as a follow-up task after all 3 are implemented (deposit, withdraw, positions) if you think that makes sense

CassioMG: Although thinking again about this I think having a feature flag for withdraw could also be important, so we avoid users interacting with the contract

aristidesstaffieri: yeah I think its a valid concern. One thing that seems potentially inconsistent is that in other cases we choose to not stop users from interacting with the network, like in the case a transaction is flagged then we choose to warn users but let them continue if they choose to. I think this could be a similar case. I think what I can do here is just add the feature flag anyway and then we can choose to use it or not as a case for it arrives(or not). I will add that and report back.

I assume we would use the same feature flag for deposit and withdrawl?

CassioMG: Humm great point on having a consistent UX compared to Blockaid warnings, I haven't thought from this point of view 🤔 I think that's fair

But for some reason I feel like this feature will bring more attention and have more users actively using it with higher amounts so it seems a bit more sensitive, but that's just a hunch

In case we go for the feature flags, I'd suggest having 1 for earn_deposit and 1 for earn_withdraw so we have a bit more control, also considering we will have more protocols in the future I think it's possible there could be a situation where we want to disable only one of the features. Wdyt?

aristidesstaffieri:
> > ### `scanTx` result is computed but never read
> > **Nit / question** — no security impact, **not a merge blocker**.
> > **TL;DR:** The Earn deposit flow runs a Blockaid scan on the prepared transaction and stores the verdict on the simulation payload, but nothing downstream ever reads it — the review sheet doesn't surface it and the Confirm button doesn't consider it. Flagging it in case wiring up the warning UI was intended and got left behind; if it was deliberate (the scan and simulate endpoints share a host, so the verdict adds little here, and the deposit target is a pinned constant rather than anything caller-supplied), then the round-trip itself is probably worth dropping.
> > Detailed explanation (for agents)
> 
> great catch, it seems like this actually needs design input so I will look into that and get back to this thread.

@CassioMG update: this is now implemented according to the new design layout in 92e5f562a25032e42b208ac0140f3887ec20a9b3

extension/src/popup/components/account/AccountHeader/index.tsx
aristidesstaffieri: yeah that sounds good I will add both flags

extension/src/popup/helpers/blendDeposit.ts
aristidesstaffieri: added this in a03d32c4e7f62d61fa2ef5cd4417404cc141ab30

extension/src/popup/components/earn/EarnAmount/hooks/useSimulateEarnDeposit.tsx
aristidesstaffieri: while it was safe, it was inconsistent with other calls so I used the `internal` label in db451df2c60cdd1fc38368760309642b3e02e8f5

aristidesstaffieri:
> ### `earn.max_amount_selected` fires for every percentage shortcut, not just Max
> **Suggestion (analytics correctness)** — wrong-by-name event data from the moment the flow ships; cheap to fix now, awkward to fix once dashboards are built on it. **Not a merge blocker.**
> 
> **TL;DR:** All four percentage shortcuts on the deposit amount screen — 25%, 50%, 75% and Max — emit the same event named "max amount selected". The event does carry a `percent` property, so the underlying data is recoverable, but anything keyed on the event name alone (dashboards, funnels, alerting) will count a 25% tap as Max usage. The Send flow shipped this exact behaviour, got called out for it in review, and was subsequently fixed to emit only on the 100% tap — this is the same pattern going in the other direction.
> 

this is addressed in c60e904a4bbb80577307290f5fa980383b196cb6

aristidesstaffieri:
> ### Earn submit hook reports `status: "success"` when the submission fails
> **Bug (latent)** — the hook's own state is wrong on every failed deposit, but a second check in the consuming component currently keeps the wrong value off the screen. **Not a merge blocker.**
> 
> **TL;DR:** When the on-chain submission is rejected, the deposit hook records the attempt as a success anyway — it tracks the failure metric, then falls through and reports success regardless. Nothing is visibly broken today, because the screen decides what to render by also consulting the Redux submission status, which is correctly set to error. The risk is that the hook's own signal is wrong: anyone who later trusts it directly — a refactor, a new consumer, a test exercising the hook in isolation — silently gets "the deposit succeeded" for a deposit that failed.
> 

this is addressed in 55b1057da0b1d1b3e76762f52c453819304650fb

extension/src/popup/components/earn/helpers/formatPoolStats.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Compact USD for pool-scale figures — "$50.05M" rather than "$50,050,000".
+ *
+ * Returns "--" for null, which means the pool oracle has no fresh price. That
+ * is distinct from a real zero, which formats as "$0.00".
+ */
+export const formatCompactUsd = (value: number | null): string => {
+  if (value === null) {
+    return "--";
```
CassioMG: I think we have an existing constant for this on the codebase

extension/src/popup/components/earn/helpers/formatPoolStats.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+/**
+ * A rate as a percentage — 0.1694 becomes "16.94%".
+ *
+ * Null means no fresh oracle price and renders "--"; a genuine zero renders
+ * "0.00%". Never conflate the two.
+ */
+export const formatRate = (rate: number | null): string =>
+  rate === null ? "--" : `${new BigNumber(rate).times(100).toFormat(2)}%`;
```
CassioMG: same here for the "--" constant

extension/src/popup/components/earn/helpers/formatPoolStats.ts
aristidesstaffieri: indeed there is, thanks. This was updated in a22cfd87e846a54c9734b1d8472514ca554a0fd5

extension/src/popup/components/earn/helpers/formatPoolStats.ts
aristidesstaffieri: updated in a22cfd87e846a54c9734b1d8472514ca554a0fd5

PR #2977, opened 2026-08-24
chore(deps): migrate to @stellar/stellar-sdk 17.0.0

aristidesstaffieri:
## Summary

Bumps `stellar-sdk` and `stellar-sdk-next` from `16.0.0-rc.1` to `17.0.0` across `extension`, `@shared/api`, `@shared/constants` and `@shared/helpers`.

v17 rebuilds the `xdr` namespace as classes — unions become discriminated variants, enums become singletons, `int64` becomes `bigint` — and returns `Uint8Array` instead of `Buffer` from every byte-returning API. Most of the diff is mechanical call-site updates. 

## What's in this PR

- `*/package.json`, `yarn.lock` — SDK bump, `engines.node` → `>=22.12.0`, drops the direct `@stellar/js-xdr` dependency and `extension/src/types/js-xdr.d.ts`.
- `extension/src/popup/helpers/soroban.ts` — largest rewrite. Unions switch on `.type` with property arm access; `xdr.expectUnionVariant` preserves the previous throw-on-wrong-arm behavior.
- `signTransaction/Operations/index.tsx`, `.../KeyVal/index.tsx` — same treatment for host functions, contract executables, contract-id preimages and claim predicates.
- `@shared/api/helpers/soroban.ts` — `parseWasmXdr` reads contract specs via `xdr.decodeStream` instead of a hand-rolled `XdrReader`.
- `handlers/signBlob.ts`, `handlers/signAuthEntry.ts`, `popup/helpers/hardwareConnect.ts` — re-wrap `Buffer` at the two boundaries that require it.
- `jest.config.js` — add `@exodus/bytes` and `smol-toml` to the babel transform allowlist. v17's CJS entry `require()`s them, and Jest cannot load ESM through its CJS module registry.
- Tests across all workspaces — method renames, `xdr` construction updates, and mock fixtures that now have to carry real bytes.
- Also replaces `native.constructor === "Uint8Array"` (never true) in `scValByType` with an `instanceof` check, which now matters because `scValToNative` returns `Uint8Array` for non-UTF-8 string/symbol payloads.

## Notes for review

- `SignAuthEntryResponse` / `SignBlobResponse` stay `Buffer | null` on purpose. dApps on `freighter-api` <4.0.0/<4.2.0 receive the raw value over `runtime.sendMessage`, and `Buffer` and `Uint8Array` do not JSON-serialize identically. The handlers wrap `Keypair.sign()` so the public API stays byte-identical.
- `@ledgerhq/hw-app-str` still types `signHash`/`signTransaction` as `Buffer`, so `tx.hash()` / `tx.signatureBase()` are re-wrapped at that call.
- `simulateTransaction` now sends `useUpgradedAuth: true` by default. Freighter's own flows use source-account credentials so the recorded arm is unchanged, but custom-network RPCs now receive a field they did not before.
- Asset codes render via `assetCode.toJson()` — `XdrValue.toString()` base64-encodes.
- `scValByType` returns `null` rather than `undefined` for `scvLedgerKeyContractInstance`, matching the sibling void arm.

## Test plan

- [x] `yarn test:ci` — 208 suites / 1617 tests green
- [x] `yarn build:freighter-api` and `yarn build:extension` (incl. eslint) pass
- [x] `yarn test:e2e` — all 25 non-integration Playwright suites green
- [x] Live testnet round-trip on v17: simulate → assemble → sign → submit
- [x] CI green


@shared/api/helpers/soroban.ts
```diff
@@ -509,43 +510,36 @@ export const getLedgerEntries = async (
...
+    WebAssembly.Module.customSections(wasmModule, "contractspecv0")[0],
   );
 
-  const specs = [];
-  do {
-    specs.push(xdr.ScSpecEntry.read(reader));
-  } while (!reader.eof);
+  // v17: `xdr.ScSpecEntry.read(reader)` is gone along with the exported
+  // js-xdr Reader; `decodeStream` walks the back-to-back entries instead.
+  const specs = xdr.decodeStream(xdr.ScSpecEntry, specSection);
```
Ryang-21: BTW `Spec` can process contract wasms. So you can call `new contract.Spec(wasmBytes)`

aristidesstaffieri: thanks @Ryang-21 TIL I updated it to use `Spec.fromWasm` in e16cc6acfc990fabf9f67468024ee47c2fb5ee5a

piyalbasu:
### Render-time `TypeError` for revoke-signer-sponsorship ops with hash-based signers

**Low-likelihood bug** — reachable only for an uncommon op + signer-type combination, and an ErrorBoundary contains it, so **not a merge blocker** — but the error screen it triggers replaces the very view where the user decides whether to sign.

**TL;DR:** The buffer-formatting helper now asserts it receives raw bytes, but for one operation type the SDK hands it an already-hex-encoded string at runtime (the SDK's own type declarations disagree with its runtime here, and an existing type cast defeats the compiler check). Before this PR that mismatch silently rendered wrong text; after it, the signing-review screen throws while rendering. Worth fixing now since the same fix also closes the long-standing wrong-hex display bug.

**Steps to reproduce:**
1. Ask Freighter to sign a transaction containing a `revokeSponsorship` operation that revokes a **signer** sponsorship where the signer key is a `sha256Hash` or `preAuthTx` key.
2. Open the sign-transaction review screen.
3. The operation details fail to render (`TypeError: Expected \`Uint8Array\`, got \`string\``) and the ErrorBoundary error screen is shown instead of the review UI.

---

<details>

**Root cause:** `formattedBuffer` was switched to `xdr.encodeBytes(data, "hex")`, which delegates to `uint8ArrayToHex` from `uint8array-extras` — whose first statement is `assertUint8Array()`:

https://github.com/stellar/freighter/blob/56e9ef8871cc342d5fcdfb8eb2dad10242a263b7/extension/src/popup/helpers/formatters.ts#L146-L149

But in `@stellar/stellar-sdk@17.0.0`, the `revokeSponsorshipSigner` arm of `Operation.fromXDRObject` calls `convertXdrSignerKeyToObject`, which sets `signer.sha256Hash` / `signer.preAuthTx` to **hex strings** (via `uint8ArrayToHex(...)`), not `Uint8Array` — verified against the published 17.0.0 tarball. The SDK's `.d.ts` declares these fields as `Uint8Array`, so the types disagree with the runtime, and the `signer as unknown as Signer` cast at the call site prevents the typechecker from flagging it:

https://github.com/stellar/freighter/blob/56e9ef8871cc342d5fcdfb8eb2dad10242a263b7/extension/src/popup/components/signTransaction/Operations/index.tsx#L817-L823

https://github.com/stellar/freighter/blob/56e9ef8871cc342d5fcdfb8eb2dad10242a263b7/extension/src/popup/components/signTransaction/Operations/KeyVal/index.tsx#L360-L375

Pre-PR, `Buffer.from(<hex string>)` accepted the string and silently produced wrong hex (the display bug flagged on #2838 and never fixed). This PR converts that silent corruption into a thrown render.

The `setOptions` path (`KeyValueSigner`) is **not** affected: SDK 17's `setOptions` arm sets `signer.preAuthTx = key.preAuthTx.toBytes().slice()`, a real `Uint8Array`. `convertXdrSignerKeyToObject` is reached only from the revoke-sponsorship arm.

**Deterministic repro:** `uint8ArrayToHex("<64-char hex string>")` → ``TypeError: Expected `Uint8Array`, got `string` `` — or drive `formattedBuffer` in a unit test with the `signer.sha256Hash` produced by `Operation.fromXDRObject` on a `revokeSignerSponsorship` op.

1. *Cheap guard:* make `formattedBuffer` accept `Uint8Array | string` — if given a string, it is already hex from the SDK, so truncate/display it directly (skip `xdr.encodeBytes`). This also fixes the #2838 wrong-hex display, since the string is the correct hex the old `Buffer.from(<string>)` path mangled.
2. *Root cause:* normalize at the revoke-sponsorship call site in `Operations/index.tsx` — convert the SDK's hex-string `sha256Hash`/`preAuthTx` fields explicitly (e.g. via `hexToUint8Array`) before passing them down, and drop the `as unknown as Signer` cast so the compiler can catch the next SDK drift.

</details>

aristidesstaffieri:
> ### Render-time `TypeError` for revoke-signer-sponsorship ops with hash-based signers
> **Low-likelihood bug** — reachable only for an uncommon op + signer-type combination, and an ErrorBoundary contains it, so **not a merge blocker** — but the error screen it triggers replaces the very view where the user decides whether to sign.
> 
> **TL;DR:** The buffer-formatting helper now asserts it receives raw bytes, but for one operation type the SDK hands it an already-hex-encoded string at runtime (the SDK's own type declarations disagree with its runtime here, and an existing type cast defeats the compiler check). Before this PR that mismatch silently rendered wrong text; after it, the signing-review screen throws while rendering. Worth fixing now since the same fix also closes the long-standing wrong-hex display bug.
> 
> **Steps to reproduce:**
> 
> 1. Ask Freighter to sign a transaction containing a `revokeSponsorship` operation that revokes a **signer** sponsorship where the signer key is a `sha256Hash` or `preAuthTx` key.
> 2. Open the sign-transaction review screen.
> 3. The operation details fail to render (`TypeError: Expected \`Uint8Array`, got `string``) and the ErrorBoundary error screen is shown instead of the review UI.
> 

good catch thanks! `formattedBuffer` now accounts for this, tested manually as well in the playground with and without pre-auth. 7cfc48a751bb3a546cfbf559dfb76765a0bdbfbf

piyalbasu (approved):

PR #2991, opened 2026-09-01
Don't attach volume telemetry to pre-submission failures

JakeUrban:
## Summary

Follow-up to #2984. Don't attach volume telemetry to a pre-submission failure event.

Found while implementing the same telemetry on mobile ([stellar/freighter-mobile#996](https://github.com/stellar/freighter-mobile/pull/996)) and checking the two clients against each other.

### DIFF
--- a/extension/src/helpers/confirmationPriceSnapshot.ts
+++ b/extension/src/helpers/confirmationPriceSnapshot.ts
@@ -39,12 +39,17 @@ export interface ConfirmationSnapshotHandle {
 }
 
 /**
- * Issues ONE price fetch covering every leg's canonical id, started at
- * confirmation and never blocking signing/submission — callers do not await
- * this. `cachedDisplayPrices` is the price map already held for the on-screen
- * fiat estimate, captured by the caller at this same moment: it must reflect
- * "the price already shown to the user for this transaction", not whatever
- * the cache holds later when `resolve()` is called.
+ * Issues ONE price fetch covering every leg's canonical id, started once
+ * signing has succeeded and immediately before submission — as close to the
+ * transaction's execution as the flow allows — and never blocking submission,
+ * since callers do not await this. `cachedDisplayPrices` is the price map
+ * already held for the on-screen fiat estimate, captured by the caller at
+ * this same moment, and is the fallback `resolve()` closes on when the fetch
+ * has not landed by terminal status. Because the snapshot starts after
+ * signing, that map is the display cache as of the start of submission rather
+ * than as of the confirm tap — a password prompt or hardware approval in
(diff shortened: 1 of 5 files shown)


CassioMG:
### Two doc comments invalidated by the snapshot-timing change

**Docs only** — no behavior impact; **not a merge blocker**.

**TL;DR:** Moving the price snapshot to after signing leaves two load-bearing comments describing the old flow. The snapshot helper's own doc still says the fetch is "started at confirmation", while this PR's call site now correctly says the opposite ("after signing succeeded and immediately before submission") — a reader of the helper alone gets the wrong contract. And the failure screen's comment explaining why it emits no telemetry still claims the central emit site always has the price snapshot and transaction result in scope, which is no longer true for the new pre-submission failure path that emits with neither. Both are two-line comment fixes; these docs were relied on heavily during the parent PR's review, so keeping them truthful is worth the diff.

---

<details>

**(a) `confirmationPriceSnapshot.ts` module doc** still reads "Issues ONE price fetch covering every leg's canonical id, started at confirmation and never blocking signing/submission" — its sole caller now starts it post-signing. The `cachedDisplayPrices` sentence ("captured by the caller at this same moment... 'the price already shown to the user'") also loosened: post-signing, the display cache may have refreshed during a password prompt or hardware approval, so the fallback is no longer guaranteed to be confirm-click state:

https://github.com/stellar/freighter/blob/21959712979dd1a01aeea2b2b69f5e9f64533f88/extension/src/helpers/confirmationPriceSnapshot.ts#L41-L50

The updated call-site comment it now contradicts:

https://github.com/stellar/freighter/blob/21959712979dd1a01aeea2b2b69f5e9f64533f88/extension/src/popup/components/InternalTransaction/SubmitTransaction/hooks/useSubmitTxData.tsx#L258-L266

**(b) `SubmitFail/index.tsx` centralization comment** — the conclusion (single emit site, no double-counting) still holds, but the supporting rationale "it already has the confirmation price snapshot and the transaction result in scope" is false for the new `!isSigned` branch, which emits `paymentFailed`/`swapFailed`/`collectibleSendFailed` before any snapshot is started and without a transaction result:

https://github.com/stellar/freighter/blob/21959712979dd1a01aeea2b2b69f5e9f64533f88/extension/src/popup/components/InternalTransaction/SubmitFail/index.tsx#L31-L40

**Suggested fix:** reword (a) to "started once signing succeeds, immediately before submission, never blocking submission" and adjust the `cachedDisplayPrices` sentence accordingly; reword (b)'s rationale to note the pre-submission failure path deliberately emits without snapshot/result.

</details>

CassioMG (approved):
This LGTM - other than the stale-docs comment

PR #2994, opened 2026-09-02
fix(icons): show USDT0's icon, for new and existing users

piyalbasu:
## TL;DR

USDT0 shows no icon in the wallet, and once it has failed for a user it never comes back — not on reload, not on restart.

Two things cause it. USDT0 is carried by two of the default asset lists: Soroswap, with a direct HTTPS icon, and LOBSTR, with an `ipfs.io` gateway URL. **That gateway serves the image to `curl` but returns 403 to a browser User-Agent, so an `` can never load it.** Icon lookup was keeping whichever list it read last, which handed the browser the one URL it cannot fetch. And when that image failed, the failure was recorded to disk as "this asset has no icon", which lookup reads as "already tried, don't look again" — a verdict that was only ever meant to last a session.

This makes the **first** matching list win, since the list order is the user's priority order, and makes the cache stop reporting those stale "no icon" records. **Users already stuck recover on their first load after upgrading** — no migration needed: the fresh lookup overwrites the old entry.

Four lines of production code.

### Verifying

1. On `master`, hold `USDT0:GATISXX6BZ6NC7IKQBY37CJD4SOZL3CYZJWXEDG6JVIY4WBS6KXJHN6Q` on Mainnet and open the account view — no icon, and it stays missing across reloads.
2. Check the stored value in the service worker console:
   ```js
   chrome.storage.local.get("cachedAssetIcons", (r) =>
     console.log(r.cachedAssetIcons["USDT0:GATISXX6BZ6NC7IKQBY37CJD4SOZL3CYZJWXEDG6JVIY4WBS6KXJHN6Q"])
   );
   ```
   → `null`
3. Load this branch and open the account view. The icon appears, and the same check returns `"https://docs.usdt0.to/downloads/usdt0/Symbol_USDT0_Secondary.png"` — the stale record is gone from disk.

Confirmed end to end in a local build against the prod backend.

---

<details>
<summary>Implementation details (for agents)</summary>

**Root cause**

1. `getIconFromTokenLists` iterated the lists in order, but its `break` only exited the inner *record* loop. The outer loop over lists ran to completion, so the **last** matching list overwrote every earlier match. Ordering is meant to be priority (stellar.expert → Soroswap → LOBSTR), so USDT0 resolved to LOBSTR's `ipfs.io` URL instead of Soroswap's `docs.usdt0.to`.
2. `https://ipfs.io/ipfs/bafkreifaalohkkikosp27qp6qczwaffwseejvoaafkgv7ahyrqvejmpqou` returns **200 to curl and 403 to a browser User-Agent** — isolated to the UA alone, other headers make no difference. So the `` always failed.
3. `onError` → `retryAssetIcon` writes `iconUrl: null` to `CACHED_ASSET_ICONS_ID`, then consults only `getIconUrlFromIssuer` (home domain → SEP-1 TOML).
4. That fallback can never succeed for this asset: Horizon reports `home_domain: null` for the issuer, with `thresholds {0,0,0}` and a single signer at weight 0 — the account is permanently locked and **can never set one**. A valid `stellar.toml` with the right `image` does exist at `usdt0.to`, but the wallet has no way to discover that domain.
5. `getAssetIcons` skips any asset whose cached value is `null`, with a comment saying the null "is only stored in Redux, so we will re-try on next app reload". It wasn't — `retryAssetIcon` persisted it, so the skip outlived the session. With no icon there is no ``, so nothing fires the `onError` that would have retried: the asset could not recover on its own.

**What changed**

Label the outer loop and break out of it on the first match:
https://github.com/stellar/freighter/blob/6701092003ce12ae819f11efa281c970d551a720/%40shared/api/helpers/getIconFromTokenList.ts#L20-L55

Leave null entries out when reading the persisted icon cache:
https://github.com/stellar/freighter/blob/6701092003ce12ae819f11efa281c970d551a720/extension/src/background/messageListener/handlers/getCachedAssetIconList.ts#L1-L24

Filtering on **read** rather than deleting on write is what makes this a four-line fix instead of a storage migration: a stale entry becomes an ordinary cache miss, and `getIconFromTokenLists` already writes what it finds, so the fresh lookup overwrites it on disk. It also makes `retryAssetIcon`'s existing `iconUrl: null` message finally mean what it intended — clear this entry — with no change to that handler.

The in-session memo is deliberately untouched. It lives in Redux (`saveIconsForBalances`), which is where `getAssetIcons`' comment always claimed it lived; filtering at the storage boundary separates the two cleanly.

**Verification**

- New tests: the first matching list wins by issuer and by contract ID (both previously returned the later list's icon); the cache read drops nulls, passes real icons through, and handles an empty cache.
- Full suite: 1773 passed, 0 failed, and `tsc --noEmit` clean — identical to `master`.
- Manually confirmed against the prod backend: stored `null` → icon renders → stored value replaced with the working URL.

**Follow-ups / out of scope**

- **LOBSTR's list entry is broken.** It ships an `ipfs.io` URL that no browser can load. Worth reporting upstream — ordering saves us here only because Soroswap happens to list USDT0 too. It is the only such entry across all three default lists today (stellar.expert serves everything from `meta.stellar.expert`, all browser-loadable).
- **Nothing checks that a chosen icon URL actually loads.** Ordering fixes the known case; a list shipping a URL browsers reject is only caught by loading it. Explored in #2993 and deliberately not carried here — it needs a per-URL verdict cache and concurrency control to stay off the balances render path, which is a much larger change than a hotfix should carry.
- **`AssetIcon` latches its error state.** Once an image fails, the row keeps the broken-image glyph for the life of the component even if a working URL arrives, so a successful retry is invisible until the popup is reopened. Pre-existing and unrelated to these two lines, but it is why the existing retry path appears to do nothing.
</details>


extension/src/background/messageListener/handlers/getCachedAssetIconList.ts
```diff
@@ -10,6 +10,15 @@ export const getCachedAssetIconList = async ({
...
-    icons: assetIconCache,
+    // A null entry records that a lookup came up empty, and getAssetIcons reads
+    // one as "already tried, don't look again". That was only ever meant to
+    // last a session, but this cache is on disk, so the verdict outlived it: an
+    // asset whose icon failed once stayed iconless for good, with no way back —
+    // no icon means no <img>, so nothing fires the error handler that would
+    // have retried. Leaving nulls out makes them ordinary cache misses, and the
+    // fresh lookup overwrites the stale entry.
+    icons: Object.fromEntries(
+      Object.entries(assetIconCache).filter(([, iconUrl]) => iconUrl),
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 676006d3b — `lookupIconData` now passes `iconsSelector` instead of `{}`, so the session nulls do what they were meant to.

One correction on the framing though: this is not caused by the null filter. `getAssetIcons` records nothing in its result for an asset it skips, so a cached `null` and an absent entry are indistinguishable to this hook — `!icons[asset]` is true either way, the asset lands in `assetsWithoutIcons`, and `lookupIconData` runs. The repeat lookup happened identically before this PR; filtering the persisted nulls did not change how often that hook looks up.

Still worth fixing, so it is fixed here, with a test that pins the lookup pass receiving the session map.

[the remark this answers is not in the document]
piyalbasu: Reverted in dc29db9cb — the fix was unsafe, and I should have checked further before making it.

The session icon map is shared with a path that looks less hard. History's `getIconUrl` consults the token lists and, when they are cached — which the Account view guarantees via `saveTokenLists` — never reaches the `else if` issuer-toml branch, then records `iconUrl || null` for whatever it did not find. That null goes into the same Redux map via `saveIconsForBalances`.

So an asset whose icon exists only in its issuer's `stellar.toml` gets marked "no icon" by a lookup that never tried the toml, and honouring the map here made the Account view believe it — the asset then renders iconless for the rest of the session. That is a session-scoped version of the latched failure this PR exists to fix, so it made things worse in the same class. A transient error inside `getIconUrlFromIssuer` latches identically.

Filtering the nulls out before passing the map would be safe but pointless: it is exactly the iconless assets this was meant to stop re-looking-up.

The repeat lookups are real, but pre-existing and not introduced here — `getAssetIcons` records nothing for an asset it skips, so a cached null and an absent entry are indistinguishable to this hook and both already triggered the lookup. Fixing it properly means stopping the history path recording a negative verdict for a lookup it deliberately skipped, which is a change to another view and belongs outside a hotfix.

extension/src/background/messageListener/handlers/getCachedAssetIconList.ts
```diff
@@ -10,6 +10,34 @@ export const getCachedAssetIconList = async ({
...
+    // retried. Dropping nulls here turns them back into ordinary cache misses,
+    // and the fresh lookup overwrites the stale entry.
+    //
+    // TODO: this is a read-side workaround, not the real fix. It hides bad data
+    // rather than stopping it being written, and it permanently rules out ever
+    // storing a legitimate null here — which we may well want, since an asset
+    // with genuinely no icon currently re-runs the whole lookup chain (token
+    // lists, then Horizon, then the issuer's stellar.toml) on every cold popup
+    // open. A durable negative cache with a TTL is the sane answer to that, and
+    // this filter would silently swallow it.
```
JakeUrban: What is a "cold popup open"? What is the consequence of re-running the lookup chain for nulls besides USDT0?

@shared/api/helpers/getIconFromTokenList.ts
```diff
@@ -20,7 +20,11 @@ export const getIconFromTokenLists = async ({
 }) => {
   let verifiedToken = {} as AssetListReponseItem;
   let canonicalAsset = undefined as string | undefined;
-  for (const data of assetsListsData) {
+  // The lists arrive in the user's configured order, which is a priority order:
+  // the first list carrying the asset wins. Without the labeled break below the
+  // outer loop runs to completion and the LAST matching list silently overwrites
+  // earlier, higher-priority entries.
+  listLoop: for (const data of assetsListsData) {
```
JakeUrban, line 27: Just to confirm, soroswap's list is higher-priority than Lobstr's?

piyalbasu: Correct. As noted below, this is a bit of a bandaid fix that relies on ordering.

extension/src/background/messageListener/handlers/getCachedAssetIconList.ts
piyalbasu: "cold popup open" just means opening the extension for the first time (so no ephemeral data, like balances, has been cached). I can make this wording a bit more clear.

The consequence of re-running the lookup chain is just a couple of seconds of the loading icon state as we retry the icon (non-blocking). Previously, we wouldn't bother retrying and just say "this previously was not found" and show the default icon

JakeUrban (approved):

PR #3001, opened 2026-09-08
Activate and transact without first holding XLM

adrianvrj:
Receiving a stablecoin on Stellar requires first holding a different one:
1.5 XLM locked per user, paid either by a user who doesn't know what a lumen
is or by an app that never gets it back.

## What this unlocks

- A user holding **0 XLM** with an inbound claimable balance can create their
  account, open the trustline and claim it in **one signature**, without leaving
  Freighter. Today the wallet's only possible answer is "go acquire XLM first".
- Fees and base reserves are paid from the asset the user was actually sent
  (USDC, EURC, PYUSD, AQUA and others on a liquidity-verified allowlist) through
  a strict-receive path payment in the same transaction. The user never has to
  acquire a second asset to use the first one.
- First-run onboarding stops depending on an on-ramp selling ~1.5 XLM, an amount
  no on-ramp is designed to sell.
- The user chooses which of their held assets pays, and sees the exact ceiling
  (`sendMax`) before signing.

## What this does not change

- Opt-in and conditional: the flow only surfaces when the account is unfunded
  and a supported asset is inbound. Every existing path is untouched.
- No change to key handling or the signing pipeline.
- Freighter re-derives the server-built transaction locally before prompting,
  (operation list, sponsorship sandwich, fee destination, `destAmount`, `sendMax`
  and every path hop) against the request the wallet itself made, so a hostile
  or compromised service cannot alter user intent.
- The sponsor signs only the outer fee bump. `SetOptions`, signers and
  thresholds are not expressible in the operation set, so it cannot take over an
  account.

## Test plan
- [ ] Unfunded account with a Testnet USDC claimable balance: Activate uses the inbound token, one signature, 0 spendable XLM.
- [ ] Classic send of a held token with 0 spendable XLM: fee asset picker, quote, sign, submit.
- [ ] Classic swap of a held token with 0 spendable XLM: same picker, Max leaves enough of the fee token.
- [ ] Funded account with enough XLM: existing send/swap/onboarding paths unchanged.

https://github.com/user-attachments/assets/0c731093-8787-4494-a03e-1bae3fb5ea0a


adrianvrj:
hi @aristidesstaffieri I am working on a solution so users do not have to hold XLM for account activation or paying fees, I think this is something worth taking a look since it is a major UX improvement.

The sdk used is published and maintained by our team and we are happy to discuss if any changes are needed :)

aristidesstaffieri:
hey @adrianvrj thanks for your work on this PR. For external integrations, we need to put the feature through our product lifecycle which will require input from internal/external stakeholders. I suggest opening an issue to accompany this PR with the details of the integration so that we can get attention on this issue before we take a look at the code here. 

Let me know if you have any questions.

adrianvrj:
> hey @adrianvrj thanks for your work on this PR. For external integrations, we need to put the feature through our product lifecycle which will require input from internal/external stakeholders. I suggest opening an issue to accompany this PR with the details of the integration so that we can get attention on this issue before we take a look at the code here. 
> 
> 
> 
> Let me know if you have any questions.

Tysm, will do!

PR #3009, opened 2026-09-11
Skip the Blockaid bulk scan when no scannable assets are present

### LINKED ISSUE #2905: Dont call blockaid on empty token list
Latest

### What did you do?
Observed error from blockaid because we are calling for a scan with an empty token list

No blockaid related errors, should guard against this case

### What did you see instead?
API error, increasing rate limit frequency

devtechedge:
## What and why

In \`makeDisplayableBalances\`, the \`scan-asset-bulk\` request went out whenever the account was on mainnet, even when every balance was native XLM or an LP share and nothing was appended to \`asset_ids\`. The empty token list made the Blockaid call error and consumed rate limit for nothing.

The v2 balance path (\`addBlockaidScanResults\`) already skips the scan when the scannable id list is empty; this mirrors that guard on the standalone path by checking \`url.searchParams.has("asset_ids")\` before fetching.

Fixes #2905

## Testing

The touched helper is covered by existing suites; the behavior change only affects accounts with no scannable assets, which previously errored into the catch block.

### DIFF
--- a/@shared/helpers/stellar.ts
+++ b/@shared/helpers/stellar.ts
@@ -84,12 +84,16 @@ export const makeDisplayableBalances = async (
       }
     }
 
-    try {
-      const response = await fetch(url.href);
-      const data = await response.json();
-      blockaidScanResults = data.data.results;
-    } catch (e) {
-      console.error(e);
+    // Skip the scan entirely when there is nothing to scan: an empty token
+    // list makes the Blockaid call error out and wastes rate limit.
+    if (url.searchParams.has("asset_ids")) {
+      try {
+        const response = await fetch(url.href);
+        const data = await response.json();
+        blockaidScanResults = data.data.results;
+      } catch (e) {
+        console.error(e);
+      }
     }
   }
 


devtechedge:
Hi - gentle check-in on this Blockaid empty-list guard whenever you have a moment.

In standalone `makeDisplayableBalances`, the `scan-asset-bulk` request still went out on mainnet even when every balance was native XLM or an LP share and nothing was appended to `asset_ids`, so the empty token list errored and burned rate limit for nothing (`Fixes #2905`).

This mirrors the v2 `addBlockaidScanResults` guard in `@shared/helpers/stellar.ts` by checking `url.searchParams.has("asset_ids")` before fetching.

The behavior change only affects accounts with no scannable assets, which previously fell into the catch path.

I am glad to add the regression test Copilot suggested for a mainnet account with only native XLM and/or LP shares, and to rebase onto current `master` if that would help (the branch is a little behind).

Whenever it fits your queue is fine - still ready from my end.

